Issue #72, opened 2012-10-25
Unicode description fail


warvariuc:
In some cases folding a description doesn't work correctly causing

`DatabaseError: invalid byte sequence for encoding \"UTF8\": 0xd00d\n\n"`

A simple test to reproduce:

```
# coding: utf-8

import icalendar

event = icalendar.cal.Component()

event.set('description', "АБВГДЕЁЖЗИЙКЛМНОПРСТУФХЦЧШЩЬЫЪЭЮЯ")

calendar = icalendar.Calendar.from_ical(
    "BEGIN:VCALENDAR\r\n"
    "VERSION:2.0\r\n"
    "PRODID:'-//max m//icalendar.mxm.dk/'\r\n"
    "END:VCALENDAR\r\n"
)

calendar.add_component(event)

content = calendar.to_ical()

print content
content.decode('utf-8')
```

PR #73, opened 2012-10-25
Fix for long string folding with unicode chars

warvariuc:
https://github.com/collective/icalendar/issues/72

Reverted old algorithm (instead of 'wrap')for folding long strings - not unicode text is not folded in the middle of a multibyte sequence.
Removed some tests which were not run because of bad names (not started with 'test', like 'XtestTrailingNewline'). I put '1/0' in them to be sure they are not used.
Update tests for checking long line folding.

### DIFF
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -446,9 +446,10 @@ class Contentline(str):
         "Turns a tuple of parts into a content line"
         (name, params, values) = parts
         try:
-            if values and not isinstance(values, basestring):
-                if hasattr(values, 'to_ical'):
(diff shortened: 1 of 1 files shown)


src/icalendar/parser.py
```diff
@@ -414,7 +446,7 @@ def from_parts(parts):
         "Turns a tuple of parts into a content line"
         (name, params, values) = parts
         try:
-            if values and not isinstance(values, basestring):
+            if values and not isinstance(values, str):
```
thet: @rnixx can you please review at this change, since you have refactored unicode support?

PR #78, opened 2012-11-22
Fix icalendar "time" type support

rdunklau:
I'm not sure if the proposed change will put others at risk of regressions, since the time data type is not used in the types_factory. 

All tests pass, and (rudimentary) new tests have been added to cover the fix.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,7 +7,7 @@ Changelog
 * Supports to_ical() on date/time properties for dates prior to 1900.
   [cdevienne]
 * Fix time supports. vTime events can be instantiated with a datetime.time
-  object, and do not inherit from datetime.time itself, but from vDatetime.
+  object, and do not inherit from datetime.time itself.
 
 
 3.1


src/icalendar/prop.py
```diff
@@ -416,7 +416,7 @@ class vDDDTypes:
 
     def __init__(self, dt):
         "Returns vDate from"
-        if type(dt) not in (datetime, date, timedelta):
+        if type(dt) not in (datetime, date, timedelta, time):
             raise ValueError ('You must use datetime, date or timedelta')
```
thet, line 420: ... or time

src/icalendar/prop.py
```diff
@@ -1150,7 +1155,7 @@ def from_ical(ical):
     from_ical = staticmethod(from_ical)
 
 
-class vTime(time):
+class vTime:
     """A subclass of datetime, that renders itself in the iCalendar time
```
thet, line 1159: A subclass of time

Issue #79, opened 2012-12-04
It is illegal to have a double quote character inside parameter values


warvariuc:
```
# coding: utf-8

import icalendar

content = """\
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//Apple Inc.//iCal 5.0.1//EN
CALSCALE:GREGORIAN
BEGIN:VEVENT
END:VEVENT
END:VCALENDAR
"""

cal = icalendar.Calendar.from_ical(content)
attendee = icalendar.vCalAddress('test-cal@bk.ru')
attendee.params['PARTSTAT'] = 'DECLINED'
attendee.params['CN'] = "Арамис\", д’Аламеда"
cal.subcomponents[0].add('ATTENDEE', attendee)
print cal.to_ical()
```

Result:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//Apple Inc.//iCal 5.0.1//EN
CALSCALE:GREGORIAN
BEGIN:VEVENT
ATTENDEE;CN="Арамис", д’Аламеда";PARTSTAT=DECLINED:test-ca
 l@bk.ru
END:VEVENT
END:VCALENDAR
```

Note the CN parameter.

http://stackoverflow.com/questions/3014137/can-should-ical-property-parametrs-be-escaped

It is illegal to have a double quote " character inside parameter values, so they should be deleted (or otherwise removed from the parameter value).

What do you think?

PR #81, opened 2013-01-15
Quoting parameter values

warvariuc:
Please see the changes and tell me what you think about them:

In some place Python 2.6's str.format is used, so i also polished some other places:
- use decorator @staticmethod
- removed double spaces
- put two spaces before comments in the same line wit code
- exceptions raising in form Exception()
- using ternary "operator": result = True if condition else False
- if using a newer version of Python, do not load unittest2

Added space, single-quote char and other chars to the list of chars that fors parameter value to be quoted when generating icalendar content.

Before generating ical content all double quotes from parameter values are replaced with single quotes, as it is illegal to have double quotes in parameter values (https://github.com/collective/icalendar/issues/79)
Added tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,6 @@
 .DT_Store
 .coverage
 .tox/
-.project
-.pydevproject
-.settings/
 bin/
 build/
 coverage-*


src/icalendar/tests/test_timezoned.py
```diff
@@ -1,11 +1,15 @@
...
 import pytz
 import datetime
 import dateutil.parser
 import os
 
+import unittest
+if not hasattr(unittest.TestCase, 'assertIsNotNone'):
+    import unittest2 as unittest
+
+
```
garbas, line 12: i believe we should import from untitest2 first and if that fails then we import unittest

``` python
try:
  import unittest2 as unittest
except:
  import unittest
```

src/icalendar/tests/test_time.py
```diff
@@ -1,8 +1,12 @@
-import unittest2 as unittest
 import icalendar
 import datetime
 import os
 
+import unittest
+if not hasattr(unittest.TestCase, 'assertIsNotNone'):
+    import unittest2 as unittest
+
+
```
garbas, line 9: same here

src/icalendar/tests/test_encoding.py
```diff
@@ -1,12 +1,16 @@
...
 import pytz
 import datetime
 import os
 
 
+import unittest
+if not hasattr(unittest.TestCase, 'assertIsNotNone'):
+    import unittest2 as unittest
+
+
```
garbas, line 13:
``` python
try:
  import unittest2 as unittest
except:
  import unittest
```

src/icalendar/tests/test_cases.py
```diff
@@ -1,8 +1,12 @@
 # -*- coding: utf-8 -*-
-import unittest2 as unittest
 import icalendar
 import os
 
+import unittest
+if not hasattr(unittest.TestCase, 'assertIsNotNone'):
+    import unittest2 as unittest
+
+
```
garbas, line 9:
``` python
try:
  import unittest2 as unittest
except:
  import unittest
```

src/icalendar/tests/test_timezoned.py
warvariuc, line 12: What if you have an old Python version and you don't have `unittest2` installed? You end up having `unittest` successfully imported, but which doesn't have required methods like `assertIsNotNone`?

garbas, line 12: Quoting warvariuc (2013-01-15 17:09:19)

> In src/icalendar/tests/test_timezoned.py:
> 
> >  import icalendar
> >  import pytz
> >  import datetime
> >  import dateutil.parser
> >  import os
> > 
> > +import unittest
> > +if not hasattr(unittest.TestCase, 'assertIsNotNone'):
> > -    import unittest2 as unittest
> >   +
> >   +
> 
> What if you have an old Python version and you don't have unittest2 installed?
> You end up having unittest successfully imported, but which doesn't have
> required methods like assertIsNotNone?

currently unittest2 is always installed (as per setup.py).
it should be made conditional to be only installed for python 2.6 and lower.

## 

Rok Garbas - http://www.garbas.si

.gitignore
```diff
@@ -6,6 +6,9 @@
 .DT_Store
 .coverage
 .tox/
+.project
+.pydevproject
+.settings/
```
thet: i don't care much about these additions, but IMO these should go into the global .gitignore.

src/icalendar/parser.py
```diff
@@ -121,20 +123,25 @@ def paramVal(val):
...
-        validator = QUNSAFE_CHAR
+    validator = QUNSAFE_CHAR if quoted else UNSAFE_CHAR
     if validator.findall(value):
-        raise ValueError, value
+        raise ValueError(value)
+
+
+# chars presence of which in parameter value will be cause the value
+# to be enclosed in double-tuotes
+QUOTABLE = re.compile("[,;: ’']")
```
thet, line 142: can you point me, where this rule is defined? i can't find it.
while it obviously makes sense, we should follow the standard as close as possible.
RFC5545, section 3.2, paragraph 2 says just that property values containing any of: ;:,
must be double-quoted.

thet, line 142: ok, 3.2.2 gives an example with double quoted property values with spaces in it...

src/icalendar/tests/test_timezoned.py
warvariuc, line 12: When i install icalendar using pip directly from github, `unitest2` is not installed automatically. If i clone the source manually into a folder, obviously i have to install `unittest2` manually if i have an older version of Python.

```
vic@vic-desktop ~/projects/icalendar/src (master) » mkvirtualenv test
New python executable in test/bin/python
...
[test] vic@vic-desktop ~/projects/icalendar/src (master) » pip install git+git://github.com/collective/icalendar.git
Downloading/unpacking git+git://github.com/collective/icalendar.git
...
[test] vic@vic-desktop ~/projects/icalendar/src (master) » pip install nose
Downloading/unpacking nose
...
[test] vic@vic-desktop ~/projects/icalendar/src (master) » cd ~/projects/.venv/test/lib/python2.7/site-packages/icalendar
[test] vic@vic-desktop ~/projects/.venv/test/lib/python2.7/site-packages/icalendar » nosetests
E.EE.EE
======================================================================
ERROR: Failure: ImportError (No module named unittest2)
...
----------------------------------------------------------------------
Ran 7 tests in 0.039s

FAILED (errors=5)
```

Also, i think "explicit is better then implicit", better to not imply that unittest2 is installed, imo.

src/icalendar/parser.py
warvariuc, line 142: That's why i wanted your comments on these changes, which are arguable. As far as i understand there are characters which MUST be double-quoted, but actually you can double-quote all parameter values, just to be sure to client breaks on other symbols (Outlook?)...

thet, line 142: makes sense.
i've reported the SPACE case to http://www.rfc-editor.org/errata_search.php?rfc=5545
let's see what they say...
but you're right - it doesn't break anything because everything can be double quoted. so I'm OK with merging.

.gitignore
warvariuc: Thank for the note - i din't think about this possibility. I'll remove this.

src/icalendar/tests/test_timezoned.py
thet, line 12: thats because unittest2 is in the "test" extra of extras_require. IMO, it makes no sense to install unittest2 on production environments where not tests will ever run. on development machines you can install it like so:
$ pip install "icalendar [test]"

assuming your pip installation isn't years-old (https://github.com/pypa/pip/issues/7)

the above doesn't work when installing directly from a git repository.

src/icalendar/parser.py
thet, line 142: my errata to the spec was rejected: http://www.rfc-editor.org/errata_search.php?eid=3457
still, we can double-quote if we want.

src/icalendar/parser.py
```diff
@@ -121,20 +123,25 @@ def paramVal(val):
...
+
 
-QUOTABLE = re.compile('[,;:].')
```
warvariuc, line 137: Please note that i removed the trailing dot in the pattern. Am i wrong? Why was it there?

thet, line 137: that would just not match at the end of a line, since the dot does match any character except a newline. but it doesn't make any sense, does it? since the tests pass, the change should be ok.

PR #86, opened 2013-02-01
Added a JSONEncoder subclass that permits the Calendar tree to be encoded as JSON

rbpasker:
here's a working example

``` python

from json import JSONEncoder
import json
from datetime import datetime
from icalendar import Calendar, vDDDTypes, Event, ICalendarEncoder
import pprint

cal = Calendar.from_ical(open('test.ics','rb').read())

for event in cal.walk(name="VEVENT"):
    suspect = json.dumps(event, cls=ICalendarEncoder)
    working = json.loads(suspect)

pprint.pprint(working)

```

### DIFF
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -57,6 +57,9 @@ from icalendar.parser import escape_char
 from icalendar.parser import unescape_char
 from icalendar.parser import tzinfo_from_dt
 
+from json import JSONEncoder
+import json
+
 
 SequenceTypes = [TupleType, ListType]
 
@@ -1553,3 +1556,19 @@ class TypesFactory(CaselessDict):
         type_class = self.for_property(name)
         decoded = type_class.from_ical(value)
         return decoded
+
+
+class ICalendarEncoder(JSONEncoder):
+    def default(self, obj, markers=None):
+        
+        try:
+            if obj.__module__.startswith("icalendar.prop"):
+                return (obj.to_ical())
+        except AttributeError:
+            pass
+
+        if isinstance(obj, datetime):
+            return (obj.now().strftime('%Y-%m-%dT%H:%M:%S'))
+        
+        return JSONEncoder.default(self,obj)    
+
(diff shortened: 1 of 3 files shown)


src/icalendar/prop.py
```diff
@@ -1553,3 +1556,19 @@ def from_ical(self, name, value):
...
+        try:
+            if obj.__module__.startswith("icalendar.prop"):
+                return (obj.to_ical())
+        except AttributeError:
+            pass
+
+        if isinstance(obj, datetime):
+            return (obj.now().strftime('%Y-%m-%dT%H:%M:%S'))
+        
+        return JSONEncoder.default(self,obj)
```
thet, line 1573: wouldn't a call to super be more appropriate here?

src/icalendar/prop.py
```diff
@@ -1553,3 +1556,19 @@ def from_ical(self, name, value):
         type_class = self.for_property(name)
         decoded = type_class.from_ical(value)
         return decoded
+
+
+class ICalendarEncoder(JSONEncoder):
+    def default(self, obj, markers=None):
+
```
thet, line 1563: markers is not used

src/icalendar/prop.py
```diff
@@ -1553,3 +1556,19 @@ def from_ical(self, name, value):
         type_class = self.for_property(name)
         decoded = type_class.from_ical(value)
         return decoded
+
+
+class ICalendarEncoder(JSONEncoder):
```
thet, line 1561: A different name would be better. Interfaces are always named ISomething  by convention. ICalendarEncoder looks like an interface.

rbpasker, line 1561: Sure, go ahead and fix it

src/icalendar/prop.py
rbpasker, line 1573: sure, go ahead and fix it

untitaker:
Closing due to inactivity, if there's continued interest in fixing up this PR please open up a new one!

PR #98, opened 2013-03-20
Please comment

warvariuc:
I did some changes and i need your feedback on them.

```
Now not accepting unescaped new line characters in a content line.
When folding a line - preserve space at the end, if present.
Modified the tests accordingly.
```

The changes are arguable, so what do you think?

### DIFF
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -16,11 +16,11 @@ from .parser_tools import (
     data_encode
 )
 
+
 def escape_char(text):
     """Format value according to iCalendar TEXT escaping rules.
     """
     assert isinstance(text, basestring)
-    # TODO: optimize this
     # NOTE: ORDER MATTERS!
     return text.replace('\N', '\n')\
                .replace('\\', '\\\\')\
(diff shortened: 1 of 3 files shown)


src/icalendar/parser.py
```diff
@@ -291,6 +258,8 @@ class Contentline(unicode):
     """
     def __new__(cls, value, strict=False, encoding=DEFAULT_ENCODING):
         value = to_unicode(value, encoding=encoding)
+        assert u'\n' not in value, ('Content line can not contain unescaped '
```
thet, line 261: wouldn't that match also `u'\\n'`?

warvariuc, line 261:
```
>>> u'\n' in u'first line\nsecond line'
True
>>> u'\n' in u'first line\\nsecond line'
False
>>> 
```

thet, line 261: tnx

src/icalendar/parser.py
```diff
@@ -64,50 +63,20 @@ def foldline(text, length=75, newline='\r\n'):
...
-#                replace_whitespace=False
-#                )
-#            )
+    assert isinstance(line, unicode)
+    assert u'\n' not in line
+
+    ret_line = u''
+    byte_count = 0
+    for char in line:
+        char_byte_len = len(char.encode('utf-8'))
```
rnixx, line 72: please use DEFAULT_ENCODING instead of 'utf-8'

PR #138, opened 2014-05-29
Fix tests for latest pytz

untitaker:
See https://bugs.launchpad.net/pytz/+bug/1319939

### DIFF
--- a/src/icalendar/tests/test_timezoned.py
+++ b/src/icalendar/tests/test_timezoned.py
@@ -36,8 +36,10 @@ class TestTimezoned(unittest.TestCase):
         ev1 = cal.walk('VEVENT')[0]
         self.assertEqual(
             ev1.decoded('DTSTART'),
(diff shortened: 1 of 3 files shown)


src/icalendar/tests/test_timezoned.py
```diff
@@ -36,8 +36,10 @@ def test_create_from_ical(self):
         ev1 = cal.walk('VEVENT')[0]
         self.assertEqual(
             ev1.decoded('DTSTART'),
-            datetime.datetime(2012, 2, 13, 10, 0, 0,
-                              tzinfo=pytz.timezone('Europe/Vienna')))
+            pytz.timezone('Europe/Vienna').localize(
+                datetime.datetime(2012, 2, 13, 10, 0, 0)
+            )
```
thet, line 41: +1! actually, using localize is necessary, otherwise we don't even get DST normalization.

PR #147, opened 2014-09-05
Doc fixes

t-8ch:
Commit 029b70e is more like a proof of concept, it has some issues noted in its commit message.

### DIFF
--- a/src/icalendar/cal.py
+++ b/src/icalendar/cal.py
@@ -386,7 +386,7 @@ class Component(CaselessDict):
     def to_ical(self, sorted=True):
         '''
         :param sorted: Whether parameters and properties should be
-        lexicographically sorted.
+                       lexicographically sorted.
         '''
 
         content_lines = self.content_lines(sorted=sorted)


docs/api.rst
```diff
@@ -0,0 +1,50 @@
+API Reference
+-------------
+
+icalendar.parser
+++++++++++++++++
+
+.. automodule:: icalendar.parser
```
untitaker: I don't think we should expose the `parser` module. At least don't put it first into the list.

docs/api.rst
```diff
@@ -0,0 +1,50 @@
...
+++++++++++++++++++++++
+
+.. automodule:: icalendar.caselessdict
+   :members:
+
+
+icalendar.compat
+++++++++++++++++
+
+.. automodule:: icalendar.compat
```
untitaker: Same for this.

docs/api.rst
```diff
@@ -0,0 +1,50 @@
...
++++++++++++++
+
+.. automodule:: icalendar.cal
+   :members:
+
+
+icalendar.caselessdict
+++++++++++++++++++++++
+
+.. automodule:: icalendar.caselessdict
```
untitaker: Same for this.

PR #154, opened 2014-12-07
[RFC] creating timezones from VTIMEZONE components

### LINKED ISSUE #44: ignores VTIMEZONE
If an iCal file has something like:    (this is from Outlook 2010)

```
BEGIN:VTIMEZONE
TZID:Eastern Standard Time
BEGIN:STANDARD
DTSTART:16011104T020000
RRULE:FREQ=YEARLY;BYDAY=1SU;BYMONTH=11
TZOFFSETFROM:-0400
TZOFFSETTO:-0500
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:16010311T020000
RRULE:FREQ=YEARLY;BYDAY=2SU;BYMONTH=3
TZOFFSETFROM:-0500
TZOFFSETTO:-0400
END:DAYLIGHT
END:VTIMEZONE
```

individual events are supposed to be able to reference it by name, by TZID:

```
DTSTART;TZID="Eastern Standard Time":20120130T100000
```
instead, this package depends entirely on pytz for timezone lookup, ignoring timezones that are defined in the file.

geier:
I'd like to run this by you before it is ready, so everybody can voice their conceptual problems with this approach early. Once this is ready it should close #44.

TZIDs known to the TZDB, are left alone and data from the TZDB is used as before. Every TZ is only created once and registered in a global timezone dict.

Some caveats:
 * we are currently relying (especially the tests) on pytz internals
 * this only works when VTIMEZONE components come before DATE-TIME components that make use of them (while this seems to be the standard way to generate icalendars, there is no word about this in the RFC). To circumvent this, we would probably need some architectural changes, like  two rounds of parsing.
 * tz creation feels a bit brute-forced into cal.py, but there are so many exceptions in from_ical() already, that I don't feel  too guilty about introducing one more.

TODO
--------

* [x] check for the edge case of DST moving a timezone over the date line
* [x] some rounding issues (see to_tz() )
* [x] more tests
* [x] deal with timezones without TZNAME (they are optional)


CHANGES.rst
```diff
@@ -2,11 +2,13 @@
 Changelog
 =========
 
-3.8.5 (unreleased)
+3.9.0 (unreleased)
 ------------------
 
 - Documentation fixes.
   [t-8ch, thet]
+- Creating timezone objects from VTIMETZONE components [geier]
```
untitaker: typo

src/icalendar/cal.py
```diff
@@ -483,6 +491,89 @@ class Timezone(Component):
...
+
+            for transtime in transtimes:
+                transitions.append((transtime, offsetfrom, offsetto, tzname))
+
+            if component.name == 'STANDARD':
+                dst[tzname] = 0
+            elif component.name == 'DAYLIGHT':
+                dst[tzname] = 1
+            else:
+                raise ValueError('unknown component name')
```
untitaker: I think the for-loop should continue and just ignore unknown subcomponents.

src/icalendar/cal.py
```diff
@@ -483,6 +491,89 @@ class Timezone(Component):
     required = ('TZID', 'STANDARD', 'DAYLIGHT',)
     singletons = ('TZID', 'LAST-MODIFIED', 'TZURL',)
 
+    def to_tz(self):
```
untitaker, line 548: Right now this function seems really hard to grasp for me. I suggest moving the bodies of the two big for-loops into their own functions so that it's clear which state is mutated.

src/icalendar/cal.py
```diff
@@ -483,6 +491,89 @@ class Timezone(Component):
...
+                # when the first transition is to dst, we didn't find anything in
+                # the past, so we have to look into the future
+                if not dst_offset:
+                    for index in range(num, len(transitions)):
+                        if not dst[transitions[index][3]]:  # [3] is the name
+                            dst_offset = osto - transitions[index][2]  # [2] is osto
+                            break
+            transition_info.append((osto, dst_offset, name))
+
+        cls = type(zone, (DstTzInfo,), dict(
```
untitaker, line 591: I kind-of understand what this does, but it still seems like magic to me, a subclass seems cleaner?

geier, line 591: I agree, that this is kind of magic, I had to read [this](http://www.jeffknupp.com/blog/2013/12/28/improve-your-python-metaclasses-and-dynamic-classes-with-type/) to understand how it works. But this is actually [how pytz's timezones are made](http://bazaar.launchpad.net/~stub/pytz/devel/view/head:/src/pytz/tzfile.py#L120), too.

geier:
Does anyone have the time to have a look at this again?

geier:
I just discovered that there is at least one more issues to deal with, that is timezones without a TZNAME property.

thet:
i'll give review as soon as possible. we're using something similar in
plone.app.event and i'd be happy to get that functionality out of there
to icalendar, where it belongs.

thet:
is merged.
 thanks for this big improvement!

thet:
https://github.com/collective/icalendar/commit/e882e0379ed58cbb8fce8af38adaa5bd862b2b3a

geier:
Thanks! I'll be on standby for all the incoming bug reports...

thet:
we already got a test failure within a different test setup:
http://jenkins.plone.org/job/plone-5.0-python-2.7/lastCompletedBuild/testReport/icalendar.tests.test_timezoned/TestTimezoneCreation/test_create_america_new_york/
@geier do you have an idea why this happens? is the test unstable against different sorting methods on different machines? or does the timezone or date setting on the server affect test results?

untitaker:
That link 404s for me.

thet:
another one: http://jenkins.plone.org/job/plone-5.0-python-2.7/528/testReport/icalendar.tests.test_timezoned/TestTimezoneCreation/test_create_america_new_york/

geier:
:hankey: I'll have a look

geier:
the obvious explanation is the server running with a really old version of the tzdb, do you have any idea what operating system (and which version) this is running on?

thet:
@tisto @gforcada which operating system in which version is our jenkins server on? and how is are the icalendar tests run?

gforcada:
Operating system is ``Ubuntu 14.04.2 LTS`` on all nodes and master.

As for how the icalendar tests are run ... could you elaborate which check do you want us to do on the server?

If it's timezone the timezone, here you go:
- all nodes are in ``Tue Mar 24 16:52:33 PDT 2015`` (Pacific Daylight Time)
- master is in ``Wed Mar 25 00:52:44 CET 2015`` (Central European Time)

Is that what you were looking for?

thet:
@gforcada thanks, maybe these infos can help when debugging the test failures.
regarding on how the tests are run - i meant, how the icalendar tests are started, if there is some special jenkins configuration on icalendar. but i found it, it's simply in https://github.com/plone/buildout.coredev/blob/5.0/tests.cfg#L47

thet:
by the way, the Plone jenkins jobs are green again, because we're simply depending on the previous released version now.

geier:
This makes less and less sense (as I'm testing on `Ubuntu 14.04.2 LTS` myself).

@gforcada is there a way to run a test on a branch (or another repository)? (so I don't pollute icalendar's history while trying to fix this)

regebro:
The failing list starts with a transition *from* DST, which makes no sense, so it's definitely not a change in the Timezone rules, but a bug in how they are extracted.

dateutil.rrule() does not include the event instance on the start date given (dstart). That seems to me on a superficial look of the code to be the error here, but I could be wrong.

geier:
If I'm not reading py.test's output wrong, the *bad* list is the one from pytz, meaning there is an extra transition before the first one described by the ``RRULE``.

gforcada:
@geier you can run the pull request job as much as you want: http://jenkins.plone.org/job/pull-request/ login with your github account, set this pull request URL on the parameter and wait around 1 hour... (sorry all tests are run sequentially, we will improve that in the future...)

regebro:
Aha, I'm actually reading the test now, and the tests takes the transition times from pytz and then just takes the 95th transition and onwards. You can't do that, you need to filter them on transition time.

This is because some distributions skip some of the first time zone transitions for certain time zones. In America/New_York I seem to remember that it's the local time definition that is skipped, so that's exactly one transition that is ignored, which fits with this.

"Why?" You say. "WHY!?" Do you *really* want to know? Can you handle the truth?

geier:
@gforcada Thank you!

@regebro that's what I feared and what I still find strange (as I'm typing this from the very same operating system and the test runs fine here)

regebro:
Well, it's not strange, it's just how it is. Split on the date instead of just doing [:95] and you should be fine.

geier:
this should be fixed with 31e8af3631b143d1da895f3a402039f34ec1c9cb

PR #157, opened 2015-03-21
Recurring rules and trailing semicolons

sleeper-public:
I ran iCalendar on a calendar from CalReply.com (SyFy are using them for their episode reminders - http://on.syfy.com/) and got hit with this:

```
  File "/usr/lib/python3.4/site-packages/icalendar/prop.py", line 669, in from_ical
    key, vals = pairs.split('=')
ValueError: need more than 1 value to unpack
During handling of the above exception, another exception occurred:
  ...
  File "/usr/lib/python3.4/site-packages/icalendar/cal.py", line 350, in from_ical
    vals = factory(factory.from_ical(vals))
  File "/usr/lib/python3.4/site-packages/icalendar/prop.py", line 673, in from_ical
    raise ValueError('Error in recurrence rule: %s' % ical)
ValueError: Error in recurrence rule: FREQ=YEARLY;BYMONTH=11;BYDAY=1SU;
```
I don't know if this complies with the standards to have a trailing semicolon on the RRULE records or if you've ran into this before, but here's a patch anyway.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ Changelog
 - Documentation fixes.
   [t-8ch, thet]
 
+- Made RRULE tolerant of trailing semicolons.
+  [sleeper]
+
 
 3.8.4 (2014-11-01)
 ------------------
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -665,11 +665,10 @@ class vRecur(CaselessDict):
             return ical
         try:
             recur = cls()
-            if ical[-1] == ';':
-                ical = ical[0:-1]
             for pairs in ical.split(';'):
-                key, vals = pairs.split('=')
-                recur[key] = cls.parse_type(key, vals)
+                if len(pairs) >= 3:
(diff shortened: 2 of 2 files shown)


untitaker:
I've got no problem with the proposal (since it's obvious what a correct value should look like), but your fix of preprocessing the value is incredibly inefficient, I'd rather just skip empty groups (inside the for-loop)

sleeper-public:
If you would prefer it that way that's fine. My logic was that this must be a fringe case and checking the last character before we begin and replacing it if necessary would cost less cycles in the long run than checking every iteration's length. I'll update it.

untitaker:
Good point, I haven't considered that.

I think we can get the best of both worlds by simply catching the exception that occurs when trying to split an invalid pair.

sleeper-public:
Good idea, I think that's the best solution too.

untitaker:
Thanks! I manually melded your commits into one and pushed it to master.

sleeper-public:
Thanks, no worries.

PR #161, opened 2015-05-23
fix vPeriod.__repr__

spacekpe:
Without this fix following code does not work:

p = vPeriod((datetime.datetime(2016, 5, 20, 9, 0), datetime.datetime(2016, 5, 20, 10, 0)))
print(p)


untitaker:
Thanks!

PR #163, opened 2015-06-13
Improve foldline() performance

emfree:
We encountered surprising slowness when working with iCalendar objects that
have large inline attachments (e.g., base64-encoded images). This commit
improves the runtime of `foldline()` from ~7 minutes (!) to 500ms when
invoked on a 1MB string of non-ASCII data, and to 17ms on a 1MB ASCII string.


thet:
Great, thanks! I'll cut a release this week.

PR #170, opened 2015-11-28
Added display method

TomTry:
Added display method to Calendar class as suggested in the docs

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ Changelog
 
 New:
 
-- *add item here*
+- Added display method to pretty print Calendar objects.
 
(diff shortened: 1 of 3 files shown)


untitaker:
"as suggested in the docs"?

TomTry:
Here:

http://icalendar.readthedocs.org/en/latest/usage.html#overview

Is it not helpful?

stlaz:
That's not really suggested, that's a strange example that's not really helpful I think? Is there a problem with just using `to_ical()`? I don't think there is.

untitaker:
I'll close this. Thanks for the PR though!

PR #171, opened 2015-11-28
Adds __version__ attribute

TomTry:


untitaker:
A shorter version of this can be found in https://github.com/mitsuhiko/click/blob/master/setup.py

:+1: on the idea.

thet:
fixed via https://github.com/collective/icalendar/pull/172

Issue #151, opened 2014-11-19
__version__ attribute


scls19fr:
Hello,

it will be nice to add a `__version__` attribute
http://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

Kind regards

kprussing:
Just as a heads up, the "correct" way to get the version when using `setuptools` is to use the companion package `pkg_resources`

    >>> import pkg_resources
    >>> pkg_resources.get_distribution('icalendar').version
    '3.8.4'

I can't seem to find the appropriate link in the `pkg_resources` documentation, but [here is a Stack Overflow answer](http://stackoverflow.com/questions/8880661/getting-package-version-using-pkg-resources) covering it.

untitaker:
I'll close this since as mentioned by @kprussing it's no longer modern to have such an attribute.

PR #172, opened 2015-11-29
Fix issue #151

TomTry:
Added __version__ to init.py in a more succinct way.


thet:
tnx! i've added a setup.cfg with a statement to tell zest.releaser where to automatically set the version. we @ plone community use zest.releaser a lot.

PR #173, opened 2015-12-02
defined test_suite in setup.py

geier:


thet:
We need the test module/subpackage to return a unittest.TestSuite instance - according to:

> # test_suite

>     A string naming a unittest.TestCase subclass (or a package or module containing one or more of them, or a method of such a subclass), or naming a function that can be called with no arguments and returns a unittest.TestSuite. If the named suite is a module, and the module has an additional_tests() function, it is called and the results are added to the tests to be run. If the named suite is a package, any submodules and subpackages are recursively added to the overall test suite.

>     Specifying this argument enables use of the test command to run the specified test suite, e.g. via setup.py test. See the section on the test command below for more details.

thet:
can't remember, what i wanted to say with my last comment.
merged.
tnx!

Issue #174, opened 2015-12-08
[RFC] handling of broken icals


geier:
I would like to introduce some kind of handling of for broken/invalid icalendar files and would like your input on how to build this feature.

Some ideas off the top of my head:

three different (possible) modes:
 * ignore [default]: just ignore every error (as we do currently)
 * raise an exception the first time an invalid component is discovered
 * collect all invalid component properties, add a `broken` or `valid` property to each component

stlaz:
It seems that the third option is the closest to the solution they use in libical where components consist of lists of properties. If an error occurred during parsing they add an error property accordingly.

See https://github.com/libical/libical/blob/master/src/libical/icalparser.c#L562.

thet:
I your option 3 and the approach of @stlaz . I think this is fixed by: https://github.com/collective/icalendar/pull/175  - if you diagree, please repoen.

niccokunzmann:
I commented here: https://github.com/collective/icalendar/issues/315#issuecomment-1281476631 - I would do both - add a broken/valid property and not raise an exception but a [warning](https://docs.python.org/3/library/warnings.html).

geier:
I don't really care about this issue, but we probably still should implement the suggested features (warnings and some kind of validate function).

Issue #168, opened 2015-11-18
Parsing invalid icalendars fails without any warning


geier:
Example:

```python
import icalendar

event_str = """
BEGIN:VCALENDAR
BEGIN:VEVENT
DTEND:20150905T100000Z
DTSTART:20150905T090000Z
X-APPLE-RADIUS=49.91307046514149
UID:123
END:VEVENT
END:VCALENDAR
"""

calendar = icalendar.Calendar.from_ical(event_str)
print(calendar.to_ical())
```
returns
```
BEGIN:VCALENDAR
END:VCALENDAR
```

but probably should throw an exception or at least a warning instead of ignoring the invalid VEVENT silently. Also see https://github.com/geier/khal/issues/294 .

stlaz:
I do not think this should be referenced to that specific issue. The issue here (at least as described in the topic) is much more dire. Let me give you a nice example of where the icalendar library fails miserably without throwing an exception or giving any means to detect an error.

Suppose you have the following file (note the EXDATE date):
```
BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART;TZID=Europe/Vienna:20120327T100000
DTEND;TZID=Europe/Vienna:20120327T180000
RRULE:FREQ=WEEKLY;UNTIL=20120703T080000Z;BYDAY=TU
EXDATE;TZID=Europe/Vienna:20121329T100000
DTSTAMP:20130716T120638Z
END:VEVENT
END:VCALENDAR
```

Now, do the following in a prompt:
```python
import icalendar
cal = icalendar.Calendar.from_ical(open('icaltest', 'r').read())

print cal.to_ical()
BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART;TZID=Europe/Vienna:20120327T100000
DTEND;TZID=Europe/Vienna:20120327T180000
DTSTAMP:20130716T120638Z
EXDATE;TZID=Europe/Vienna:201213
RRULE:FREQ=WEEKLY;UNTIL=20120703T080000Z;BYDAY=TU
END:VEVENT
END:VCALENDAR

```
As you can see, not only from_ical does not fail, after calling to_ical, we get another (different) invalid iCalendar string.

geier:
While your example is not pretty, at least the event is still there...
[...]

PR #175, opened 2015-12-18
Fixes bug when date-time is recognized as time

stlaz:
Date-time was recognized incorrectly as a date or time. This resulted
in wrong representation of some iCalendar strings.

Also adds "errors" list in Component for saving error strings from parsing.

https://github.com/collective/icalendar/issues/174
https://github.com/collective/icalendar/issues/168
=============================================================
This is my first fork and pull request ever, hopefully I did it right.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ New:
 
 Fixes:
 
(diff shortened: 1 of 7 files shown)


thet:
tnx for the contribution!

Issue #141, opened 2014-06-13
boolean behavior of Component


untitaker:
bool(component) is false if the component doesn't have properties of its own, regardless of whether it has subcomponents or not.

untitaker:
Fixed by #179

PR #179, opened 2016-01-21
IndexError fixes + bool(Component) to return True

stlaz:
Hi,

Found an IndexError in Component.from_ical() at https://github.com/collective/icalendar/blob/master/src/icalendar/cal.py#L350 so I replaced it with ValueError instead.

Also added \_\_bool\_\_() in Component so the instances of the class return True on bool() check (https://github.com/collective/icalendar/issues/141). Without it, the check at https://github.com/collective/icalendar/blob/master/src/icalendar/cal.py#L317 might always fail and raise Exception even for Components with ignore_exceptions == True.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ New:
 
 Fixes:
 
+- Fixed possible IndexError exception during parsing of an ical string.
+  [stlaz]
+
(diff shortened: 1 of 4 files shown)


thet:
Tnx for sorry for the delay!

stlaz:
Thank you for your time :)

Issue #178, opened 2016-01-21
Invalid/unknown component name being substituted during parsing


stlaz:
Hi,

I encountered a strange behavior in a case where there are invalid component names in an iCalendar string.

```python
>>> # the following probably shouldn't pass
>>> vcal = icalendar.Calendar.from_ical('BEGIN:THISWRONG\nSTRANGEPROP:12345\nEND:THISWRONG')
>>> # but it did
>>> vcal.to_ical()
'BEGIN:VCALENDAR\r\nSTRANGEPROP:12345\r\nEND:VCALENDAR\r\n'
```

As you can see, the parser accepted the component "THISWRONG" without any warning/error. On top of that, it represents it as a VCALENDAR component instead. This does not seem to be a correct behavior at all. It is caused by:

https://github.com/collective/icalendar/blob/master/src/icalendar/cal.py#L328

Also notice the property "STRANGEPROP" from the example. The case here is similar as "STRANGEPROP" is definitely not defined in iCalendar RFC. At least here it stays the same "STRANGEPROP" property, though.

I understand that this behavior might be there for the sake of extensibility, but at least in the first case it is not handled correctly at all. Perhaps a pluggable interface would help here but until then at least the first case should be fixed.

Opinions/comments are welcome :)

thet:
The first case, where THISWRONG is substituded with VCALENDAR shouldn't happen, I agree. The second one in OK for me. icalendar should be able to parse properties like: ``X-LIC-LOCATION:America/New_York``, which are also non-standard.

stlaz:
You're right, the second case is probably OK as everyone should be able to catch the invalid property later if that is required.

I might see what I can do about the other issue.

thet:
Fixed here: https://github.com/collective/icalendar/pull/180

PR #180, opened 2016-01-22
Fixes unknown component represented as known one

stlaz:
from_ical() caused unknown an component to be represented as a known
one, losing the name information about the original component

https://github.com/collective/icalendar/issues/178

===================
The fix was rather trivial in the end

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ New:
 
 Fixes:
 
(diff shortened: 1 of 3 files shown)


thet:
Fixed via: https://github.com/collective/icalendar/commit/d9264e90c0d85bb63d6c27389eeab497e4d0973f
Tnx!

Issue #176, opened 2015-12-23
Typo and discrepancies with RFC ?


fgatbdx:
I am not python expert neither a developer but for a small personnal project I looked at icalendar project. I am not 100% sure if my findings are relevant. My comments are related to ical.ics:
- in class Event(Component), the class variable singletons shows twice the property 'DTSTART'. Should not be an issue.
- in class FreeBusy(Component), the class variable singletons shows a property 'DURATION', as far as I can see in the RFC, this property is not applicable to VFREEBUSY component?
- in classes, TimezoneStandard(Component) and TimezoneDaylight(Component), the property 'EXDATE' is not specified, while it seems to be specified in RFC?
- in class Alarm(Component), the property 'DESCRIPTION' is missing, while it seems to be specified in RFC?
- In various classes, One could see the 'EXRULE' property which seems to be outdated in the RFC. Maybe it is for backward compatibility.

stlaz:
Could you see the patch at https://github.com/stlaz/icalendar/commit/56fb1f12e3435833920e9fab3f37384962fdff3f and check, if it would make things better? No developer experience is really needed :)

geier:
I do agree with all these changes, but as this is backwards incompatible I would defer this patch until the next major release.

fgatbdx:
Sounds good !

stlaz:
That's probably a good idea. I shall fix the typo in the patch as well. What is the ETA of the next major release?

thet:
For a new major release we need some better sorted out roadmap. Currently, the README says only:
```
Roadmap

    4.0: API refactoring
``` 
... I've added that, but that's a bit sparse.
Let's start a discussion about a roadmap for a 4.0 release: https://github.com/collective/icalendar/issues/186

thet:
Fixed with: https://github.com/collective/icalendar/pull/183
IMO PR #183 is a bugfix and doesn't really break the programmatic API of icalendar. I'll include this for the next release.

fgatbdx:
As far as I see all comments from initial post have been taken into account.
Many thanks !

Issue #28, opened 2012-02-09
Strict mode for reading/writing ical


hukka:
Currently icalendar happily reads and writes data without all the fields required by the RFC. There should be a strict mode available, that does error checking.

geier:
yes, [RFC 5545](https://tools.ietf.org/html/rfc5545)

PR #183, opened 2016-02-11
Updated components to better conform to RFC 5545

stlaz:
https://github.com/collective/icalendar/issues/176

======
Should we ever get the strict mode described in https://github.com/collective/icalendar/issues/28, the properties description here needs to be correct. RFC5545 is not so easily readable so some of the properties might still be missing even in this patch. I should hope not, though.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Changelog
 New:
 
 - *add item here*
+- Updated components description to better comply with RFC 5545.
+  [stlaz]
 
 Fixes:
 
--- a/src/icalendar/cal.py
+++ b/src/icalendar/cal.py
@@ -432,7 +432,7 @@ class Component(CaselessDict):
 
 
 #######################################
-# components defined in RFC 2445
+# components defined in RFC 5545
 
 class Event(Component):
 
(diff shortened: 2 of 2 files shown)


thet:
Looks good to me!

mauritsvanrees:
Coredev 5.0 tests are broken after this: http://jenkins.plone.org/job/plone-5.0-python-2.7/4630/

There are two failures in plone.app.event, for example this one:

```
menu item "RRULE:FREQ=DAILY;COUNT=3;INTERVAL=1" missing or out of order

  File "/usr/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/home/jenkins/workspace/plone-5.0-python-2.7/src/plone.app.event/plone/app/event/tests/test_icalendar.py", line 84, in test_event_ical
    'END:VCALENDAR')
  File "/home/jenkins/workspace/plone-5.0-python-2.7/src/plone.app.event/plone/app/event/tests/test_icalendar.py", line 37, in checkOrder
    'menu item "%s" missing or out of order' % item
  File "/usr/lib/python2.7/unittest/case.py", line 424, in assertTrue
    raise self.failureException(msg)
```

Can you have a look?

mauritsvanrees:
BTW, coredev 5.1 is also broken, in the same tests.

thet:
Tnx for the pointer! that was just different ordering of the output, which we checked. Fixed in both 5.0 and 5.1

PR #185, opened 2016-02-14
Reintroduce cal.Component.is_broken

### LINKED ISSUE #184: incompatible changes in 3.9.2
With 70a7b5a16748afbf0d48ca180c2b7613fdd7e7d0 we introduced some  backwards incompatible changes:

 * `Component.is_broken` got replaced with `Component.errors`
 * events with a `RDATE;VALUE=PERIOD:19970101T180000Z/19970102T070000Z,19970109T180000Z/PT5H30M` component still had an `RDATE` with `VALUE=PERIOD` param before, now they are `RDATE:None`

While I do agree with both changes, I think they should have been deferred to the 4.0.0 release. Because we don't have `VALUE=PERIOD` anyway I think we can leave this one as it is, but I believe we should bring back `Component.is_broken` for the 3.9.3 release.

geier:
cal.Component.is_broken got removed with 3.9.2 but should not have been
removed before 4.0.0

This should (partly) fix #184

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@ New:
 Fixes:
 
 - Fix testsuite for use with ``dateutil>=2.5``. Refs #195.
+- Reintroduce cal.Component.is_broken that was removed with 3.9.2 [geier]
 
 
 3.9.2 (2016-02-05)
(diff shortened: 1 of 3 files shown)


untitaker:
Could be a property that checks bool value of `self.errors`.

stlaz:
I would rather like to see it a property dependant on self.errors, too.
You could do something like:
~~~python
class Component:
    @property
    def is_broken(self):
        return bool(self.errors)
~~~

geier:
changed as suggested

thet:
LGTM

PR #191, opened 2016-03-10
Introduce Hypothesis for testing

untitaker:
This reveals a failure: The character `\u2028` in a value is silently discarded. A quick
look at the iCalendar RFC didn't make it obvious that this character wouldn't
be allowed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,8 +15,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
-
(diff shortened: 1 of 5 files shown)


untitaker:
I'd like to get a resolution on this. Right now icalendar's behavior is not acceptable, but it's not clear what the behavior should look like.

untitaker:
I've read the RFC again and it appears that the unicode line separator must not be treated as one. I fixed that bug, so this PR is now ready for review.

Issue #187, opened 2016-02-17
Property values ignoring property parameters


stlaz:
It seems that the values of a property don't respect the property's parameters.

Example:
~~~
BEGIN:VEVENT
DTSTAMP:20150217T095800
EXDATE;VALUE=DATE:20150217T095800
END:VEVENT
~~~

This is not a valid RFC5545 string (see section [3.3](http://tools.ietf.org/html/rfc5545#section-3.3)), yet it is accepted.

This would probably only be the case for vDDDTypes and vDDDLists classes. I see two possible ways to perform the check:
1. Have a check(param) method for the properties where needed. Should be called in the component's from_ical() after the property was parsed. In case of an error, the original value is stored in the parsed dictionary.
2. Check values against parameters during parsing - in the property's from_ical() call. The property's parameter would have to be passed to its from_ical() function. It could be easier for the parser to decide which type should be its output.  In case of an error, the original value would not be stored but it would appear in the component's errors list.

Perhaps 1. is the better way. Solving this should help properly representing RDATE with PERIOD values as discussed in https://github.com/collective/icalendar/issues/184.

stlaz:
Actually, I changed - 2 (with some modifications) is probably the better solution. It seems that all properties have default data types assigned in the TypesFactory class. Except for properties that may have date-time type values for some reason (probably because these are in most cases allowed to have their value changed from datetime to date or some other value type).

As a result, it seems that the change would require some more thought. My idea is to make TypesFactory.types_map values tuples instead of strings as it is now. The index zero member of the tuple would be the default value type, other members of the tuple would be other possible value types that are currently supported. Then, TypesFactory.for_property() could then take the VALUE parameter that was specified at the property to decide which of value types should be used and whether it is a valid value type.

PR #192, opened 2016-04-13
Strong types + no longer ignore the VALUE parameter

stlaz:
Hi,

In this maybe a bit controversial patch, I introduce strong types to the python-icalendar library. Strong types are required in the RFC 5545 as discussed in the issue #187.

This implementation allows the parser not to ignore the VALUE parameter of properties anymore. As a positive side effect of that, it also allows the VALUE parameter to be recognized even for unknown properties thus representing their content properly.

Contrary to my original thought, I am keeping vDDDTypes as default type for vDDDList should none other be set for it. Therefore, this patch makes use of #189 pull request and should it be accepted the other pull request may be closed.

Also, even though the tests pass, there might be bugs so I beg for a careful review.


stlaz:
For some reason, the tests fail on https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_timezoned.py#L138, which I believe is not my doing.

geier:
I'm unsure if this should be the only behaviour or if we should introduce a lax- or strict- mode parsing. As this is breaks backwards compatibility (at least in the current form or with strict-mode parsing as the default) we should probably discuss if we want to break compatibility in master or create a new branch for that.

I might to some actual code review later.

stlaz:
Well, it might be difficult to separate lazy and strict modes just here as you would get different behaviors for unknown properties based on strictness of parsing the input.

However, it might be helpful to further improve error handling so that when exception is raised during parsing of a property (and ignore_exceptions is true) this property would still be assigned the original value in the representation although as a text so that the user knows there was an error (from the component .errors) and is able to handle it appropriately. 

Also, thanks for the review in advance :)

geier:
I haven't got anything of meaning to add to this, looks good.

So instead, I'll start bikeshedding on line breaks:
I find lines like those rather ugly:

```
vals = factory(factory.from_ical(vals,
                                 params['TZID'])
```
and personally prefer:
```
vals = factory(factory.from_ical(
     vals,  params['TZID'])
```

or
```
vals = factory(
    factory.from_ical(vals, params['TZID']))

stlaz:
Thanks, that's actually pretty valid point, I had trouble tetrissing it in. I haven't had much time recently to fix it but will hopefully get to it soon.

stlaz:
http://docs.plone.org/develop/coredev/docs/contributors_agreement_explained.html states that the agreement does not cover github/collective/ code.
Also, the Changelog verifier seems broken.

thet:
@stlaz @geier @untitaker sorry guys for not being part of the discussion. I can't really help out in the moment. I trust your mutual reviews and try to follow what happens.

thet:
Something else: @stlaz I added you as a collaborator to this repository with read/clone/push rights. If you dont mind. That way, you're able to merge others pull requests.

stlaz:
Thank you. It's quite a responsibility and I'll still keep the non-trivial pushes to those of you who maintain the repo but may merge the simpler things when necessary :)

gforcada:
@stlaz both checks (changelog entry and contributors agreement) should be fixed. If you make a change on the commits we will see it.

Thanks for noticing and reporting it!

thet:
@stlaz please rebase your PR with master.

does this introduce a backwards incompatibility, so that this should go into a 4.0 release? if not, i'd go for a 3.11 release.

stlaz:
Rebased.
There are some changes in the TypesFactory class that might be considered breaking bacward compatibility maybe.
Also, some strings with wrong VALUE type at a property will not be accepted now even though they would have been before these changes (note that the property's VALUE parameter would have been completely ignored before). This may require even better error handling that would not remove the original value of a property on parse fail but rather force it as a vText/string.

stlaz:
I added all the changes to https://github.com/collective/icalendar/pull/196 which is now rebased on the branch which is designated for 4.0 development.

PR #196, opened 2016-08-12
Error handling

stlaz:
Hello,
Long time no patch from me so I decided it might be good to finally improve error handling. I built the error handling improvements on top of https://github.com/collective/icalendar/pull/192 as I believe this patch may finally get it pushed.


stlaz:
I will leave running your Jenkins jobs on you if you don't mind.

gforcada:
@stlaz I just did (starting the jenkins jobs), you can find a guide on how to do that yourself next time [here](http://jenkinsploneorg.readthedocs.io/en/latest/run-pull-request-jobs.html)

stlaz:
It would seem those tests failed if I see the right jobs.
http://jenkins.plone.org/job/pull-request-5.0/1349/
http://jenkins.plone.org/job/pull-request-5.1/536/
Do you know what causes the failure? The tests in python-icalendar, the only repository I really care about, pass. If those tests that do not pass are valid could you then add them to this repository instead?

stlaz:
@gforcada I believe the Internet revealed the string that should have made the test fail which is the one in test_portal_ical on [Plone:tests.test_icalendar/](http://pydoc.net/Python/plone.app.event/2.0a13/plone.app.event.tests.test_icalendar/). However, it does not seem to fail for me.
@geier or @thet, do you guys think you could look into it and possibly (N)ACK these changes for pushing? Thanks :+1:

src/icalendar/prop.py
```diff
@@ -258,12 +267,16 @@ def to_ical(self):
         dts_ical = (dt.to_ical() for dt in self.dts)
         return b",".join(dts_ical)
 
-    @staticmethod
-    def from_ical(ical, timezone=None):
+    @classmethod
+    def from_ical(cls, ical, unit_type=None, timezone=None):
```
geier: I think we should keep the old order of arguments, this breaks backwards compatibility otherwise.

stlaz: You're right, that wasn't clever at all, don't know why I did that. Should be fixed now.

stlaz:
@gforcada I fixed the issue pointed out by @geier in a comment and decided to go ahead and run a Jenkins job according to the guide. Allowed the application to read whatever it wanted to read and I suddenly get "stlaz is missing the Overall/Read permission" on http://jenkins.plone.org/. This seems a bit funny to me because I have more read permissions as an anonymous user (when I'm not logged in).

gforcada:
@stlaz probably that's related to you not being in github.com/plone organization I think... I will run the jobs for you

gforcada:
http://jenkins.plone.org/job/pull-request-5.0/1410/
http://jenkins.plone.org/job/pull-request-5.1/602/

stlaz:
Still not sure why the IcalendarTestExportDX tests are failing, I was thinking they would be caused by the backward compatibility issues which I was hoping I fixed.
I shall look more into the decoded() method, I think possibly there might be bug in it not checking the VALUE parameter. However, if you look into the exception, you see that it's actually correct, '20130404' is indeed wrong datetime format.

geier:
I've created https://github.com/geier/icalendar/tree/fix_failures to help pinpoint those errors.

src/icalendar/prop.py
```diff
@@ -402,6 +429,10 @@ def from_ical(ical, timezone=None):
                 tzinfo = _timezone_cache.get(timezone, None)
 
         try:
+            if ical[8] != 'T':
```
geier: > check for length here

ignore that

stlaz: I will swap this check with the length check to make it a bit cleaner.

stlaz:
Thanks @geier, that helped! I was able to find the issues with the decoded() method. Still not able to reproduce the  IcalendarTestExportDX issue.
Btw: There's an error in those tests you uploaded. The error is on lines 63, 64 of _plone.app.event.icaltest.ics_ and on lines 61, 62 of _src/icalendar/tests/plone.app.event.icaltest_2.ics_. The 'DTSTART' and 'DTEND' property type defaults to 'DATE-TIME' in RFC 5545 (not sure if this is a change to the previous version) therefore 'VALUE=DATE' has to be defined.

thet:
@gforcada @stlaz I think people contributing to icalendar shouldn't deal with Plone's CI suite.
icalendar is used quite a lot out of the Plone context. Passing travis tests is all what's required.

@gforcada what is needed to remove the Plone CI integration?

stlaz:
@thet: This time it helped to reveal a bug so I was happy we had it there :) But you, Plone, can always run those tests yourself and open an Issue if something is broken by my patches, mark me as an Assignee and I shall have a look at it. That seems a bit more reasonable than having CI tests that the community can't run anyway.

geier:
@stlaz sorry, somehow I didn't see your replies, I know these the test data isn't RFC5545 compliant, but it's what the plone people use, so I'm sure they wouldn't appreciate it if we broke this in a non-major release. 

And yes, this is somewhat inconsistent with the behavior of icalendar in other places, like only accepting timeshifts for VTIMEZONE components <= 24h. I'm actually in favor of some kind of `strict` or `lenient` mode for icalendar.

stlaz:
@geier If I recall correctly, originally this patch enabled lazier approach to parsing iCalendar strings, leaving strange values at the given property as strings, while also keeping the property parameters. This should allow the user of the parser to evaluate the value according to their needs instead of parsing it in a wrong way and possibly dropping some of the parameters on the way.
That being said, wouldn't it be time to get the major version features prepared so that new major version could be released? :)

geier:
> That being said, wouldn't it be time to get the major version features prepared so that new major version could be released? :)

I fully agree, but at the moment I do not have a lot of time for this.

src/icalendar/prop.py
```diff
@@ -238,13 +238,22 @@ def from_ical(cls, ical):
...
             dt_list = [dt_list]
         vDDD = []
         tzid = None
+        if dt_list:
+            # we have some values, all should have the same type
+            ltype = type(dt_list[0])
         for dt in dt_list:
-            dt = vDDDTypes(dt)
+            # raise ValueError if type of the input values differs
+            if not isinstance(dt, ltype):
```
geier, line 253: Wouldn't this test fail if the first element in the list is a date and the later ones are datetime? I believe we will need to actually do a `if type(dt) != type(ltype)` here.

geier:
I have used khal's testsuite as an additional test for this PR and while I have found an issue, minimal example as follows:

```python
import icalendar
import datetime as dt
event = icalendar.Event()                                                                                                                                                        
event.add('DTSTART', dt.date.today())                                                                                                                                            
event.to_ical()
```
leads to an AttributeError:
```
  File "/usr/home/cg/workspace/icalendar/src/icalendar/cal.py", line 450, in to_ical
  File "/usr/home/cg/workspace/icalendar/src/icalendar/cal.py", line 439, in content_lines
  File "/usr/home/cg/workspace/icalendar/src/icalendar/cal.py", line 432, in content_line
  File "/usr/home/cg/workspace/icalendar/src/icalendar/parser.py", line 304, in from_parts
    values = values.to_ical()
  File "/usr/home/cg/workspace/icalendar/src/icalendar/prop.py", line 404, in to_ical
    tzid = tzid_from_dt(dt)
  File "/usr/home/cg/workspace/icalendar/src/icalendar/parser.py", line 52, in tzid_from_dt
    if hasattr(dt.tzinfo, 'zone'):
AttributeError: 'datetime.date' object has no attribute 'tzinfo'
```

under

geier:
Documentation issue: accessing a DURATION's timedelta changed from `event['DURATION'].dt` to `event['DURATION'].td`.

src/icalendar/prop.py
stlaz: Hm, you're right, it would seem that we need to be even more strict here. Thanks for pointing that out with the `date`/`datetime` example, I did not realize.

stlaz:
The documentation issue did not appear in this PR though, did it?

geier:
I believe it does, the following snippet works on master but not with this PR:

```python
import icalendar
import datetime as dt
event = icalendar.Event()
event.add('DURATION', dt.timedelta(hours=2))
print(event['DURATION'].dt)
```

stlaz:
Hello,
Sorry I did not update the PR recently, I was on a vacation away from my laptop. I have some work to do this week but i hope I'll get to this by the end of it.

I believe the `dt` -> `td` doc issue is caused by the `duration` property actually correctly resolving to the `vDuration` type so it's indeed just a doc issue.

I need to look more closely on the first issue though. It's apparent that `datetime.date` object would not have `tzinfo` attribute, I'll just need to see how much we should care :)

stlaz:
I am sorry I did not get enough time for this for the past few months, a lot happened. I should hope to get to this real soon, hopefully.

A note to the failure with the simple example `event.add('DTSTART', dt.date.today())`. This will create an invalid event since `DTSTART` expects `DATE-TIME` value by default. This should be handled more properly during `to_ical` I guess.

PR #201, opened 2016-11-01
Handle UnicodeEncodeError in Error messages. Depending on the input, …

digsim:
…I might be that instead of collection the error message, python throws a new UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 96: ordinal not in range(128). So we use the repr() built-in function to prevent this from happening.

I got this error when having something like this in my ics file. 
X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS="Roadstar 16\\n12764 Happyville\\nDenmark";X-APPLE-MAPKIT-HANDLE=CAESARoSCWYTYFhHQBEGfw4hQCIBDQoHRGVubWFyaxJES0hhcHB5dmlsbGUqSGFwcHl2aWxsZTIHSGFwcHl2aWxsZToEMTI3NjRCDQpSb2Fkc3RhcloCMTZiUm9hZHN0YXIgMTYBEU1vcmRvcgENCk1vcmRvcioSUm9hZHN0YXIgMTYyUm9hZHN0YXIgMTYxMjc2NCBIYXBweXZpbGxlMgdEZW5tYXJrOThA=;X-APPLE-RADIUS=49.91305
 866584698;X-APPLE-REFERENCEFRAME=1;X-TITLE=:geo:52.382762,7.528319

This kind of entries are created by Apple's Calendar application and encode maps information.
The problem occurs when parsing the 
X-APPLE-MAPKIT-HANDLE=CAESARoSCWYTYFhHQBEGfw4hQCIBDQoHRGVubWFyaxJES0hhcHB5dmlsbGUqSGFwcHl2aWxsZTIHSGFwcHl2aWxsZToEMTI3NjRCDQpSb2Fkc3RhcloCMTZiUm9hZHN0YXIgMTYBEU1vcmRvcgENCk1vcmRvcioSUm9hZHN0YXIgMTYyUm9hZHN0YXIgMTYxMjc2NCBIYXBweXZpbGxlMgdEZW5tYXJrOThA=
part.

When calling key, val = q_split(param, '=') this raises a ValueError: too many values to unpack (expected 2) because of the trailing '=' sign of Base64 encoded Values. Maybe this case can be handled better by the code, yet the fix for the error message in this pull-request would still be nice to have if something else causes an error.

The patch should work with python 2 & 3.


digsim:
Will first have a look at the test error and re-submit another pull-request.

PR #202, opened 2016-11-01
Handle UnicodeEncodeError

digsim:
Apple's calendar application seems to write location stuff into the X-APPLE-STRUCTURED-LOCATION field. Among them, some base64 encoded string ending with a '=' sign. This confuses parser.py when doing `key, val = q_split(param, '=')` and results in a `ValueError: too many values to unpack (expected 2)`. 

With Python2, when the resulting ValueError is converted into a string, a new error is thrown `UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 96: ordinal not in range(128)`. The `repr()` functions works around this issue by creating a representable string of the ValueError.

Since repr(e) rather prints the string `'ValueError("Content line cou......")'` instead of just `'Content line cou....'`, I had to adapt the `test_issue_104__ignore_exceptions`. If we don't desire this behaviour we need Python2 and 3 differentiation in cal.py like:

```
if sys.version_info >= (3, 0):
    component.errors.append((None, str(e)))
else:
    component.errors.append((None, repr(e.message)))
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,22 @@
 Changelog
 =========
 
+3.11.1 (unreleased)
+-------------------
+
+Breaking changes:
+
+- *add item here*
+
+New features:
+
+- *add item here*
+
+Bug fixes:
+
+- *add item here*
+
+
 3.11 (2016-11-18)
 -----------------
 
--- a/src/icalendar/__init__.py
+++ b/src/icalendar/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-__version__ = '3.11'
+__version__ = '3.11.1.dev0'
 
 from icalendar.cal import (
     Calendar,


geier:
Great that you have been looking at this error, this [came up in khal recently](https://github.com/pimutils/khal/issues/467). I'll have a look asap.

digsim:
The patch doesn't solve the https://github.com/pimutils/khal/issues/467 error itself :-/ , but at least icalendar stops crashing with Python2.7.  The X-APPLE-STRUCTURED-LOCATION  information is still dropped. Solving https://github.com/pimutils/khal/issues/467 would require some work in the `q_split` function. Maybe something like this could solve the problem.
```
def q_split(st, sep=','):
    """Splits a string on char, taking double (q)uotes into considderation.
    """
    result = []
    cursor = 0
    length = len(st)
    inquote = 0
    firstsep = True
    for i in range(length):
        ch = st[i]
        if ch == '"':
            inquote = not inquote
        if not inquote and ch == sep and firstsep:
            firstsep = False
            result.append(st[cursor:i])
            cursor = i + 1
        if i + 1 == length:
            result.append(st[cursor:])
    return result
```

stlaz:
Please don't use repr() of the error, that's just weird.

src/icalendar/cal.py
```diff
@@ -333,7 +333,7 @@ def from_ical(cls, st, multiple=False):
                 component = stack[-1] if stack else None
                 if not component or not component.ignore_exceptions:
                     raise
-                component.errors.append((None, str(e)))
+                component.errors.append((None, repr(e)))
```
stlaz: This is wrong. You can add e.message instead. You'll need to modify it at https://github.com/collective/icalendar/blob/master/src/icalendar/cal.py#L385 as well.

src/icalendar/tests/apple_xlocation_test.py
```diff
@@ -0,0 +1,14 @@
...
+import icalendar
+import os
+import pytz
+
+class TestEncoding(unittest.TestCase):
+
+    def test_apple_xlocation(self):
+        directory = os.path.dirname(__file__)
+        data = open(os.path.join(directory, 'x_location.ics'), 'rb').read()
+        cal = icalendar.Calendar.from_ical(data)
```
geier: is there some assert missing here? If not, please clarify in a comment.

digsim: No Assert necessary, without the repr() the test will simply fail. Thus if the test does not fail, then we are happy

src/icalendar/cal.py
digsim: AFAIK `e.message` is only available in Python2 but not in Python3

```
$ $ python
Python 2.7.10 (default, Oct 23 2015, 19:19:21)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.5)] on darwin
>>> a = ValueError("Hi")
>>> a
ValueError('Hi',)
>>> a.message
'Hi'
>>> ^D

$ $ python3
Python 3.5.2 (default, Jun 29 2016, 13:43:58)
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
>>> a = ValueError("Hi")
>>> a
ValueError('Hi',)
>>> a.message
AttributeError: 'ValueError' object has no attribute 'message'
>>>
```

stlaz: Right, sorry, here I go suggesting Python 3 compatibility with stuff getting deprecated since Python 2.6.
The obvious solution then is to use six:

``` python
import six

if six.PY3:
    unicode = str
...
component.errors.append((None, unicode(e)))
```

digsim: nice, didn't know about six. Will change the code like you suggest

stlaz:
Please use force pushes to synchronize pull requests: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

digsim:
There were no changes in upstream, my fork is up-to-date. AFAIK there is nothing to rebase.
Am I missing something?

stlaz:
Yes, you are. I think your changes should be in 1 commit, maybe 2 commits at most. I see 8 commits where you revert some of the changes in some of your previous commits in this single PR. That's not how pull requests work.

digsim:
Done. All changes in one commit

src/icalendar/cal.py
```diff
@@ -333,6 +334,8 @@ def from_ical(cls, st, multiple=False):
                 component = stack[-1] if stack else None
                 if not component or not component.ignore_exceptions:
                     raise
+                if six.PY3:
+                    unicode = str
```
stlaz: This should be at the beginning of the file.

src/icalendar/tests/apple_xlocation_test.py
```diff
@@ -0,0 +1,18 @@
...
+class TestEncoding(unittest.TestCase):
+
+    def test_apple_xlocation(self):
+        """
+        This test doesn't need any assert. If the tests runs wihtout throwing
+        any errors, we're happy
+        """
+        directory = os.path.dirname(__file__)
+        data = open(os.path.join(directory, 'x_location.ics'), 'rb').read()
+        cal = icalendar.Calendar.from_ical(data)
```
stlaz: Add newline to the end of the file.

stlaz: Also, check whether expected == got, that's where you need the assert.

src/icalendar/cal.py
```diff
@@ -17,9 +17,13 @@
 from icalendar.timezone_cache import _timezone_cache
 
 import pytz
+import six
 import dateutil.rrule
 from pytz.tzinfo import DstTzInfo
 
+if six.PY3:
+    unicode = str
```
geier: I don't think we are using six until now, but instead we have some of our own definitions in compat.py, please use that instead. Also, I'm somewhat confused why this works now, I don't see any other mention of `unicode` in the file?

digsim: I need some time to look at your definitions in compat. Since you suggested to use six.PY3, I thought this is fine. I adapted the test (put some accentuated characters in it, to force encoding problems) and also corrected cal.py to use `unicode` (happened while refactoring)

geier: Sorry if I suggested six, probably wasn't my brightest hour.

geier: and that should be as simple as doing `from icalendar.compat import unicode_type` and than use unicode_type instead of unicode.

stlaz: @geier: I did, my apologies.

digsim: no worry, but I'm out of time today. Will look at it by tomorrow.

digsim:
Is something wrong with my changes?

src/icalendar/tests/apple_xlocation_test.py
```diff
@@ -0,0 +1,20 @@
...
+class TestEncoding(unittest.TestCase):
+
+    def test_apple_xlocation(self):
+        """
+        Test if error messages are encode properly.
+        """
+        try:
+            directory = os.path.dirname(__file__)
+            data = open(os.path.join(directory, 'x_location.ics'), 'rb').read()
+            cal = icalendar.Calendar.from_ical(data)
```
stlaz: Could you just please add the assertion that `errors` of the given component contain what you expect them to contain? Given that the binary format may get fixed, there would be nothing in `errors` (as no error would happen) and this test would need fixing otherwise it would be pointless.

PR #204, opened 2016-11-22
Handle UnicodeEncodeError in Error messages. Depending on the input, …

digsim:
It might be that instead of collecting the thrown error message, python throws a new UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 96: ordinal not in range(128). So we encode the error message to unicode before adding it to the stack of already collected error messages.

Since this took more time than projected, I had to close the old pull request an rearrange the code properly on a branch. Sorry for the delay.


geier:
@digsim Thanks for taking care of this and going the extra mile here.

I'm satisfied with the PR as is. @stlaz Could you also have another look?

digsim:
No problem. I understand the importance of code quality.

stlaz:
Hi, the code looks fine, thanks! Sorry I got to it so late, I only have little time lately.

digsim:
No problem. Took me also some time to do the final changes. Any idea why the pypy3 build is red?

geier:
Looks like the creation of the virtualenv failed, it's not related to your PR.

PR #207, opened 2016-12-30
[RFC] Allow `=` in parameter values.

### LINKED ISSUE #197: parameters with binary values including =
We currently do not support VEVENTS with binary (hex-encoded) parameter-values (ending in `==`) like this:

```
BEGIN:VEVENT
DTSTART:20130416T100000Z
DTEND:20130416T110000Z
DTSTAMP:20130416T092616Z
UID:123
SUMMARY:wichtiger termin 1
X-APPLE-STRUCTURED-LOCATION;X-BINARY=abc==:geo:48.206686,16.363235
END:VEVENT
```

I'm not really sure that's allowed, I'll have to have a another look in the RFC, but I believe we should support it anyway, because Apple seems to produce those VEVENTs, see geier/khal#467

geier:
Some parameter values (e.g., BASE64 encoded binary data often ends with one or two equal signs) may contain an equal sign (`=`). The current implementation splits key-value pairs at all equal signs, which leads to errors. Especially icalendar files generated by Apple's software often feature BASE64 encoded binary data in parameter values.

This patch introduces a new parameter `maxsplit` to icalendar.parser.q_split() which works similar as python's string.split(sep, maxsplit) which we then use to split parameter key-value pairs only at the first equal sign.

This patch fixes #197.


digsim:
Regarding the test introduced in #204 one could argue to remove this test now. Its only purpose was to pinpoint a problem, which is getting fixed by this commit making this test obsolete. However, I would rather argue that the test should stay. I would therefore suggest to change line 20 into
`self.assertEqual(len(event.errors), 0, 'Got too many errors'`
and remove lines 21 and 22. 

This second approach would then verify that:
1. icalendar is able to parse a file containing binary information (**already possible without this patch**) and
2. icalendar can do this without producing any errors (**new with this patch**).

geier:
I have modified the old test and added a new one that tests the fix introduced in #204.

geier:
Looks like I forgot to push those changes (done now). The failing test is pypy3, which we currently do not support, I have opened #206 to deal with that.

src/icalendar/tests/test_icalendar.py
```diff
@@ -247,7 +275,50 @@ def test_q_split(self):
...
+        Test if error messages are encode properly.
+        """
+        try:
+            directory = os.path.dirname(__file__)
+            data = open(os.path.join(directory, 'x_location.ics'), 'rb').read()
+            cal = icalendar.Calendar.from_ical(data)
+            for event in cal.walk('vevent'):
+                self.assertEqual(len(event.errors), 0, 'Got too many errors')
+
+        except UnicodeEncodeError as e:
```
untitaker: Why catch the error at all?

geier: Honestly: no idea, I've just re-used the old test, but sure, we can remove it.

geier:
I removed those try/excepts.

untitaker:
LGTM

digsim:
That should be ok. Back then, the except was used to let the test fail.

geier:
I'd like to get someone else's review on this. @stlaz, @thet or @regebro, if you find the time, I'd really appreciate your input.

stlaz:
Putting it on my TODO list

regebro:
"=" is a SAFECHAR and should be allowed in unquoted values, so this is indeed a bug.
Only comment I have is that passing in maxsplit=0 doesn't return the expected result (no splits). It's a sufficiently silly case that you can choose not to support it.

untitaker:
Apparently I can push commits to this PR, so I fixed this!

untitaker:
Also Jenkins is utterly broken and emails me because of that.

regebro:
Well, maxsplit=0 should reasonably not split at all. maxsplit=0 being the same as maxsplit=1 is even more confusing than it being the same as maxsplit=-1. :-)

untitaker:
`maxsplit=0` returns a list with len <= 1, `maxsplit=1` a list with len <= 2, just like `str.split` does.

untitaker:
Ah, I see it fails in cases where there's nothing between the separator.

untitaker:
@regebro so that was a bug, fixed!

src/icalendar/tests/test_icalendar.py
```diff
@@ -247,7 +275,46 @@ def test_q_split(self):
         self.assertEqual(q_split('Max,Moller,"Rasmussen, Max"'),
                          ['Max', 'Moller', '"Rasmussen, Max"'])
 
+    def test_q_split_bin(self):
+        from ..parser import q_split
+        for s in ('X-SOMETHING=ABCDE==', ',,,'):
+            for maxsplit in range(3):
```
geier: I'd suggest `range(-1, 3)` here.

stlaz:
I hope I am not missing anything but is the `q_split()` function really needed? Why not either use the Python's native `str.split()` or if we really really don't like the "." notation, map it in a function like this:
```python
def q_split(st, sep=",", max_split=0):
    st.split(sep, max_split)
```
The `q_split()` function looks like a hack from Python-prehistoric times to me.

geier:
Because it doesn't split if we are in between "s.

Quoting Stanislav Láznička (2017-01-19 18:49:27)

stlaz:
@geier Thank you, I found that out right after I wrote that comment so I removed it :) Been in work for too long today.

src/icalendar/tests/test_icalendar.py
untitaker: done

regebro:
Ah, pft, I apparently didn't see the initial "if maxsplit==0:" check. All good.

src/icalendar/tests/test_icalendar.py
```diff
@@ -247,7 +275,46 @@ def test_q_split(self):
...
+            self.assertEqual(len(event.errors), 1, 'Not the right amount of errors.')
+            error = event.errors[0][1]
+            self.assertTrue(error.startswith(u'Content line could not be parsed into parts'))
+
+    def test_apple_xlocation(self):
+        """
+        Test if we support base64 encoded binary data in parameter values.
+        """
+        directory = os.path.dirname(__file__)
+        data = open(os.path.join(directory, 'x_location.ics'), 'rb').read()
```
untitaker: Please use a context manager here to fix the resource warning.

untitaker:
This conflicts with master, otherwise LGTM

src/icalendar/tests/test_icalendar.py
geier: That's the way it's done throughout the test suite. I believe it's better to be consistent than to be right (in this case).

untitaker: It's no longer done this way, see #213

geier: oh, it isn't any more... alright, will fix

geier:
done

geier:
I can force push into the branch, and your approval doesn't get auto-revoked!?

geier:
anybody got any idea why python 2.6 now fails on travis with

    import unittest2 as unittest
    E   ImportError: No module named unittest2
?

geier:
As this group maintained, I'm not going to merge this myself, but I'd be grateful if somebody else would do it or voice their concerns.

I believe the current failure with python2.6 can safely be ignored, as it fails on master, too :grimacing:

hobarrera:
Any update on this? Looks ready, right?

untitaker:
@regebro could you take a quick look at this again?

geier:
Thanks everyone who contributed!

PR #208, opened 2017-01-19
Remove custom merge strategy for changelog

untitaker:
This causes old PRs to add the changelog entry to already released
versions. It happened to me while rebasing #207

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-CHANGES.rst merge=union


untitaker:
perhaps cc @mauritsvanrees who was the original author of this commit

mauritsvanrees:
The reason for the union merge, is to avoid this situation:

- PR 1 has a changelog entry
- PR 2 has a changelog entry
- PR 1 gets merged
- You look at PR 2 and see that there is a merge conflict. From looking at github you have no idea if this is a simple conflict in the changelog, which you could quickly fix while merging manually, or a nasty conflict elsewhere that may take far longer.  So you ignore the pull request for now, or ask the author to rebase, and everything takes longer.

That's the theory. And usually it seems to go fine.

But in practice, I have seen problems like you mention. Or the changelog entry will end up under the correct header, but without enough white space, or with problems that break the ReStructuredText rendering on PyPI. So I have also become less enthused about this.

So let's merge this and see what the effects are.
Later we can maybe do this for other packages as well.

untitaker:
Yeah, given that you can now resolve merge conflicts via GitHub's web UI I
think this may become less painful anyway.

PR #211, opened 2017-01-24
Drop requirement for setuptools package from setup.py

jdufresne:
setuptools is shipped with Python, no need to include it as a requirement. It
is bad form to include it as a requirement as pip will try to pull in the
latest version.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,8 @@ Changelog
 
 Breaking changes:
 
-- *add item here*
+- Removed ``setuptools`` as a dependency as it was only required by setup.py
+  and not by the package.
 
 New features:
 
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ with Python.
     :Homepage: http://icalendar.readthedocs.org
     :Code: http://github.com/collective/icalendar
     :Mailing list: http://github.com/collective/icalendar/issues
-    :Dependencies: `setuptools`_ and since version 3.0 we depend on `pytz`_.
+    :Dependencies: `python-dateutil`_ and `pytz`_.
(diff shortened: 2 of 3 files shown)


untitaker:
Yeah that wouldn't work anyway since we already import setuptools at the top of `setup.py`.

I'd like to hear somebody else's opinion on this though.

untitaker:
In any case this is probably worth a changelog entry.

jdufresne:
I have added a note to the changelog. Please let me know if you'd like any wording adjusted. All feedback welcome, thanks.

kevin-brown:
I'm 👍 on this change, for the reasons mentioned in the intro post.

> Yeah that wouldn't work anyway since we already import setuptools at the top of setup.py.

This applies for cases where `setup.py` is executed during installation, which is all of the cases for `icalendar`.

untitaker:
Ok, thanks!

mauritsvanrees:
I agree with the change, but I think there are a few mistakes in the reasoning here, so let me add some comments.

- `setuptools` is *not* shipped with Python. Maybe in some versions, but certainly not all.
- Just because `setuptools` is imported in the top of the file does not mean that adding it in `install_requires` is useless: as far as I know, packaging tools can actually handle this.
- But `setuptools` is usually only needed when you have a namespace package. That is not the case here, so removing setuptools is fine.

BTW, I wouldn't call this a 'breaking change'.  See this question and answer on [semver.org](http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) that says an update in dependencies does not mean you are suddenly backwards incompatible. There will be cases where it *is* better to do that, but not here. 
So I have removed that sub header and filed the changelog entry under 'Bug fixes'.

I have released version 3.11.3 with this change.  Thank you!

PR #214, opened 2017-02-15
On python2.6 install unittest2 for tox.

geier:


untitaker:
This is already specified in icalendar's `setup.py` (see also the entry `icalendar [test]` directly below). I think it would make sense to move all test dependencies into tox

geier:
even stranger that it doesn't work on travis any more.

PR #220, opened 2017-03-22
Fix handling of VTIMEZONEs with subcomponents with the same DTSTARTs

geier:
For VTIMEZONE subcomponents without a TZNAME we would construct a
TZNAME from their shared TZID and DTSTART, if their DTSTART was equal,
this would lead to the same TZNAME and with that to a broken timezone
object.

~~The new method of finding a unique TZNAME is pretty brain dead, but it's guaranteed that it is unique. Suggestions (and code) for better methods are welcome, feel free to push to this branch (if you can).~~


geier:
Test on python2.6 fail, see #214.

PR #222, opened 2017-04-20
If not strict, don't break on UTCOffsets > 24 hours.

### LINKED ISSUE #155: Offsets larger than 24h?
At https://github.com/geier/khal/issues/140 there is a user whose offset values are not accepted by icalendar. I have no idea what is going on in that iCalendar file. Anybody got any ideas?

geier:
closes #155 

Question: does this break backwards compatibility? It now accept some .ics files it didn't accept before if strict=False, which is the default.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,9 @@ Bug fixes:
 - Make every from_ical() method accept an optional `strict=False` parameter
   [geier]
 
+- Don't break on UTCOffsets > 24 hours, even though they make no sense, at least
+  one major calendar service provider uses those [geier]
+
 - *add item here*
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,9 @@ Bug fixes:
 - Fix handling of VTIMEZONEs with subcomponents with the same DTSTARTs.
   [geier]
 
+- Make every from_ical() method accept an optional `strict=False` parameter
+  [geier]
+
 - *add item here*
 
 
--- a/src/icalendar/cal.py
+++ b/src/icalendar/cal.py
@@ -317,12 +317,12 @@ class Component(CaselessDict):
         return properties
 
     @classmethod
-    def from_ical(cls, st, multiple=False):
+    def from_ical(cls, st, multiple=False, strict=False):
         """Populates the component recursively from a string.
         """
         stack = []  # a stack of components
         comps = []
-        for line in Contentlines.from_ical(st):  # raw parsing
+        for line in Contentlines.from_ical(st, strict=strict):  # raw parsing
             if not line:
                 continue
 
@@ -378,9 +378,9 @@ class Component(CaselessDict):
                                   'FREEBUSY', 'RDATE', 'EXDATE')
                 try:
                     if name in datetime_names and 'TZID' in params:
-                        vals = factory(factory.from_ical(vals, params['TZID']))
+                        vals = factory(factory.from_ical(vals, params['TZID'], strict=strict))
                     else:
-                        vals = factory(factory.from_ical(vals))
+                        vals = factory(factory.from_ical(vals, strict=strict))
                 except ValueError as e:
                     if not component.ignore_exceptions:
                         raise
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -378,7 +378,7 @@ class Contentlines(list):
         return b'\r\n'.join(line.to_ical() for line in self if line) + b'\r\n'
 
     @classmethod
-    def from_ical(cls, st):
+    def from_ical(cls, st, strict=False):
         """Parses a string into content lines.
         """
         st = to_unicode(st)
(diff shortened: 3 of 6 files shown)


untitaker:
Should icalendar attempt to fix those offsets?

untitaker:
PR LGTM by the way.

stlaz:
I wonder if we might somehow either get rid of `ignore_exceptions`, merge it with `strict` or not use `strict` at all and just do with `ignore_exceptions`.

Note that I made a proposal for the latest case in https://github.com/collective/icalendar/pull/196, specifically here: https://github.com/stlaz/icalendar/blob/b257aa5f7c3c499ecca8c56b4ff3b792504eeed5/src/icalendar/cal.py#L385 where I convert the incoming string to a `vText` property if it's invalid but we want to ignore exceptions.

geier:
@untitaker I'd say no, at least not for the moment. icalendar doesn't fix any issues with broken/invalid icalendars at the moment, so why start here?

@stlaz I believe it's a good idea to merge `ignore_exceptions` and `strict`, but as usual, it would break backwards compatibility.

Perhaps we should start preparing the next major release in a new branch, into which we could for example merge your two open PRs for a start.

stlaz:
My point is - what does "non-strict" mean for us?

Look at the changes you've done - you are passing a new argument to each `from_ical()` method but you only implement non-strict behavior for one special case - greater offsets. Why haven't you done this for the rest of the properties in `prop.py`? I believe it's because it's simply not possible. What if boolean properties get completely wrong string as a value? What about completely wrong dates?

To me, `ignore_exceptions=True` should have the same functionality as your proposed `strict=False` because that's what you want to do. You don't throw an error if the supplied value string of a property is wrong, you fall back to a different representation because you don't know what to do with what you received - and that is just ok. I believe the fallback representation should be `vText`. This has several advantages:
1. it is still possible to reconstruct the original string
2. the user is able to handle our representation of his supplied string as they want
3. the user of the library is able to find out that they are creating a non-standard iCalendar string and that they should probably change it because different softwares may not understand it

So the TLDR version:
We should not add `strict` argument to `from_ical()` methods but use current `ignore_exceptions` instead.

I agree that we should maybe start preparing for the new major release, new branch sounds like a good idea.

stlaz:
One more note, though. It might be worth to move `ignore_exceptions` in a way that we are able to handle strict/non-strict parsing directly when calling `from_ical()`, then your proposal to add `strict` as an argument would be more or less OK, except that we would set `ignore_exceptions` accordingly during parsing instead of implementing special cases.

geier:
I used `strict` because that was used in `parser.Parameters.from_ical()` and I didn't see `cal.Component.ignore_exceptions`. But I believe the current way of making icalendar ignoring exceptions by explicitly setting `Component.ignore_exceptions = True` isn't really great (or is there another way I'm not seeing?). I'd much prefer being able to use `icalendar.Calendar.from_ical(icalstr, ignore_exceptions=True)`.

stlaz:
You're right, there's no way of doing that and I too believe that should be changed. I think we could do with calling it `strict` in the call and we could do with adding this argument just to `Component.from_ical()`, right?
Then what I'd do is to replace `if ignore_exceptions` conditions with `if not strict`. This is however backward incompatible. I wonder whether we need to worry much about that in this case. Probably so but given that the way it could have been used was so strange and that its values for some components seem to be more or less randomly chosen... But then again, if we removed it we would remove the possibility to perform strict parse only for some components.

geier:
>  But then again, if we removed it we would remove the possibility to perform strict parse only for some components.

Hmm, didn't think of that.

What would your preferred way of dealing with this be then?

stlaz:
If we want to keep that use case, we should first remove `ignore_exceptions = True` from class `Event` so that it's false everywhere by default. Then we could add `strict=True` to `Component.from_ical()`, which will make strict parsing the default for every instance of `Component` subclass. When adding a component to the component stack in `Component.from_ical()`, we already have such an instance. We can therefore do `if strict=False: component.ignore_exceptions=True`, keeping the original use-case but also adding the possibility to do non-strict parsing as a whole.

The behavior would change for the `Event` component, of course, I hope we can do at least that.

If we can agree on this and on the development branch, I can rebase my pending pull requests on the new branch, adding this as well.

geier:
I'll close this one in favor of the much simpler #235.

@stlaz please create a dev branch and rebase your PRs against it.

While we are at it, I would really like to get rid of all the bytes in icalendar and just deal with unicode/str, do you think we could make to_ical() and from_ical() return and accept unicode or would this be to much of a break and should I create new methods and add deprecation warnings to the old ones?

stlaz:
According to https://tools.ietf.org/html/rfc5545#section-3.1.4 it makes sense to have the string returned by python-icalendar be of a `unicode` (python2) or `string` (python3) type. Note that this is a backward-incompatible change (people are probably `.decode()`-ing our stuff) and should thus be included in a the major version we'll start preparing.
I guess we could make `from_ical()` accept only unicode/str with the backward incompatibility also in mind.
Note that bytes should probably stay as the correct inner object representation at places like `vBinary`, for example.

I will create the dev branch some time this week.

Issue #237, opened 2017-08-09
UnicodeEncodeError if TZID parameter is non-ascii


clivest:
```python
# -*- coding: utf-8 -*-
from icalendar import Calendar
from datetime import datetime
import pytz

ics = u"""BEGIN:VCALENDAR
METHOD:REQUEST
PRODID:Microsoft Exchange Server 2010
VERSION:2.0
BEGIN:VTIMEZONE
TZID:(UTC-03:00) Brasília
BEGIN:STANDARD
DTSTART:16010101T235959
TZOFFSETFROM:-0200
TZOFFSETTO:-0300
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=3SA;BYMONTH=2
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:16010101T235959
TZOFFSETFROM:-0300
TZOFFSETTO:-0200
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=2SA;BYMONTH=10
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
ORGANIZER:MAILTO:b@example.com
ATTENDEE:MAILTO:a@example.com
DESCRIPTION;LANGUAGE=pt-BR:something something
UID:something
SUMMARY;LANGUAGE=pt-BR:Something
DTSTART;TZID=\"(UTC-03:00) Brasília\":20170511T133000
DTEND;TZID=\"(UTC-03:00) Brasília\":20170511T140000
CLASS:PUBLIC
PRIORITY:5
DTSTAMP:20170509T131359Z
TRANSP:OPAQUE
STATUS:CONFIRMED
SEQUENCE:0
END:VEVENT
END:VCALENDAR"""

cal = Calendar.from_ical(ics)
dtstart = cal.walk(name='VEVENT')[0].decoded("DTSTART")
expected = pytz.timezone('America/Sao_Paulo').localize(datetime(2017, 5, 11, 13, 30))
assert dtstart == expected
```

Currently fails with an error due to the timezone TZID parameter having a non-ASCII character:
```
Traceback (most recent call last):
  File "/tmp/t.py", line 42, in <module>
    cal = Calendar.from_ical(tz)
  File "/usr/lib/python2.7/site-packages/icalendar/cal.py", line 347, in from_ical
    _timezone_cache[component['TZID']] = component.to_tz()
  File "/usr/lib/python2.7/site-packages/icalendar/cal.py", line 555, in to_tz
[...]

PR #238, opened 2017-08-17
Fix non-ascii TZID and TZNAME handling in python 2

clivest:
Following on from #237, I realised this problem only occurs when using python 2. I've done the inexact conversion to str only when running in python 2. This retains the full zone name in python 3 but unfortunately makes timezone handling slightly different when running in different python versions. What do you think?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Fix non-ASCII TZID and TZNAME parameter handling #237 [clivest]
 
 
 3.11.6 (2017-08-04)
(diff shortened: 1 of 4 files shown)


clivest:
Sure, thanks!

> I'm so ready for dropping python 2 support

Yeah. Shame some of us still have to live in the past

geier:
Since nobody else had a look and it's been months, I just merged this. Thanks again for your effort!

clivest:
@geier Thanks for merging! Is there plans for another release?

geier:
@clivest I have know idea how releases are done, sorry. But they are made rather regular.

thet:
@clivest @geier release done: https://pypi.python.org/pypi/icalendar/3.12

geier:
Thanks!

Quoting Johannes Raggam (2017-11-07 10:15:25)

clivest:
Thanks

PR #240, opened 2017-09-20
Update install.rst

Ekran:
change to pip install instead of not referenced setup.py

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ New features:
 
 Bug fixes:
 
+- Wiki: update install instruction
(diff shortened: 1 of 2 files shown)


untitaker:
I think this makes sense. It was intended that you clone this repository, which includes a setup.py though.

Ekran:
It seems I can not run Jenkins jobs... "Zugriff verweigert Ekran fehlt das Recht „Allgemein/Read“" (access denied, missing rights)

untitaker:
Thanks! Don't worry about Jenkins, travis is good enough.

PR #242, opened 2017-09-28
Accept Windows timezone identifiers as valid.

geier:
Currently we accept Olson timezone identifiers (tzids) as valid, even
when no matching VTIMEZONE component is present.  Windows tzids are
different from Olson tzids, but also far spread.  Some software
(Microsoft Exchange) produces .ics files using those Windows tzids with
missing VTIMEZONE components. While in general our stance is to only
support standard compliant .ics files and try to get the issuer of the
invalid .ics files to fix their software, I believe our chances here are
very slim.  On the other hand, a lot of those .ics files exist in the
wild and not supporting those Windows tzids is a source of major
inconvenience for our users.  We should there accept Windows tzids as we
do for Olson tzids.  The unicode consortium has a suggested mapping from
Windows tzids to Olson tzids [0], which we now make use of.

[0] http://www.unicode.org/cldr/charts/29/supplemental/zone_tzid.html


geier:
Alternatively, and perhaps additionally, users could provide a callback function that deals with unrecognized timezone identifiers.
Usage scenarios would be asking the end user for an alternative timezone, inserting a default timezone or raising an exception.

thet:
The reason behind supporting only Olson was not to have to parse the VTIMEZONE components and use these timezone definitions to build up own Python tzinfo objects - so mostly lazyness.
As I found out, we have VTIMEZONE parsing capability since 3.9, so the original reasoning is already overhauled.

This PR looks good to me! Is it ready to merge?

geier:
@thet: yes, this should be ready. I pushed with an updated commit message, I had a non-ascii point in there before (•).

PR #244, opened 2017-10-22
Fix ResourceWarnings in setup.py when Python warnings are enabled

jdufresne:
Always explicitly close files. Use a context manager to make this easier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,10 +13,11 @@ with open('src/icalendar/__init__.py', 'rb') as f:
 
 
 shortdesc = 'iCalendar parser/generator'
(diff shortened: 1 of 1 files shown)


untitaker:
Idk, could you change this to something like:

```
for fname in ('README.rst', 'CHANGES.rst', ...):
    with codecs.open(fname, ...):
        ...
```

jdufresne:
Made suggested update.

jdufresne:
The commit is marked as a failure by "Changelog verifier". Am I expected to add an entry for this? I don't think it really affects users when upgrading.

untitaker:
Thanks! No worries.

PR #245, opened 2017-10-22
Fix invalid escape sequences in string and bytes literals

jdufresne:
Invalid escape sequences have been deprecated in Python 3.6. See:

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases.

When warnings are enabled, this appears as:

```
DeprecationWarning: invalid escape sequence ...
```

Sequences discovered through test suite.

### DIFF
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -39,11 +39,11 @@ def unescape_char(text):
                    .replace(u'\\;', u';')\
                    .replace(u'\\\\', u'\\')
     elif isinstance(text, compat.bytes_type):
-        return text.replace(b'\N', b'\n')\
+        return text.replace(b'\\N', b'\\n')\
                    .replace(b'\r\n', b'\n')\
                    .replace(b'\n', b'\n')\
-                   .replace(b'\,', b',')\
-                   .replace(b'\;', b';')\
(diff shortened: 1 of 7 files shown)


src/icalendar/parser.py
```diff
@@ -109,8 +109,8 @@ def param_value(value):
 # Could be improved
 
 # [\w-] because of the iCalendar RFC
-# \. because of the vCard RFC
-NAME = re.compile('[\w\.-]+')
+# . because of the vCard RFC
```
untitaker, line 112: This change is invalid... `.` matches any character in regex.

jdufresne: The dot is inside square brackets, so it loses its special meaning. In general, I find fewer escape sequences easier to read. From [re docs](https://docs.python.org/3/library/re.html)

> Special characters lose their special meaning inside sets. For example, [(+*)] will match any of the literal characters '(', '+', '*', or ')'.

If you still prefer the backslash, let me know. I can add it back in.

untitaker:
You're right, sorry about that!

PR #247, opened 2017-10-22
Clean up trailing and leading white space throughout project

jdufresne:
Makes future diffs cleaner for editors that automatically clean up white space.


untitaker:
Thanks!

PR #248, opened 2017-11-07
Add pip cache support to Travis CI configuration

jdufresne:
Can slightly improve build times and reduce load on PyPI servers.

For documentation on cache support, see:

https://docs.travis-ci.com/user/caching/#pip-cache


jdufresne:
I think this change is too trivial for a changelog entry.

PR #246, opened 2017-10-22
Drop support for Python 2.6

jdufresne:
I would humbly like to suggest icalendar drop support for Python 2.6.

The last release of Python 2.6 was 2013-10-29, almost 4 years ago. It is no longer receiving security or bug fixes.

https://www.python.org/dev/peps/pep-0361/

The pip project itself has recently dropped support for 2.6. At the beginning of the year, their numbers estimated that Python 2.6 accounts for ~2% of their downloads.

pypa/pip#4343

For projects that still use Python 2.6, they can continue to pip install an older version.

I've tried my best to remove as much 2.6 specific code as I can, including the `'Programming Language :: Python :: 2.6'` trove classifier from setup.py. I've also removed Travis CI testing, which should result
in slightly faster testing and fewer wasted resources.

Changed:

* Update trove classifiers, Travis test matrix, and tox test matrix
* Use set literals
* Use dict comprehension
* Remove dependencies on ordereddict and unittest2
* Remove `u` string prefix, use more modern unicode literals instead

PR #249, opened 2017-11-09
Quote all version strings in Travis CI configuration

jdufresne:
Per the feedback from @polyzen in PR #246, in YAML, numeric bare strings are interpreted as floats, not strings. Versions are strings and should not be coerced to a float. For example, 7.10 isn't the same as 7.1.

For more details, see Travis issue:

https://github.com/travis-ci/docs-travis-ci-com/pull/1537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: python
 cache: pip
 python:
-  - 2.7
-  - 3.4
-  - 3.5
-  - 3.6
-  - pypy
-  - pypy3
+  - "2.7"
+  - "3.4"
+  - "3.5"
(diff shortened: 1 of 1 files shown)


polyzen:
I have inquired as to whether single quotes should be enforced in the upstream issue: https://github.com/travis-ci/docs-travis-ci-com/issues/1540

thet:
Looks good to me. For consistency reasons I'd enforce only only one type of quotes - single or double. Since double quotes are more flexible in  YAML, I prefer double quotes (tnx @polyzen for the info about this). Even though that discussion does not matter much in the travis file.

PR #250, opened 2017-11-19
Move tests out of install package to top level directory

jdufresne:
Cleanly separates tests from the package itself. Prevents the tests being installed to site-packages. Tests are still distributed with the source distribution by `MANIFEST.in`.


geier:
I'm +1 on this, but I remember reading some article from the plone community (which I now cannot find anymore) advising on the current layout. Instead I found [this article](http://plone-best-practices-discussion.readthedocs.io/en/latest/package_layout/), which also advises the layout you proposed, but I don't know if is "official" in any way.

geier:
PS: I believe this change does warrant a changelog entry.

thet:
There are at least two downsides with this:

- For installed packages, tests cannot be run anymore. In Plone we are running all the tests for all the critical dependencies. But icalendar might not belong to these critical dependencies - it also has no dependencies to Plone core packages.

- Tests, which also serve as documentation, are not available via omelette anymore. https://github.com/collective/collective.recipe.omelette is a buildout recipe which symlinks all dependencies into a easy-to-navigate directory.

I guess it's OK for icalendar to move out the tests from the install directory, but I'd want to clear things out with the Plone community first. Also see: https://github.com/plone/bobtemplates.plone/issues/242

jdufresne:
Thanks for reviewing.

> I believe this change does warrant a changelog entry.

Added in the latest revision.

jdufresne:
Just some background on why I created this change request.

I noticed that icalendar was one of _very few_ packages that included tests installed to my virtualenv. I prefer to keep dependencies and dead code to a minimum. Just less code surface area to consider.

I don't want other packages' tests available to any part of my production system. Should some module or test discovery system wander into icalendar's tests, I prefer they never be executed. Not including them is the best way to ensure this.

In my opinion, tests are a development tool, not a production tool. Tests can still be run either through the source distribution or a git clone.

jdufresne:
I rebased to resolve merge conflicts.

Issue #255, opened 2018-01-12
Command line interface.


jfjlaros:
I wrote a simple command line interface for use in terminal based e-mail clients like mutt. Would you be interested in a pull request?

thet:
sounds interesting. what does the CLI do?
In general, PR's are alyways welcome.

jfjlaros:
It writes a human readable summary (organiser, attendees, start and end time, description, etc.) of an event to a stream (stdout by default). I have been using a similar tool for years to read invitations in mutt.

When I found your library, it seemed like a good idea to toss out my parser and rewrite this tool.

thet:
closed via  #256

PR #256, opened 2018-01-13
Cli

jfjlaros:
Simple command line interface for use in terminal based e-mail clients, also see issue #255.

### DIFF
--- a/src/icalendar/cli.py
+++ b/src/icalendar/cli.py
@@ -82,7 +82,7 @@ def main():
         formatter_class=argparse.RawDescriptionHelpFormatter,
         description=__doc__)
     parser.add_argument(
-        '-v', action='version',
+        '-v', '--version', action='version',
         version='{} version {}'.format(parser.prog, __version__))
 
     # This seems a bit of an overkill now, but we will probably add more
--- a/src/icalendar/cli.py
+++ b/src/icalendar/cli.py
@@ -51,19 +51,6 @@ def _format_attendees(attendees):
     return _format_name(attendees)
 
 
-def _optional(event, key):
-    """Wrapper to prevent a KeyError.
-
-    :arg Event event: An event.
-    :arg str key: Attribute of the event.
-
-    :returns any: The attribute or an empty string.
-    """
-    if key in event:
-        return event[key]
-    return ''
-
-
 def view(input_handle, output_handle):
     """Make a human readable summary of an iCalendar file.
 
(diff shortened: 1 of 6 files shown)


src/icalendar/cli.py
```diff
@@ -0,0 +1,125 @@
...
+    :returns str: Human readable summary.
+    """
+    cal = Calendar.from_ical(input_handle.read())
+
+    for event in cal.walk():
+        if event.name == 'VEVENT':
+            break
+
+    output_handle.write(_template.format(
+        organiser=_format_name(event['organizer']),
```
PabloCastellano: Looks like the organizer is optional. In my case using a google calendar I got:

```Traceback (most recent call last):
  File "/home/pablo2/.virtualenvs/icalendar/bin/icalendar", line 11, in <module>
    load_entry_point('icalendar==4.0.1.dev0', 'console_scripts', 'icalendar')()
  File "/home/pablo2/.virtualenvs/icalendar/lib/python2.7/site-packages/icalendar-4.0.1.dev0-py2.7.egg/icalendar/cli.py", line 118, in main
    args.func(**{k: v for k, v in vars(args).items()
  File "/home/pablo2/.virtualenvs/icalendar/lib/python2.7/site-packages/icalendar-4.0.1.dev0-py2.7.egg/icalendar/cli.py", line 81, in view
    organiser=_format_name(event['organizer']),
  File "/home/pablo2/.virtualenvs/icalendar/lib/python2.7/site-packages/icalendar-4.0.1.dev0-py2.7.egg/icalendar/caselessdict.py", line 42, in __getitem__
    return super(CaselessDict, self).__getitem__(key.upper())
KeyError: u'ORGANIZER'
```

I would remove the `_optional` wrapper and just use `event.get('organizer', '')`

src/icalendar/cli.py
```diff
@@ -0,0 +1,125 @@
...
+        description=_optional(event, 'description')).encode('utf-8'))
+
+
+def main():
+    """Main entry point."""
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=__doc__)
+    parser.add_argument(
+        '-v', action='version',
```
PabloCastellano: I would add `--version` as well

src/icalendar/cli.py
```diff
@@ -0,0 +1,125 @@
...
+def view(input_handle, output_handle):
+    """Make a human readable summary of an iCalendar file.
+
+    :arg stream handle: Open readable handle to an iCalendar file.
+
+    :returns str: Human readable summary.
+    """
+    cal = Calendar.from_ical(input_handle.read())
+
+    for event in cal.walk():
```
PabloCastellano: Here you can use `cal.walk(name='VEVENT')` and remove the check below.

However an ical file can contain multiple events and here you are considering only the first one. Is it the expected behavior? I would show all the events instead.

src/icalendar/cli.py
jfjlaros: Fixed in a18a8e4.

src/icalendar/cli.py
jfjlaros: Added in b684257.

src/icalendar/cli.py
jfjlaros: Fixed in 70b15dc.

jfjlaros: Thank you for your useful suggestions. I incorporated all of them.

src/icalendar/cli.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def view(input_handle, output_handle):
+    """Make a human readable summary of an iCalendar file.
+
+    :arg stream handle: Open readable handle to an iCalendar file.
+
+    :returns str: Human readable summary.
+    """
+    cal = Calendar.from_ical(input_handle.read())
```
thet, line 62: I'd enclose the whole body in a ``with`` statement, so that the file handles are closed automatically at the end. Regarding the discussion here: https://stackoverflow.com/a/41495679/1337474 the following should work:

```
with input_handle as fih, output_handle as foh:
    ...
```

thet:
That's a great start for a CLI! Looking forward for other sub commands.

src/icalendar/cli.py
jfjlaros: I would rather keep this function free from side-effects, on the other hand, closing the file handles in `main()` will be messy...

thet: fair point

PR #262, opened 2018-05-11
Get version in docs without pkg_resources

davidfischer:
- Reads the version from `icalendar/__init__.py`
- Appends the path to icalendar to the Python path like how the sphinx-quickstart sample `conf.py` recommends

This should fix the docs build on Read the Docs which hasn't built in a [few years](http://readthedocs.org/projects/icalendar/builds/).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
 # icalendar documentation build configuration file
-import pkg_resources
 import datetime
 import os
+import sys
 
+base_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), '..')
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+
+sys.path.append(os.path.join(base_dir, 'src'))
 
 try:
     import sphinx_rtd_theme
@@ -29,7 +32,16 @@ master_doc = 'index'
 project = u'icalendar'
 this_year = datetime.date.today().year
 copyright = u'{}, Plone Foundation'.format(this_year)
-version = pkg_resources.get_distribution('icalendar').version
+
+# Read the version number
+with open(os.path.join(base_dir, 'src/icalendar/__init__.py')) as f:
+    for line in f.readlines():
+        if line.startswith('__version__'):
+            version_info = {}
+            exec(line, None, version_info)
+            version = version_info['__version__']
+            break
+
 release = version
 
 exclude_patterns = ['_build', 'lib', 'bin', 'include', 'local']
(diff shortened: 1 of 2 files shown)


davidfischer:
Any way to get this a look? Read the Docs is getting reports of broken docs because the docs haven't been built in a few years there.

docs/conf.py
```diff
@@ -29,7 +32,16 @@
...
 this_year = datetime.date.today().year
 copyright = u'{}, Plone Foundation'.format(this_year)
-version = pkg_resources.get_distribution('icalendar').version
+
+# Read the version number
+with open(os.path.join(base_dir, 'src/icalendar/__init__.py')) as f:
+    for line in f.readlines():
+        if line.startswith('__version__'):
+            version_info = {}
+            exec(line, None, version_info)
```
thet: I dislike using ``exec`` here.
Please refactor to do it like here: https://github.com/collective/icalendar/blob/master/setup.py#L7

docs/conf.py
```diff
@@ -1,11 +1,14 @@
...
 # icalendar documentation build configuration file
-import pkg_resources
 import datetime
 import os
+import sys
 
+base_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), '..')
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
+sys.path.append(os.path.join(base_dir, 'src'))
```
thet: Is this needed for a successful RTD build?

davidfischer: No, it isn't strictly. I added this because I removed the `.` line from your requirements file. I believe editing the sys path is the preferred way of adding the module to the path. Depending on your setup, you may not need this.

fuzzy76:
Is there anything particular holding this up? Working documentation would be nice...

mgrollins:
I'm getting the same bad CSS on the built Doc Pages. https://icalendar.readthedocs.io/en/latest/ won't load CSS due to utilizing an old CDN that isn't supported by RTD any longer.

https://github.com/rtfd/readthedocs.org/issues/4082#issue-322232604

roberson-io:
@davidfischer Sorry for the blast from the past, but it looks like this helpful fix you made a year ago may be stuck because the build requires an update the Changelog.  Would you mind pushing an update and seeing if that helps move this along?  I appreciate your help.

davidfischer:
@roberson-io, I created an alternative PR that may be preferable to this one https://github.com/collective/icalendar/pull/289. That one only specifies a Read the Docs config file rather than editing the `conf.py`. I'll make sure the tests pass over there but if that is merged this one can be closed. If instead you prefer this approach I can update this one.

mauritsvanrees:
https://icalendar.readthedocs.io is building correctly now, after PR #289 .

Issue #263, opened 2018-06-06
Categories should be comma separated not repeated


cleder:
Currently when I pass a list of values to categories it will create:
```
BEGIN:VEVENT
SUMMARY:with cats
DTSTART;VALUE=DATE:20180614
DTEND;VALUE=DATE:20180615
DTSTAMP;VALUE=DATE-TIME:20180606T160233Z
UID:http://demo.earth.local/diary/detail/588554/
CATEGORIES:Governor Meeting
CATEGORIES:PTA Meeting
CATEGORIES:School Trips
CATEGORIES:Staff Training Day
DESCRIPTION:miau
URL:http://demo.earth.local/diary/detail/588554/
END:VEVENT
```

Reading the documentation at https://www.kanzaki.com/docs/ical/categories.html it should be:

```
BEGIN:VEVENT
SUMMARY:with cats
DTSTART;VALUE=DATE:20180614
DTEND;VALUE=DATE:20180615
DTSTAMP;VALUE=DATE-TIME:20180606T160233Z
UID:http://demo.earth.local/diary/detail/588554/
CATEGORIES:Governor Meeting,PTA Meeting,School Trips,Staff Training Day
DESCRIPTION:miau
URL:http://demo.earth.local/diary/detail/588554/
END:VEVENT
```

cleder:
see also: #127 #110

Issue #260, opened 2018-04-03
Problems with latest python-dateutil


jensens:
I tried to update our Plone dependencies to *python-dateutil 2.7.2* which broke, see plone/buildout.coredev#447

I found it is an icalendar issue. Also icalendar Travis CI https://travis-ci.org/collective/icalendar/builds/361592061is red.

kyrias:
Been looking into this for a bit now, and it seems like the dateutil exception is actually lying.  The rules in question seems to have a DTSTART without a timezone and a UNTIL in UTC, but the exception claims otherwise.  Will try to ask the dateutil people about this.

kyrias:
Reading RFC 5545 a bit closer it seems that the dateutil exception is indeed misleading, and it's the test case that's wrong.

```

legrostdg:
What is the recommendation for this test? Should it be disabled? python-icalendar (and its reverse dependencies) will be removed soon from Debian... https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=908498

jensens:
This is solved for now, follow up see #274

legrostdg:
Would it be possible to do a new release with this fix? This would help Debian packaging. Thanks!

jensens:
@esteele @mauritsvanrees @thet can we have a release of the latest fixes please?

mauritsvanrees:
There you go: https://pypi.org/project/icalendar/4.0.3/

legrostdg:
Great, thanks!

PR #265, opened 2018-07-12
Categories to be comma separated list 

cleder:
see #263 for details

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Categories are comma separated not 1 per line #265. [cleder]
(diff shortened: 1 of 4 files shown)


cleder:
@esteele Any thoughts how to proceed?

esteele:
I'm ok with this PR being merged.

How can we make sure that test gets fixed at some point?

cleder:
#260 suggests that the test is actually wrong. I have never seen such complex examples in the wild and I am unlikely to put effort into testing of this edge case. 
Question is: do we NEED to cover this edge case or should it just be documented that this is currently not supported?

jensens:
I am fine with this PR for now. As software development is incremental, another PR refining this may be needed.

PR #277, opened 2018-11-11
Remove empty & unnecessary "tests_require" from setup.py

jdufresne:
Does not contain anything. tox will install the package by default.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@ for fname in ['README.rst', 'CONTRIBUTING.rst', 'CHANGES.rst', 'LICENSE.rst']:
         longdesc += f.read()
         longdesc += '\n'
 
-tests_require = []
 install_requires = [
     'python-dateutil',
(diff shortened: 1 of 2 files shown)


jdufresne:
>  But others may have icalender[test] in their requirements, for example in the Plone 5.1 core development buildout. That would break then.

I see, but why do other packages need to install icalendar's test requirements (which is nothing). Shouldn't those packages stop using it?

thet:
I've updated our core buildout and removed all ``test`` extras as a pre-requisite to removeing it from here:

https://github.com/plone/buildout.coredev/pull/678
https://github.com/plone/buildout.coredev/pull/677
https://github.com/plone/buildout.coredev/pull/676
https://github.com/plone/buildout.coredev/pull/675

mauritsvanrees:
@thet I have merged your PRs. (Except 5.0, which is not needed anymore.)

PR #281, opened 2018-12-26
Replace bare 'except:' with 'except Exception:'

jdufresne:
Catching all exceptions is generally considered a bad practice under
most circumstances as it will also catch KeyboardInterrupt and
SystemExit. These special cases should be raised to the interpreter to
allow the Python process to exit.

This fix complies with pycodestyle's error code E722:

https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes

> do not use bare except, specify exception instead

### DIFF
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -391,7 +391,7 @@ class Contentlines(list):
                         line in NEWLINE.split(unfolded) if line)
             lines.append('')  # '\r\n' at the end of every content line
             return lines
-        except:
+        except Exception:
             raise ValueError('Expected StringType with content lines')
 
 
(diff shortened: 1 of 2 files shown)


kevin-brown:
Would it make more sense to catch specific errors, rather than catching `Exception` (the base class for 90% of exceptions)?

I always interpreted E722 as "be more specific with your except clause if you can" because of how it's [specified in PEP 8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

> When catching exceptions, mention specific exceptions whenever possible instead of using a bare `except:` clause

FayasNoushad (approved):
LGTM

PR #292, opened 2019-07-10
Clean up deprecated Python 2.7 code

aleksihakli:
As Python 2.7 deprecation is coming up and Python 3.4 is already dead I cleaned up the codebase and updated the CI and test setup so that the tests can be run cleanly for maintainability.

I also removed Python 3.5 support as Python 3.6 offers us f-strings and other nice things and everyone should be running 3.6 already. Supporting old versions just makes maintenance and development of new features more complex and most distributions should offer 3.6 in the stock packages at this time.


aleksihakli:

UniversalSuperBox:
Hi @aleksihakli,

Thank you for your interest!

I'm new around here, but I don't think we can drop Python2 support just yet. `icalendar` is part of the install dependencies for Plone, and Plone still supports Python2. Specifically, Plone versions lower than 5.2 only support Python 2.7, Plone 5.2 and up support Python 2.7, 3.6, and 3.7.

https://docs.plone.org/manage/installing/requirements.html#python

aleksihakli:
@UniversalSuperBox are you still planning to support Python 2.7 after EOL or are you deprecating the legacy Python code in the new versions? Plone can always depend on an older version of the library since there shouldn't be any major breakage in the iCalendar specification.

aleksihakli:
PS. Python 3.4 is also EOL and should be considered unsafe to use or support.

aleksihakli:

WhyNotHugo:
Is support for Python 2 still important for icalendar / plone?

I'd like to contribute support for zoneinfo, which is part of stdlib and has been backported to many older pythons, but not to unsupported pythons, so this becomes a bit of a blocker. This is mostly since my own applications now depend on `tzinfo` _and_ `zoneinfo`, and I'd like to be able to stop pulling in the former, since it's just duplicate functionality.

Do you have any timeline of when you'll stop supporting unmaintained versions of Python?

BTW, have you considered simply pinning icalendar version used in versions of Plone that support Python 2? Or do you actively push new features to it too?

PR #294, opened 2019-09-11
Don't Require pytz

speedplane:
In this pull request, we don't explicitly require pytz unless the ical events use timezones. pytz is a huge library, and many often try to avoid it if they do not need it. In this case, if you're just using the library to generate all-day events, it isn't strictly necessary.


speedplane:

niccokunzmann:
@speedplane, we are in the process of setting up maintenance for this repository again, see https://github.com/collective/icalendar/discussions/360. It will take some time but I hope, we will get further towards integrating your changes. Thanks a lot for the effort and patience already!
If you like to run the latest tests for this PR, you can rebase/merge master.

niccokunzmann:
```
from icalendar.parser import tzid_from_dt
```

Other modules might need modification now, after that long time.
Also, adding a test would be good: run a new file that imports icalendar, parses a file and checks that pytz is absent.

niccokunzmann:
I will close this for now as it did not receive any attention towards changes. It inspired other issues! It can be opened again.

PR #296, opened 2019-11-12
Calm down fuzzing

UniversalSuperBox:
The fuzzing configuration introduced by https://github.com/collective/icalendar/pull/191 *did* find a novel issue in its first run. However, it is unable to run on travis-ci due to taking too long without output. I tried running it locally to make sure it wasn't Travis being Travis, and sure enough, on my i9-9880H, the operation took well over 20 minutes before I stopped it.

Because of this, I've reduced the number of examples that Hypothesis will run. This will catch fewer errors but help us get PRs back on track.

Ping @untitaker, who introduced this change originally

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Reduce Hypothesis iterations to speed up testing, allowing PRs to pass
+  [UniversalSuperBox]
 
 
 4.0.3 (2018-10-10)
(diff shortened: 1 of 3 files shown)


UniversalSuperBox:
It appears that [deadline](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.deadline) is for individual examples rather than the entire test case.

wbob:
On a tangent: Hypothesis allows to use [profiles](https://hypothesis.readthedocs.io/en/latest/settings.html#settings-profiles), as @untitaker uses in [vdirsyncer](https://github.com/pimutils/vdirsyncer/blob/ccafa115962f6c57d61869641e9e4da5b1e3650a/tests/conftest.py#L32). Travis can then be instructed to use a smaller example size, while per default outside CI more samples are generated.
When I stumbled on this issue, I even went down to `max_examples=10**3` and set `suppress_health_check` to avert a failing test as of slow generation (see HypothesisWorks/hypothesis#2009 for a similar case).

PR #299, opened 2020-01-15
Default type vText when converting vRecur to ical

kam193:
Use default type `vText` when converting recurrence definition to ical string. 

This is similar to `from_ical` behavior, that allows to set anything as recurrence key and treats it as `vText`; but at the same time, `to_ical` has strictly limited list of supported types and no default type, so the rule `x == vDataType.from_ical(VDataType(x).to_ical())` (from https://github.com/collective/icalendar/blob/master/src/icalendar/prop.py#L33) is broken. 

Furthermore, with this we can use non-standard recurrence keys (like `BYEASTER` supported by `deteutil`). 

This is a fix without breaking changes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ New features:
 Bug fixes:
 
 - Fixed a docs issue related to building on Read the Docs [davidfischer]
+- Use ``vText`` as default type, when convert recurrence definition to ical string. [kam193] 
 
 
 4.0.4 (2019-11-25)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ Bug fixes:
 
 - Fixed a docs issue related to building on Read the Docs [davidfischer]
 - Use ``vText`` as default type, when convert recurrence definition to ical string. [kam193] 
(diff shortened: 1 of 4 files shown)


kam193:

kam193:
It looks like broken CI on Plone and too long working `hypothesis` :/

kam193:

mauritsvanrees:
I don't know enough (basically nothing) about `icalendar`. I just occasionally make a release. I cannot review this.

tuergeist:

kam193:

kam193:

jensens:
Even I know not much about the internals of the package. But I would say `vText` as default makes sense. Out of curiosity: When is there no explicit value set? I would expect it to be there.

kam193:
The problem is when type is out of the standard range: `icalendar` library explicitly supports only limited number of types, but there is a lot of useful extending standard types supported by `dateutil` (the library used here to parse rrules).

Now, when we pass to `icalendar` a string with those non-standard types, it works (because `from_ical` has default type `vText`), but if we need this string again (this is my use case) there is an exception, because `to_ical` hasn't any default type :(

kam193:
Thanks!

Issue #303, opened 2020-02-13
icalendar fails parsing some RRULE based timezone definitions


simon-budig:
Some timezone definitions fail parsing. One example is the testcase "test_create_america_new_york" in the current testsuite, which is right now marked unittest.expectedFailure.

Underlying reason is that evaluating the RRULE in some cases is done with mixed offset-awareness, datetime.rrule can't handle that (see e.g. https://github.com/dateutil/dateutil/issues/851),
I've prepared a pull request to fix this. The patch evaluates the rrule with the relevant TZOFFSETFROM from the timezone description. This avoids this problem.

thet:
Closed by: https://github.com/collective/icalendar/pull/304

mauritsvanrees:
I have released 4.0.7 with the fix.
Thanks!

PR #304, opened 2020-02-13
Correct the handling of rrules within timezone definitions.

simon-budig:
Fix for https://github.com/collective/icalendar/issues/303

This also reactivates the testcase that used to fail.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- fixed rrule handling, re-enabled test_create_america_new_york()
 
 
 4.0.4 (2019-11-25)
--- a/src/icalendar/cal.py
+++ b/src/icalendar/cal.py
@@ -17,7 +17,7 @@ from icalendar.prop import vText, vDDDLists
 from icalendar.timezone_cache import _timezone_cache
 
 import pytz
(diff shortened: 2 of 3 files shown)


simon-budig:

d7415:
@simon-budig Thanks for this - I have some downstream users (pimutils/khal#923) reporting success with this patch.
Have you had any more time to look at making Jenkins and Travis happy?

simon-budig:
I am sorry, I have no idea what made Jenkins/Travis fail at that time. I suspect that the patch has bit-rottet a bit in the last 6 months. So any help with bringing this up-to-date and pass the automated tests would be appreciated.

thet:

Actually we probably should just ignore the failing Plone Jenkins jobs here, as Plone depends on icalendar but not the other way around.. I'm thinking of removing the Plone Jenkins jobs after the current run.

d7415:
It looks like most of the Plone Jenkins is happy now, and I'm not convinced that the [failing one](https://jenkins.plone.org/job/pull-request-5.2/1514/testReport/Products.CMFPlone.tests.test_robot/RobotTestCase/Scenario_Location_query_Custom/) is related.

Is it worth rerunning Travis too as those failures seem to be timeouts?

thet:
Looks good to me. Merging.

PR #307, opened 2020-03-03
replace substrings of type byte instead of string

mariushegele:
`to_ical()` returns a `bytes` object, not a string 

"The methods on bytes and bytearray objects don’t accept strings as their arguments, just as the methods on strings don’t accept bytes as their arguments."
https://docs.python.org/3/library/stdtypes.html#bytes-and-bytearray-operations


mariushegele:

PR #340, opened 2021-11-22
Use GitHub Actions. Drop 3.4, 3.5, PyPy2

mauritsvanrees:


mauritsvanrees:
This needs help. Locally on my Mac, `tox` goes fine. But on GHA I get errors:

```
21
____ ERROR collecting build/lib/icalendar/tests/hypothesis/test_fuzzing.py _____
22
ImportError while importing test module '/home/runner/work/icalendar/icalendar/build/lib/icalendar/tests/hypothesis/test_fuzzing.py'.
23
Hint: make sure your test modules/packages have valid Python names.
24
Traceback:
25
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/importlib/__init__.py:127: in import_module
26
    return _bootstrap._gcd_import(name[level:], package, level)
27
build/lib/icalendar/tests/hypothesis/test_fuzzing.py:3: in <module>
28
    from hypothesis import given, settings
29
E   ModuleNotFoundError: No module named 'hypothesis'
30
```

From looking at `tox.ini`, this file is only meant to be run on 2.7 and 3.6, where `hypothesis` is installed. I don't know why we would want to skip this test on other Python versions.
Locally, the test is probably not even found, so there are no complaints.

The rest of the errors is like this:

```
____________ ERROR collecting src/icalendar/tests/test_encoding.py _____________
31
import file mismatch:
32
imported module 'icalendar.tests.test_encoding' has this __file__ attribute:
33
  /home/runner/work/icalendar/icalendar/build/lib/icalendar/tests/test_encoding.py
34
which is not the same as the test file we want to collect:
35
  /home/runner/work/icalendar/icalendar/src/icalendar/tests/test_encoding.py
36
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

PR #352, opened 2022-04-08
Fix ical printing doc

NicoHood:
There more be more places in the doc, where this is required. I was not 100% sure, so I only fixed this obvious, confirmed bug.


niccokunzmann:
@NicoHood, same as with the other PR #355

NicoHood:
I dont understand your question. You've linked a comment and this only covers documentation, not the code itself.

niccokunzmann (approved):

PR #358, opened 2022-06-23
Document Setup process for development

niccokunzmann:
This pull request aims at clarifying how to setup icalendar for development.

PR #359, opened 2022-06-24
use tox to build the documentation

niccokunzmann:
This is a start to build the documentation using tox.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,7 @@ jobs:
         - ["3.9",   "py39"]
         - ["3.10",  "py310"]
         - ["pypy3", "pypy3"]
+        - ["3.10",  "docs"]
 
     runs-on: ubuntu-latest
     name: ${{ matrix.config[1] }}
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -29,3 +29,10 @@ To build the documentation follow these steps:
 You can now open the output from ``_build/html/index.html``. To build the
 presentation-version use ``make presentation`` instead of ``make html``. You
 can open the presentation at ``presentation/index.html``.
+
+You can also use ``tox`` to build the documentation:
+
+.. code-block:: bash
+
+    cd icalendar
+    tox -e docs
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 # to run for a specific environment, use ``tox -e ENVNAME``
 [tox]
-envlist = py27,py36,py37,py38,py39,py310,pypy3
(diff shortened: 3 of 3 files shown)


niccokunzmann:
I would like to use tox extensively in documentation and such so that testing the package becomes easy and contributors are inclined to contribute tested code. This is a patch for tox and documentation building.

mauritsvanrees (approved):
LGTM and it works when I try it locally.
There is conflict though, because I merged your other PR #358 just now. Can you rebase or merge?

mauritsvanrees:
> Can you rebase or merge?

Never mind, I have fixed the conflict in the GitHub UI. Tests should be green again and then someone can merge.

mauritsvanrees:
I have a weird problem now locally. `tox -e py38` works. `tox -e docs` works. After this, `tox -e py38` fails (same for the other Python versions. There are two kinds of errors.

```
__________________________________________________________________________________ ERROR collecting build/lib/icalendar/tests/hypothesis/test_fuzzing.py ___________________________________________________________________________________
ImportError while importing test module '/Users/maurits/community/plone-coredev/6.0/src/icalendar/build/lib/icalendar/tests/hypothesis/test_fuzzing.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/Users/maurits/.pyenv/versions/3.8.12/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
build/lib/icalendar/tests/hypothesis/test_fuzzing.py:3: in <module>
    from hypothesis import given, settings
E   ModuleNotFoundError: No module named 'hypothesis'
__________________________________________________________________________________________ ERROR collecting src/icalendar/tests/test_encoding.py ___________________________________________________________________________________________
import file mismatch:
imported module 'icalendar.tests.test_encoding' has this __file__ attribute:
  /Users/maurits/community/plone-coredev/6.0/src/icalendar/build/lib/icalendar/tests/test_encoding.py
which is not the same as the test file we want to collect:
  /Users/maurits/community/plone-coredev/6.0/src/icalendar/src/icalendar/tests/test_encoding.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

The fix is: remove the `build` directory that was created by `tox -e docs`. Can you confirm this? Would you know a way around this?
Actually, I guess in the Makefile I can simply remove the `build` directory at the end, because this is not needed. Only `docs/_build/html` is needed. I will do that.

niccokunzmann:
> Actually, I guess in the Makefile I can simply remove the build directory at the end, because this is not needed. Only docs/_build/html is needed. I will do that.

Ok.

PR #362, opened 2022-07-11
[WIP] use github-actions for CI

geier:


mauritsvanrees:
I have just found out what was wrong with my own PR from November, a bad way of specifying the `test` extra in `tox.ini`.
https://github.com/collective/icalendar/pull/340 is now almost green. Need to check one problem with coveralls on Py 2.7 only. I will probably disable it.

mauritsvanrees:
Done. I think this PR can be closed. If you think there some things in here are still improvements, opening a new PR is probably better.

I have released 4.1.0 and 5.0.0a1. The only current difference is that 4.1.0 (new branch 4.x) only drops some Python versions from the tests, and on 5.0.0a1 (master) they are really officially not supported anymore. We could drop more versions there if we want. Now at least both branches have a reasonable tox and gh-actions.

PR #363, opened 2022-08-07
Stop Build form Failing on Python 3.4

QuantumNovice:


.github/workflows/tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install tox
+        pip install -r requirements.txt
```
niccokunzmann, line 45: Are these special requirements for the tests only?
Is there a link to the failing tests?

niccokunzmann: Hm. This stops it failing on Python3.4 but Python3.4 is not in the matrix.... Would this go to the 4.x branch?

QuantumNovice: Link to failing tests https://travis-ci.org/github/collective/icalendar/jobs/774624957

QuantumNovice: > Are these special requirements for the tests only?

I just added a generic `requirement.txt` file. 
`typing` module is a requirement for some tests in python3.4.

niccokunzmann:
@QuantumNovice Thanks! I am wondering: according to 

> Branch 4.x is for Python 2.7, 3.6-3.10, and PyPy3. Those are tested. Technically it should still work on 3.4, 3.5 and PyPy2 and let's try not to break this. - https://github.com/collective/icalendar/discussions/360#discussioncomment-3120708

4.x is the branch to put the tests for 3.4 and 3.5.
I see that testing these is missing here: https://github.com/collective/icalendar/blob/4.x/.github/workflows/tests.yml

Would you like to make a PR to the 4.x branch enabling the testing of 3.4, 3.5 and maybe pypy2?

Also a hint: If `typing` is missing for some installations, then the way to install it would be here, like this:

https://github.com/collective/icalendar/blob/db7874f352291a0ffd02a0350cacd81ed67ca0f8/tox.ini#L9

Thanks for keeping an eye out! Also, we do not use Travis any more but GitHub Actions. I replaced the badges to make that clear here: https://github.com/collective/icalendar/pull/364

If you were to do that, that would probably be quite supporting as the project is looking for a maintainer. #360

geier:
We do not support python 3 versions < 3.6 any more, so I don't think we should be accommodating to EOLed python versions.

QuantumNovice:
I see I will do as requested.

PR #366, opened 2022-08-13
Require Python 3.8.

mauritsvanrees:
Nothing has changed in the code. We can clean it up in a later PR.


mauritsvanrees:
There is an error on PyPy3. I saw something locally as well, but it was not clear. On [gh-actions it is clear](https://github.com/collective/icalendar/runs/7819322381?check_suite_focus=true#step:6:20):

    ERROR: Package 'icalendar' requires a different Python: 3.6.12 not in '>=3.8'

So if we drop even Python 3.6 support, this results in dropping PyPy3 support? I am hardly ever using PyPy, but it would be nice if that keeps working. Does anyone know in what way to handle this best?

docs/Makefile
```diff
@@ -45,7 +45,8 @@ clean:
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@echo
-	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
+	@echo "Build finished. The HTML pages are in docs/$(BUILDDIR)/html."
+	rm -rf build
```
niccokunzmann, line 49: Yep, that is what you spoke about here:
https://github.com/collective/icalendar/pull/359#issuecomment-1214150269

tox.ini
```diff
@@ -1,16 +1,17 @@
 # to run for a specific environment, use ``tox -e ENVNAME``
 [tox]
-envlist = py27,py37,py38,py39,py310,pypy3,docs
+envlist = py38,py39,py310,pypy3,docs
+# Note: the 'docs' env creates a 'build' directory which may interfere in strange ways
+# with the other environments.  You might see this when you run the tests in parallel.
+# See https://github.com/collective/icalendar/pull/359#issuecomment-1214150269
```
niccokunzmann, line 6: Is that still the case now that you delete it?

niccokunzmann (changes requested):

niccokunzmann:
Hm. the build fails.

niccokunzmann:
I think, I would also add some documentation on the different branches and their purpose.

niccokunzmann:
I installed PyPy and I do not know where this error comes from:
```
$ pypy
Python 3.9.12 (05fbe3aa5b0845e6c37239768aa455451aa5faba, Mar 29 2022, 08:15:34)
[PyPy 7.3.9 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] on linux
>>>> import sys
>>>> sys.version
'3.9.12 (05fbe3aa5b0845e6c37239768aa455451aa5faba, Mar 29 2022, 08:15:34)\n[PyPy 7.3.9 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)]'
```

Maybe, does the test env install an old version of PyPy?

niccokunzmann:
I do not know how to fix that yet, see https://github.com/collective/icalendar/runs/7821956469?check_suite_focus=true#step:3:8

niccokunzmann:
If #372 is merged, this can be merged with green tests. @mauritsvanrees

tox.ini
mauritsvanrees: I think so, when `docs` finishes before one of the other jobs is properly started.

PR #368, opened 2022-08-13
describe purpose of the branches master and 4.x

niccokunzmann:
@mauritsvanrees @geier
I created this PR to make the purpose of the branches clear.
There is some difference between the branch purpose as pointed out in https://github.com/collective/icalendar/discussions/360:

> Increase master branch version to 5 (or 6 if needed). This should keep working on Plone 6.0 (which is near beta), which means Python 3.7-3.10.

and this PR https://github.com/collective/icalendar/pull/366

> [Require Python 3.8.](https://github.com/collective/icalendar/pull/366/commits/c73cab2c101beaf2a13c7e8d7037da3340123033)

How to approach this?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ Since we pour more `effort into maintaining and developing icalendar <https://gi
 we split the project into two:
 
 - `Branch ``4.x`` <https://github.com/collective/icalendar/tree/4.x>`__ with maximum compatibility to Python versions ``2.7`` and ``3.4+``, ``PyPy2`` and ``PyPy3``.
-- `Branch ``master`` <https://github.com/collective/icalendar/>`__ with the compatibility to Python versions ``3.8+`` and ``PyPy3``.
+- `Branch ``master`` <https://github.com/collective/icalendar/>`__ with the compatibility to Python versions ``3.7+`` and ``PyPy3``.
 
 We expect the ``master`` branch with versions ``5+`` receive the latest updates and features,
 and the ``4.x`` branch the subset of security and bug fixes only.
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,22 @@ files.
 .. _`pytz`: https://pypi.org/project/pytz/
 .. _`BSD`: https://github.com/collective/icalendar/issues/2
 
+Versions and Compatibility
+--------------------------
+
+``icalendar`` is a critical project used by many. It has been there for a long time and maintaining
+long-term compatibility with projects conflicts partially with providing and using the features that
+the latest Python versions bring.
+
+Since we pour more `effort into maintaining and developing icalendar <https://github.com/collective/icalendar/discussions/360>`__,
+we split the project into two:
+
+- `Branch ``4.x`` <https://github.com/collective/icalendar/tree/4.x>`__ with maximum compatibility to Python versions ``2.7`` and ``3.4+``, ``PyPy2`` and ``PyPy3``.
+- `Branch ``master`` <https://github.com/collective/icalendar/>`__ with the compatibility to Python versions ``3.8+`` and ``PyPy3``.
+
+We expect the ``master`` branch with versions ``5+`` receive the latest updates and features,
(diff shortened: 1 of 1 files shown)


geier:
To me it sounds good, icalendar 4.x on 3.7-3.10, icalendar >= 5.0 on python >= 3.8

geier:
Thinking about this again:

Perhaps we should make a 5.x release (or at least a branch for now) specifically for Plone 6 with Python 3.7-3.10 support. This will be much easier to support then the 4.x branch, which still has Python 2 and Python 3.4-3.6 support.

Perhaps even hold out on dropping 3.7 support until it is really needed.

mauritsvanrees:
In [this discussion comment](https://github.com/collective/icalendar/discussions/360#discussioncomment-3389429) I wrote:

> For me it is okay to drop Python 2.7 and 3.7 from the master branch. I have changed Plone 5.2 to use branch 4.x. And on latest Plone 6.0 I have added a pin for Python 3.7 to use the latest 4.x version as well.

I dropped 3.7 because of what @geier [wrote](https://github.com/collective/icalendar/discussions/360#discussioncomment-3353528):

> I'd like to not only drop python 3.6 but also 3.7 (because python 3.8 makes type hinting really a lot easier IMHO)

In the spirit of moving forward, this would be good. But I am not well versed in type hinting yet. And since this has been dubbed as a critical project, perhaps we should not abandon older Python versions until they are really not supported anymore.
From a Plone standpoint, it should be fine to drop 3.7. I am happy to let you two decide on this.

mauritsvanrees (approved):

niccokunzmann:
How is it if we support 3.7 for a bit and if we see a merge request failing because of 3.7 incompatibility, we drop 3.7 explicitely?

README.rst
```diff
@@ -38,6 +38,22 @@ files.
...
+
+``icalendar`` is a critical project used by many. It has been there for a long time and maintaining
+long-term compatibility with projects conflicts partially with providing and using the features that
+the latest Python versions bring.
+
+Since we pour more `effort into maintaining and developing icalendar <https://github.com/collective/icalendar/discussions/360>`__,
+we split the project into two:
+
+- `Branch ``4.x`` <https://github.com/collective/icalendar/tree/4.x>`__ with maximum compatibility to Python versions ``2.7`` and ``3.4+``, ``PyPy2`` and ``PyPy3``.
+- `Branch ``master`` <https://github.com/collective/icalendar/>`__ with the compatibility to Python versions ``3.7+`` and ``PyPy3``.
```
niccokunzmann, line 52: This mentions `3.7+` now.

niccokunzmann:
I like documentation! It allows us to have a hands-on discussion and join forces into one direction :)

niccokunzmann:
@geier - if you are fine with these changes, I think, it is up to you to merge them. Then, it is clear how to maintain the repository and how to direct contributions. (branch + version) I'd be ok with both 3.8+ and 3.7+. Also, I think we have enough time to migrate to 3.8 and zoneinfo - 10 months - when we merge it with 3.7+.

niccokunzmann:
We merged https://github.com/collective/icalendar/pull/366 and thus the changes here do not reflect what is written - but I do documentation first - the documentation guides the expectation and the expectation the code. I will make changes for 3.7 based on #366 once it is merged.

mauritsvanrees (approved):
LGTM now.

geier:
Sorry about doing a u-turn there @mauritsvanrees .

Didn't see that you had already dropped 3.7. I don't mind either way.

niccokunzmann:
@geier

Who would you like to review?

I see this:
image

I think mauritsvanrees gave a review and I made the commit and I do not know who else is left to review/take the decision.
You can merge from my perspective. We can drop 3.7 any time but add it in will be hard, maybe.

geier:
I actually wanted to review, self assigned but then got interrupted... I'll merge this now.

PR #370, opened 2022-08-13
use PyPy3.9 explicitely in GitHub actions

niccokunzmann:
This is a try for #366


niccokunzmann:
This was just a try to install a later PyPy version.

PR #381, opened 2022-08-19
test Python 3.7 with GitHub actions

niccokunzmann:
this updates the tests to match the Python version described in the README

PR #376, opened 2022-08-15
Remove any syntax required for python < 3.8

WhyNotHugo:
`> find . -iname "*.py" -exec pyupgrade --py38-plus {} \;`

PR #384, opened 2022-08-19
Drop syntax that was only required for py < 3.7

WhyNotHugo:
find . -iname "*.py" -exec pyupgrade --py37-plus {} \;

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # icalendar documentation build configuration file
 import pkg_resources
 import datetime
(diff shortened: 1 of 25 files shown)


niccokunzmann:
@WhyNotHugo It would go to the master branch - the 4.x branch keeps backward compatibility for fixes that people might want to apply.

niccokunzmann:
I will notify you when the PR #381 is merged because that will create the tests so that we can test and merge fast instead of all this back and forth. Sorry for that!

niccokunzmann:
merged as #376

WhyNotHugo:
Oh, I think I misunderstood the intent of 4.x.

Issue #315, opened 2020-07-15
unique identifier must be included in VEVENT, VTODO, and VJOURNAL


joooeey:
### Issue

As per [RFC 5545](https://tools.ietf.org/html/rfc5545#section-3.8.4.7) every VEVENT, VTODO, and VJOURNAL entry must have a unique identifier (UID). This module does not currently implement this requirement and hence produces output incompatible with RFC 5545 if the user does not manually supply an UID.

### Proposed solution 1

Every time an event is initialized with `icalendar.Event()`, the UID field is added automatically and populated with a unique identifier. Then the user can still specify their own UID.

The question is how to get a unique identifier. Perhaps do it similar to the example given in RFC 5545: concatenate a timestamp, a sufficiently large pseudo-random string, sequential number, and "@collective.github.io".

### Proposed solution 2

Emit a warning or error if the user does not specify the UID property.

joooeey:
I overlooked that we already have a way to get a UID:

```
from tools import UIDGenerator
uid = UIDGenerator.uid("@collective.github.io")
```

joooeey:
Anyways, the issue appears to be in `cal.py` in the last line of the `Event` class: `ignore_exceptions = True`. This looks intended. Why do we do this?

I think this flag applies to both parsing and file generation. But then I think there's never a reason to ignore exceptions when generating a file? It certainly shouldn't be the default!

niccokunzmann:
@jacadzaca did some work on this in #390.

I have suggestions of how to go about this:

1. Add a `is_valid_rfc_5455()` to the classes with requirements. These functions document what is required to be RFC 5545 compatible. They might also check for invalid attribute names. Also the function doc string links to (2).
2. How to generate valid components should be stated in the documentation. One subsection per component as an example.
3. Make `to_ical()` raise a warning **once** with the docstring content/a link to the documentation if `is_valid_rfc_5455()` returns false. This should not break production code and also guide while developing.

@joooeey @jacadzaca do you have thoughs on this?

joooeey:
This sounds good. Better than my original idea of putting the responsibility of choosing UIDs on the `icalendar` software.

Regarding 3, I don't think a warning is enough in the long run. I propose making this a deprecation warning which will be an error a few versions down the line as @geier suggested [here](https://github.com/collective/icalendar/pull/390#issuecomment-1224911142). Compatibility is crucial.

niccokunzmann:
[...]

Issue #223, opened 2017-04-22
(Auto) fixing invalid icalendar (components)


geier:
As already recently discussed in #222, should we try to fix broken icalendars and/or icalendar components?

The issue in #222 were VTIMEZONE's OFFSETs > 24h which should be minutes (e.g., `TZOFFSETFROM:+5744` =>  `TZOFFSETFROM:+005744`) (I'm inclined to say no, as it's hard to guess what was really meant).

Further issues are VEVENTs without a UID (see https://github.com/pimutils/khal/issues/632) or DTSTART == DTEND.

stlaz:
I don't think we should try to be clever with invalid inputs. After all, we are but a parser, it's our duty to warn the user that their input is not correct, and fail in that case.
We can have some error recovery in case the parsing should not be strict but the best thing we can do is to leave the input untouched, not trying to fix it. The user should fix it themselves in such case. A warning that their input is not correct should still be issued, though.

gasull:
> I don't think we should try to be clever with invalid inputs.

Many [HTML parsers](https://en.wikipedia.org/wiki/Comparison_of_HTML_parsers), like HTML Tidy, attempt to fix invalid HTML, or have the option to do so.  `icalendar` should do the same.

PR #390, opened 2022-08-23
Add missing mandatory fields to user generated components

jacadzaca:
This pull requests attempts to solve #315 by exposing a new ComponentFactory that's supposed to be 'user-firendly'.  I added some examples in the README.

I think it wouldn't be correct to make the old ComponentFactory return components with all the required fields, since that might  overwrite some properties during parsing that we wouldn't want to overwritten, e.g [DTSTAMP](https://www.rfc-editor.org/rfc/rfc5545#section-3.8.7.2) on events. Moreover, if parsed components don't provide them, I don't think a parser should fix that implicitly, but rather throw an error.

Lastly, I need some help with figuring out what should be put in the DTSTART field inside the TimeZoneStandard/TimeZoneDaylight component, the RFC says that it ought to be 'the effective onset date and local time for the time zone sub-component', but I have no clue how to get the value using the libraries.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,9 +66,8 @@ Related projects
 Examples
 ================
 Create a calendear and add an event to it (master branch)
-..example-code::
 
-    .. codeblock:: python
+.. code-block:: python
 
     #!/usr/bin/env python3
     from dateutil import tz
@@ -100,15 +99,15 @@ Create a calendear and add an event to it (master branch)
         main()
 
 
-    You can view it with the handy CLI tool by:
+You can view it with the handy CLI tool by:
 
-    .. codeblock:: bash
+.. code-block:: text
 
     python SCRIPT-NAME.py > sample.ics && pythom -m icalendar.cli sample.ics
 
-    Create a ics file with 5 alarms, each 10 minutes apart
+Create a ics file with 5 alarms, each 10 minutes apart
 
-    .. codeblock:: python
+.. code-block:: python
 
     #!/usr/bin/env python3
     from datetime import datetime, timedelta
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -61,3 +61,72 @@ Related projects
...
+
+
+Examples
+================
+Create a calendear and add an event to it (master branch)
+
+.. code-block:: python
+
+    #!/usr/bin/env python3
+    from dateutil import tz
```
geier, line 73: We are already using pytz and hopefully soon using tzinfo, let's not start using dateutil's tz implementation as well.

Also, at least in this snippet, we don't really need it, or am I missing something?

geier:
Hi @jacadzaca ,

I think the issue you tackle here is really important (making it easier, that users don't produce invalid icalendar files here).

I would prefer another approach here, that -- after perhaps one or two major release versions  -- assures that users cannot produce invalid icalendar files. 

The first versions should be opt-in, e.g. adding a validate=True parameter to icalendar.to_ical(), but later versions should either refuse to produce invalid files or add the required properties.

We had i previous discussion on this before which I now can't find.

Anyway, perhaps there are other opinions.

geier:
> Lastly, I need some help with figuring out what should be put in the DTSTART field inside the TimeZoneStandard/TimeZoneDaylight component, the RFC says that it ought to be 'the effective onset date and local time for the time zone sub-component', but I have no clue how to get the value using the libraries.

Creating proper VTIMEZONE component is pretty hard, there is currently no code at all for it in icalendar. There is a "solution" in khal, but it's not pretty, prone to breakage and only works for pytz timezones.

jacadzaca:
I agree that the goal is to stop users from creating bad icals. I think what way we choose to do it should depend on whether we want/need to refactor the parsing/marshalling code - to me the code is a little all over the place right now, thus implementing your idea might prove difficult

If we'd be happy with leaving the code as is, specific functions for creating components (and then documenting that they are *the* way) is enough, otherwise your idea seems reasonable. 

If icalendar were to stop users from creating bad icals altogether, I think the proper way would be to refuse to produce output - I don't know how we could guess what timezone to put into the Timezone object or what's the correct default trigger for an alarm.

jacadzaca:
> We had i previous discussion on this before which I now can't find.

The discussions:
 - #174 
 - #223

niccokunzmann:
@geier I would assign this to you as it takes more time for me to work myself into this matter.

niccokunzmann:
@jacadzaca for the progress: If you like me to look at this, I will but I would like you to tell me.

jacadzaca:
@niccokunzmann Since the VTIMEZONE component creation isn't correct and generating RFC complaint components isn't difficult to do by hand, I think we shouldn't merge it. Later, we can move the code into the examples section as per #443.

 What do you think?

PR #391, opened 2022-08-29
add __eq__ to icalendar.prop.vDDDTypes

### LINKED ISSUE #308: Missing equivalence member function for icalendar.prop.vDDDTypes
While comparing two vevents, I have to check if RECURRENCE-ID are the same in addition to UID.
However, `vevent1["RECURRENCE-ID"]==vevent2["RECURRENCE-ID"] `always returns false, as only the object ids are compared and not the content.

It would be helpful if `icalendar.prop.vDDDTypes` would implement `__eq__(self, other)`.

jacadzaca:
this fixes #308

### DIFF
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -323,6 +323,11 @@ class vDDDTypes:
         else:
             raise ValueError(f'Unknown date type: {type(dt)}')
 
+    def __eq__(self, other):
+        if isinstance(other, vDDDTypes):
+            return self.params == other.params
+        return False
+
     @classmethod
     def from_ical(cls, ical, timezone=None):
         if isinstance(ical, cls):
--- a/src/icalendar/tests/test_unit_prop.py
+++ b/src/icalendar/tests/test_unit_prop.py
@@ -112,6 +112,27 @@ class TestProp(unittest.TestCase):
         # Bad input
         self.assertRaises(ValueError, vDDDTypes, 42)
 
+    def test_vDDDTypes_equivalance(self):
+        from ..prop import vDDDTypes
+        from copy import deepcopy
+
+        vDDDTypes = [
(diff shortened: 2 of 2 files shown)


src/icalendar/prop.py
```diff
@@ -323,6 +323,11 @@ def to_ical(self):
         else:
             raise ValueError(f'Unknown date type: {type(dt)}')
 
+    def __eq__(self, other):
```
niccokunzmann, line 326: If you override `__eq__`, please also override `__hash__` or something like this or the dictionary mapping will fail.

niccokunzmann:
Great, thanks, I pointed out a Python convention.
Thanks for testing!

src/icalendar/prop.py
jacadzaca: [The 'standard' way of implementing hash](https://docs.python.org/3/reference/datamodel.html#object.__hash__) requires fields compared in eq to be hashable, vDDDTypes.param is of type Parameter and that's a subclass of CaselessDict, neither are hashable. Should I make CaslelssDict/Parameter hashable (for example as per [this](https://stackoverflow.com/questions/1151658/python-hashable-dicts#answer-1151705)) and then implement vDDDTypes's hash in the standard way or would you recommend some other solution?

niccokunzmann: Oh, I actually do not know what would be good to do in this case. I would not like to drag it out because of uncertainty. It solves an issue already. That counts. How would you go forward?

niccokunzmann: Personally, after reading tbe stackoverflow answer, I think making them hashable is not worth making everything hashable as they are not supposed to be used as keys anyway and also easily converted to keys if necessary.
It is good to me now and thanks for having the discussion!

Issue #399, opened 2022-09-09
Inconsistent behavior when parsing calendars?


jacadzaca:
This snippet:
```python
from icalendar import Calendar

with open('some_calendar.ics') as f:
    calendar = Calendar.from_ics(f.read())
    print(calendar.to_ical())
```

when passed this broken calendar:
```
BEGIN:VCALENDAR
VERSION:2.0
METHOD:PUBLISH
BEGIN:VEVENT
DTSTART:20140401T000000Z
DTEND:20140401T010000Z
DTSTAMP:20140401T000000Z
SUMMARY:Broken Eevnt
CLASS:PUBLIC
STATUS:CONFIRMED
TRANSP:OPAQUE
END:VEVENT
X
END:VCALENDAR
```

outputs `ValueError: Content line could not be parsed into parts: 'X': Invalid content line`

but when it's passed this broken calendar:
```
BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART:20150905T090000Z
DTEND:20150905T100000Z
UID:123
X-APPLE-RADIUS=49.91307046514149
END:VEVENT
END:VCALENDAR
```

outputs `b'BEGIN:VCALENDAR\r\nBEGIN:VEVENT\r\nDTSTART:20150905T090000Z\r\nDTEND:20150905T100000Z\r\nUID:123\r\nEND:VEVENT\r\nEND:VCALENDAR\r\n'`

icalendar should either always throw an exception on broken calendar or always skip the broken property.

icalendar is from master branch
Related tests: [test_issue_168](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_fixed_issues.py#L347) and [test_issue_104](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_fixed_issues.py#L214)

niccokunzmann:
Hm. I can not see where the second example is invalid. Can you point out the line?

jacadzaca:
[...]

Issue #397, opened 2022-09-08
test_issue_335_identify_UTC fails?


jacadzaca:
Running test_issue_335_identify_UTC fails on my machine. Seems to be a timezone issue, since the diff shows that the dates are an hour apart. Any idea why that is?

[Full testlog](https://github.com/collective/icalendar/files/9528969/testlog.txt)

jacadzaca:
I think the issue is in  [line 506](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_fixed_issues.py#L506), since [astimezone actually converts the given datetime to the timezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone)

If I change line line 505 to `dt = datetime.datetime(2021, 11, 17, 15, 9, 15, tzinfo=zone)` and line 506 to `myevent.add('dtstart', dt)` the tests pass, so it must be it

niccokunzmann:
Hm. Is this changed in the tests in the master branch? I would close it then. The test code might be wrong.

niccokunzmann:
Ah. OK. The app I am using does not show the link to the merged PR. Alright.

PR #400, opened 2022-09-09
test_fixed_issues restructure

jacadzaca:
Restructure tests inside test_fixed_issues along the lines of #374. I tried to keep tests for similar assertions in one file.

Before considering merging, please take a look at ~~#397~~ and #399


jacadzaca:
Test fails because `pytz.timezone('America/Sao_Paulo')` is not considered the same as `(UTC-03:00) Brasília`, [although it is ( principal city) ](https://en.wikipedia.org/wiki/UTC%E2%88%9203:00#As_standard_time_(year-round)). ZoneInfo and dateutil.tz detect that.

src/icalendar/tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+EVENTS_FOLDER = os.path.join(HERE, 'events')
+CALENDARS_FOLDER = os.path.join(HERE, 'calendars')
+
+TIMEZONES = DataSource(TIMEZONES_FOLDER, icalendar.Timezone.from_ical)
+EVENTS = DataSource(EVENTS_FOLDER, icalendar.Event.from_ical)
+CALENDARS = DataSource(CALENDARS_FOLDER, icalendar.Calendar.from_ical)
+
+@pytest.fixture
+def timezones():
+    return TIMEZONES
```
niccokunzmann, line 44: Sometimes, elements might be changed by tests and then, we have side effects.
Personally, I would put `DataSource(TIMEZONES_FOLDER, icalendar.Timezone.from_ical)` into the function body.

Does that impact execution time?

I would say that we can use `__getattr__` to set the attributes and then, they will only loaded when needed. What do you think?

src/icalendar/tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    return EVENTS
+
+@pytest.fixture
+def calendars():
+    return CALENDARS
+
+@pytest.fixture
+def calendars_folder():
+    return CALENDARS_FOLDER
+
```
niccokunzmann, line 57: Nice, I like this file!

src/icalendar/tests/events/issue_112_missing_tzinfo_on_exdate.ics
```diff
@@ -1,30 +1,3 @@
-BEGIN:VCALENDAR
-PRODID:-//Google Inc//Google Calendar 70.9054//EN
```
niccokunzmann, line 2: Why is is removed is probably in the commit...

src/icalendar/tests/test_api.py
```diff
@@ -0,0 +1,25 @@
...
+        'geo:-33.868900,151.207000',
+        parameters={
+            'VALUE': 'URI',
+            'X-ADDRESS': '367 George Street Sydney CBD NSW 2000',
+            'X-APPLE-RADIUS': '72',
+            'X-TITLE': '367 George Street'
+        }
+    )
+
+    assert event.to_ical() == events.issue_116_117_add_x_apple_structured.raw_ics
```
niccokunzmann, line 24: Cool!

src/icalendar/tests/test_encoding.py
```diff
@@ -88,3 +96,48 @@ def test_unicode_parameter_name(self):
...
             + b'END:VEVENT\r\nEND:VCALENDAR\r\n'
         )
+
+@pytest.mark.parametrize('event_name', [
+    # Non-unicode characters in summary
+    'issue_64_event_with_non_unicode_summary',
+    # Unicode characters in summary
+    'issue_64_event_with_unicode_summary',
+    # chokes on umlauts in ORGANIZER
+    'issue_101_icalendar_chokes_on_umlauts_in_organizer'
```
niccokunzmann, line 106: Cool. Very clear which issue is referenced. This was important to me and it seems to work nicely.

src/icalendar/tests/conftest.py
niccokunzmann: This is how I do it:

https://github.com/niccokunzmann/python-recurring-ical-events/blob/master/test/conftest.py#L49

... not quite like I said ...

src/icalendar/tests/test_encoding.py
```diff
@@ -88,3 +96,48 @@ def test_unicode_parameter_name(self):
...
+    '''Issue #64 - Event.to_ical() fails for unicode strings
+       Issue #101 - icalendar is choking on umlauts in ORGANIZER
+
+    https://github.com/collective/icalendar/issues/64
+    https://github.com/collective/icalendar/issues/101
+    '''
+    event = getattr(events, event_name)
+    assert event.to_ical() == event.raw_ics
+
+@pytest.mark.parametrize('timezone', [
```
niccokunzmann: I would like to change this and add a parametrized fixture that takes the time zone string as an argument, so we make sure we have one place in which we list all the different time zone implementations.

https://github.com/niccokunzmann/python-recurring-ical-events/blob/7f6aeed4729b97981c22ef0f481029cca225ae76/test/conftest.py#L120

In conftest: 

```
# for parametrizing fixtures, see https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures
@pytest.fixture(params=[pytz.timezone, tz.gettz, zoneinfo.ZoneInfo], scope="module")
def TZ(request):
    return request.param()
```

Here:

```
def test_parses_event_with_non_ascii_tzid_issue_237(events, TZ):
...
    expected = datetime.datetime(2017, 5, 11, 13, 30, tzinfo=TZ('America/Sao_Paulo'))
```

1. This way, we can add a timezone implementation easily in one place for all tests.
2. We can use this fixture in the future, if we were to specify a timezone parameter for the parsing of the ics files.

What do you think?

src/icalendar/tests/test_issue_104_mark_events_broken.py
```diff
@@ -0,0 +1,22 @@
...
+    https://github.com/collective/icalendar/issues/104
+    '''
+    assert events.issue_104_mark_events_broken.is_broken # TODO: REMOVE FOR NEXT MAJOR RELEASE
+    assert events.issue_104_mark_events_broken.errors == [(None, "Content line could not be parsed into parts: 'X': Invalid content line")]
+
+def test_dont_ignore_exceptions_on_broken_calendars_issue_104(calendars_folder):
+    '''Issue #104 - line parsing error in a VCALENDAR
+    (which doesn't have ignore_exceptions). Should raise an exception.
+    '''
+    calendar_path = os.path.join(calendars_folder, 'issue_104_broken_calendar.ics')
```
niccokunzmann, line 20: This would use the fixture.

src/icalendar/tests/test_issue_335_identify_UTC.py
```diff
@@ -0,0 +1,13 @@
...
+import pytest
+from dateutil import tz
+from datetime import datetime
+
+try:
+    import zoneinfo
+except ModuleNotFoundError:
+    from backports import zoneinfo
+
+from icalendar import Event
```
niccokunzmann, line 12: WIP, this is empty.

src/icalendar/tests/test_parsing.py
```diff
@@ -0,0 +1,163 @@
...
+    https://github.com/collective/icalendar/issues/53
+    '''
+    assert timezones.issue_53_tzid_parsed_properly['tzid'].to_ical() == b'America/New_York'
+
+@pytest.mark.parametrize('timezone', [
+    pytz.utc,
+    zoneinfo.ZoneInfo('UTC'),
+    pytz.timezone('UTC'),
+    tz.UTC,
+    tz.gettz('UTC')])
```
niccokunzmann, line 34: calls for a general fixture in conftest: utc

src/icalendar/tests/test_parsing.py
```diff
@@ -0,0 +1,163 @@
...
+    timezone = getattr(timezones, timezone_file)
+    assert timezone.to_ical() == timezone.raw_ics
+
+def test_decode_rrule_attribute_error_issue_70(events):
+    # Issue #70 - e.decode("RRULE") causes Attribute Error
+    # see https://github.com/collective/icalendar/issues/70
+    recur = events.issue_70_rrule_causes_attribute_error.decoded('RRULE')
+    assert isinstance(recur, vRecur)
+    assert recur.to_ical() == b'FREQ=WEEKLY;UNTIL=20070619T225959;INTERVAL=1'
+
```
niccokunzmann, line 163: I really like the test structure, the different fixtures and the way that there are a few assertions always at the end.

niccokunzmann (changes requested):
So, I like where the code is heading!

There is one problem that I see.... To review these conditions seem to be important:

- [ ] that no test is left out
- [ ] that tests that were changed correspond to the test that was there before
- [ ] that nothing was added that was not tested before

Since this is a very big PR, this will take a while and any new PR to tests will create merge conflicts.

How can we tackle that?

To make sure that this is clear: **I really like the code that I see!**
I find it hard to review.

I see two options:
1. You wait until I reviewed this
     - I will go through the commits to see that the changes are consistent refectorings
     - I will go though each test in the original code (master) and then try to find them in the PR
2. You split it up:
    - Creating the base (conftest.py first), wait for it to be merged
    - Create a PR for tests by cherry-picking

Also a question: Is this possible to port to 4.x?

What are your thoughts on this?

src/icalendar/tests/test_encoding.py
niccokunzmann: https://github.com/collective/icalendar/runs/8273351016?check_suite_focus=true#step:6:58 would be solved by passing a parameter to the parser of which implementation to use.

src/icalendar/tests/test_issue_335_identify_UTC.py
jacadzaca: I think I forgot to remove this file, when I moved test 335 to test_parsing. I will remove it

src/icalendar/tests/events/issue_112_missing_tzinfo_on_exdate.ics
jacadzaca: Seems like a trip up on my side, weird that the test still passes.

src/icalendar/tests/conftest.py
jacadzaca: To be honest, I didn't quite grasp the piece of code you've linked when I was writing my conftest. You're right that side effects are possible. I think implementing simple lazy loading shouldn't be too hard. I can override __getattr__ or use the `@property` decorator, like you've said.

src/icalendar/tests/test_encoding.py
jacadzaca: Should that be in the commit with the test or a separate one?

jacadzaca:
@niccokunzmann Now that I think about it, it really is a big PR. Tomorrow-ish, I will implement the suggestion you've made and cherry pick it into few smaller ones (option 2). Thanks :)

angatha:
@jacadzaca nice work that you provide to this package 👍

PR #401, opened 2022-09-11
test restructure - add conftest

jacadzaca:
splitting into smaller PRs per #400


niccokunzmann:
Thanks, I like it!

niccokunzmann:
@jacadzaca thanks for putting in so much work - sorry I was absent for a bit - my mails did not come in as the mail provider changed the imap server name.

PR #403, opened 2022-09-11
Test restructure - test_restructure_issue_55

jacadzaca:
splitting as per #400


src/icalendar/tests/test_parsing.py
```diff
@@ -1,3 +1,15 @@
+'''Tests checking that parsing works'''
+import pytest
+
+@pytest.mark.parametrize('timezone_file', [
+    # Issue #55 - Parse error on utc-offset with seconds value
+    # see https://github.com/collective/icalendar/issues/55
+    'issue_55_parse_error_on_utc_offset_with_seconds',
+])
+def test_timezones_to_ical_is_inverse_of_from_ical(timezones, timezone_file):
+    timezone = getattr(timezones, timezone_file)
```
niccokunzmann: If you add a __getitem__ method - should be the same as __getattr__, then this is a one-liner.
Would you like to do that?

niccokunzmann:
merging as the tests work now

niccokunzmann:
Thanks again for this PR!

PR #406, opened 2022-09-11
Test restructure issue 70

jacadzaca:
splitting as per #400


src/icalendar/tests/test_parsing.py
```diff
@@ -1,3 +1,13 @@
+'''Tests checking that parsing works'''
+from icalendar import vRecur
+
+def test_decode_rrule_attribute_error_issue_70(events):
+    # Issue #70 - e.decode("RRULE") causes Attribute Error
```
niccokunzmann, line 5: Ususally, this would appear in a string - why is that different?

niccokunzmann:
Thanks!

PR #407, opened 2022-09-11
Test restructure index error

jacadzaca:
splitting as per #400


niccokunzmann:
This build failed but it seems it has nothing to do with this commit. However, to keep the master branch green, I will not merge it for now. I looked into it: The only notable difference is that the Sphinx package has a new version. This new version might cause the problem.
Thus, I restart the master branch build to see if it works with the new version of Sphinx. 
Old Sphinx: Sphinx==5.1.1
New Sphinx: Sphinx==5.2.0.post0

niccokunzmann:
#419

niccokunzmann:
re-running tests...

niccokunzmann:
Thanks!

PR #410, opened 2022-09-11
Test restructure issue 104

jacadzaca:
splitting as per #400


angatha (approved):

src/icalendar/tests/test_fixed_issues.py
```diff
@@ -185,54 +185,6 @@ def test_issue_101(self):
...
-DTSTAMP:20140401T000000Z
-SUMMARY:Broken Eevnt
-CLASS:PUBLIC
-STATUS:CONFIRMED
-TRANSP:OPAQUE
-X
-END:VEVENT"""
-        event = icalendar.Calendar.from_ical(ical_str)
-        self.assertTrue(isinstance(event, icalendar.Event))
-        self.assertTrue(event.is_broken)  # REMOVE FOR NEXT MAJOR RELEASE
```
niccokunzmann, line 72: Uh, this is interesting .. a TODO for the next Major release which we are drafting

src/icalendar/tests/test_issue_104_mark_events_broken.py
```diff
@@ -0,0 +1,23 @@
...
+
+@pytest.mark.parametrize('calendar_file',
+        (('issue_104_broken_calendar.ics'),),
+        indirect=True)
+def test_dont_ignore_exceptions_on_broken_calendars_issue_104(calendar_file):
+    '''Issue #104 - line parsing error in a VCALENDAR
+    (which doesn't have ignore_exceptions). Should raise an exception.
+    '''
+    with pytest.raises(ValueError), open(calendar_file) as f:
+        Calendar.from_ical(f.read())
```
niccokunzmann: If you replace this line 23 with `calendars.issue_104_broken_calendar` - it should raise the exception and the parametrization can be deleted.
The refactoring seems clean from my perspective.

Would you like to implement this change?

niccokunzmann:
Hm, it looks like this needs a `calendars` fixture like the `events` fixture.
https://github.com/collective/icalendar/actions/runs/3147557141/jobs/5117173509#step:6:41

src/icalendar/tests/test_issue_104_mark_events_broken.py
jacadzaca: If we do that, the code actually throws a `KeyError`

Since `issue_104_broken_calendar` is broken, the `ValueError` we look for would be raised on the line 22 of conftest and caught on the line 25, so the `calendars` object produced by calling `calendars.issue_104_broken_calendar` wouldn't have the property, therefore the `KeyError` would be caused by the line 27 of conftest :)

niccokunzmann:
@jacadzaca - could you review this and merge if that is ok with you?

jacadzaca:
Looks good for me :+1:

PR #417, opened 2022-09-20
refactor test_issue_178

jacadzaca:
splitting as per #400


src/icalendar/tests/calendars/issue_178_component_with_invalid_name_represented.ics
```diff
@@ -0,0 +1,2 @@
+BEGIN:MYCOMP
```
angatha, line 1: @jacadzaca I try to understand, what you mean by ["RFC states that there can be no custom copoments I think"](https://github.com/collective/icalendar/discussions/374#discussioncomment-3572199) at #374. iana-comp and x-comp are allowed within the VCALENDAR component. Then I would agree that this test case and the MYCOMPTOO one are not that clear since `icalendar.Calendar.from_ical` should parse it as a VCALENDAR (or the doc is misleading since `from_ical(...)` works indeed more like a static method than a class method.). 

On the other hand, being able to parse just parts of an .ics file sounds like a great feature and i would suggest dropping that. And since custom components are within the boundaries of RFC5545, I would keep supporting them, even if parsing them is not as obvious as parsing a calendar.

A last point about these test cases. They may not be a good choice. (And for some seconds, I thought this was the reason for the comment in #374.) MYCOMP, MYCOMPTOO and UNKNOWN are neither x-names nor IANA registered identifiers so these tests test for something that is not required by RFC5545. I suggest that the tests are updated to use x-names so something actually specified is tested. Maybe not in this PR since it is refactoring but I just raised my voice so these observation is not lost. Ignoring this, the refactoring looks really nice!

Some things that may be worth discussing:
1. Add documentation about parsing custom components
2. Split this into two test one for VCALENDAR paring and one for custom components.
3. (using x-names)

jacadzaca: Yeah, I worded myself too vaguely. I was confused by the weird choice of test cases, so I'm all for changing them. I'm not sure what you're trying to say in the second paragraph, is `being able to parse just parts of an .ics` a great feature or should we drop it? I'm not even sure if we can `parse just parts of an .ics` right now, since there's not lazy parsing. Could you please clarify?

angatha: I think it is a great feature. Take the content of `src/icalendar/tests/calendars/issue_178_custom_component_inside_other.ics`. We can still parse this, even if it is not an `icalstream`.

jacadzaca: Maybe file an issue/start a discussion with the points you've raised? I think they're very valid.

niccokunzmann: Yes, issues are good to keep the TODOs recorded.

niccokunzmann:
@jacadzaca @angatha I will basically assign this PR to you two. @jacadzaca has merge ability so when you @angatha agree, you can merge it from my perspective. You can discuss what you like to see as optimal and of cause mention me whenever.
What do you think?

angatha:
I will approve this PR and add issues for the raised questions. Even for the x-name suggestion, a new issue sounds clearer because it separates the refactoring and other adjustments.

angatha (approved):

niccokunzmann:
@angatha could you create an issue and link it here?

PR #418, opened 2022-09-20
refactor test_issue_237

jacadzaca:
splitting as per #400. Any idea why the tests pass in #400 but not here?


niccokunzmann:
So, I see [this](https://github.com/collective/icalendar/actions/runs/3090206021/jobs/4998711463#step:6:79):

```
AssertionError: assert datetime.datetime(2017, 5, 11, 13, 30) == 
                       datetime.datetime(2017, 5, 11, 13, 30, tzinfo=zoneinfo.ZoneInfo(key='America/Sao_Paulo'))
```
That is an okay fail - but why is there no time zone in it?

Hm. Maybe a merge of master changes something?
Why would the event contain a time zone? There should be an error, I think.... The time zone has an unknown identifier.
Also: If it is a test refactoring - where are the tests that get deleted like in all the other PRs?

Maybe that helps.. let's find out together!

jacadzaca:
The TZID property must be [specified before use](https://www.rfc-editor.org/rfc/rfc5545#section-3.2.19), so we need to parse a whole calendar. Now, we're back to the problem [with pytz](https://github.com/collective/icalendar/pull/400#issuecomment-1242119293), to which @niccokunzmann proposed [a solution](https://github.com/collective/icalendar/pull/400#discussion_r967359497) that I'm not sure how to implement

niccokunzmann:
```
expected = datetime.datetime(2017, 5, 11, 13, 30, tzinfo=timezone('America/Sao_Paulo'))
```
I think, looking at the output:
```
assert datetime.datetime(2017, 5, 11, 13, 30, tzinfo=<DstTzInfo '(UTC-03:00) Brasília' Brasília standard-1 day, 21:00:00 STD>)
                     == datetime.datetime(2017, 5, 11, 13, 30, tzinfo=<DstTzInfo 'America/Sao_Paulo' LMT-1 day, 20:54:00 STD>)
                                                                                                                ^^^^^
```
This is the problem one has with pytz: You can not pass a time zone to the datetime. You must call `tz.localize` on the datetime object as far as I remember. There was an article about why to avoid pytz and this is one of the reasons - it does not (because old) obey how one would want to use the tz object nowadays. The offset is false because that was the offset before UTC alignment of the time zones. - all from memory this one is.

Could you try running `localize()` on this dt object?

niccokunzmann:
I think, my comment there is that I would like to see that icalendar can actually choose a tz implementation and I would like to have all tests run across all different tz implementations.

I would pass it to `from_ical(..., timezone=pytz.tzinfo)`
Creating an issue ...

niccokunzmann:
@jacadzaca can you merge master again into this one? I can not merge it, yet and I am too tired to resolve the conflict.

src/icalendar/tests/test_fixed_issues.py
```diff
@@ -42,47 +42,3 @@ def test_issue_116(self):
...
-                    'END:VTIMEZONE',
-                    'BEGIN:VEVENT',
-                    'DTSTART;TZID=\"(UTC-03:00) Brasília\":20170511T133000',
-                    'DTEND;TZID=\"(UTC-03:00) Brasília\":20170511T140000',
-                    'END:VEVENT',
-                    'END:VCALENDAR',
-                    ]
-
-        cal = icalendar.Calendar.from_ical('\r\n'.join(ical_str))
-        self.assertEqual(cal.errors, [])
```
niccokunzmann, line 75: We are missing this assertion. Apart from that: I trust that you copied it alright and the other assertions seem transferred.

niccokunzmann (changes requested):

src/icalendar/tests/test_fixed_issues.py
jacadzaca: I readded the assertion in test_parses_event_with_non_ascii_tzid_issue_237

niccokunzmann (approved):

niccokunzmann:
@jacadzaca Thanks for wroking together on this one!

PR #422, opened 2022-09-26
Revert "fix Sphinx build"

### LINKED ISSUE #419: docs build fails on Sphinx==5.2.0.post0
Currently, the build of the docs fails.
https://github.com/collective/icalendar/actions/runs/3119899747/jobs/5060141897
The only difference from before is that the Sphinx version changed.

Possible Solution:
- [x] I would introduce version pinning for that matter. - #420
- [x] This needs to be reported to Sphinx/read-the-docs-theme - https://github.com/readthedocs/sphinx_rtd_theme/issues/1343

niccokunzmann:
This reverts commit f4b50d6b9621f71f5fb1c128dd010789093603e8.

This is to check if the usual, non-pinned installation works like it did before.

fixes #419

### DIFF
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -1,4 +1,3 @@
-Sphinx>=1.2.3 
-Sphinx<=5.1.1
+Sphinx>=1.2.3
 sphinx_rtd_theme
 .


niccokunzmann:
@jacadzaca would you like to review & merge?

PR #356, opened 2022-06-15
Fix #355 url escaping

NicoHood:
@niccokunzmann

See #355

Issue #355, opened 2022-06-12
Unwanted unescaping of http url strings


NicoHood:
Hi!
I have an url in my ical event description that is already html encoded. Here is an example:

```
https://www.facebook.com/events/756119502186737/?acontext=%7B%22source%22%3A5%2C%22action_history%22%3A[%7B%22surface%22%3A%22page%22%2C%22mechanism%22%3A%22main_list%22%2C%22extra_data%22%3A%22%5C%22[]%5C%22%22%7D]%2C%22has_source%22%3Atrue%7D
```

This is how it looks in the ical file:
```
DESCRIPTION:https://www.facebook.com/events/7561195021867
 37/?acontext=%7B%22source%22%3A5%2C%22action_history%22%3A[%7B%22surface%22
 %3A%22page%22%2C%22mechanism%22%3A%22main_list%22%2C%22extra_data%22%3A%22%
 5C%22[]%5C%22%22%7D]%2C%22has_source%22%3Atrue%7D
```

But some characters now get unescaped for some unknown reason:
```
https://www.facebook.com/events/756119502186737/?acontext=%7B%22source%22:5,%22action_history%22:[%7B%22surface%22:%22page%22,%22mechanism%22:%22main_list%22,%22extra_data%22:%22\%22[]\%22%22%7D],%22has_source%22:true%7D
```

It turns out, that this code causes the issue:
https://github.com/collective/icalendar/blob/master/src/icalendar/parser.py#L273

It converts `%3A` to `:` which in my case is NOT wanted. The url is broken then.

Why was this html unescape introduced and how can we fix that?

NicoHood:
Wouldn't it make more sense to escape them with the character function, with backslashes instead of html codes?
https://github.com/collective/icalendar/blob/master/src/icalendar/parser.py#L25

NicoHood:
I see what the issue is: `escape_string` is used as a workaround to hide `\:` characters and only search for "real" : splitting characters. However, what nobody thought of was, that this will modify the final output, if there already were escaped characters. This solution is quite of wonky and error prone and should be fixed. The splitting should be parsed more smart, and not by escaping/unescaping which will always lead to errors, if implemented like that.

NicoHood:
This is my **untested** recoding of this function. It completely removed the escaping functions, as they are not required anymore.

What I am not sure about is, if inside quotes, backslashes should also quote the quote itself. Example:
`"this text is quote and it even contains a \" quote mark"`

[...]

PR #426, opened 2022-10-03
add tests for issue #355

niccokunzmann:
see https://github.com/collective/icalendar/pull/356
see https://github.com/collective/icalendar/issues/35

@NicoHood - I created this PR to make sure the tests work.

### DIFF
--- a/src/icalendar/tests/test_issue_355_url_escaping.py
+++ b/src/icalendar/tests/test_issue_355_url_escaping.py
@@ -1,0 +1,11 @@
+"""Tests for Issue 355.
+
+see https://github.com/collective/icalendar/issues/355
+"""
+
+def test_facebook_link_is_correctly_parsed(events):
+    """The facebook link must not be damaged.
+
+    see https://github.com/collective/icalendar/pull/356#issuecomment-1222626128
+    """
+    events.issue_355_url_escaping["DESCRIPTION"] == "https://www.facebook.com/events/756119502186737/?acontext=%7B%22source%22%3A5%2C%22action_history%22%3A[%7B%22surface%22%3A%22page%22%2C%22mechanism%22%3A%22main_list%22%2C%22extra_data%22%3A%22%5C%22[]%5C%22%22%7D]%2C%22has_source%22%3Atrue%7D"
--- a/src/icalendar/tests/events/issue_355_url_escaping.ics
+++ b/src/icalendar/tests/events/issue_355_url_escaping.ics
@@ -1,0 +1,8 @@
+BEGIN:VEVENT
+DTSTAMP:20220822T164528Z
+UID:1661186704812-62650@ical.marudot.com
+DTSTART;TZID=Europe/Berlin:20220822T120000
+DTEND;TZID=Europe/Berlin:20220822T120000
+SUMMARY:test
+DESCRIPTION:https://www.facebook.com/events/756119502186737/?acontext=%7B%22source%22%3A5%2C%22action_history%22%3A[%7B%22surface%22%3A%22page%22%2C%22mechanism%22%3A%22main_list%22%2C%22extra_data%22%3A%22%5C%22[]%5C%22%22%7D]%2C%22has_source%22%3Atrue%7D
+END:VEVENT


src/icalendar/tests/test_issue_355_url_escaping.py
```diff
@@ -0,0 +1,11 @@
...
+
+see https://github.com/collective/icalendar/issues/355
+"""
+
+def test_facebook_link_is_correctly_parsed(events):
+    """The facebook link must not be damaged.
+
+    see https://github.com/collective/icalendar/pull/356#issuecomment-1222626128
+    """
+    events.issue_355_url_escaping["DESCRIPTION"] == "https://www.facebook.com/events/756119502186737/?acontext=%7B%22source%22%3A5%2C%22action_history%22%3A[%7B%22surface%22%3A%22page%22%2C%22mechanism%22%3A%22main_list%22%2C%22extra_data%22%3A%22%5C%22[]%5C%22%22%7D]%2C%22has_source%22%3Atrue%7D"
```
NicoHood: Maybe a return is missing here?

NicoHood: or an assert?

niccokunzmann: yep... well spotted! That was a mistake!

niccokunzmann:
@NicoHood I cherry-picked your commit from #356 and there is a test failing for a content line.
```
    def test_escaping(self):
        # verify that escaped non safe chars are decoded correctly
        NON_SAFE_CHARS = ',\\;:'
        for char in NON_SAFE_CHARS:
            cn_escaped = "Society\\%s 2014" % char
            cn_decoded = "Society%s 2014" % char
            vevent = Event.from_ical(
                'BEGIN:VEVENT\r\n'
                'ORGANIZER;CN=%s:that\r\n'
                'END:VEVENT\r\n' % cn_escaped
            )
>           self.assertEqual(vevent['ORGANIZER'].params['CN'], cn_decoded)
E           AssertionError: ['Society\\', ' 2014'] != 'Society, 2014'

```

It looks like a string gets split into parts if it has a space....

src/icalendar/tests/test_issue_355_url_escaping.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+
+def test_facebook_link_is_correctly_parsed(events):
+    """The facebook link must not be damaged.
+
+    see https://github.com/collective/icalendar/pull/356#issuecomment-1222626128
+    """
+    assert events.issue_355_url_escaping["DESCRIPTION"] == "https://www.facebook.com/events/756119502186737/?acontext=%7B%22source%22%3A5%2C%22action_history%22%3A[%7B%22surface%22%3A%22page%22%2C%22mechanism%22%3A%22main_list%22%2C%22extra_data%22%3A%22%5C%22[]%5C%22%22%7D]%2C%22has_source%22%3Atrue%7D"
+
+def test_other_facebook_link_is_correctly_parsed(events):
```
NicoHood, line 13: what is this test used for? It is completely identical to the first one!?

src/icalendar/tests/events/issue_355_url_escaping_2.ics
```diff
@@ -0,0 +1,8 @@
+BEGIN:VEVENT
```
NicoHood, line 1: why was this file added? It seems completely identical to your test file

NicoHood:
The string does not get split into parts if it has a space. You can clearly see the space in your result. It is split, when a backslash is being escaped. I currently try to understand why, but the test itself looks like a horrible syntax. Can we please fix it and make it use python fstrings?

NicoHood:
Alright, so I can see what is happening here.

This is the failing test:
https://github.com/collective/icalendar/blob/de7a96addfbae1a21524d096302b95f36a69ed13/src/icalendar/tests/test_property_params.py#L74

And this is the function that splits up our string:
https://github.com/collective/icalendar/blob/832a16639b0df854e269b10b76ab3c2a7f15cccf/src/icalendar/parser.py#L227

It will recognize `CN=Society\, 2014` as the input string and split it accordingly by the comma. That is how multiple parameters should be split.

The RFC section that describes this process is this one:
https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html

It seems that I have not implemented it correctly and for me it is still a brainfuck to understand how it should be implemented. Which strings can be quoted and which not. I think this needs more detailed parsing, but we can get to that.

NicoHood:
So I looked at the mentioned spec and I could not find any part, where escaping of those characters in the test are described. I am wondering where this was defined, in which spec? Otherwise I am unable to fix anything here.

This is what I understood from the spec:
* The name can only contain alphanumeric characters
* Multiple parameters can follow
* Those parameters can be quoted with double quotes (case sensitive) or without quotes (case insensitive)
* No matter what the parameters cannot single escape comma, backslash, etc. by using backslashes. Only with double quotes they can be used.
* After the parameters follows the value

It seems in my implementation I made the mistake to quote those characters by using a backslash.

But the test also looks incorrect. If those parameter value are not quoted (which they are not) they should be split by a comma. And that is what the code does now.

I have to think about this some more, but there seems to be a lot wrong. Before my PR but also with. We need to document the code much more, it is totally confusing and not understandable.

NicoHood:
Maybe we should just take the abnf and convert it to a regex? It might be a bad idea, I dont know yet, but the abnf itself should be as close to the spec as we can get. https://github.com/aas-core-works/abnf-to-regexp

src/icalendar/tests/test_issue_355_url_escaping.py
niccokunzmann: `_2` - there is the second example

niccokunzmann:
Hm... Which interface do we have to fulfill for parsing these parameters?
Would you like to help create a list of examples? Then, we can feed them into a test function... If the test is wrong in testing what it does, we can change it but I would like to be more precise than the test before.

> This is what I understood from the spec:

Can you link that here?

Here is a list of tests that can be written and parametrized with different values to make sure it is working.

- The name can only contain alphanumeric characters
    - [ ] test valid names
    - [ ] test invalid characters (hypothesis)
-   Multiple parameters can follow
    - [ ] test 0
    - [ ] test 1
    - [ ] test 2
    - [ ] test with broken one in the middle
-    Those parameters can be quoted with double quotes (case sensitive)
    - [ ] test case sensitivity
    - [ ] test conversion to ical again to have quotes
- without quotes (case insensitive)
    - [ ] test case insensitivity
    - [ ] test conversion to ical again to have no quotes
- [ ] (L1) No matter what the parameters cannot single escape comma, backslash, etc. by using backslashes.
- [ ] (L2) Only with double quotes they can be used.
-    After the parameters follows the value
    - [ ] empty value, other values

(1) One proposal would be that I write tests and you can use TDD to work a bit more on the code.

(2) Another proposal is to just fix this little test to be accurate: Is the problem that the current implementation does not distinct between (L1) and (L2) above?

NicoHood:
Yeah, if you can write tests first, that would help me a lot. This way we can talk clearer about the spec and how it should look like.

niccokunzmann:
I think, I need the examples mentioned. From what you write, I guess that writing tests is out of what you usually do or out of the comfort zone. The code that you wrote however is much more complicated than writing the test case for it. Of you give it a try, I am happy to guide you. The next steps seem to be gathering examples. Very likely they are already mostly implemented. Also, if the test case is wrong, we should change it. Do you think it is wrong? It seems so as it has no " in it. So, what would be your course of action here?

NicoHood:
I will have a look at it this week. I will let you know of my progress

PR #430, opened 2022-10-03
Document maintenance

niccokunzmann:
This is a contribution towards the clarity of maintaining and releasing.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,4 +12,5 @@ Contents
     api
     cli
     credits
+    maintenance
     license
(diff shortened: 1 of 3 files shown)


jacadzaca:
https://github.com/orgs/collective/teams/icalendar-admin and https://github.com/orgs/collective/teams/icalendar-contributor give a 404 as of now, is that expected?

niccokunzmann:
@jacadzaca I guess that is because they are only visible to admins or organization people.

PR #438, opened 2022-10-07
Refactor test_encoding create_from_ical

jacadzaca:
Restructure `create_from_ical` from `test_encoding` along the lines of https://github.com/collective/icalendar/discussions/374


jacadzaca:
Merging #439 first may lead to less merge conflicts

src/icalendar/tests/test_encoding.py
```diff
@@ -7,30 +7,32 @@
...
+
+@pytest.mark.parametrize('test_input, field, expected_value', [
+    ('event_with_unicode_fields', 'SUMMARY', 'Non-ASCII Test: ÄÖÜ äöü €'),
+    ('event_with_unicode_fields', 'DESCRIPTION', 'icalendar should be able to handle non-ascii: €äüöÄÜÖ.'),
+    ('event_with_unicode_fields', 'LOCATION', 'Tribstrül'),
+    # Non-unicode characters in summary
+    # https://github.com/collective/icalendar/issues/64
+    ('issue_64_event_with_non_unicode_summary', 'SUMMARY', 'åäö'),
+    # Unicode characters in summary
+    ('issue_64_event_with_unicode_summary', 'SUMMARY', 'abcdef'),
```
angatha: I noticed that `src/icalendar/tests/events/issue_64_event_with_non_unicode_summary.ics` contains (as the name suggest) not unicode characters but this tests for `'åäö'` and then the unicode test only checks ascii characters. This should be the other way around? Why are the tests not failing?

jacadzaca: I think by `non-unicode` I've meant `non-ascii`, so I'll fix that.

src/icalendar/tests/test_encoding.py
```diff
@@ -1,51 +1,26 @@
...
-
-    https://github.com/collective/icalendar/issues/64
-    https://github.com/collective/icalendar/issues/101
-    '''
-    event = getattr(events, event_name)
-    assert event.to_ical() == event.raw_ics
-
+def test_event_from_ical_respects_unicode(test_input, field, expected_value, events):
+    event = events[test_input]
+    event[field].to_ical().decode('utf-8') == expected_value
```
angatha: This needs an `assert`.

src/icalendar/tests/test_encoding.py
angatha: The test where actually wrong. `issue_64_event_with_non_ascii_summary.ics` was called `issue_64_event_with_unicode_summary.ics` and contained
```
BEGIN:VEVENT
SUMMARY:åäö
END:VEVENT
```

and originally the test was `('issue_64_event_with_unicode_summary', 'SUMMARY', 'abcdef'),`, where summary clearly didn't match. I looked into the action log to figure out, why the action wasn't failing. After pulling the code myself to debug it, I noticed that they where indeed not equal, but you forgot to assert the equality.

Can you please add the assert for this test.

angatha (approved):

niccokunzmann:
Thanks for this one @jacadzaca  and @angatha  for the review! It is nice to see that you are bringing the project forward without the maintainers being needed and I liked seeing the review process - that builds trust in me that the changes have a conversation and more eyes looking at them :)

Issue #425, opened 2022-09-30
help getting values from ics file


C-Powers:
Hi there, I'm using this package to parse a file. I looked in the docs, but there doesn't seem to be much documentation on how to do this. This is what I've got, and it kinda works, but I'm not getting the native python types out of it. 

I have vText to string, vDDDTypes to datetime, and vCalAddress to string that I need to parse.

Thanks for the help!

As an example:

```
gcal = Calendar.from_ical(content)
out: dict[str, Union[str, datetime]] = {}
for component in gcal.walk():
    if component.name == "VEVENT":
        print(component.get("UID"), type(component.get("UID")))
        print(
            vText.from_ical(component.get("UID")),
            type(vText.from_ical(component.get("UID"))),
        )
        

"""
prints:
[radact]@google.com <class 'icalendar.prop.vText'>
[radact]@google.com <class 'icalendar.prop.vText'>
"""
```

C-Powers:
ah, I think it's `to_ical` instead of `from_ical`.

C-Powers:
Sorry for the false alarm, I had to rubber duck myself, but this may be nice for someone in the future. 

```
gcal = Calendar.from_ical(content)
    out: dict[str, Any] = {}
    for component in gcal.walk():
        if component.name == "VEVENT":
            out["id"] = vText.to_ical(component.get("UID")).decode("utf-8")

            out["start"] = vDDDTypes.to_ical(component.get("DTSTART")).decode("utf-8")  # type: ignore
            out["end"] = vDDDTypes.to_ical(component.get("DTEND")).decode("utf-8")  # type: ignore

            out["summary"] = vText.to_ical(component.get("SUMMARY")).decode("utf-8")
            out["description"] = vText.to_ical(component.get("DESCRIPTION")).decode(
                "utf-8"
            )
            out["location"] = vText.to_ical(component.get("LOCATION")).decode("utf-8")
            attendees = []
            for a in component.get("ATTENDEE"):
                attendees.append(vCalAddress.to_ical(a).decode("utf-8"))
            out["attendees"] = attendees

            out["google_meet_url"] = vText.to_ical(
                component.get("X-GOOGLE-CONFERENCE")
            ).decode("utf-8")
            ```

abe-101:
@C-Powers Can you provide a example ical file that you parsed with the above script?

I'd like to add it as an example to the docs

Issue #421, opened 2022-09-26
Problem with Event rrule format


tyctor:
trying to create recuring event, but export to ical fails:

```
event = icalendar.Event()
event.add("rrule", {'FREQ': 'MONTHLY', 'WKST': 'MO', 'BYDAY': '+1TU,+3TU'})
event.to_ical()

...
ValueError: Expected weekday abbrevation, got: +1TU,+3TU
```
but icalendar validator on https://icalendar.org/validator.html claims that this rrule is valid

niccokunzmann:
Thanks for reporting this! You would probably expect the event to be created. What would the valid rrule string or ics string look like?

tyctor:
hi, 
valid rrule string should be
`RRULE:FREQ=MONTHLY;WKST=MO;BYDAY=+1TU,+3TU`

in words:
`Every month on the 1st Tuesday and 3rd Tuesday`

tyctor:
oh, understand, value must be list 

this is ok:
```
event = icalendar.Event()
event.add("rrule", {'FREQ': 'WEEKLY', 'WKST': 'MO', 'BYDAY': ['+1TU', '+3TU']})
event.to_ical()

b'BEGIN:VEVENT\r\nRRULE:FREQ=WEEKLY;BYDAY=+1TU,+3TU;WKST=MO\r\nEND:VEVENT\r\n'
```

Issue #341, opened 2021-11-30
 how to automatically create event on google calendar


paulorobsant:
I would like to know if there is any way for me to create an event within google calendar without the need for the user to enter the email and accept the event I sent.

jacadzaca:
This issue will be closed as it's out of scope of this module. I think that the functionality can be achieved by __using__ icalendar, so  if still need a hand implementing it, you can start a [disccusion here](https://github.com/collective/icalendar/discussions) or ask for help on some other programming forum. If you choose to start a discussion, please link it in a comment, so people facing a similar problem have easier time finding it.

abe-101:
Consider looking into the google calendar [API](https://developers.google.com/calendar/api/guides/overview) .

Issue #317, opened 2020-12-30
how convert an ics to md, if (many) vevents don't have category


DoctorSubtilis:
I get an error message when I try to convert an ics calendar, whose (many) vevents are without category.   
The message is  

      File "/usr/local/lib/python3.8/dist-packages/ics/component.py", line 54, in _populate
      raise ValueError(
     ValueError: A VEVENT must have at most one CATEGORIES

I use a Python 3 script to convert ics to md:

    from ics import Calendar

    md_template = """
    ### {start}
    *{name}*: {desc}""" 

    with open("diario.ics") as f:
       cal = Calendar(f.read())

    print("## Diario")
    for e in sorted(cal.events):
    print(md_template.format(start=e.begin.strftime("%Y-%m-%e"), name=e.name, 
                             desc=e.description))

There is a workaround to avoid this behavior, without add manually a category to every vevent (quite boring)?

jacadzaca:
I presume you wanted to achieve something like this...

```python
from icalendar import Calendar

md_template = """
### {start}
*{name}*: {desc}"""

with open("sample.ics") as f:
   cal = Calendar.from_ical(f.read())

print("## Diario")
for event in cal.walk('vevent'):
    print(md_template.format(start=e.decoded('dtstart').strftime("%Y-%m-%e"),
                             name=event.get('summary'),
                             desc=event.get('description')))
```

PR #439, opened 2022-10-07
refactor test_create_to_ical from test_encoding to test_examples.py

jacadzaca:
Refactor `test_create_to_ical`, `test_create_event_simple` and `test_unicode_parameter_name` from test_encoding along the lines of https://github.com/collective/icalendar/discussions/374, move them to `test_examples.py`.


src/icalendar/tests/test_examples.py
```diff
@@ -0,0 +1,41 @@
+'''tests ensuring that *the* way of doing things works'''
```
niccokunzmann, line 1: The docstring does not explain anything.

niccokunzmann (commented):
I do not know what to do with it. I would need to reconnect to the value of refactoring this. It seems to be an ok refactoring. It can be merged as I have nothing against it except from the docstring. I mean in short: it does not hurt and I have nothing against it.

jacadzaca:
I think creating a single file with tests that showcase practical icalendar usage is important, so as to expand the [examples section](https://icalendar.readthedocs.io/en/latest/usage.html#example).  We know that documentation is somewhat lacking in examples, because there have been issues where people ask how to do X, e.g. #317, #341, #317, #421, #425.

niccokunzmann:
@jacadzaca, could you create an issue for this? I would find it nice if

1. The documentation with the examples shows up on readthedocs
2. That the code is actually running.

As a suggestion, one could use [doctest](https://docs.python.org/3/library/doctest.html)
I do think that this issue is a very good first-timer issue because of its repetitiveness, the simplicity and the range of skills you build up while contributing to it. I would really love to see it. If you do not like to make this issue, I will do it.

What do you think?

jacadzaca:
I'm off to bed now, but I'll gladly file it tomorrow :)

jacadzaca:
@niccokunzmann is #443 OK?

niccokunzmann:
I commented. Could I edit it to make it more beginner friendly?

niccokunzmann:
So, yes, thanks that is quite a good start. I could not have listed all the issues.

jacadzaca:
Feel free to edit it :)

PR #440, opened 2022-10-07
restructure test_issue_82

jacadzaca:
splitting as per #400, I forgot about this one


src/icalendar/tests/test_parsing.py
```diff
@@ -133,3 +134,23 @@ def test_no_tzid_when_utc(utc, date, expected_output):
...
+def test_creates_event_with_base64_encoded_attachment_issue_82(events):
+    '''Issue #82 - vBinary __repr__ called rather than to_ical from
+                   container types
+    https://github.com/collective/icalendar/issues/82
+    '''
+    b = vBinary('text')
+    b.params['FMTTYPE'] = 'text/plain'
+    event = Event()
+    event.add('ATTACH', b)
+    assert event.to_ical() == events.issue_82_expected_output.raw_ics
```
niccokunzmann, line 156: I think, we could double these kind of tests:
to_ical should be the raw one and parsing the raw one should lead to an object that is equal.

niccokunzmann:
Thanks! It is nice to work through all the tests to get an overview over what is happening in the module.

src/icalendar/tests/test_parsing.py
niccokunzmann: I tried my idea and it can be ignored. It did not work out as useful.

Issue #443, opened 2022-10-09
Enriching the examples section on readthedocs


jacadzaca:
The project can be improved with useful examples. The `usage.rst` file contains examples already. However, more examples will improve the clarity. Please use one of the issues mentioned below to extract an example and make a pull request.
Issues/tests that contain possible examples:
- [ ] #85 
- [x] #210 
- [x] #267 
- [x] #301 
- [ ] #317 
- [ ] #341 
- [ ] #342 
- [ ] #421 
- [ ] #425 
- [ ] https://github.com/collective/icalendar/pull/390 - generating RFC compliant TIMEZONE components
- [ ] #130 
- [ ] test_property_params and test_parse_and_access_property_params from [test_property_params.py](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_property_params.py)
- [ ] test_tools_UIDGenerator from [test_unit_tools](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_unit_tools)
- [ ] test_issue_116 from [test_fixed_issues](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_fixed_issues.py)
- [x] #510

## See Also
- This issues arises from #439.
- [icalendar examples on readthedocs](https://icalendar.readthedocs.io/en/latest/usage.html#example).
- When #445 is merged, the examples will be tested with `doctest` so we can make sure that they are actually working.
- #626

niccokunzmann:
I think, adding the examples to a file for them or creating a file for them then allows running Python's doctest on them. So anyone who creates a pull request with an example of how to use icalendar will receive help so that the code is run by the test cases.

I do not have an overview over different kinds of examples that are already there. We do not want to duplicate code if not necessary and also make sure that the existing examples get tested and are where a beginner expects to see them.

I would like to see examples in the future also of how to read icalendar files and how to use the command line.

abe-101:
Having used this library for a recent project of mine I feel the need for this initiative :smile: 
and I'm happy to help out

abe-101:
Is there a specific strategy or structure we want with the examples in the docs?
Or do we want to add any useful (but random) example?

stevepiercy:
I audited the checklist, and moved remaining items by editing the checklist in #626. Closing.

PR #445, opened 2022-10-10
test .py and .rst files with doctest

niccokunzmann:
This runs doctest on .py and rst files.
See https://github.com/collective/icalendar/issues/443

The tests should be failing at the moment.
To fix a doctest, please create a pull request to the `doctest` branch.

In contrary to #443, this runs doctest as part of the tests to make sure that the documentation is valid for all Python versions.


niccokunzmann:
If this is open for another 4 days, I would like to merge it to have progress.

ale-rt:
It looks quite good to me but the branch needs a rebase.

tox.ini
```diff
@@ -10,6 +10,7 @@ usedevelop=True
 deps =
     pytest
     coverage
+    hypothesis
```
ale-rt, line 13: I think this one should do in the `tests_require` list that you can find in the `setup.py` file: https://github.com/collective/icalendar/blob/68dc8b18caf0806d2fa1a8e0f06027329ad96c4c/setup.py#L20

Then the `requirements_docs.txt` file should install `. [test]` (untested)

niccokunzmann: Thanks for your reply! Hm I do not know why hypothesis is even in there.

Here, it uses `use_debelop=True` and therewith does a `pip install .`.
We do have different tests running that have different dependencies. Using use-develop is what we do here. I think, changing this would go into an other PR as it is independent of doctest. If you think, that change is worth experimenting with, you can create a PR or an issue, I would say. I do not really see a difference at the moment.

What do you think?

niccokunzmann:
Thanks for the review @ale-rt.

Issue #446, opened 2022-10-11
Easy Deployments


niccokunzmann:
For my projects, I like it if I create a tag and then this tag gets deployed automatically after all the tests were run.

![image](https://user-images.githubusercontent.com/564768/195097037-5024676f-d601-4ba9-b33b-4d8b860749fc.png)

Here is a successful deployment:
https://github.com/niccokunzmann/python-recurring-ical-events/actions/runs/3227154888/jobs/5281596597

I do like this method because
- it allows for automated testing of the package before deploy
- it speeds up new releases and this makes it easier to integrate new, small changes

The flow would look like this:

1. Anyone with write access follows the [deployment documentation](https://icalendar.readthedocs.io/en/latest/maintenance.html#new-releases) to create a new tag.
2. Maintainers are notified for approval of this tag.
3. If one maintainer approves, the deployment goes to PyPI

@mauritsvanrees Since you run the tests for the Plone community each time before a release, are there requirements that should be taken into account for deployments, more tests to run?

What are your thoughts on this?

Related:
- #429

mauritsvanrees:
It *would* be good if the `plone.app.event` tests would be run against the development version of this package. This is the only package that uses `calendar`. I am making a PR.

niccokunzmann:
This can be closed when
- [ ] #447 is merged
- [ ] #448 is merged
- [ ] action secrets for the `PyPI` environment have been added
- [ ] it is documented that
    - maintainers need to approve the deploy
    - how maintainers are added to the environment
- [ ] a new release has been made and is on PyPI using this method, see #429

niccokunzmann:
This can be closed as all closing criteria are fulfilled.

PR #447, opened 2022-10-11
Run the plone.app.event tests to see if our dev version causes problems

mauritsvanrees:
See https://github.com/collective/icalendar/issues/446

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,7 @@ jobs:
         - ["3.10",  "py310"]
         - ["pypy-3.9", "pypy3"]
         - ["3.10",  "docs"]
+        - ["3.10",  "plone"]
         - ["3.11.0-rc.1",  "py311"]
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Tests are green.

Could someone double check if this is good from the Plone side? Maybe @ale-rt or @gforcada ? Note that I already have a PR to [remove `icalendar` from the Plone core dev tests](https://github.com/plone/buildout.coredev/pull/817).

niccokunzmann (approved):

niccokunzmann:
@mauritsvanrees, it is fine to merge with me. We can revert it or improve it once you have feedback from the Plone community.
Also: I wonder: Is the master there ever red? What is would be the policy if these tests are red? Do not ship & do not merge?

mauritsvanrees:
I think the plone.app.event tests that we are running here are quite stable. If you see it go red when it is highly unlikely to be a problem in icalendar, you can ping me.

I would say: if the check here is red, and you merge, then the checks on master will be red. So best not merge. But at least I would say, do not release then.

Issue #314, opened 2020-07-11
AssertionError: VTIMEZONEs sub-components' DTSTART must be of type datetime, not date


jhwright:
I hope this is an ignorant newb question, but when I run this code in Python 3.75, vscode:
   
```from icalendar import Calendar
    import requests
    SkeddaBridgeURL = 'https://bridgeartspace.skedda.com/ical?key=457bda7'
    f=requests.get(SkeddaBridgeURL)
    c = Calendar.from_ical(f.text)
```
I get this error:

```Exception has occurred: AssertionError
VTIMEZONEs sub-components' DTSTART must be of type datetime, not date
  File "/Users/jeff/Scrape/Code/skedda.py", line 6, in <module>
    c = Calendar.from_ical(f.text)
```
But when I run this URL through the ical validator at https://icalendar.org/validator.html, I get no errors.

I don't feel qualified to call this a bug, but maybe it is!

niccokunzmann:
@jhwright I created a pull request to reproduce the behavior but I could not reproduce it. Also when I look at the ics file:
https://github.com/collective/icalendar/pull/451/files#diff-210a5b8e4260d128b739ad607c14af0e279aa9ef7da36dc2b75f737608c9d6c6R7
I can see that they are now using datetime.

jhwright:
@niccokunzmann im not using this code any more. Thanks for looking.

PR #451, opened 2022-10-13
add icalendar file from issue #314 and parse it

niccokunzmann:
This is a pull request to test and fix https://github.com/collective/icalendar/issues/314.


jacadzaca:
I don't find a test without an assertion correct.

niccokunzmann:
Maybe we just close this and the issue.

niccokunzmann:
The tests show that the example cannot be reproduced any more.

But, it might be worth checking that the standard does not allow DATE as type. Then, this PR can be modified.

niccokunzmann:
Well, that might be an issue in the future. I think, we can close this and the issue as the faulty implementation removed cause of the error. Thanks for the review!

Issue #156, opened 2015-01-30
Better support for RDATE with VALUE=PERIOD


geier:
We currently do not support RDATE;VALUE=PERIOD: properly. Below are some examples more or less taken from the RFC.

```
In [1]: import icalendar

In [2]: event_str = """BEGIN:VEVENT
SUMMARY:RDATE period
DTSTART:19961230T020000Z
DTEND:19961230T060000Z
UID:rdate_period
RDATE;VALUE=PERIOD:19970101T180000Z/19970102T070000Z
END:VEVENT
"""

In [3]: event = icalendar.Event.from_ical(event_str)

In [4]: print(event.to_ical())
BEGIN:VEVENT
SUMMARY:RDATE period
DTSTART:19961230T020000Z
DTEND:19961230T060000Z
UID:rdate_period
RDATE;VALUE=PERIOD:19970101T180000Z
END:VEVENT
```

another example:

```
In [5]: event_str = """BEGIN:VEVENT
   ...: SUMMARY:RDATE period
   ...: DTSTART:19961230T020000Z
   ...: DTEND:19961230T060000Z
   ...: UID:rdate_period
   ...: RDATE;VALUE=PERIOD:19970101T180000Z/19970102T070000Z,199709T180000Z/PT5H30M
   ...: END:VEVENT
   ...: """

In [6]: event = icalendar.Event.from_ical(event_str)

In [7]: 

In [7]: print(event.to_ical())
BEGIN:VEVENT
SUMMARY:RDATE period
DTSTART:19961230T020000Z
DTEND:19961230T060000Z
UID:rdate_period
RDATE;VALUE=PERIOD:19970101T180000Z,19970109T180000Z
END:VEVENT
```

niccokunzmann:
I added #456 to test that the examples are parsed and validly put back into ICS.
To close this issue, we also need to make sure that the values stored in the RDATE are valid

niccokunzmann:
@geier, there is an error on one of the examples: 
```
199709T180000Z/PT5H30M # given
1997__09T180000Z/PT5H30M # month expected
```

geier:
thanks for taking care of this.

PR #456, opened 2022-10-13
add tests for RDATE with PERIOD value

niccokunzmann:
contributes to https://github.com/collective/icalendar/issues/156

This tests that the RDATE of type PERIOD is put back into ICS


niccokunzmann:
For some reason, the duration looses its minutes:
https://github.com/collective/icalendar/actions/runs/3244748413/jobs/5321333702#step:6:60

PR #462, opened 2022-10-18
fix setuptools deprecation warnings

mgorny:
Fixes two setuptools deprecation warnings:

1. "The license_file parameter is deprecated, use license_files instead."
2. "Installing 'icalendar.tests.calendars' as data is deprecated, please list it in `packages`." and alike

The second one I've fixed the "easier" way by using `find_namespace_packages()`. If you prefer, I can change that to adding `__init__.py` files to all icalendar/tests subdirectories.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ Changelog
 
 Minor changes:
 
-- ...
+- fixed setuptools deprecation warnings [mgorny]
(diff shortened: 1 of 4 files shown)


niccokunzmann:
Thanks for the PR!

mgorny:
Thanks!

PR #463, opened 2022-10-19
Do not create a universal wheel for the release.

mauritsvanrees:
Very minor thing: I see that release [5.0.0 has a universal wheel](https://pypi.org/project/icalendar/5.0.0/#files), but the package is Python 3 only.
And `zest.releaser.create-wheel` is not needed, since we already have a `bdist_wheel` section.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,10 +7,9 @@ ignore =
 
 [zest.releaser]
 python-file-with-version = src/icalendar/__init__.py
-create-wheel = yes
 
 [bdist_wheel]
-universal = 1
+universal = 0
 
 [metadata]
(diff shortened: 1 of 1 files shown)


niccokunzmann:
@mauritsvanrees did you consider this line?

https://github.com/collective/icalendar/blob/252864124208d5723e240643e94aa0bededdc430/.github/workflows/tests.yml#L98

niccokunzmann:
I am merging this, a changelog entry is a TODO..

mauritsvanrees:
This works, in release 5.0.1 we have a correct wheel: https://pypi.org/project/icalendar/5.0.1/#files

PR #471, opened 2022-10-23
Fix 464

### LINKED ISSUE #464: [BUG] RDATE can become None if value is invalid
## Describe the bug

The RDATE is None if it has an invalid PERIOD value in a list and that has an effect on the to_ical fundtion: None is in the ical, making it invalid RFC.

## To Reproduce

```
>>>> import icalendar
>>>> s = b"""BEGIN:VEVENT
.... SUMMARY:RDATE period
.... DTSTART:19961230T020000Z
.... DTEND:19961230T060000Z
.... UID:rdate_period
.... RDATE;VALUE=PERIOD:19970101T180000Z/19970102T070000Z,199709T180000Z/PT5H30M
.... END:VEVENT"""
>>>> event = icalendar.Event.from_ical(s)
>>>> print(event.to_ical().decode())
BEGIN:VEVENT
SUMMARY:RDATE period
DTSTART:19961230T020000Z
DTEND:19961230T060000Z
UID:rdate_period
RDATE:None                <------- error!
END:VEVENT
```
## Expected behavior**

I would expect
1. That the error is recorded in the event
2. that this is not None as a value

## Environment

 - [x] OS: Ubuntu                              
 - [x] Python version: 3.9           
 - [x] `icalendar` version:  5.0.0    

## Additional context

- [x] I tested it with the latest version `pip3 install https://github.com/collective/icalendar.git`

jacadzaca:
fix #464


angatha (approved):

src/icalendar/tests/events/issue_464_invalid_rdate.ics
```diff
@@ -0,0 +1,7 @@
+BEGIN:VEVENT
+SUMMARY:RDATE period
+DTSTART:19961230T020000Z
+DTEND:19961230T060000Z
+UID:rdate_period
+RDATE;VALUE=PERIOD:19970101T180000Z/19970102T070000Z,199709T180000Z/PT5H30M
```
niccokunzmann, line 6: So, the problem we have here is that one element in this list is actually ok and the other one is not. How can that be reflected?

Is it ok to skip the whole RDATE ... probably is for now. It should be rare to have an invalid rdate.

src/icalendar/tests/test_components_break_on_bad_ics.py
```diff
@@ -18,3 +18,11 @@ def test_dont_ignore_exceptions_on_broken_calendars_issue_104(calendars):
...
     with pytest.raises(ValueError):
         calendars.issue_104_broken_calendar
+
+def test_rdate_dosent_become_none_on_invalid_input_issue_464(events):
+    '''Issue #464 - [BUG] RDATE can become None if value is invalid
+    https://github.com/collective/icalendar/issues/464
+    '''
+    assert events.issue_464_invalid_rdate.is_broken
+    assert ('RDATE', 'Expected period format, got: 199709T180000Z/PT5H30M') in events.issue_464_invalid_rdate.errors
+    assert not b'RDATE:None' in events.issue_464_invalid_rdate.to_ical()
```
niccokunzmann, line 28: Cool! The test is alright.

niccokunzmann (approved):

niccokunzmann:
Thanks!

PR #473, opened 2022-10-29
Add max error length

### LINKED ISSUE #472: [BUG] Calendar.from_ical() issue
Hello, 
When I try to import an ics file (28Mb) with the following code, the calendar is displayed at the screen without any print() command.

## To Reproduce
```
import icalendar
file="Calendar-2022-10-21-164146"
with open(f'{file}.ics', 'rb') as file:
    ics_text = file.read()
    cal = Calendar.from_ical(ics_text)
```
Output:

```
it's a dump of the file
```

## Environment

 - [x] OS:                               Mac OS Monterey
 - [x] Python version:            Python 3.10 
 - [x] `icalendar` version:      5.0.1

How to browse a large file ?

Thanks

jacadzaca:
fix #472


src/icalendar/cal.py
```diff
@@ -390,14 +390,22 @@ def from_ical(cls, st, multiple=False):
...
         if len(comps) < 1:
-            raise ValueError(f'Found no components where '
-                             f'exactly one is required: '
-                             f'{st!r}')
+            raise ValueError(cls._format_error(
+                'Found no components where exactly one is required', st))
         return comps[0]
 
+    def _format_error(error_description, bad_input, elipsis='[...]'):
+        # there's three character more in the error, ie. ' ' x2 and a ':'
```
niccokunzmann, line 401: I think, a docstring would be nice, also linking to tje issue.

src/icalendar/cal.py
```diff
@@ -390,14 +390,22 @@ def from_ical(cls, st, multiple=False):
...
+        max_error_length = 100 - 3
```
niccokunzmann, line 402: There is code smell called magic number. This number is in a constant. So that is nice  removing the code smell.

niccokunzmann:
Thanks! New version?

jacadzaca:
Maybe let's merge #479 into master first?

src/icalendar/cal.py
jacadzaca: I think a docstring would be redundant and the issue's linked in the commit message.

niccokunzmann:
That might take a bit of time. Whenever you or I or anyone else feel like it, a new version can be generated.

src/icalendar/tests/test_components_break_on_bad_ics.py
```diff
@@ -26,3 +26,14 @@ def test_rdate_dosent_become_none_on_invalid_input_issue_464(events):
...
+
+@pytest.mark.parametrize('calendar_name', [
+    'big_bad_calendar',
+    'small_bad_calendar',
+    'multiple_calendar_components',
+])
+def test_error_message_doesnt_get_too_big(calendars, calendar_name):
+    with pytest.raises(ValueError) as exception:
+        calendars[calendar_name]
+    assert len(str(exception).split(': ')[1]) <= 100
```
angatha, line 38: That should probably be `.split(': ', 1)` so a second `': '` does not hide later parts of the message from the assertion.

PR #479, opened 2022-10-29
Issue 318 skip value parameter for default date time1

jacadzaca:
make @niccokunzmann's test green :)


jacadzaca:
oh, i've only ran pytest. How do I fix the plone tests?

niccokunzmann:
I think, they will need modification. They probably test some code with explicit date-time value. Their job, I would say. Their code should not break.

It should not prevent progress to have their tests run. The question is if we actually do something bad. Maybe not!

PR #480, opened 2022-10-30
Adjust assertion of max_error_length test

angatha:
Some message parts may be hidden from the assertion because of multiple splitting. See #473 and [this](https://github.com/collective/icalendar/pull/473/files)for context.

### DIFF
--- a/src/icalendar/tests/calendars/pr_480_summary_with_colon.ics
+++ b/src/icalendar/tests/calendars/pr_480_summary_with_colon.ics
@@ -1,0 +1,7 @@
+BEGIN:VCALENDAR
+BEGIN:VEVENT
+SUMMARY:Example calendar with a ': ' in the summary
+END:VEVENT
+BEGIN:VEVENT
+SUMMARY:Another event with a ': ' in the summary
+END:VEVENT
--- a/src/icalendar/tests/test_components_break_on_bad_ics.py
+++ b/src/icalendar/tests/test_components_break_on_bad_ics.py
@@ -31,6 +31,7 @@ def test_rdate_dosent_become_none_on_invalid_input_issue_464(events):
     'big_bad_calendar',
     'small_bad_calendar',
     'multiple_calendar_components',
+    'pr_480_summary_with_colon',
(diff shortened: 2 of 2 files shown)


jacadzaca (approved):

jacadzaca:
Good catch :)

jacadzaca:
@angatha could you provide an example file ics with a ': ' and add it to the parameter list? e.g:

```
BEGIN:VCALENDAR
BEGIN:VEVENT
SUMMARY:Example calendar with a ': ' in the summary
END:VEVENT
BEGIN:VEVENT
SUMMARY:Another event with a ': ' in the summary
END:VEVENT
[...]
END:VCALENDAR
```

jacadzaca (changes requested):
@angatha could you provide an example file ics with a ': ' and add it to the parameter list? e.g:

```
BEGIN:VCALENDAR
BEGIN:VEVENT
SUMMARY:Example calendar with a ': ' in the summary
END:VEVENT
BEGIN:VEVENT
SUMMARY:Another event with a ': ' in the summary
END:VEVENT
[...]
END:VCALENDAR
```

angatha:
I can do that. The one that you provided, is a good reference, but I'm missing the advantage of that.

Assuming the state of #473, that test would succeed. And assuming the state before #473, where long messages where generated, that test will still pass, since the error is in the test itself.

The fact, that the test succeeds, indicates the bug inside the test but no one will notice this since all tests passed.

jacadzaca:
Oh, you're right.

jacadzaca (approved):

angatha:
But it doesn't hurt :)

jacadzaca:
maybe we should assert that the message is split as we think in the test? or would that be testing tests esentialy?

niccokunzmann:
Tests are different from contracts or verifications or proofs of programs. They are a heuristic. If they fail, something is wrong if they don't, we do not know. I think, we are hitting that here.

niccokunzmann:
It looks good to me  merging.

PR #483, opened 2022-10-30
Update parser.py

pronoym99:
1. Convert ```format``` strings to python3.6+ compliant ```f-strings``` wherever possible.
2. Replace ```for i in range()``` construct with an equivalent ```enumerate``` in the ```q_split``` function.

### DIFF
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -6,6 +6,7 @@ It is stupid in the sense that it treats the content purely as strings. No type
 conversion is attempted.
 """
 
+from icalendar.prop import vText
 from icalendar.caselessdict import CaselessDict
 from icalendar.parser_tools import DEFAULT_ENCODING
 from icalendar.parser_tools import SEQUENCE_TYPES
@@ -51,12 +52,13 @@ def tzid_from_dt(dt):
     if hasattr(dt.tzinfo, 'zone'):
         tzid = dt.tzinfo.zone  # pytz implementation
     elif hasattr(dt.tzinfo, 'key'):
-        tzid = dt.tzinfo.key # ZoneInfo implementation
+        tzid = dt.tzinfo.key  # ZoneInfo implementation
     elif hasattr(dt.tzinfo, 'tzname'):
         # dateutil implementation, but this is broken
         # See https://github.com/collective/icalendar/issues/333 for details
         tzid = dt.tzinfo.tzname(dt)
     return tzid
+
 
 def foldline(line, limit=75, fold_sep='\r\n '):
     """Make a string folded as defined in RFC5545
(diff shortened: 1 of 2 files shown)


src/icalendar/parser.py
```diff
@@ -158,8 +159,7 @@ def q_split(st, sep=',', maxsplit=-1):
     length = len(st)
     inquote = 0
     splits = 0
-    for i in range(length):
-        ch = st[i]
+    for i, ch in enumerate(st):
```
niccokunzmann, line 162: Nice one spotted!

niccokunzmann:
Nice thanks! I like it that you update this old code base. It needs tjat love :)

jacadzaca:
@niccokunzmann shouldn't we squash all these `Update changelog` commits into one, so as to keep the commit log clean? or do we just let it slide?

niccokunzmann:
I value contribution and engagement over perfection, people over code - manifesto of agile software development. If we can have both, this is great. We can merge with a squash commit.
What would you do? I mean: I have my approach and I really like to hear from you! I do not expect you to do the same and diversity is strength.

niccokunzmann:
Uh. I did a merge here and a squash in the other....   Yes, I try to keep this in mind a bit more.

jacadzaca:
@niccokunzmann I also think that engagement is better than perfection, but the stuff around coding (documentation, commit logs etc) shouldn't be played down. For example, when a potential contributor sees spam in the commit history they might feel discouraged, or when we'll need to revert some change, fragmenting it over multiple (>3) commits will make the process hurdle-some.

 I squashed @pronoym99 commit in [my fork](https://github.com/collective/icalendar/compare/master...jacadzaca:icalendar:master) but for some reason can't PR the changes :(

pronoym99:
True that @jacadzaca. From my side as well, I tried squashing the commits: both from the Git Kraken GUI as well as from the terminal as I promised. However, I am unable to see the changes in the commits tab for this PR. The only thing I did was I ended up polluting my local clone quite a bit 😆. To be very precise, I don't know what exactly took place in [#8b7b224](https://github.com/collective/icalendar/pull/483/commits/8b7b2245acf08b50f2a5d8c185d18887990bec3e) when I did it via the Kraken GUI and when I did exactly the same using the terminal in [#0c3800c](https://github.com/collective/icalendar/pull/483/commits/0c3800cd602c1cebc2d5adc0f5fe0b0580749466). Therefore, I was expecting a squash and merge operation from the maintainers' side.

jacadzaca:
Yeah, git's difficult sometimes. I should have said that you should do a `git rebase -i` first and then a `git push --force`, but I thought that you'd get it right given time and went to bed.

Issue #318, opened 2021-02-17
Skip VALUE properties when using the default value


dilyanpalauzov:
As outlined at https://github.com/indico/indico/issues/4783 when the iCalendar library is used to generate iCalendar data it includes the default VALUE property parameter.  This increases the amount of exchanged data without having added value.

In the example, generated by iCalendar:
```
DTSTART;VALUE=DATE-TIME:20190616T050000Z
DTEND;VALUE=DATE-TIME:20190620T153000Z
DTSTAMP;VALUE=DATE-TIME:20210217T082500Z
```
`;VALUE=DATE-TIME` should be absent.

niccokunzmann:
@dilyanpalauzov , я писал три теста для этого в #450 и если вы хотите, можете составить PR и участвовать в Hacktoberfest.

I wrote three tests for this in #450 and if you like, you can create a pull request and participate in Hacktoberfest.

PR #492, opened 2022-11-10
implement hash and __eq__ for vDDDTypes

### LINKED ISSUE #487: [REGRESSION] DDDType not hashable in icalendar 5
Compare this:

```
>>> import icalendar
>>> import datetime
>>> icalendar.__version__
'4.0.9'
>>> icalendar.vDDDTypes(datetime(2022,10,10)).__hash__()
TypeError: 'module' object is not callable
>>> icalendar.vDDDTypes(datetime.datetime(2022,10,10)).__hash__()
8729746249635
>>> icalendar.vDDDTypes(datetime.datetime(2022,10,10)).__hash__
<method-wrapper '__hash__' of vDDDTypes object at 0x7f08d6c97cd0>
>>>
```

with this:

```
>>> import icalendar
>>> import datetime
>>> icalendar.__version__
'5.0.1'
>>> icalendar.vDDDTypes(datetime.datetime(2022,10,10)).__hash__()
TypeError: 'NoneType' object is not callable
>>> icalendar.vDDDTypes(datetime.datetime(2022,10,10)).__hash__
```

I noticed tests failing because of this in another project - https://github.com/niccokunzmann/python-recurring-ical-events/issues/99

niccokunzmann:
This adds tests and an implementation of `__hash__` for vDDDTypes to fix #487.

- add `__hash__`
- add `.dt` comparison to `__eq__` so that the equality is really tested.

### DIFF
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -320,7 +320,7 @@ class vDDDTypes:
 
     def __eq__(self, other):
         if isinstance(other, vDDDTypes):
-            return self.params == other.params
+            return self.params == other.params and self.dt == other.dt
         return False
 
     def __hash__(self):
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -323,6 +323,9 @@ class vDDDTypes:
             return self.params == other.params
         return False
 
+    def __hash__(self):
+        return hash(self.dt)
+
     @classmethod
     def from_ical(cls, ical, timezone=None):
         if isinstance(ical, cls):
--- a/src/icalendar/tests/test_unit_prop.py
+++ b/src/icalendar/tests/test_unit_prop.py
@@ -6,8 +6,11 @@ from icalendar.parser import Parameters
 import unittest
 from icalendar.prop import vDatetime
 from icalendar.windows_to_olson import WINDOWS_TO_OLSON
-
+import pytest
 import pytz
+from ..prop import vDDDTypes
+from copy import deepcopy
+
 
 
 class TestProp(unittest.TestCase):
@@ -111,27 +114,6 @@ class TestProp(unittest.TestCase):
 
         # Bad input
         self.assertRaises(ValueError, vDDDTypes, 42)
-
-    def test_vDDDTypes_equivalance(self):
-        from ..prop import vDDDTypes
-        from copy import deepcopy
-
-        vDDDTypes = [
-                vDDDTypes(pytz.timezone('US/Eastern').localize(datetime(year=2022, month=7, day=22, hour=12, minute=7))),
-                vDDDTypes(datetime(year=2022, month=7, day=22, hour=12, minute=7)),
-                vDDDTypes(date(year=2022, month=7, day=22)),
-                vDDDTypes(time(hour=22, minute=7, second=2))]
-
-        for v_type in vDDDTypes:
-            self.assertEqual(v_type, deepcopy(v_type))
-            for other in vDDDTypes:
-                if other is v_type:
-                    continue
-                self.assertNotEqual(v_type, other)
-
-            # see if equivalnce fails for other types
-            self.assertNotEqual(v_type, 42)
-            self.assertNotEqual(v_type, 'test')
 
     def test_prop_vDate(self):
         from ..prop import vDate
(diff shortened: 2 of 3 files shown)


src/icalendar/tests/test_unit_prop.py
```diff
@@ -518,6 +500,38 @@ def test_prop_TypesFactory(self):
         )
 
 
+
+vDDDTypes_list = [
+    vDDDTypes(pytz.timezone('US/Eastern').localize(datetime(year=2022, month=7, day=22, hour=12, minute=7))),
```
jacadzaca, line 504: Shouldn't this use the `in_timezone` fixture from the `conftest.py`?

src/icalendar/tests/test_unit_prop.py
```diff
@@ -6,8 +6,11 @@
 import unittest
 from icalendar.prop import vDatetime
 from icalendar.windows_to_olson import WINDOWS_TO_OLSON
-
+import pytest
 import pytz
+from ..prop import vDDDTypes
```
jacadzaca: This could have been imported on line 7

src/icalendar/prop.py
```diff
@@ -320,9 +320,12 @@ def to_ical(self):
 
     def __eq__(self, other):
         if isinstance(other, vDDDTypes):
-            return self.params == other.params
+            return self.params == other.params and self.dt == other.dt
```
jacadzaca, line 323: If we only hash `dt` then we shouldn't check for `params`' equality in `__eq__`, because we can have a mismatch between hashes' equality and equivalence, e.g.

```python
x = vDDDTypes(datetime(year=2022, month=7, day=22, hour=22, minute=7, second=2))
y = vDDDTypes(datetime(year=2022, month=7, day=22, hour=22, minute=7, second=2))
y.params['value'] = 'DATE-TIME'
x == y # False
hash(x) == hash(y)  # True
```
while [x is clearly equivalent to y](https://github.com/collective/icalendar/issues/318).

src/icalendar/tests/test_unit_prop.py
```diff
@@ -518,6 +500,38 @@ def test_prop_TypesFactory(self):
...
+    deepcopy,
+    identity,
+    hash,
+])
+@pytest.mark.parametrize("v_type", vDDDTypes_list)
+@pytest.mark.parametrize("other", vDDDTypes_list)
+def test_vDDDTypes_equivalance(map, v_type, other):
+    if v_type is other:
+        assert map(v_type) == map(other), f"identity implies equality: {map.__name__}()"
+        assert not (map(v_type) != map(other)), f"identity implies equality: {map.__name__}()"
```
jacadzaca, line 525: This check seems redundant.

src/icalendar/tests/test_unit_prop.py
```diff
@@ -518,6 +500,38 @@ def test_prop_TypesFactory(self):
...
+    else:
+        assert map(v_type) != map(other), f"expected inequality: {map.__name__}()"
+        assert not (map(v_type) == map(other)), f"expected inequality: {map.__name__}()"
```
jacadzaca, line 528: This one also seems redundant.

jacadzaca (changes requested):

src/icalendar/prop.py
jacadzaca: Since the params are equal if dts are, we can remove `params` from the `__eq__`. We should private `params` then though.

src/icalendar/tests/test_unit_prop.py
niccokunzmann: I tried to add not too much. Hm. Since many of the arguments do not use a tz, that would create redundant tests.

src/icalendar/tests/test_unit_prop.py
niccokunzmann: There is `__neq__`, is there not?  This would test both. Maybe I remember wrongly.

src/icalendar/prop.py
niccokunzmann: Equality implies hash equality. Hash equality does not imply equality.

niccokunzmann: Do you propose another test case? There can be special cases covered with default values etc. I would not do that here.

niccokunzmann: Thanks for the review! What do you think is necessary to cover still to get the issue resolved?
What will improve maintenance?
What would be nice to have in general?

src/icalendar/tests/test_unit_prop.py
jacadzaca: Oh, I wasn't aware of that.

src/icalendar/prop.py
jacadzaca: Yes, I think there should be a test that checks that datetime vDDDTypes are equal if both have same `dt` but one has the the `params['value']` set to `DATE-TIME` and another dose not.

jacadzaca: > Equality implies hash equality. Hash equality does not imply equality.

You're right -- I didn't read the docs page carefully enough

jacadzaca: > Yes, I think there should be a test that checks that datetime vDDDTypes are equal if both have same `dt` but one has the the `params['value']` set to `DATE-TIME` and another dose not.

We can do that in another PR tho

src/icalendar/tests/test_unit_prop.py
jacadzaca: Could you explain how would they be redundant? I don't quite get that.

src/icalendar/tests/test_unit_prop.py
angatha: It's actually `__ne__`.

src/icalendar/tests/test_unit_prop.py
angatha: A second or third case with timezone should do the job too. One with same timezone and one with different time zone but same instant in time.

niccokunzmann: In ae9ac71 I added another time zone. I think, this covers the case:
- time zone differs
- time zone different from no time zone

Since no time zone is also a time zone in my view, all the cases would be covered. What do you think?

PR #501, opened 2022-12-15
Fix 500

### LINKED ISSUE #500: [BUG] Issues using vBoolean and others for parameters
## Describe the bug
I'm new to using this library so it's possible I'm misunderstanding how these property types are being used.  I was trying to add an `rsvp` parameter to an attendee as a `vBoolean` and got a stacktrace.

The [docs talk about doing it this way](https://icalendar.readthedocs.io/en/latest/usage.html#example):
> Automatic encoding is not yet implemented for parameter values, so you must use the ‘v*’ types you can import from the icalendar package (they’re defined in icalendar.prop):

Looking through the code, it seems like this issue exists for any of the property types that are not based on `str`, `list`, or `tuple`.  `list` and `tuple` go through `q_join()` while everything else goes through `dquote()` which requires it to have a `str.replace()` method.

https://github.com/collective/icalendar/blob/16f9015e16a0376721e66367d72323e7fd676900/src/icalendar/parser.py#L100-L105

## To Reproduce

```python
from icalendar import Event, vBoolean, vCalAddress
attendee = vCalAddress(f'mailto:someone@example.com')
attendee.params['rsvp'] = vBoolean(True)
event = Event()
event.add('attendee', attendee, encode=0)
print(event.to_ical().decode())
```
Output:

(file paths reduced to make it easier to read)
```
  File "<console>", line 1, in <module>
  File "icalendar/cal.py", line 430, in to_ical
  File "icalendar/cal.py", line 419, in content_lines
  File "icalendar/cal.py", line 412, in content_line
  File "icalendar/parser.py", line 315, in from_parts
    return cls(f'{name}:{values}')
  File "icalendar/parser.py", line 218, in to_ical
    value = param_value(value)
  File "icalendar/parser.py", line 105, in param_value
    return dquote(value)
  File "icalendar/parser.py", line 144, in dquote
    val = val.replace('"', "'")
AttributeError: 'vBoolean' object has no attribute 'replace'
```

## Expected behavior
```
BEGIN:VEVENT
ATTENDEE;RSVP=TRUE:mailto:someone@example.com
END:VEVENT
```

## Environment

 - [x] OS:  MacOS 13.1                              
 - [x] Python version:  3.9.13           
 - [x] `icalendar` version:  5.0.3    

## Additional context

- [ ] I tested it with the latest version `pip3 install https://github.com/collective/icalendar.git`

jacadzaca:
fix #500

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,10 @@ New features:
 
 Bug fixes:
 
-- ...
+- vBoolean can now be used as an parameter
+  Ref: #501
+  Fixes: #500
+  [jacadzaca]
 
 
 5.0.3 (2022-11-23)
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -102,7 +102,10 @@ def param_value(value):
     """
     if isinstance(value, SEQUENCE_TYPES):
         return q_join(value)
-    return dquote(value)
+    elif isinstance(value, str):
+        return dquote(value)
+    else:
+        return dquote(value.to_ical().decode(DEFAULT_ENCODING))
 
 
 # Could be improved
--- a/src/icalendar/tests/test_issue_500_vboolean_for_parameter.py
+++ b/src/icalendar/tests/test_issue_500_vboolean_for_parameter.py
@@ -1,0 +1,9 @@
+from icalendar import Event, vBoolean, vCalAddress
+
+def test_vBoolean_can_be_used_as_parameter_issue_500(events):
+    '''https://github.com/collective/icalendar/issues/500'''
+    attendee = vCalAddress(f'mailto:someone@example.com')
+    attendee.params['rsvp'] = vBoolean(True)
+    event = Event()
+    event.add('attendee', attendee)
+    assert event.to_ical() == events.event_with_rsvp.raw_ics
--- a/src/icalendar/tests/events/event_with_rsvp.ics
+++ b/src/icalendar/tests/events/event_with_rsvp.ics
@@ -1,0 +1,3 @@
+BEGIN:VEVENT
+ATTENDEE;RSVP=TRUE:mailto:someone@example.com
+END:VEVENT
--- a/tox.ini
+++ b/tox.ini
@@ -32,7 +32,7 @@ commands_pre =
 # Install Plone and explicitly the single package that uses icalendar, plus the test runner.
     pip install Plone plone.app.event[test] zope.testrunner -c https://dist.plone.org/release/6.0-dev/constraints.txt
 # Install the dev version of the package, mostly so we can safely point to the path with the tests.
-    pip install -e "git+https://github.com/plone/plone.app.event.git#egg=plone.app.event"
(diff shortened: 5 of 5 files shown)


jacadzaca:
@mauritsvanrees plone tests failed

mauritsvanrees:
> @mauritsvanrees plone tests failed

@jacadzaca This is a problem with `tox`, where 4.0.0 was recently released. On tox 3 the job works fine, but on 4 we need a small change:

```
-    pip install -e "git+https://github.com/plone/plone.app.event.git#egg=plone.app.event"
+    pip install -e "git+https://github.com/plone/plone.app.event.git\#egg=plone.app.event"
```

I have added a commit to your branch.

CHANGES.rst
```diff
@@ -18,7 +18,10 @@ New features:
 
 Bug fixes:
 
-- ...
+- vBoolean can now be used as an parameter
```
tisdall, line 21: I think the fix covers all the property types...  `vInt`, `vTime`, etc

src/icalendar/tests/test_issue_500_vboolean_for_parameter.py
```diff
@@ -0,0 +1,9 @@
+from icalendar import Event, vBoolean, vCalAddress
+
+def test_vBoolean_can_be_used_as_parameter_issue_500(events):
```
tisdall, line 3: Should there be tests for some of the other types or is that excessive?  Basically everything with an `.to_ical()` method works and I'm assuming there's test already for those methods.

tisdall (commented):
Thanks for fixing this!

CHANGES.rst
jacadzaca: You're probably right, but I don't put things in the changelog if there isn't a corresponding test for the entry.

src/icalendar/tests/test_issue_500_vboolean_for_parameter.py
jacadzaca: There are tests for [proper encoding](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_unit_prop.py) for other v types, but I don't think there are any for checking if they can be used as a param. If you've got the time to file a PR with a parametrization of this test for other v types I'll gladly accept it :) if you don't, I will try to update the test sometime next week :)

CHANGES.rst
niccokunzmann: If you put it in the changelog, it creates an expectation and then people come back if this is not the case. But since it is in the documentation, the expectation is there either way.

src/icalendar/parser.py
```diff
@@ -102,7 +102,10 @@ def param_value(value):
     """
     if isinstance(value, SEQUENCE_TYPES):
         return q_join(value)
-    return dquote(value)
+    elif isinstance(value, str):
+        return dquote(value)
+    else:
+        return dquote(value.to_ical().decode(DEFAULT_ENCODING))
```
niccokunzmann, line 108: The None return is interesting. When does it hit?

niccokunzmann:
I see the tests running and from the conversation there arises no problem. I merge.

PR #503, opened 2022-12-19
add missing import pytz in example

abe-101:
The example uses the pytz module:
`event.add('dtstart', datetime(2005,4,4,8,0,0,tzinfo=pytz.utc))`
But doesn't import it


jacadzaca (approved):

jacadzaca:
Every little bit helps, thanks :)

PR #504, opened 2022-12-19
display file content of example

abe-101:
I suggest it might be a nice idea to display the content of the file generated in the example

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -317,6 +317,26 @@ Write to disk::
   522
   >>> f.close()
 
+Print out the calendar::
+
+  >>> print(cal.to_ical().decode("utf-8")) # doctest: +NORMALIZE_WHITESPACE
+  BEGIN:VCALENDAR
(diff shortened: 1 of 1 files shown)


jacadzaca:
Could you try to make the [expected output be tested](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_with_doctest.py)? This way the documentation wont become stale.

abe-101:
> Could you try to make the [expected output be tested](https://github.com/collective/icalendar/blob/master/src/icalendar/tests/test_with_doctest.py)? This way the documentation wont become stale.

Done.
doctest are new to me but I think it should be good now

jacadzaca (approved):

jacadzaca:
Thanks!

jacadzaca:
Personally, I'd make the commit message more detailed, e.g. `show generated example's content on the usage page`, but it's a nitpick.

Issue #268, opened 2018-08-13
Add support for BYWEEKDAY


jonocodes:
dateutils.rrule supports the byweekday parameter already:
http://dateutil.readthedocs.io/en/stable/rrule.html

Could this be added to icalendar?

niccokunzmann:
@natashamm Thanks for working on this issue! I wonder: Do you think that this issue can be closed now? Is there anything else to do?

natashamm:
The other issue I was thinking of is: what if someone generates an RRULE with BYDAY and BYWEEKDAY, do we want to throw an error or automatically replace one with the other? There is nothing in RFC 5545 about that of course since it only considers BYDAY.

niccokunzmann:
@natashamm @jonocodes https://github.com/collective/icalendar/pull/511 was merged and released in v5.0.5. Can you check if this can be closed?

natashamm:
> Since it is an assumption of the algorithm, I would say an assertion would do it.

My bad - do you mean an assumption in https://dateutil.readthedocs.io/en/stable/rrule.html?

niccokunzmann:
No. If we have, these two arguments that mean the same and the value differs, what do we do? Which one is used? rrule can only take one of them, I assume. What are your thoughts?

niccokunzmann:
Ok, thanks! I had a look. It seems that dateutil will handle the parsing of an rrule string and as such the question of how to convert this parameter is not necessary to answer in icalendar. Thanks for the work!
If there is any doubt whether this should be closed, please comment/re-open!

PR #511, opened 2023-02-06
Issue 268 - support for BYWEEKDAY

natashamm:
Add support for BYWEEKDAY in the RRULE based on https://github.com/collective/icalendar/issues/268

Background: [dateutil](https://dateutil.readthedocs.io/en/stable/rrule.html) renamed BYDAY to BYWEEKDAY to "avoid the ambiguity of that keyword". Note that RFC 5545 uses BYDAY only.

From https://github.com/collective/icalendar/issues/268

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ Changelog
 
 Minor changes:
 
-- ...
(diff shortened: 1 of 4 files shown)


niccokunzmann:
Thanks a lot!

Issue #348, opened 2022-01-23
Exception when there's no ':' when parsing value


N0ury:
For the following ics file, I have the error:
```
  File "/usr/local/lib/python3.9/site-packages/icalendar/cal.py", line 330, in from_ical
    name, params, vals = line.parts()
  File "/usr/local/lib/python3.9/site-packages/icalendar/parser.py", line 349, in parts
    values = unescape_string(st[value_split + 1:])
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```
Here's the ics file:
```
BEGIN:VCALENDAR
CALSCALE:GREGORIAN
METHOD:PUBLISH
VERSION:2.0
PRODID:-//Sixt//RAC//EN
BEGIN:VFREEBUSY
FREEBUSY;FBTYPE=FREE:20190624T063000Z/20190624T163000Z
ORGANIZER;CN=Sixt SE
UID:SIXT_9879691160
DTSTAMP:20190612T104813Z
END:VFREEBUSY
BEGIN:VEVENT
CLASS:PUBLIC
DESCRIPTION:\nVotre véhicule avec le numéro de réservation xxxxxxxxxxx sera à votre disposition le 24.06.2019 à 08:30 heures (heure locale). Retour prévu : Ferney-Voltaire AP de Genève Cointrin Sect(F), le 24.06.2019 à 18:30 heures.\nVous trouverez des informations relatives à la modification ou à  l'annulation de votre réservation sous http://www.sixt.fr\n
DTSTART:20190624T063000Z
DTSTAMP:20190612T104813Z
DTEND:20190624T163000Z
LOCATION:Sixt Genève Aéroport secteur France, AP de Genève Cointrin Sect(F), 01210 Ferney-Voltaire, FR
PRIORITY:5
TRANSP:TRANSPARENT
SEQUENCE:0
STATUS:CONFIRMED
SUMMARY:Sixt : détails de votre réservation
UID:SIXT_9879691160
BEGIN:VALARM
TRIGGER:-PT30M
ACTION:DISPLAY
[...]

PR #502, opened 2022-12-16
Update parser.py

N0ury:
issue #348

PR #514, opened 2023-04-12
N0ury master

niccokunzmann:
This updates https://github.com/collective/icalendar/pull/502 to have a merge with master.


niccokunzmann:
Interesting ... There are new Jenkins jobs appearing. I can understand the changelog one. I will experiment with the others.

niccokunzmann:
This PR does indeed fix #348. However, I wonder for the tests how the values can be accessed:

```
E           TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```
Is now:
```
>       assert freebusy["ORGANIZER"]["CN"] == "Sixt SE"
E       TypeError: string index indices must be integers or slices, not str
```
for this line:
```
ORGANIZER;CN=Sixt SE
```

niccokunzmann:
I fixed the test with the help of the documentation. The current error is that the parameter is missing a character. 

```
AssertionError: assert 'Sixt S' == 'Sixt SE'
```
On it ...

niccokunzmann:
@mauritsvanrees The plone PR tests were added and I do not know if they work for this module. They just appeared. I will run them but I do not know if they actually mean something. Could you check that they use the code from this PR when they run?

niccokunzmann:
@mauritsvanrees The 3.8 test fails.
https://jenkins.plone.org/job/pull-request-6.0-3.8/2576/robot/report/robot_report.html#totals
I am not really clear about why and how this bug fix contributed to that. I will wait a bit for a reply. If I merge this, then I will wait a bit with the release if the tests fail.

mauritsvanrees:
I see the Jenkins tests pass on Python 3.9, 3.10 and 3.11 and only fail on 3.8.  The two failures are in robot framework tests, and there can be some instability in there.  Looks unrelated.

But really, the Jenkins jobs should not run.  We have a GitHub Actions test in this repo that uses Plone, and that should be enough.  I think I or someone else has pushed a button to automatically re-enable these tests on lots of repositories.
I have deleted some web hooks in the settings.  The Jenkins tests should not be started for new PRs now.

For the record, the Jenkins jobs *do* use the branch from this PR.  See the [3.8 console](https://jenkins.plone.org/job/pull-request-6.0-3.8/2576/consoleFull):

```
mr.developer: Cloned 'icalendar' with git using branch 'N0ury-master' from 'https://github.com/niccokunzmann/icalendar.git'.
```

So: should be fine to merge and release, but I will leave both to you.  Thanks!

niccokunzmann:
@mauritsvanrees Thanks for the clarification! I like the changelog test by the way. That is a handy reminder. If you like to enable this again, that should be ok to try out.

Issue #518, opened 2023-05-25
RRULE BYDAY=xMO with x >=10 raises ValueError with to_ical()


semiprime:
According to RFC-5545, section 3.8.5.3, the BYDAY component of a yearly RRULE can take values like:
```
RRULE:FREQ=YEARLY;BYDAY=xMO
```
to mean xth Monday of the year. They explicitly give an example with x=20.

However, in icalendar calling to_ical() on such an event with x>=10 raises a ValueError.

## To Reproduce

```
import icalendar
from datetime import date

event1 = icalendar.Event()
event1.add('SUMMARY', 'Ev 1')
event1.add('DTSTART', date(2023,2,27))
event1.add('RRULE', {'FREQ':['YEARLY'], 'BYDAY':['9MO']}) # 9th Monday of year
print(event1.to_ical()) # works

event2 = icalendar.Event()
event2.add('SUMMARY', 'Ev 2')
event2.add('DTSTART', date(2023,3,6))
event2.add('RRULE', {'FREQ':['YEARLY'], 'BYDAY':['10MO']}) # 10th Monday of year
print(event2.to_ical()) # Raises ValueError with icalendar 5.0.5 & 4.1.0
```

This gives:
```
...
  File ".../python3.9/site-packages/icalendar/prop.py", line 581, in __new__
    raise ValueError(f'Expected weekday abbrevation, got: {self}')
ValueError: Expected weekday abbrevation, got: 10MO
```

## Environment
 - OS:   Slackware 15.0
 - Python version:  3.9.16
 - icalendar version:  5.0.5 or 4.1.0 (installed from PyPI)

## Proposed fix
The problem seems to be the regular expression `WEEKDAY_RULE` in file `prop.py`, which only accepts 0/1 characters for the `relative` component. A simple fix would therefore be to allow {0,2} characters:
```
diff --git a/src/icalendar/prop.py b/src/icalendar/prop.py
index ac87d0b..e00e805 100644
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -69,7 +69,7 @@ DATETIME_PART = f'(?:{DATE_PART})?(?:{TIME_PART})?'
 WEEKS_PART = r'(\d+)W'
 DURATION_REGEX = re.compile(r'([-+]?)P(?:%s|%s)$'
                             % (WEEKS_PART, DATETIME_PART))
[...]

PR #524, opened 2023-05-31
4.x branch: Fix for issue 518 (to_ical() with RRULE BYDAY values>=10)

semiprime:
Issue 518 fix for the 4.x branch (already fixed in master branch). Issue: https://github.com/collective/icalendar/issues/518

The fix itself is the same as in the master branch. The test code has been adapted because the corresponding test code in 4.x uses the unittest module, rather than pytest as in the master branch.

All tests pass (run with setup.py test). Without the fix, the new tests give 3 failures.

Justification for inclusion in 4.x branch: It's a crashing bug. The fix is low risk (never say zero risk!).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@ Changelog
 4.1.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Bug fixes:
+
+- to_ical() now accepts RRULE BYDAY values>=10 (Issue #518)
 
 
 4.1.0 (2022-07-11)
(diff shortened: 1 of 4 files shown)


semiprime:
Looking at the test logs, failures seem to be py35, py36, pypy3 because the version of Python wasn't available. Other tests were canceled.

FWIW, I initially tested with Python 3.9.16 (x64) (pytest & unittest versions) . After the automatic test failures, I tested on Python 2.7.18 (x86) and Python 3.5.3 (Arm64) (both unittest only). All these tests passed.

niccokunzmann:
Hm. The old Python version numbers have changed.

Would you be willing to adjust them in the  file?
Here are the ones: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
3.6.15
3.5.9
3.4.10

This way, the tests suit can run. Otherwise, all following changes might fail, too.

Thanks so much for backporting!

niccokunzmann:
#529 includes this commit to merge it with CI.

PR #528, opened 2023-08-28
No longer run the plone.app.event tests.

mauritsvanrees:
Also, run the Python 3.11 tests with the latest Python 3.11, instead of a release candidate.

In the last two weekly cron tests, the plone tests fail with:

    ModuleNotFoundError: No module named 'plone.app.event'

I see the same locally, and I have no idea why.

This seems a good time to stop running the `plone.app.event` tests within `icalendar`.  We ran them until now to prevent changes from creeping in that negatively affect Plone, which is where this module was born.  But so far, nothing like that has happened.  And occasionally these tests break because of reasons that have nothing to do with `icalendar`, so they distract from maintenance of this package.  So let's stop doing this.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,8 +23,7 @@ jobs:
         - ["3.10",  "py310"]
         - ["pypy-3.9", "pypy3"]
         - ["3.10",  "docs"]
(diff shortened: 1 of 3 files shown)


niccokunzmann:
thanks!

PR #538, opened 2023-08-30
Update readthedocs config

niccokunzmann:
see https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os
> The configuration key build.os is required to build your documentation.

image


niccokunzmann:
This will be opened again as a PR from this repo so we can test it with readthedocs.

Issue #165, opened 2015-07-13
Problem parsing a file with event recurring on weekdays


dmedvinsky:
The offending file:

```
BEGIN:VCALENDAR
METHOD:REQUEST
PRODID:Microsoft CDO for Microsoft Exchange
VERSION:2.0
BEGIN:VTIMEZONE
TZID:GMT +0100 (Standard) / GMT +0200 (Daylight)
BEGIN:STANDARD
DTSTART:16010101T030000
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
RRULE:FREQ=YEARLY;WKST=MO;INTERVAL=1;BYMONTH=10;BYDAY=-1SU
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:16010101T020000
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
RRULE:FREQ=YEARLY;WKST=MO;INTERVAL=1;BYMONTH=3;BYDAY=-1SU
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
DTSTAMP:20150703T071009Z
DTSTART;TZID="GMT +0100 (Standard) / GMT +0200 (Daylight)":20150703T100000
SUMMARY:Sprint 25 Daily Standup
DTEND;TZID="GMT +0100 (Standard) / GMT +0200 (Daylight)":20150703T103000
RRULE:FREQ=DAILY;UNTIL=20150722T080000Z;INTERVAL=1;BYDAY=MO, TU, WE, TH, FR
 ;WKST=SU
END:VEVENT
END:VCALENDAR
```

Trying to parse it:

```
icalendar.Calendar.from_ical(content).subcomponents
[...]

PR #545, opened 2023-08-30
Check that event is included in parsing

niccokunzmann:
This fixes https://github.com/collective/icalendar/issues/165


niccokunzmann:
This is ready for review :)

Issue #152, opened 2014-11-22
No documentation about parsing


enkidulan:
First of all thank you for this awesome package.
I have read documentation and haven't found any information about how to parse ical files, and that is quite confusing. It would be just great if you add some parsing examples into docs. I found a lot of examples how to use icalendar here -  http://stackoverflow.com/questions/3408097/parsing-files-ics-icalendar-using-python.
Thank you.

danstowell:
FWIW I put together a simple script using icalendar to render ical to html. It may be useful as a source of example material https://github.com/danstowell/calendrino

jnikula:
Please try to find some time to put together even rudimentary documentation about parsing. Basically I've initially disregarded the whole package because I've thought there isn't proper support for parsing. The tests aren't documentation, and they aren't even good examples of how this should be used.

I'm trying to like this, but it's tough going having to read the source to get this right...

nicbou:
This is still an issue in 2021. The documentation is very sparse, and doesn't cover basic use cases like parsing an iCalendar file (for instance from a Google Calendar export).

jnikula:
> This is still an issue in 2022.

FTFY. ;)

nicbou:
Ha! Good catch!

I got it working. Here's a little code sample to get you started. It should be [here](https://github.com/nicbou/timeline/blob/master/backend/source/archive/models/icalendar.py) when the code is committed and pushed some time in the next few days.

```
with open(ics_file, 'r') as file:
    calendar = Calendar.from_ical(file.read())
    for event in calendar.walk('VEVENT'):
        event_metadata = {
            'event': {
                'end_date': datetime_to_json(event['DTEND'].dt)
            }
        }

        if event.get('LOCATION'):
            event_metadata['location'] = {
                'name': event['LOCATION']
            }

        if event.get('DTSTAMP'):
            event_metadata['event']['creation_date'] = datetime_to_json(event['DTSTAMP'].dt)

        # ...
```

PR #547, opened 2023-08-30
Add documentation about parsing

niccokunzmann:
This intents to fix https://github.com/collective/icalendar/issues/152.

I put the example into the first page because this is where people are looking. Also, this is more interesting than a version discussion from my view.


niccokunzmann:
This is ready for review :)

jacadzaca:
maybe we should also add examples on how to edit/create ics files?

PR #550, opened 2023-09-01
fix #526

### LINKED ISSUE #526: [BUG] Two `Calendar` objects with different events should not `==` each other
## Describe the bug
Given two `Calendar` objects being loaded from a .ics file (one fresh, one cached):

```py
calendar_a: icalendar.cal.Calendar
calendar_b: icalendar.cal.Calendar
```

I would expect (and attempted to rely upon) the `==` operator to reliably test whether they're the same or not (so I can avoid emitting "changes" when no changes have occurred).

`calendar_a == calendar_b` in all cases, regardless of whether their contents differ.

## Expected behavior
The `==` operator correctly returns `True` or `False` depending on content.

jacadzaca:
fix #526


jacadzaca:
@yassirasad and @Lx does this fix work for you folks?

src/icalendar/tests/test_unit_cal.py
```diff
@@ -455,3 +457,37 @@ def test_cal_strict_parsing(self):
...
+    'issue_526_calendar_with_events',
+    'issue_526_calendar_with_different_events',
+])
+@pytest.mark.parametrize('other_calendar', [
+    'issue_156_RDATE_with_PERIOD_TZID_khal',
+    'issue_156_RDATE_with_PERIOD_TZID_khal_2',
+    'issue_178_custom_component_contains_other',
+    'issue_178_custom_component_inside_other',
+    'issue_526_calendar_with_events',
+    'issue_526_calendar_with_different_events',
```
niccokunzmann: If the lists are always supposed to be equal, they could be in one variable.

src/icalendar/cal.py
```diff
@@ -436,6 +436,22 @@ def __repr__(self):
         subs = ', '.join(str(it) for it in self.subcomponents)
         return f"{self.name or type(self).__name__}({dict(self)}{', ' + subs if subs else ''})"
 
+    def __eq__(self, other):
```
niccokunzmann, line 439: As a rule, overriding eq should also override hash. But I think, as long as the equality gets reduced, equality still implies hash equality.

src/icalendar/cal.py
```diff
@@ -436,6 +436,22 @@ def __repr__(self):
...
+        if not properties_equal:
+            return False
+
+        # The subcomponents might not be in the same order,
+        # neither there's a natural key we can sort the subcomponents by nor
+        # are the subcomponent types hashable, so  we cant put them in a set to
+        # check for set equivalence. We have to iterate over the subcomponents
+        # and look for each of them in the list.
+        for subcomponent in self.subcomponents:
+            if subcomponent not in other.subcomponents:
```
niccokunzmann, line 453: this is a subset test. a == b does not imply b == a.

niccokunzmann, line 453: Copy the calendar and remove an event and test again a==b and b==a.

niccokunzmann: `all(component in other.subcomponents for component in self.subcomponents) and all(component in self.subcomponents for component in other.subcomponents)` but this has O(2^n) which is bad.

ah!
return len(self.subcomponents) == len(other.subcomponents) 
should do the same.

jacadzaca: Good catch!

src/icalendar/cal.py
jacadzaca: I think that implementing a `__hash__` for sub-class of CaselessDict is not straightforward as per [this](https://github.com/collective/icalendar/pull/391/files#r957502946}

niccokunzmann:
I think it is perfect.

niccokunzmann:
I guess, we can create a new release.

PR #552, opened 2023-09-15
Update about.rst

bobflorian:
Updated about.rst to reflect the current state and remove the notion that this repo is outdated.


niccokunzmann:
I would rather merge this than wait but also we have a check for the changelog. Would you like to add to that?

PR #554, opened 2023-09-17
Fix `vText.__repr__` `BytesWarning`

scop:
Exposed by running in `python3 -b` mode.

### DIFF
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -719,7 +719,7 @@ class vText(str):
         return self
 
     def __repr__(self):
-        return f"vText('{self.to_ical()}')"
+        return f"vText('{self.to_ical()!r}')"
 
     def to_ical(self):
         return escape_char(self).encode(self.encoding)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Minor changes:
 - Move pip caching into Python setup action.
 - Check that issue #165 can be closed.
 - Updated about.rst for issue #527
+- Avoid ``vText.__repr__`` BytesWarning.
 
 Breaking changes:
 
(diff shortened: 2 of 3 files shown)


src/icalendar/prop.py
```diff
@@ -719,7 +719,7 @@ def __new__(cls, value, encoding=DEFAULT_ENCODING):
         return self
 
     def __repr__(self):
-        return f"vText('{self.to_ical()}')"
+        return f"vText('{self.to_ical()!r}')"
```
scop, line 722: The `!r` approach preserves existing behavior. Depending on desired outcome, another fix would be
```suggestion
        return f"vText('{self.to_ical().decode(self.encoding)}')"
```

jacadzaca:
hey, big thanks for the PR! Since the ical text is stored as bytes internally, I think having bytes explicitly in `__repr__` is correct behavior. However, I think your suggestion would be an ideal implementation of an `__str__` :) What do you think?

A `__str__` / `__repr__` implementation for other v elements would also be very welcome, since they all are lacking one :)

Also, please fill out the [CHANGELOG](https://github.com/collective/icalendar/blob/master/CHANGES.rst) and add yourself to [CREDITS](https://github.com/collective/icalendar/blob/master/docs/credits.rst) as per [CONTRIBUTING](https://github.com/collective/icalendar/blob/master/CONTRIBUTING.rst)

src/icalendar/prop.py
jacadzaca: I like the `decode` version better

scop:
> Since the ical text is stored as bytes internally, I think having bytes explicitly in `__repr__` is correct behavior.

Agree, that's (and for no behavioral changes) I opted for the `!r` option.

> However, I think your suggestion would be an ideal implementation of an `__str__` :) What do you think?

Kind of, but then again I don't know if the `vText('...')` "wrapper" belongs in it. Anyway, I'd like to leave that and addition of any additional `__str__`/`__repr__` implementations out from this PR to keep it in scope. And frankly, leave all that to someone else to take care of :)

> Also, please fill out the CHANGELOG and add yourself to CREDITS

Done.

jacadzaca:
Thanks for the PR :)

niccokunzmann:
This is part of release 5.0.8

PR #556, opened 2023-09-19
Add timezone to period value types

niccokunzmann:
The time zone was missing from PERIOD values.

While working on https://github.com/niccokunzmann/python-recurring-ical-events/issues/113, I noticed that I was not able to retrieve the time zone for the start and end of the PERIOD value. Thus, I created this PR.


jacadzaca:
thanks :)

PR #561, opened 2023-09-27
cli: Display datetimes in local timezone

vimpostor:
This makes up for a much nicer UX for users of the cli utility.
Before this patch the start and end datetimes were always displayed with a fixed timezone (the timezone of the calendar creator).
This meant that users had to manually calculate the time in their own timezone.

With this patch the times are printed in the expected format, i.e. in the user's own local timezone.

It seems that in 7a8d584b an attempt was already made to implement this, but unfortunately `start.astimezone(start.tzinfo)` is a no-op and does absolutely nothing. Instead no arguments have to be passed to the `astimezone()` function to convert to the user's timezone, see https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone

The unit test has been adapted to check for the output in the correct timezone.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,9 @@ Changelog
 
 Minor changes:
 
-- ...
+- The cli utility now displays start and end datetimes in the user's local timezone.
+  Ref: #561
+  [vimpostor]
 
 Breaking changes:
 
--- a/src/icalendar/cli.py
+++ b/src/icalendar/cli.py
@@ -52,10 +52,10 @@ def view(event):
         end = event.decoded('dtend', default=start)
     duration = event.decoded('duration', default=end - start)
     if isinstance(start, datetime):
-        start = start.astimezone(start.tzinfo)
+        start = start.astimezone()
     start = start.strftime('%c')
     if isinstance(end, datetime):
-        end = end.astimezone(end.tzinfo)
+        end = end.astimezone()
     end = end.strftime('%c')
 
     return f"""    Organizer: {organizer}
--- a/src/icalendar/tests/test_cli_tool.py
+++ b/src/icalendar/tests/test_cli_tool.py
@@ -1,6 +1,11 @@
 import unittest
 
+from datetime import tzinfo, datetime
 from icalendar import Calendar, cli
+try:
+    import zoneinfo
+except ModuleNotFoundError:
+    from backports import zoneinfo
 
 INPUT = '''
 BEGIN:VCALENDAR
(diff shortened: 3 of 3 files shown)


CHANGES.rst
```diff
@@ -6,7 +6,7 @@ Changelog
 
 Minor changes:
 
-- ...
+- The cli utility now displays start and end datetimes in the user's local timezone.
```
niccokunzmann, line 9: Nice, could you reference the PR here and also your name?
Would you like to be part of the contributors' list?

niccokunzmann (commented):
The changes look good to me. I am wondering though: It never mentions the time zone. How is this tool expected to work: should you always see the times in the local time zone? Should you see it in the event's time zone? What if there are many time zones - then it does not make sense to see e.g. 19:00  as it could mean a lot.

vimpostor:
> I am wondering though: It never mentions the time zone. How is this tool expected to work: should you always see the times in the local time zone?

Yes

> Should you see it in the event's time zone?

No, that wouldn't make any sense. If someone shares their calendar in Google Calendar with me, I expect Google Calendar to show me all events with the time of my own timezone, not the timezone of the other person.
Similarly, the CLI tool should show the time always in the timezone of the user.

> What if there are many time zones - then it does not make sense to see e.g. 19:00 as it could mean a lot.

I don't know what you mean with many time zones, the time is always displayed in one timezone, the local timezone of the user. Theoretically we could append the timezone description to the time, but that would be pretty unusual. Usually you only add the timezone description, if it is not already in the local timezone of the user.

By the way before this patch the CLI tool also never showed the timezone info, it just displayed no timezone despite the timezone of the displayed time being not the local timezone. That was pretty surprising behaviour.


Just as an example, to make sure we are on the same page:
Imagine you live in the `Europe/Berlin` timezone and some other person lives in the `America/Recife` timezone. This other person now shares a calendar with you with an event starting at 9:30 AM in the other person's timezone.
This means that the raw `.ics` file contains something like: `DTSTART;TZID=America/Recife:20231002T093000`.

Now **before this patch**, the CLI tool would show this for the user living in the `Europe/Berlin` timezone:

```
Starts     : Mon Oct  2 09:30:00 2023
``` 

Note how there is no timezone info, so the user would mistakenly assume that the event starts at 9:30 in their own timezone (since they don't know what the timezone of the other person is). But in reality the event starts 14:30 in their own timezone.

Now **after this patch**, the CLI tool outputs the expected:
```
    Starts     : Mon Oct  2 14:30:00 2023
```

Which is the correct time in the timezone of the user.

CHANGES.rst
vimpostor: OK I referenced the PR in the changes file.

I don't think I need to be in the contributors' list for this small change, unless you really want me to.

niccokunzmann:
This makes sense to me, too. Thanks for providing this!

vimpostor:
And in case you were wondering where the hell people are using the icalendar CLI tool in production, many people use it in CLI mail clients (e.g. mutt) to display sent calendar events inline directly in the mail without having to leave the terminal.

That's also why the local timezone makes sense.

niccokunzmann:

PR #566, opened 2023-09-30
fix handling of single strings passed to vCategory

### LINKED ISSUE #322: Single string not given in a list for the field CATEGORIES results in a list of characters
Dear developers,

If a single string is given as the argument for a category, each character of the string is considered as a single category. Here is an example:
```
#### Python code:
from icalendar import Calendar, Event

cal = Calendar()

event = Event()
event.add('summary', 'Event with bare string as argument for categories')
event.add('categories', "Lecture")
cal.add_component(event)

event = Event()
event.add('summary', 'Event with a one element list of strings for categories')
event.add('categories', ["Lecture"])
cal.add_component(event)

print(cal.to_ical().decode("utf-8"))

#### Output:
BEGIN:VCALENDAR
BEGIN:VEVENT
SUMMARY:Event with bare string as argument for categories
CATEGORIES:L,e,c,t,u,r,e
END:VEVENT
BEGIN:VEVENT
SUMMARY:Event with a one element list of strings for categories
CATEGORIES:Lecture
END:VEVENT
END:VCALENDAR
```

From the documentation I found that the problem might be caused within the definition of class vCategories:
```
class vCategory(object):

    def __init__(self, c_list):
        if not hasattr(c_list, '__iter__'):
            d_list = [c_list]
```
Here d_list seems to be never used later on. The if statement has therefore no effect.

Best,
JT

jacadzaca:
fix #322


CHANGES.rst
```diff
@@ -20,7 +20,9 @@ New features:
 
 Bug fixes:
 
-- ...
+- CATEGORIES field now accepts strings properly
```
niccokunzmann: > CATEGORIES field works if there is only one entry and this entry is a string

Is that what it does?

niccokunzmann:
Thanks, this looks quite nice!

niccokunzmann:
Uh. Conflicts!

PR #567, opened 2023-10-06
Use non legacy timezone name

jspricke:


niccokunzmann:
Thanks!

niccokunzmann:

PR #568, opened 2023-10-09
Add some __eq__

jspricke:
Needed as 088eea7 adds comparison of elements this makes sure that they
are compared by value instead of by reference.

This party fixes https://github.com/niccokunzmann/x-wr-timezone/issues/8. I
think the unit tests should be moved here as well.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,8 @@ Bug fixes:
   [jacadzaca]
 - Use non legacy timezone name.
   Ref: #567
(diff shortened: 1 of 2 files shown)


niccokunzmann:
Hi, thanks for this! Would you like to add unit tests. too?

niccokunzmann:
I think, it would be good to open an issue to discuss value based equality everywhere.

niccokunzmann:

PR #569, opened 2023-10-09
OSS-Fuzz Integration

ennamarie19:
Hello,
This is a follow-on from my Issue regarding OSS-Fuzz integration. This PR introduces the source for the fuzzer and a YAML config to fuzz each PR for a set duration. I would greatly appreciate it being merged in!

I will continue monitoring OSS-Fuzz for bugs and fix any I can in future PRs. Max M also has access to the OSS-Fuzz dashboard to review any security-relevant crashes that may come up.

Please let me know if anything else is needed.

Thank you!


niccokunzmann:
Hi, thanks for che PR. To merge it I would like the tests to run.

src/icalendar/fuzzing/build.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -eu
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
niccokunzmann, line 4: Since this file is under the apache license, we need to check that we do not ship it or adjust the license information.

niccokunzmann:
The other tests fail with
```
 import atheris
E   ModuleNotFoundError: No module named 'atheris'
```

It might be needed to either install the dependency or exclude the fuzzing test file from being collected when we test the documentation. Either is fine.

src/icalendar/fuzzing/build.sh
ennamarie19: The license is mandated by OSSFuzz for these files, is there an action that you would like to me take to prevent the fuzzing directory from being shipped?

ennamarie19:
@niccokunzmann Hey Nicco, I added a changelog entry and omitted the fuzzing directory from the doctest!

Thank you!

src/icalendar/fuzzing/build.sh
niccokunzmann: I think, yes that would be a fix.

niccokunzmann:
Thanks! I am a bit confused not seeing che other tests run also. This might not be your code though... I am investigating...

src/icalendar/fuzzing/build.sh
ennamarie19: I'm not familiar with how to prevent that. Any suggestions?

niccokunzmann: I can not test it at the moment but this is what I found: https://stackoverflow.com/questions/43365915/python-packaging-exclude-directory-from-bdist-wheel

niccokunzmann: I added it in https://github.com/collective/icalendar/pull/573

niccokunzmann:
https://github.com/collective/icalendar/pull/573 is a follow-up pull request that also does not distribute the fuzzing files.

ennamarie19:
﻿Thank you so much Jaca!!On Oct 25, 2023, at 12:25 PM, Jaca ***@***.***> wrote:﻿
Merged #569 into master.

ennamarie19:
Thank you so much Nicco! Really appreciate your help with getting this situated :) I will keep you posted as to what, if anything, is uncovered. Thank you again! On Oct 25, 2023, at 7:28 AM, Nicco Kunzmann ***@***.***> wrote:﻿
#573 is a follow-up pull request that also does not distribute the fuzzing files.

Issue #570, opened 2023-10-12
Value based equality


niccokunzmann:
It is to me easier to expect two calenders to equal if they have the same components and values.
#568 shows that some components or attributes use object identity when compared where value equality is expected. I open this issue to discuss this and wonder if we could transit to a unified approach to compare values. What would be the challenges?
What are your thoughts?

niccokunzmann:
I think that for a start, we could read in all calendar files that are there twice and compare the two. Some from https://github.com/niccokunzmann/x-wr-timezone/issues/8 might also show which elements are not equal.

niccokunzmann:
This is live in v5.0.11. Thanks for your contributions!
I will close this for now as it solves my use-cases. If any class escapes proper equality comparison, we should open another issue.

PR #575, opened 2023-11-01
Compare test calendars to find components that do not equal

niccokunzmann:
This contributes to #570.

- [x] move `.ics` files into their folders
- [x] create tests from the example files to check which values do not compare
- [x] test inequality of the components/properties
- [x] implement an equality function


niccokunzmann:
The tests run here but weirdly they do not on my computer. There is just one failing!

```
    def test_parsed_calendars_are_equal_if_from_same_source(ics_file):
        """Ensure that a calendar equals the same calendar."""
        same_calendar = ics_file.__class__.from_ical(ics_file.raw_ics)
>       assert same_calendar == ics_file
E       assert VCALENDAR({'VERSION': vText('b'2.0''), 'PRODID': vText('b'icalendar-2023'')}, VEVENT({'UID': vText('b'ical-jacadzaca-3''), 'SUMMARY': vText('b" Some very different event ':'"'), 'DTSTART': vDDDTypes(2021-11-01 16:00:00+01:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTEND': vDDDTypes(2021-11-01 16:30:00+01:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTSTAMP': vDDDTypes(2021-10-04 15:02:45+00:00, Parameters({}))}), VEVENT({'UID': vText('b'ical-jacadzaca-4''), 'SUMMARY': vText('b' Some very different other event''), 'DTSTART': vDDDTypes(2021-11-01 16:40:00+01:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTEND': vDDDTypes(2021-11-01 16:50:00+01:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTSTAMP': vDDDTypes(2021-10-04 15:02:45+00:00, Parameters({}))})) == VCALENDAR({'VERSION': vText('b'2.0''), 'PRODID': vText('b'icalendar-2023'')}, VEVENT({'UID': vText('b'ical-jacadzaca-3''), 'SUMMARY': vText('b" Some very different event ':'"'), 'DTSTART': vDDDTypes(2021-11-01 16:00:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTEND': vDDDTypes(2021-11-01 16:30:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTSTAMP': vDDDTypes(2021-10-04 15:02:45+00:00, Parameters({}))}), VEVENT({'UID': vText('b'ical-jacadzaca-4''), 'SUMMARY': vText('b' Some very different other event''), 'DTSTART': vDDDTypes(2021-11-01 16:40:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTEND': vDDDTypes(2021-11-01 16:50:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), 'DTSTAMP': vDDDTypes(2021-10-04 15:02:45+00:00, Parameters({}))}))
E         Common items:
E         {'PRODID': vText('b'icalendar-2023''), 'VERSION': vText('b'2.0'')}
E         Full diff:
E           {
E            'PRODID': vText('b'icalendar-2023''),
E            'VERSION': vText('b'2.0''),
E           }

src/icalendar/tests/test_equality.py:19: AssertionError
------------------------------------------------------------- Captured stdout setup --------------------------------------------------------------
issue_526_calendar_with_different_events

```

Difference when the same calendar is read twice from the same source:
```
'DTSTART': vDDDTypes(2021-11-01 16:00:00+01:00, Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'})), ==
'DTSTART': vDDDTypes(2021-11-01 16:00:00,       Parameters({'TZID': 'Western/Central Europe', 'VALUE': 'DATE-TIME'}))
```

niccokunzmann:
Please review this. I would like to make a release after this merge - it is even worth releasing before that if it takes longer.

One fix is that `==` was implemented but `!=` still used the old behavior and with that `a == b` did not mean `not a != b` or vice versa. That should be a big problem if you wanted to compare calendars.

src/icalendar/tests/conftest.py
```diff
@@ -71,3 +80,19 @@ def utc(request):
 ])
 def in_timezone(request):
     return request.param
+
+
+@pytest.fixture(params=[
```
angatha, line 85: Looks fine but `@pytest.mark.parametrize` may be the cleaner solution: https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

src/icalendar/tests/test_timezoned.py
```diff
@@ -47,7 +49,7 @@ def test_create_from_ical_zoneinfo(self):
 
     def test_create_from_ical_pytz(self):
         directory = os.path.dirname(__file__)
```
angatha: `directory` became unused?

src/icalendar/tests/conftest.py
niccokunzmann: I used this way because I parametrize the fixture: https://docs.pytest.org/en/7.3.x/how-to/fixtures.html#fixture-parametrize - the parametrize function is for the test functions.

src/icalendar/tests/test_timezoned.py
niccokunzmann: Thanks for spotting this!

src/icalendar/prop.py
```diff
@@ -280,7 +294,19 @@ def from_ical(ical):
         return out
 
 
-class vDDDTypes:
+class DtEqualityMixin:
```
jacadzaca: Shouldnt mixins always call `super.__eq__`? If I need a class to be both a DtEqualityMixin and a EqualityMixin then I'd presume that two  of its objects are equal if they're equal in terms of the two mixins. If classes implementing both mixins dont make sense, should we call theses classes mixins?

jacadzaca: Tbh, I'd in favor of removing the mixins all together. What if later down the line the classes `__eq__` methods might diverge? I feel that this is over-engineering.

src/icalendar/tests/conftest.py
angatha: I understand :)

angatha (approved):

niccokunzmann:
@angatha, if you like, you can also merge! (No pressure - I just like to ship all the contributions) Also, you can start a new release if you like. <- Please see these as invitations!

jacadzaca (changes requested):

angatha:
> @angatha, if you like, you can also merge! (No pressure - I just like to ship all the contributions) Also, you can start a new release if you like. <- Please see these as invitations!

Thank you :) I haven't done a release and are bussy right now. But @jacadzaca requested a change anyway.

src/icalendar/prop.py
niccokunzmann: >  should we call theses classes mixins?

Uh - looking at the code now, I would not call them mixins any more. At the time of writing, I was unsure if they would be mixins.

- I think, that I can just change the vBinary and vCategory `__eq__`.
- `DtEqualityMixin` is a base class for all the different time types to make them comparable.

niccokunzmann: There is a common superclass for the time types because vDDDTypes needs to equal the special classes, too.

@jacadzaca have a look again. What do you think?

jacadzaca: I like that much better :) good job!

jacadzaca (approved):

PR #579, opened 2023-11-05
OSS-Fuzz: Migrating corpus to upstream

ennamarie19:
This PR moves the corpus directory to be hosted on icalendar rather than on OSS-Fuzz.


ennamarie19:
The CIFuzz failures are expected, as this is currently a breaking change. Once this is merged, OSSFuzz will merge in the PR that moves the corpus out of their source control. After that, not further PRs should be needed to OSSFuzz (besides adding / removing emails from CC lists)

niccokunzmann:
Hi, thanks!

I think, this line needs `*.ics` also in it i Manifest.in: `recursive-exclude src/icalendar/fuzzing *.py *.sh`
If you add a changelog entry and merge master, we can see if it works. Where is OSS FUZZ told to use this data and not the one provided in the OSS FUZZ repository?

ennamarie19:
> Hi, thanks!
> 
> I think, this line needs `*.ics` also in it i Manifest.in: `recursive-exclude src/icalendar/fuzzing *.py *.sh` If you add a changelog entry and merge master, we can see if it works. Where is OSS FUZZ told to use this data and not the one provided in the OSS FUZZ repository?

Look at the Dockerfile hosted in the OSS-Fuzz projects directory for icalendar and the build.sh hosted in the fuzzing directory in collective. These two files work together to define where to pull the corpus from. We currently have an open [PR](https://github.com/google/oss-fuzz/pull/11183) with OSS-Fuzz to use the fuzzing corpus hosted in collective/icalendar. IOnce you merge, that will fix the nightly build for OSS-Fuzz.

ennamarie19:
@niccokunzmann I added a line to the changelog but the verifier is saying I didn't due to, I'm guessing, a conflict resolution I had to make after the push. This should be good to go now!

ennamarie19:
> @niccokunzmann I added a line to the changelog but the verifier is saying I didn't due to, I'm guessing, a conflict resolution I had to make after the push. This should be good to go now!

@niccokunzmann anything else we need to do?

niccokunzmann:
Hi, I think, I would merge it without the  changelog verifier passing as I see the entry.

One thing I am wondering about:
The test files seem to be the same files as in the tests/calendars directory - is that so?

Then, I had a look at the build.sh file.
https://github.com/collective/icalendar/blob/5c42e8e5231d6816a8e6218b346989b308985cf3/src/icalendar/fuzzing/build.sh#L25
It seems that the build.sh file is supposed to fill the corpus. As such, I am wondering - would a simple copy command be sufficient and we do not have to add these files as a copy? Are they even copies or are they slightly modified..?

There are some calendar files which are specifically broken:
https://github.com/collective/icalendar/blob/5c42e8e5231d6816a8e6218b346989b308985cf3/src/icalendar/tests/conftest.py#L90
However, we could name them `broken_...` and then they can be excluded - if that is an issue.

I am merging this because of the value added.

PR #581, opened 2023-11-05
Add fixes for OSS-Fuzz bugs and test cases

ennamarie19:
This PR introduces the fixes to address two bugs discovered by OSS Fuzz and adds two test cases to ensure that they are not reintroduced down the line.

The bugs are as follows:
1) Type mismatch in vDDDlists.to_ical when given a date-time that resolves to a string and is joined as bytes
2) Attempting to pop from an empty stack when parsing 'BEGIN' and 'END' in to_ical

### DIFF
--- a/src/icalendar/tests/test_oss_fuzz_errors.py
+++ b/src/icalendar/tests/test_oss_fuzz_errors.py
@@ -1,5 +1,8 @@
 """This file collects errors that the OSS FUZZ build has found."""
+from datetime import time
 from icalendar import Calendar
+from icalendar.prop import vDDDLists
+
 import pytest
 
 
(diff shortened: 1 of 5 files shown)


ennamarie19:
Please merge this in first, as it should resolve pipeline errors in the other branch

src/icalendar/tests/test_oss_fuzz_errors.py
```diff
@@ -0,0 +1,18 @@
...
+import pytest
+
+
+def test_stack_is_empty():
+    """If we get passed an invalid string, we expect to get a ValueError."""
+    with pytest.raises(ValueError):
+        Calendar.from_ical("END:CALENDAR")
+
+
+def test_vdd_list_type_mismatch():
```
niccokunzmann, line 15: nice!

src/icalendar/tests/test_oss_fuzz_errors.py
```diff
@@ -0,0 +1,18 @@
...
+    """If we pass in a string type, we expect it to be converted to bytes"""
+    vddd_list = vDDDLists([time(hour=6, minute=6, second=6)])
```
niccokunzmann, line 17: I am wondering. `to_ical()` - should it always return a string or bytes? I would assume bytes. Since the library might not have had a refactoring work since Python2.7, I can understand if these get muddled up. I think, we read calendars from bytes... Let's check. I think, this is a quick fix but in the end, we might want to create type annotations and have a common interface. This might be worth creating an issue. I wonder what the other maintainers think.

niccokunzmann: Thanks for pointing that in consistency out!

PR #584, opened 2023-11-11
Added more instrumentation + caught one more exception

ennamarie19:
Another quick PR, I added another defined exception that was holding up fuzzer progress and made a tweak to see if we can get more intelligent instrumentation and coverage exploration. 

A suggestion would be collecting all raised exceptions under a common base-class, such as CalendarException, so the fuzzer can just do except CalendarException as opposed to string-matching multiple ValueErrors. This would also help users of the library in their error-handling, as well.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Minor changes:
 - Added exclusion of fuzzing corpus in MANIFEST.in
 - Augmented fuzzer to optionally convert multiple calendars from a source string
 - Added additional exception handling of defined errors to fuzzer, to allow fuzzer to explore deeper
+- Added more instrumentation to fuzz-harness
 Breaking changes:
 
 - ...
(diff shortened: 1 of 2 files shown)


niccokunzmann:
> A suggestion would be collecting all raised exceptions under a common base-class, such as CalendarException, so the fuzzer can just do except CalendarException as opposed to string-matching multiple ValueErrors. This would also help users of the library in their error-handling, as well.

tht sounds good, I would keep the base classes and put them all into exceptions.py

niccokunzmann:
you already created fixes. feel free to follow the release process to release a new version

capuanob:
> you already created fixes. feel free to follow the release process to release a new version

Good morning,

Could you explain what you mean? Are you asking us to bump the version number due to this PR?

Thank you!

niccokunzmann:
I think, there were some fixes made in other PRs.

Independent of this: If you like to release a new version, you can follow the release process: https://icalendar.readthedocs.io/en/latest/maintenance.html - solely as information, to give you a choice!

PR #587, opened 2023-11-21
Add scripts and tests to reproduce fuzzer errors in pytest

niccokunzmann:
- move fuzzing tests in to test folder
- create a script that runs the fuzzer and extracts the test case if the test fails

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ src/icalendar.egg-info/
 !.github
 !.gitignore
 venv
+/ical_fuzzer.pkg.spec
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ Minor changes:
 - Added corpus to fuzzing directory
 - Added exclusion of fuzzing corpus in MANIFEST.in
 - Augmented fuzzer to optionally convert multiple calendars from a source string
+- Add script to convert OSS FUZZ test cases to Python/pytest test cases
 - Added additional exception handling of defined errors to fuzzer, to allow fuzzer to explore deeper
 - Added more instrumentation to fuzz-harness
 - Rename "contributor" to "collaborator" in documentation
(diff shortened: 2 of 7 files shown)


niccokunzmann:
My intention is to create small pull requests like this one: https://github.com/niccokunzmann/icalendar/pull/2
Most OSS FUZZ errors depend on ICS files. These errors get reproduced though these pull requests. Then, we can fix, review and reproduce them locally with pytest. Probably, there is more to do about this. This is my current state of work.

niccokunzmann:
My idea is that I create the infrastructure to reproduce the tests easily. From there on, each time I get an email about a new error, I will create a PR with the error so it can be reproduced.

capuanob:
@niccokunzmann This is very cool! I just want to make sure of a few things

1. Did you test this locally with a checked out OSSFuzz (instructions [here](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally)) to make sure that it doesn't break the OSSFuzz integration?
2. Also, I wonder if there is any performance impact to the fuzz-testing with the addition of stdout prints? I'd imagine its minimal, but worth confirming the reported exec/s before and after the changes when running locally

niccokunzmann:
> id you test this locally with a checked out OSSFuzz (instructions [here](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally)) to make sure that it doesn't break the OSSFuzz integration?

Yes, I did that. This allowed me to extract the ICS file used.

> Also, I wonder if there is any performance impact to the fuzz-testing with the addition of stdout prints?

That is minimal (a few milliseconds if I look at it). The corpus has a few files in it. Printing these does not make is slower. Running the fuzzer and preparing the fuzzing tests take most of the time.

ennamarie19:
LGTM!

PR #590, opened 2024-01-24
Update license

Shortfinga:
Hey,

my dependencyscanner is unfortunately not able to detect your license. I guess others aren't able to detect this too so perhaps you could help us out here?

Not sure about the second commit with the license file. Not sure if you use this heading anywhere but the library github uses to detect the license of a repo now detects it with 100% accuracy...

Thank you for your work...
Max

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setuptools.setup(
     author='Plone Foundation',
     author_email='plone-developers@lists.sourceforge.net',
     url='https://github.com/collective/icalendar',
-    license='BSD',
+    license='BSD-2-Clause',
     packages=setuptools.find_namespace_packages('src', exclude=["icalendar.fuzzing"]),
     package_dir={'': 'src'},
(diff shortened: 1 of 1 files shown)


Shortfinga:
Not sure if this really deserves an Changelog entry, does it?

Shortfinga:
> Hi, we use the heading [here](https://icalendar.readthedocs.io/en/latest/license.html). Isn't the first commit enough?
Yes totally

jacadzaca:
No need for an entry in the changelog. I'm not against removing the heading if it means that the license is properly detected

jacadzaca:
> > Hi, we use the heading [here](https://icalendar.readthedocs.io/en/latest/license.html). Isn't the first commit enough?
> > Yes totally

Could you revert/rebase the second one then? Let's not change more than necessary :)

jacadzaca:
Thanks :)

PR #597, opened 2024-03-17
refactor test file for issue 116

niccokunzmann:

See also #532


jacadzaca:
thanks :)

jacadzaca:
the CI-Fuzz job failure seems to have been caused by some unrelated error. I think mergin this is OK

PR #600, opened 2024-03-17
Refactor property parameter tests

niccokunzmann:

See also #532


niccokunzmann:
@jacadzaca, thanks for reviewing all the test PRs. If you can - would you have a look at this before #601?

PR #601, opened 2024-03-17
Refactor time tests

niccokunzmann:

See also #532


jacadzaca:
@niccokunzmann there's a conflict here

PR #606, opened 2024-03-19
Create FUNDING.yml

niccokunzmann:
This adds the collective: https://opencollective.com/python-icalendar
See https://github.com/collective/icalendar/discussions/512


niccokunzmann:
Merging this for progress.

Issue #635, opened 2024-06-21
Migrate from `pkg_resources` to `importlib`-based solutions


stevepiercy:
After following [Install `icalendar` Manually](https://icalendar.readthedocs.io/en/latest/install.html#install-icalendar-manually), [Building the documentation](https://icalendar.readthedocs.io/en/latest/install.html#building-the-documentation), and attempting to build the docs, I get the same error as in https://github.com/collective/icalendar/pull/621:

```pycon
$ make html                
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v7.3.7

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/Users/stevepiercy/projects/icalendar/env/lib/python3.12/site-packages/sphinx/config.py", line 509, in eval_config_file
    exec(code, namespace)  # NoQA: S102
    ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stevepiercy/projects/icalendar/docs/conf.py", line 2, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'

make: *** [html] Error 2
```

Per https://setuptools.pypa.io/en/latest/pkg_resources.html, pkg_resources is deprecated and should be replaced with other tools in the standard library.

> Attention
> 
> Use of `pkg_resources` is deprecated in favor of [`importlib.resources`](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), [`importlib.metadata`](https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata) and their backports ([`importlib_resources`](https://pypi.org/project/importlib_resources), [`importlib_metadata`](https://pypi.org/project/importlib_metadata)). Some useful APIs are also provided by [`packaging`](https://pypi.org/project/packaging) (e.g. requirements and version parsing). Users should refrain from new usage of `pkg_resources` and should work to port to `importlib`-based solutions.

Some examples of how to do that may be found in the [Pyramid repo's PRs](https://github.com/search?q=repo%3APylons%2Fpyramid+pkg_resources&type=pullrequests).

- [x] In `docs/conf.py`, replace `pkg_resources.get_distribution` with `importlib.metadata`. #636 
- [ ] In `bootstrap.py`, there are some conditional imports of both `pkg_resources` and `setuptools`. I'm not sure what to do with these.

niccokunzmann:
[...]

PR #636, opened 2024-06-21
Allow docs to build on Python 3.12 by replacing `pkg_resources` with …

stevepiercy:
…`importlib.metadata`.

- Fix underline length in README.rst

See #635

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,9 @@ Breaking changes:
       import icalendar
       icalendar.use_pytz()
 
+- Replaced ``pkg_resources.get_distribution`` with ``importlib.metadata`` in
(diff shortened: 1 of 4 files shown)


stevepiercy:
@niccokunzmann I'm not sure where to put my change log entry. Please correct me if I put it in the wrong place. Thank you!

CHANGES.rst
```diff
@@ -31,6 +31,9 @@ Breaking changes:
       import icalendar
       icalendar.use_pytz()
 
+- Replaced ``pkg_resources.get_distribution`` with ``importlib.metadata`` in
```
niccokunzmann, line 34: I would put this under `Minor changes:`

niccokunzmann (approved):
I would move it over but it can also be moved later.

PR #637, opened 2024-06-21
Enhance documentation of API

### LINKED ISSUE #631: Documentation does not show all objects available
## Describe the bug

When looking at the API reference, then I do not see all functions provided by icalendar. Only classes from some files are shown: https://icalendar.readthedocs.io/en/latest/api.html

## Expected behavior

I would like to see all classes and functions that are publicly accessible to be listed.

stevepiercy:
Fixes #631.

Depends on #636 for change log consistency.

I also fixed a docstring that returned this error:

```pycon
/Users/stevepiercy/projects/icalendar/src/icalendar/prop.py:docstring of icalendar.prop:4: CRITICAL: Missing matching underline for section title overline.

###########################################################################
# This module defines these property value data types and property parameters
```

Sphinx treated that line as a page title, and a blank line resolved the issue.

I'm not sure whether you want that line to be a comment or displayed, though.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,5 @@
 # icalendar documentation build configuration file
-import pkg_resources
+import importlib.metadata
 import datetime
 import os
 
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -28,7 +28,7 @@
 project = 'icalendar'
 this_year = datetime.date.today().year
 copyright = f'{this_year}, Plone Foundation'
-version = pkg_resources.get_distribution('icalendar').version
+version = importlib.metadata.version('icalendar')
```
niccokunzmann, line 31: You would usually open a new branch without the old commits by switching to master before. In your case though, I guess you would like to build with Python 3.12.

niccokunzmann:
Thanks!

PR #640, opened 2024-06-21
Rename all references of RFC 2445 to RFC 5545

### LINKED ISSUE #278: Which RFC is icalendar attempting to implement?
https://github.com/collective/icalendar/blob/master/docs/usage.rst says:

> This package is used for parsing and generating iCalendar files following the standard in **RFC 2445**.

OTOH, #183 says:

> Updated components to better conform to **RFC 5545**

So, which of these two RFCs is the *icalendar* Python module actually attempting to implement? Should the docs be updated to say "5545" instead of "2445"?

niccokunzmann:
This fixes #278.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,7 +40,8 @@ New features:
 
 Bug fixes:
 
-- ...
+- Rename RFC 2445 to RFC 5545, see `Issue 278
+  <https://github.com/collective/icalendar/issues/278>`_
 
 5.0.13 (2024-06-20)
 -------------------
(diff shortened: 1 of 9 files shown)


CHANGES.rst
stevepiercy, line 1: Line 1024 was omitted, but this might need to stay, assuming the newer RFC did not yet exist.

```
- The RFC 2445 was a bit unclear on how to handle line folding when it
```

stevepiercy (approved):
LGTM, but just one question. If that one instance should remain, then this is good to merge as is.

niccokunzmann:
For me, this is part of the changelog and a true statement about the past so it does not need changing.

PR #646, opened 2024-06-22
Remove is_broken property

### LINKED ISSUE #424: remove is_broken .attribute for .errors attribute
While refactoring the tests, this appeared:

https://github.com/collective/icalendar/pull/410/files#r982917279

Events have an `.is_broken` attribute and someone would like this to be removed for the next major release.
Instead, we should use `.errors`

niccokunzmann:
Fix #424
Contribute to #630.

`.errors` would be enough.


stevepiercy (approved):
OK to merge when it passes CI.

Issue #513, opened 2023-04-06
Drop claim to support for Python below 3.7


niccokunzmann:
In the README file, we state that we also maintain downwards compatibility:

> [Branch 4.x](https://github.com/collective/icalendar/tree/4.x) with maximum compatibility to Python versions 2.7 and 3.4+, PyPy2 and PyPy3.

It is not actually true from my perspective: We have not received one PR with a fix to that branch. There were [PRs that could have been applied to master and to 4.x](https://github.com/collective/icalendar/pulls?q=is%3Apr+is%3Aclosed+label%3A%22needs+4.x+PR+too%22), too but nobody did it. There is no update except from tests since we took up maintenance again.

So, it would be right from my perspective, to drop the claim.
Here is my proposal:

-----

Since we pour more [ effort into maintaining and developing icalendar](https://github.com/collective/icalendar/discussions/360), we decided on this structure:

-    [Branch master](https://github.com/collective/icalendar/) is compatible to Python versions 3.7+ and PyPy3.
-    [Branch 4.x](https://github.com/collective/icalendar/tree/4.x) with compatibility to Python versions 2.7 and 3.4+, PyPy2 and PyPy3 remains if anyone wishes to [port back fixes](https://github.com/collective/icalendar/pulls?q=is%3Apr+is%3Aclosed+label%3A%22needs+4.x+PR+too%22).

---
What are your thoughts on this?

mauritsvanrees:
Seems good to me.

To take Plone as example: Plone 6.0 is where all the current action is, and 5.2 is in maintenance mode.  If a fix is done in 6.0 and it makes sense to apply the same fix in 5.2, then this is possible, but it requires *someone* to do it.  Usually the original creator of a fix does this, sometimes of their own free will, or because someone explicitly asks them.    Sometimes another person does this.  If it is an obvious fix and it takes less than five minutes to back port it and do a quick check, then it often happens.  But often enough, it is not so simple, and it just does not happen.

niccokunzmann:
If we approach this slowly: Our last Py2.7 release was 2023-09-06. Releases to that branch were made once a year. We might not have a contribution to that branch until 2025. in At the moment and more so in 2025, a lot of software has already migrated, 2.7 is end of life. I think, too that it is safe to remove the claim.

niccokunzmann:
[...]

PR #634, opened 2024-06-20
drop python<=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been EOSed 27 Jun 2023.
Filter all code over `pyupgracde --py38-plus`.

PR #647, opened 2024-06-22
Update documentation for versions 3.8+

niccokunzmann:
This compacts the documentation.
Mentions of old versions are removed.
A concise path from 5.x to 6.0 is outlined.
master -> main

Contributes to https://github.com/collective/icalendar/issues/513

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,7 +52,7 @@ New features:
 
 Bug fixes:
 
-- Change documentation to represent compatibility to Python 3.8+ and PyPy3
+- Change documentation to represent compatibility with Python 3.8 - 3.12, and PyPy3.
 - Rename RFC 2445 to RFC 5545, see `Issue 278
   <https://github.com/collective/icalendar/issues/278>`_
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,7 +42,7 @@ New features:
 
 Bug fixes:
 
-- ...
+- Change documentation to represent compatibility to Python 3.8+ and PyPy3
 
 5.0.13 (2024-06-20)
 -------------------
--- a/README.rst
+++ b/README.rst
@@ -166,7 +166,7 @@ you can receive all the latest updates, and switch back to earlier behavior:
     <DstTzInfo 'Europe/Vienna' CET+1:00:00 STD>
 
 Version 6 is on `branch main <https://github.com/collective/icalendar/>`_.
-It is compatible with Python versions 3.7 - 3.11, and PyPy3.
+It is compatible with Python versions 3.8 - 3.12, and PyPy3.
 We expect the ``main`` branch with versions ``6+`` to receive the latest updates and features.
 
 Related projects
--- a/README.rst
+++ b/README.rst
@@ -136,21 +136,8 @@ with the same result:
     X-DT-ZONEINFO;TZID=Europe/London:20240619T100100
     END:VEVENT
 
-
-
-Versions and Compatibility
---------------------------
-
-``icalendar`` is a critical project used by many. It has been there for a long time and maintaining
-long-term compatibility with projects conflicts partially with providing and using the features that
-the latest Python versions bring.
-
-Volunteers pour `effort into maintaining and developing icalendar
-<https://github.com/collective/icalendar/discussions/360>`_.
-Below, you can find an overview of the versions and how we maintain them.
-
-Version 6
-~~~~~~~~~
+Version 6 with zoneinfo
+~~~~~~~~~~~~~~~~~~~~~~~
 
 Version 6 of ``icalendar`` switches the timezone implementation to ``zoneinfo``.
 This only affects you if you parse ``icalendar`` objects with ``from_ical()``.
@@ -166,7 +153,7 @@ By default and since 6.0.0, ``zoneinfo`` timezones are created.
     ZoneInfo(key='Europe/Vienna')
 
 If you would like to continue to receive ``pytz`` timezones in as parse results,
-you can receive all the latest updates, and switch back to version 5.x behavior:
+you can receive all the latest updates, and switch back to earlier behavior:
 
 .. code:: python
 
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -175,22 +162,8 @@ you can receive all the latest updates, and switch back to version 5.x behavior:
     >>> dt.tzinfo
     <DstTzInfo 'Europe/Vienna' CET+1:00:00 STD>
 
-Version 6 is on `branch master <https://github.com/collective/icalendar/>`_ with compatibility to Python versions ``3.7+`` and ``PyPy3``.
```
stevepiercy: In combination with the replacing the compatibility field with a version badge, removing this line eliminates all text of Python version compatibility. That's not good. See suggestion below.

README.rst
```diff
@@ -175,22 +162,8 @@ you can receive all the latest updates, and switch back to version 5.x behavior:
...
-No updates will be released for this.
-Please use version 6 and switch to use ``zoneinfo`` as documented above.
-
-Version 4
-~~~~~~~~~
-
-Version 4 is on `branch 4.x <https://github.com/collective/icalendar/tree/4.x>`_ with maximum compatibility with Python versions ``2.7`` and ``3.4+``, ``PyPy2`` and ``PyPy3``.
-The ``4.x`` branch only receives security and bug fixes if someone makes the effort.
-We recommend migrating to later Python versions and also providing feedback if you depend on the ``4.x`` features.
+Version 6 is on `branch main <https://github.com/collective/icalendar/>`_ with compatibility to C-Python and PyPy3.
```
stevepiercy: We need to be explicit with version numbers of Python and use text. I used the trove classifiers in `setup.py` to get the specific versions. We might want to add 3.12, now that it has been released, but only if it passes tests.

```suggestion
Version 6 is on `branch main <https://github.com/collective/icalendar/>`_.
It is compatible with Python versions 3.7 - 3.11, and PyPy3.
```

stevepiercy (changes requested):

README.rst
niccokunzmann: Since #645, we have a badge to list the versions. I would like to de-duplicate this.
The intention is to support all Python releases that are not end of life. How do you think we can improve this?

stevepiercy: A badge only shows after rendered through the web. Even then it's not accessible for people with vision impairments and it lacks a context without an HTML `alt` attribute. It's not a satisfactory replacement, but it is a nice visual enhancement for people without vision impairments.

In a README.md, you should have a clear indicator of version support somewhere, whether that is explicit text, a statement to read the trove classifiers in `setup.py`, or a link to read documentation. IMO, it's most common and easiest to keep it in README.md, even though that means more maintenance and risk of duplication mistakes.

In documentation with Sphinx, you can use replacements to avoid maintenance and duplication mistakes. We do this in Plone. Example:

[conf.py](https://github.com/plone/documentation/blob/21453a6ea6c528bbe8bb0799d688d5dc2a5d1809/docs/conf.py#L348-L359)

In the documentation source files:

https://github.com/plone/documentation/blob/21453a6ea6c528bbe8bb0799d688d5dc2a5d1809/docs/install/create-project.md?plain=1#L76-L76

Which renders as https://6.docs.plone.org/install/create-project.html#pre-requisites-for-installation

If you wanted to go fully automatic, I suppose you could change the `setup(app)` to parse trove classifiers and set the result as the value for the replacement variable.

stevepiercy: > The intention is to support all Python releases that are not end of life.

This should be documented at a policy level, and implemented throughout the repo wherever necessary.

I would go a step further, though, to future proof it. For example:

"icalendar supports all [released and supported versions of Python](https://devguide.python.org/versions/#supported-versions) at the time of its own release."

Thus if you released 6.0 today, then you would support Python 3.8-3.12 only.

In a future release, if you add or drop a version, then that could be a breaking change or a feature addition. You would follow your version numbering scheme accordingly when you make a release at that time.

README.rst
```diff
@@ -178,23 +165,10 @@ you can receive all the latest updates, and switch back to version 5.x behavior:
     >>> dt.tzinfo
     <DstTzInfo 'Europe/Vienna' CET+1:00:00 STD>
 
-Version 6 is on `branch main <https://github.com/collective/icalendar/>`_ with compatibility to Python versions ``3.7+`` and ``PyPy3``.
+Version 6 is on `branch main <https://github.com/collective/icalendar/>`_.
+It is compatible with Python versions 3.7 - 3.11, and PyPy3.
```
stevepiercy: I just noticed that the issue name says 3.8+, but the trove classifiers and this line say 3.7-3.11. Which is correct?

stevepiercy (changes requested):

README.rst
niccokunzmann: For version 6, 3.8+ is correct. #634 is included in #630.

README.rst
niccokunzmann: > It is compatible with Python versions 3.7 - 3.11, and PyPy3.

The above statement is in the README now.
If anything comes up, you can still add a PR.

README.rst
stevepiercy: OK, then we will need to update #634 to include `CHANGES.rst` with the correct Python versions.

stevepiercy: But wait a minute, look at the change log entry, the issue title, and this line of text. See the discrepancy in versions? They don't align with each other.

niccokunzmann: I know. Please change it as you like it. I would like to be consistent between releases, not always. There are other PRs changing this at the same time.
It can be changed after merge, too. If you think, this is harmful, then, do not merge it. Otherwise I prefer progress over perfection.

stevepiercy: I'll correct it in a review suggestion. I prefer to get details right as soon as I realize them, else I will likely forget to deal with them later.

stevepiercy: Side note: I assume I cannot push to your remote, but I have not tried. For ease of collaboration where others can push to this remote (`collective/icalendar`), consider using this remote when you push.

CHANGES.rst
```diff
@@ -52,6 +52,7 @@ New features:
 
 Bug fixes:
 
+- Change documentation to represent compatibility to Python 3.8+ and PyPy3
```
```suggestion
- Change documentation to represent compatibility with Python 3.8 - 3.12, and PyPy3.
```

README.rst
stevepiercy: Resolves Python version inconsistency with issue title and change log entry.

```suggestion
It is compatible with Python versions 3.8 - 3.12, and PyPy3.
```

stevepiercy (changes requested):
Versions are now consistent with merge of suggestions, then OK to merge this PR.

stevepiercy (approved):
Ready to merge at your discretion @niccokunzmann.

README.rst
niccokunzmann: Ok. This is a good suggestion to use this for collaboration. I sometimes create pull requests to other people's branches. I did not want to clutter what is going on here. Maybe, if I name that branches issue-..., then it is not so much of a clutter.

niccokunzmann:
Nice, thanks for the documentation improvement!

Issue #609, opened 2024-03-27
Transition from pytz to zoneinfo


niccokunzmann:
In EOL of Python 3.8 is 31st October 2024. Then, zoneinfo is the library that is shipped with Python. There are problems using pytz as timezone implementations have evolved. My proposal to solve this issue is the following:

- [x] create `icalendar.set_default_timezone_implementation(TimeZone)` where `TimeZone` can be pytz.Timezone or zoneinfo.ZoneInfo or others, taking a string like `"UTC"` or `Europe/Berlin`.
- [x] Allow a switch from zoneinfo to pytz by testing all timezone dependent tests with both implementations
- [x] a version 6 release after October will then switch from pytz to zoneinfo by setting another default and removing pytz altogether as a dependency. Tracked here: #630

What are your thoughts? Who thinks this will break existing code? What is needed for a transition?

See also:
- https://github.com/collective/icalendar/discussions/595
- https://github.com/collective/icalendar/discussions/532

Possibly related:
- https://github.com/collective/icalendar/issues/336
- https://github.com/collective/icalendar/issues/321

Code that needs touching:
The Timezone class contains logic to parse and create timezones from icalendar specifications:
https://github.com/collective/icalendar/blob/2b1cd4e00b80905b491294987fa59ea1f861ed4f/src/icalendar/cal.py#L545

niccokunzmann:
Also:

- [ ] Pytz has certain timezone names. We should support all of them with zoneinfo.

niccokunzmann:
I created https://github.com/collective/icalendar/pull/623 to work on some improvements.

niccokunzmann:
This is complete. We do not depend on pytz any more.

:partying_face:

PR #650, opened 2024-06-24
Remove pytz dependency

niccokunzmann:
Contributes to #609.

Until now, icalendar was using pytz.
This removes the dependency pytz.
Tests run with pytz and without.
The CI also runs the tests without pytz.
The default implementation of icalendar uses zoneinfo."

Added 

```
tox -e nopytz
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 # to run for a specific environment, use ``tox -e ENVNAME``
 [tox]
-envlist = py37,py38,py39,py310,py311,pypy3,docs,nopytz
+envlist = py38,py39,py310,py311,312,pypy3,docs,nopytz
 # Note: the 'docs' env creates a 'build' directory which may interfere in strange ways
 # with the other environments.  You might see this when you run the tests in parallel.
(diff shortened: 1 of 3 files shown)


niccokunzmann:
Ready for review.

tox.ini
```diff
@@ -1,6 +1,6 @@
 # to run for a specific environment, use ``tox -e ENVNAME``
 [tox]
-envlist = py37,py38,py39,py310,py311,pypy3,docs
+envlist = py37,py38,py39,py310,py311,pypy3,docs,nopytz
```
stevepiercy: Did we add py312 in another PR? If not, then that should be a new issue. If it was, then let's merge main into this branch.

stevepiercy (changes requested):
LGTM. Just one question in the review, and need to resolve merge conflicts.

tox.ini
niccokunzmann: We did not add that. py37 should be removed. - new PR...

niccokunzmann:
@stevepiercy could you merge when the tests succeed?

PR #660, opened 2024-06-25
fix version of icalendar during install to major version

niccokunzmann:
See https://github.com/collective/icalendar/issues/630#issuecomment-2188652309

This prevents us to wonder about the next major version change and if people get confused.


niccokunzmann:
> When you release 6.x, this will not be necessary, as pip install will install the latest stable release. 

It will be useful after the 6.0.0 release so that we can upgrade to version 7 with people being conscious about the version they fix to.

>  Changing it now is premature and harmful.

I agree, let's wait until a 6.* release is made that is not alpha.

Issue #677, opened 2024-07-01
[BUG] Coveralls does not report coverage


niccokunzmann:
## Describe the bug

Coveralls is supposed to generate a pull request summary. Currently, it breaks on `main` and on the pull requests.

## To Reproduce

See `main` and the commits:

https://github.com/collective/icalendar/actions/runs/9751261758/job/26912463324#step:6:1

```
Error running coveralls: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
Received 422 submitting job via Github Actions. By default, coveralls-python uses the "github" service name, which requires you to set the $GITHUB_TOKEN environment variable. If you want to use a COVERALLS_REPO_TOKEN instead, please manually override $COVERALLS_SERVICE_NAME to "github-actions". For more info, see https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/coveralls/api.py", line 298, in submit_report
    response.raise_for_status()
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/coveralls/cli.py", line 98, in main
    result = coverallz.wear()
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/coveralls/api.py", line 275, in wear
    return self.submit_report(json_string)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/coveralls/api.py", line 301, in submit_report
    raise CoverallsException(
coveralls.exception.CoverallsException: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
/opt/hostedtoolcache/Python/3.8.18/x64/bin/coveralls
```

## Expected behavior

I would like coveralls to
- [ ] run during PR and `main` builds
- [ ] provide a comment as it did before about the test coverage

## Additional context

- #644 
[...]

PR #678, opened 2024-07-01
use coveralls' github action

niccokunzmann:
This is another try, using coveralls' github action.

Contributes to #677.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,13 +47,11 @@ jobs:
         pip install tox coveralls coverage-python-version
     - name: Test
       run: tox -e ${{ matrix.config[1] }}
-    - name: Upload coverage data to coveralls.io
-      run: coveralls --service=github || which coveralls
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        COVERALLS_REPO_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        COVERALLS_FLAG_NAME: ${{ matrix.config[1] }}
-        COVERALLS_PARALLEL: true
+    - name: Coveralls Parallel
+      uses: coverallsapp/github-action@v2
+      with:
+        flag-name: run-${{ join(matrix.*, '-') }}
+        parallel: true
 
   coverage:
     # parallel test coverage upload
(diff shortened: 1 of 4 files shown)


niccokunzmann:
@stevepiercy, this might actually be what we should do...

.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         - ["3.9",   "py39"]
         - ["3.10",  "py310"]
         - ["pypy-3.9", "pypy3"]
-        - ["3.10",  "docs"]
+#        - ["3.10",  "docs"]  # disable as readthedocs builds it
```
stevepiercy, line 25: If we do this, then there should be a cross-reference in the `.readthedocs.yaml` file to update this file when its Python version is updated. Also RTD is configured to use 3.11.

```suggestion
        - ["3.10",  "docs"]
```

.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
...
         - ["3.11",  "py311"]
```
stevepiercy, line 26:
```suggestion
#        - ["3.11",  "py311"]  # The commented Python version should be included in `.github/workflows/tests.yml`.
```

stevepiercy (changes requested):
Align Python versions.

.github/workflows/tests.yml
niccokunzmann: I do  not understand this one... the py311 tox environment should be tested.

.github/workflows/tests.yml
niccokunzmann: I disabled this because the docs generation does not create any coverage output. At the moment and with the issue https://github.com/coverallsapp/github-action/issues/216 that I opened, this makes the build run.

.readthedocs.yml
```diff
@@ -15,7 +15,7 @@ formats: all
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.11"  # This Python version should be excluded from `.github/workflows/tests.yml`.
```
niccokunzmann: no. The tests should run under this version `tox -e py311`.
But the `tox -e docs` does not generate coverage information, thus it is excluded for now.
I would revert this.

niccokunzmann: Cool! Thanks for reviewing! I think, if you have any further questions, let me know!

.github/workflows/tests.yml
stevepiercy: Oh, I misinterpreted what was going on. Nevermind. This is fine.

.github/workflows/tests.yml
stevepiercy: Same. This is fine.

stevepiercy (approved):

.readthedocs.yml
stevepiercy: Same here. This is fine.

niccokunzmann:
Thanks for the approval. I merged it. We can have a look now if this works out as we expect it.

PR #681, opened 2024-07-03
Improve rtd pull request preview git diff

stevepiercy:


stevepiercy:
A link to a preview build will not show up in this PR, but will in future PRs.

stevepiercy:
I should note that the link appears in the PR description if and only if there are changed docs when the PR is opened, and not on subsequent pushes to the PR. See https://github.com/readthedocs/actions/issues/41

niccokunzmann:
Cool!

A link shows up in the test status already:
image

.readthedocs.yml
```diff
@@ -24,7 +24,7 @@ build:
     # If there are no changes (git diff exits with 0) we force the command to return with 183.
     # This is a special exit code on Read the Docs that will cancel the build immediately.
     - |
-      if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/main -- docs/ src/icalender/*.py CHANGES.rst CONTRIBUTING.rst LICENSE.rst README.rst .readthedocs.yaml requirements_docs.txt;
+      if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/main -- . docs/ src/icalendar/*.py *.rst .readthedocs.yaml requirements_docs.txt ':!src/icalendar/fuzzing' ':!src/icalendar/tests' ':!src/icalendar/timezone';
```
niccokunzmann, line 27: if you diff `.`, then this is for everything... Ah and I see the exclude... cool. I did not know that was possible. :heart: I learned something!

stevepiercy: Without the `.`, nothing would be included. There may be other improvements, but this is what worked for me.

Issue #614, opened 2024-05-19
Link in Docs Broken


natashamm:
Under https://icalendar.readthedocs.io/en/latest/install.html#running-tests the link  https://tox.wiki/en/latest/example/general.html#selecting-one-or-more-environments-to-run-tests-against to the Tox docs is broken.

stevepiercy:
@natashamm would you please submit a pull request? We'll get it merged. Thank you!

niccokunzmann:
@NP-compete fixed this. Thanks!

PR #688, opened 2024-07-03
Fix broken link in doc.

NP-compete:
Issue: https://github.com/collective/icalendar/issues/614

----
📚 Documentation preview 📚: https://icalendar--688.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,7 @@ New features:
 
 Bug fixes:
 
-- Documentation link fix, see [issue](https://github.com/collective/icalendar/issues/614)
+- Documentation link fix, see issue https://github.com/collective/icalendar/issues/614
 
 6.0.0a0 (2024-06-03)
 --------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,7 @@ New features:
 
 Bug fixes:
 
-- ...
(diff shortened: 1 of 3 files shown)


docs/install.rst
```diff
@@ -64,7 +64,7 @@ To run all tests in all environments, simply run ``tox``
 You may not have all Python versions installed or
 you may want to run a specific one.
 Have a look at the `documentation
-<https://tox.wiki/en/latest/example/general.html#selecting-one-or-more-environments-to-run-tests-against>`__.
+<https://tox.wiki/en/4.16.0/user_guide.html#cli>`__.
```
stevepiercy: This should be the stable version so that it is always current, instead of an outdated release. Also this is not an anonymous link, so only one underscore is needed.

```suggestion
<https://tox.wiki/en/stable/user_guide.html#cli>`_.
```

CHANGES.rst
```diff
@@ -18,7 +18,7 @@ New features:
 
 Bug fixes:
 
-- ...
+- Documentation link fix, see issue https://github.com/collective/icalendar/issues/614
```
```suggestion
- Fix link to stable release of tox in documentation.
```

stevepiercy (changes requested):
@NP-compete thanks for the report and contribution. Please see my comments and suggestions.

niccokunzmann:
Thanks!

PR #695, opened 2024-07-19
Fix test_with_doctest.py to run on Windows.

devdanzin:
On Windows, test_with_doctest.py fails because file and module paths have backslashes on them instead of forward slashes or dots. This PR fixes that by using a forward slash instead of `os.path.join` and by also substituting backslashes by forward slashes to figure out module names.

Happy to address any feedback or to close if it's not a necessary change.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Changelog
 Minor changes:
 
 - add ``__all__`` variable to each modules in ``icalendar`` package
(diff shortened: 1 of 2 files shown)


devdanzin:
> Hi, thanks! Would you add a changelog entry?

Thank you for taking a look, done.

niccokunzmann:
Thanks!

Issue #629, opened 2024-06-17
 Python: Improve test coverage in 5 places for production code


niccokunzmann:
The production code is mostly covered by the tests. Sometimes, `__repr__` or if conditions are not covered though. That would be nice!

To close the issue:
- [ ] cover 5 additional spots in the tests
- [ ] create a pull request with your changes and get it merged

What will help:

- Follow the documented [development setup](https://icalendar.readthedocs.io/en/latest/install.html#development-setup)
- After running the tests, you will have a coverage report as an HTML file.
- Create a Pull Request and we will see the improvements.

devdanzin:
We are working on this. Would you like me to send a PR to add `--branch` to the coverage run options in tox.ini?

niccokunzmann:
> Would you like me to send a PR to add --branch to the coverage run options in tox.ini?

You are welcome to create PRs with improvements or create issues to discuss those first.

niccokunzmann:
@mtr-d3v Thanks for solving this in #696!

PR #696, opened 2024-07-25
Improved test coverage in 5 places for production code

mtr-d3v:
This PR fixes issue #629 by Improving test coverage in 5 places for production code

----
📚 Documentation preview 📚: https://icalendar--696.org.readthedocs.build/

### DIFF
--- a/src/icalendar/tests/prop/test_vBinary.py
+++ b/src/icalendar/tests/prop/test_vBinary.py
@@ -32,7 +32,7 @@ def test_long_data():
     
 def test_repr():
     instance = vBinary("value")
-    assert repr(instance) == "vBinary('b'value'')"
+    assert repr(instance) == "vBinary(b'value')"
     
 def test_from_ical():
     with pytest.raises(ValueError, match='Not valid base 64 encoding.'):
--- a/src/icalendar/tests/prop/test_vBinary.py
+++ b/src/icalendar/tests/prop/test_vBinary.py
@@ -32,7 +32,7 @@ def test_long_data():
     
 def test_repr():
     instance = vBinary("value")
-    assert repr(instance) == "vBinary(b'value')"
+    assert repr(instance) == "vBinary(b'dmFsdWU=')"
     
 def test_from_ical():
     with pytest.raises(ValueError, match='Not valid base 64 encoding.'):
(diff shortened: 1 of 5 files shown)


niccokunzmann:
Thanks!

While looking through it: you found mistakes in the code by improving the coverage. 🥳

could you correct that too or create an issue for it?

vCalAddress('b'value'') repr should be vCalAddress('value')

and vBinary repr probably
vBinary(b'value')

My guess is that changing python 2 to 3 did this.

mtr-d3v:
Thanks for reviewing my PR. I have corrected them.

niccokunzmann:
Could you merge master?

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -27,3 +29,13 @@ def test_long_data():
     txt_ical = b'YWFh' * 33
     assert (vBinary(txt).to_ical() == txt_ical)
     assert (vBinary.from_ical(txt_ical) == txt)
+    
+def test_repr():
+    instance = vBinary("value")
+    assert repr(instance) == "vBinary('b'value'')"
```
niccokunzmann:
```suggestion
    assert repr(instance) == "vBinary(b'value')"
```

src/icalendar/tests/prop/test_vCalAddress.py
```diff
@@ -17,3 +17,8 @@ def test_params():
 
 def test_from_ical():
     assert vCalAddress.from_ical(txt) == 'MAILTO:maxm@mxm.dk'
+    
+
+def test_repr():
+    instance = vCalAddress("value")
+    assert repr(instance) == "vCalAddress('b'value'')"
```
niccokunzmann:
```suggestion
    assert repr(instance) == "vCalAddress('value')"
```

niccokunzmann (approved):
Thanks, I made the changes clearer that I would like to see. Do you think this can be implemented in this pull request?

It is strictly not what the issue said, so I could merge it and create a first-timer issue instead.

What do you think?

mtr-d3v:
I have fixed the issue. Now you can check.

niccokunzmann:
@mtr-d3v Thanks! I just merged master so there is no conflict. It seems that the tests are updated and fail.

Could you [install tox and run the tests in your local setup?](https://icalendar.readthedocs.io/en/latest/install.html#install-tox)

I am wondering why not all the tests were running before ...

niccokunzmann:
Thank you very much!

Issue #698, opened 2024-07-25
 Python: Improve test coverage in 5 places for production code


niccokunzmann:
The production code is mostly covered by the tests. Sometimes, `__repr__` or if conditions are not covered though. That would be nice!

To close the issue:
- [ ] cover 5 additional spots in the tests
- [ ] create a pull request with your changes and get it merged

What will help:

- Follow the documented [development setup](https://icalendar.readthedocs.io/en/latest/install.html#development-setup)
- After running the tests, you will have a coverage report as an HTML file.
- Create a Pull Request and we will see the improvements.

stevepiercy:
@saisharan0103 you've made several comments across multiple issues, but no pull requests, and it's not clear to me why you've made comments that have added no value to the discussion and which smell of AI. Please feel free to open a pull request instead.

PR #699, opened 2024-07-25
Enable collection of branch coverage on test runs

devdanzin:
Having branch coverage information helps when trying to address issues like #698 and #629 (where we used it to find places to test).

This PR adds the `--branch` option to `coverage run` in tox.ini commands.

Possible downsides: test run time will likely increase, coverage numbers will probably go down. Still, the difference should not be huge and it will help improving coverage in the long term.


niccokunzmann:
Cool, thanks!

Issue #686, opened 2024-07-03
upgrade release process from setup.py to most up-to-date way


niccokunzmann:
Python 3.12 removes setuptools. It seems there is a new way forward using `pyproject.toml`.
To close this issue, release this package according to the latest documentation:
https://packaging.python.org/en/latest/guides/section-build-and-publish/

Files to touch:
- https://github.com/collective/icalendar/blob/main/.github/workflows/tests.yml
- https://github.com/collective/icalendar/blob/main/setup.py
- https://github.com/collective/icalendar/blob/main/docs/install.rst
- https://github.com/collective/icalendar/blob/main/docs/maintenance.rst

See also:
- #635

niccokunzmann:
We published v6.0.0 with pyproject.toml and hatchling.

Issue #672, opened 2024-06-30
Use code styler (ruff)


niccokunzmann:
In order for the codebase to have a common style, we can use a linter/formatter like black or ruff.

While black allows us to format the codebase, ruff can enforce more compliance like import order, no prints and more.

To close this issue:

- [x] choose ruff or black
- [x] add a badge
- [x] document how to install the formatter/linter in a pre-commit hook. [example](https://open-web-calendar.quelltext.eu/dev/#code-quality)
- [x] create CI tests to either fail if not formatted or commit formatted code. [example](https://github.com/niccokunzmann/open-web-calendar/blob/master/.github/workflows/update.yml)
- [x] format all code files
- [x] last: remove optional test result, fail on mistakes https://github.com/collective/icalendar/pull/680#discussion_r1663452989
- [x] auto format the PR with ruff and push the commit to the branch using the github token. The CI only fails if there is no push possible OR there are errors that cannot be fixed automatically.

See also:

- #395 
- https://github.com/collective/icalendar/pull/667#discussion_r1660042356

stevepiercy:
This is a good unopinionated FAQ to help choose between tools: https://docs.astral.sh/ruff/faq/#how-can-i-disableforce-ruffs-color-output

I'd go with ruff, since we don't have a linter or formatter at this point, at least none that I know of.

I'd also suggest this issue be done before 6.0.0 final, because it will likely generate a huge diff.

> document how to install the formatter/linter in a pre-commit hook. [example](https://open-web-calendar.quelltext.eu/dev/#code-quality)

I don't see a `.pre-commit-config.yaml` file, so does this statement imply that there needs to be another step for developers in the documentation to install pre-commit?

niccokunzmann:
> I don't see a .pre-commit-config.yaml file, so does this statement imply that there needs to be another step for developers in the documentation to install pre-commit?

correct.

I am ok with ruff. If I create a PR, I am very happy to have contributions to it as it will be a lot of effort. I did it recently for the Open Web Calendar.

stevepiercy:
I would also like contributions to provide a reason for using it as a comment in the configuration.

Line length on narrative text in documentation (not docstrings) does not make any sense, and it makes it difficult to do reviews due to rewrapping text to an arbitrary line length. Who cares? When rendered to HTML or other formats, line wraps go away. One sentence per line is a good guide, and what we use in Plone.

Please do begin with what you have from Open Web Calendar, and let's iterate over it. Thank you!

[...]

PR #707, opened 2024-09-04
Use pyproject.toml

niccokunzmann:
This works on #686 and #635

- [x] python -m build
- [x] plone developers email still in use for this?
- [x] setup.cfg
- [x] tox test dependencies
- [x] In the readme, link to these:
    - "CONTRIBUTING.rst",
    - "CHANGES.rst",
    - "LICENSE.rst"
- from setup.py:
    - [x] license='BSD-2-Clause', (does not exist any more: https://packaging.python.org/en/latest/specifications/pyproject-toml/#license)
    - [x] packages=setuptools.find_namespace_packages('src', exclude=["icalendar.fuzzing"]),
    - [x] package_dir={'': 'src'},
    - [x] include_package_data=True,
    - [x] zip_safe=False
    - [x] test_suite='icalendar.tests'
    - [x] long_description_content_type="text/x-rst",
- [x] ruff in tox
- [x] version.py
- [x] document `tox -e ruff` #672
- [x] replace `__version__` by version.py

I would like to get this done before we release v6

----
📚 Documentation preview 📚: https://icalendar--707.org.readthedocs.build/


setup.py
```diff
@@ -1,67 +0,0 @@
...
-        'Programming Language :: Python :: 3.10',
-        'Programming Language :: Python :: 3.11',
-        'Programming Language :: Python :: 3.12',
-        'Programming Language :: Python :: Implementation :: CPython',
-        'Programming Language :: Python :: Implementation :: PyPy',
-    ],
-    keywords='calendar calendaring ical icalendar event todo journal '
-             'recurring',
-    author='Plone Foundation',
-    author_email='plone-developers@lists.sourceforge.net',
```
niccokunzmann, line 53: @mauritsvanrees, the setup.py file states a plone developers email for the author. I would keep this for now or should we change this. Is it still in use?

niccokunzmann:
After merging this, I would create a new release.

niccokunzmann:
@jacadzaca @geier would you review?

setup.py
stevepiercy: The last message to this list was in 2021.

https://sourceforge.net/p/plone/mailman/plone-developers/

I'd say change it, as you need to subscribe to post to that list anyway. I think it is fine to use your own email, or request an email address from the Plone Foundation, such as icalendar@plone.org or nicco@plone.org. They use Google Workspace for Nonprofits, so there is no cost, other than the setup effort.

mauritsvanrees: Indeed the address is basically unused, and it only lingers in several `setup.py` files like this one.  I usually leave it, because I don't know a better one. :-). It is fine to change.

niccokunzmann: Hm. This email seems to be very much of no value. I will leave it with a comment. Thanks!

jacadzaca (approved):

niccokunzmann:
@jacadzaca Thanks!

Issue #716, opened 2024-10-04
VALARM support


niccokunzmann:
Events and TODOs can contain alarms.
These alarms can be acknowledged.

For https://github.com/niccokunzmann/python-recurring-ical-events/issues/186, in order to calculate which alarms hit within a time span, I would like to get a list from the VEVENTs and VTODOs of the corresponding alarms and whether they are snoozed.

My proposal

```python
from icalendar import Event

event = Event()

# create alarms

event.alarms.acknowledged : list[Alarm]
event.alarms.not_acknowledged : list[Alarm]
event.alarms.all : list[Alarm]
```

According to [RFC 9074](https://www.rfc-editor.org/info/rfc9074), we can snooze or acknowledge an alarm.

```python
@property
Alarm.acknowledged : datetime.datetime # UTC
```

The problem as pointed out by RFC 9074 is that the VALARM component in conjunction with RFC5545 only had no defined way to acknowledge an alarm. Thus, the attributes of event.alarms and todo.alarms will take care of compatibility issues. E.g. Thunderbird saves the dates of the acknowledged alarm time in the VEVENT. Thus this cannot just be a property of VALARM but must reside in the VEVENT.

See also
- #657 

I will open a PR.

niccokunzmann:
A problem that I see with this approach is that `icalendar` works more on the specification level than the semantic level: An event can have an RRULE and RDATE and this will not work with these.
However, this is a good start. Whoever likes, can set DTSTART of the event to another time and then check the alarms - they will only be calculated on the first, core instance of events and TODOs. 
I wonder, how we would like to handle these things in general - I like the interface of the dict to access the specified properties and subcomponents. Any special 'read' functionality can be put in such properties.

niccokunzmann:
It seems to be nicer to have another interface. The problem is this:

- we have relative alarms and we want to be able to compute them for different event starts

What is required to compute an alarm fully:

- an event or component that displays this alarm - so we know what to say
- times at which alarms were acknowledged
- alarm relationships that say if the alarm is snoozed and moved
- either absolute times of the alarm or a start+end of a component relative to which the alarm is computed

The result is:

- a time of the alarm
- a component that the user should be notified about
- a state of the alarm of whether it is active or acknowledged

niccokunzmann:
#737 creates an interface with a functionality that is requested.

Issue #496, opened 2022-11-28
Feature request: calculation of duration and/or dtend/due


tobixen:
According to the RFC, an icalendar VEVENT can have the property DURATION or the property DTEND, but it's not allowed to have both set.  I believe the intention is that an event with DTEND set one hour after the DTSTART is to be considered equivalent with an event with DURATION set to PT30M.

The same applies to tasks, they can have a DURATION or a DUE, but never both.  I suppose the intention there also is that a task with DTSTART and DURATION set should be considered equivalent with a task with DSTART and DUE set (though the tasks are a bit more muddy in my opinion - say there is three hours of paperwork that needs to be done before the end of the year, in my opinion it would make sense to set DUE to 2023-01-01, DURATION to three hours ... but one would probably want to start with it in early December, at least some days before the new years eve).

In the caldav library I now have methods `get_duration()` , `get_due()`, `set_due()` and `set_duration()`, which will attempt to do the right thing - like, `get_duration()` will yield DURATION if it's set, otherwise `DTEND-DTSTART`.  It makes sense generalize it or also add `get_dtend` and `set_dtend`.  Anyway ... this is logic that I need, but I feel it does not belong in the caldav library.  If anything, it belongs in the icalendar library - or perhaps a separate new package like icalendar_util.  (or perhaps there already exists some package like that?)

I will gladly contribute with a pull request on this one, but not without feedback that it's wanted and appropriate to put it into the icalendar library.

niccokunzmann:
I like it. Let's make it happen. Deduplicate code, TDD. I have the same in recurring-ical-events. Please also note the PERIOD value type. It is good to get this straight for everyone.

If we can pass an rdate value, replacement start or end etc to this, that would be much appreciated later. I have PERIOD value type support still open.

It would be nice to use DUE or DTEND or PERIOD for the setter depending on what is used or choose the end property all together.

I mean: while there are different wishes, making it simpler now and better later is probably good. I am happy to support.

tobixen:
The implementation details and the API should be decided upon.

It's not a lot of logic, but still ... there is some common logic that will apply both to `cal.Todo` and `cal.Event`.  Those inherits the `cal.Component` class.  The latter has 7 other child classes that don't have a DURATION (as far as I know ... don't have time to read up the RFC again on this one).

As for the implementation, I see some alternatives:

[...]

PR #733, opened 2024-10-15
VTODO - start, end, duration

niccokunzmann:
This is a contribution to #496.

@tobixen, could you help me understand this a bit better:
What I do not know about VTODO is commented [here](https://github.com/collective/icalendar/issues/496#issuecomment-2413995620), too.
I am unclear about duration computation, too if DUE only or START only is set, it seems there is no duration, then.
In this way, this is very different from an event.
Could we talk about edge cases and such? I do not use VTODO but would like to have that implemented and known so that I can work on #716 and https://github.com/niccokunzmann/python-recurring-ical-events/issues/186.

- [ ] no properties
- [ ] DTSTART only
- [ ]  DUE only
- [ ] COMPLETED only
- [ ] DUE vs. COMPLETED for duration or both
- [ ] DURATION vs. DUE in case of COMPLETED
- [ ] any other attributes that mark when a TODO starts or ends

Your feedback would be much appreciated.

### DIFF
--- a/src/icalendar/tests/test_issue_716_alarm_extension.py
+++ b/src/icalendar/tests/test_issue_716_alarm_extension.py
@@ -1,0 +1,13 @@
+"""Test the alarm classification.
+
+Events can have alarms.
+Alarms can be in this state:
+
+- active - the user wants the alarm to pop up
+- acknowledged - the use does not want that alarm to go off any more
+- snoozed - the user moved that alarm to another time
+
+The alarms can only work on the properties of the event like
+DTSTART, DTEND and DURATION.
+
+"""
--- a/src/icalendar/tests/test_issue_662_component_properties.py
+++ b/src/icalendar/tests/test_issue_662_component_properties.py
@@ -1,4 +1,5 @@
 """This tests the properties of components and their types."""
+from __future__ import annotations
 from datetime import date, datetime, timedelta
 
 import pytest
@@ -13,24 +14,34 @@ from icalendar import (
     IncompleteComponent,
     InvalidCalendar,
     Journal,
+    Todo,
     vDDDTypes,
 )
 from icalendar.prop import vDuration
 
 
-@pytest.fixture()
-def event():
+def prop(component: Event|Todo, prop:str) -> str:
+    """Translate the end property.
+
+    This allows us to run the same tests on Event and Todo.
+    """
+    if isinstance(component, Todo) and prop.upper() == "DTEND":
+        return "DUE"
+    return prop
+
+@pytest.fixture(params=[Event, Todo])
+def start_end_component(request):
     """The event to test."""
-    return Event()
+    return request.param()
 
 @pytest.fixture(params=[
         datetime(2022, 7, 22, 12, 7),
         date(2022, 7, 22),
         datetime(2022, 7, 22, 13, 7, tzinfo=ZoneInfo("Europe/Paris")),
     ])
-def dtstart(request, set_event_start, event):
+def dtstart(request, set_event_start, start_end_component):
     """Start of the event."""
-    set_event_start(event, request.param)
+    set_event_start(start_end_component, request.param)
     return request.param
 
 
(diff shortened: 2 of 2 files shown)


tobixen:
If DURATION is defined in the object and one sets a new DUE, then I'd say the DTSTART should be moved rather than changing the DURATION.

If DTSTART and DUE is defined, and one attempts setting DURATION, then I also think DTSTART also should be moved.

This fits in with my use-cases ... though perhaps it's just my use-cases, I don't have strong opinions that it has to be this way.  As said, I've chosen to define the DURATION as the time estimate, to consider DUE to be either a hard due or wishful thinking on when the task should be completed.  Assuming those writing the standard considered DUE and DTSTART+DURATION to be equivalent, DTSTART is then efficiently the last possible  moment one can start working on the task and theoretically complete it in time - assuming that the person working on the task does not need to sleep etc.

src/icalendar/tests/test_issue_716_alarm_extension.py
```diff
@@ -0,0 +1,13 @@
+"""Test the alarm classification.
+
+Events can have alarms.
+Alarms can be in this state:
+
+- active - the user wants the alarm to pop up
+- acknowledged - the use does not want that alarm to go off any more
```
stevepiercy:
```suggestion
- acknowledged - the user no longer wants the alarm
```

stevepiercy (changes requested):
Minor docstring wording change. Otherwise LGTM.

src/icalendar/cal.py
```diff
@@ -548,6 +548,48 @@ def is_datetime(dt: date) -> bool:
...
+    self.pop("DURATION")
+
+_doc_duration = """The DURATION of {component}.
+
+The "DTSTART" property for a "{component}" specifies the inclusive start of the event.
+The "DURATION" property in conjunction with the DTSTART property
+for a "{component}" calendar component specifies the non-inclusive end
+of the event.
+
+If you would like to calculate the duration a {component}s do not use this.
```
stevepiercy: I think this is what you meant. Please double-check.

```suggestion
If you would like to calculate the duration of a {component}, do not use this.
```

src/icalendar/cal.py
```diff
@@ -548,6 +548,48 @@ def is_datetime(dt: date) -> bool:
...
+Instead use duration property (lower case).
```
```suggestion
Instead use the duration property (lower case).
```

CHANGES.rst
```diff
@@ -25,14 +25,17 @@ Bug fixes:
 
 New features:
 
-- Added ``Event.end``, ``Event.start``, ``Event.dtstart``, and ``Event.dtend`` attributes. See `Issue 662 <https://github.com/collective/icalendar/issues/662>`_.
+- Added ``end``, ``start``, ``duration``, ``DTSTART``, ``DUE``, and ``DURATION`` attributes to ``Event`` components. See `Issue 662`_.
+- Added ``end``, ``start``, ``duration`` and, ``DTSTART`` attributes to ``Journal`` components. See `Issue 662`_.
```
```suggestion
- Added ``end``, ``start``, ``duration``, and ``DTSTART`` attributes to ``Journal`` components. See `Issue 662`_.
```

niccokunzmann:
@tobixen Could you give this a review?

Here, I basically use the same algorithms in Event start & end for Todo start & end.

Altough they are different, I wonder how to deal with the differences.

There are two things to notice:

- DTSTART, DUE and DURATION are the properties, allowing invalid states
-  start, end and duration are the calculated values with more verification, making sure the component is complete and not contradictory when calculated

I went through your requirements again.  

Edge case to consider: For VTODO, DURATION seems to be ok to be set by itself without DTSTART. Is that true? In that case, we are not correct here for the `Todo.duration` attribute.

niccokunzmann:
@stevepiercy, thanks for the review. I applied the changes :)

niccokunzmann:
Readthedocs: https://icalendar--733.org.readthedocs.build/en/733/

niccokunzmann:
In the end, and if it makes sense, I would like to have start, end and duration properties on all components. Missing to evaluate:

- FreeBusy

niccokunzmann:
@tobixen I will merge it for now. Please use it one day and create a PR in case you notice how to improve it.
@stevepiercy Thanks for your review. I added your suggestions.

Over a week has passed and I would like this functionality for #737.

PR #741, opened 2024-11-02
This adds generating VTIMEZONE components from tzinfo objects

### LINKED ISSUE #722: Generate VTIMEZONE components
The `icalendar` can be enhanced by allowing the generation of the VTIMEZONE component automatically. The problem we are facing is recorded here, too:

- https://github.com/mergecal/python-mergecal/issues/7
- https://github.com/niccokunzmann/x-wr-timezone/issues/4

When someone creates an `.ics` file, they want to have it easy to create a valid file #596 and that includes VTIMEZONE information. The timezone information is given by zoneinfo/tzdata and can be translated into a VTIMEZONE component. However, this is not implemented in this package. May people would benefit from that. 

Maybe, in this ecosystem, there is someone who has already implemented this. This could be evaluated:
- [ ] https://pypi.org/project/ics-vtimezones/
- [ ] https://www.djangosnippets.org/snippets/10569/
- [ ] https://github.com/libical/vzic - looks promising
    - [ ] https://www.npmjs.com/package/timezones-ical-library - does what we want just in node - https://tz.add-to-calendar-technology.com/ easy enough.

See also:
- #596
- https://github.com/mergecal/python-mergecal/issues/7
- https://github.com/niccokunzmann/x-wr-timezone/issues/4
- https://www.iana.org/time-zones

niccokunzmann:
Fixes #722 

- Add `icalendar.Timezone.from_tzinfo(ZoneInfo('Europe/Berlin'))`
- Add `icalendar.Calendar.add_missing_timezones()`
- Identify totally unknown timezones (e.g. custom ones from ics strings)
- make sure `from_ical().to_ical()` creates equivalent content 

-----

TODOs:

- [x] check that transition times are computed properly for pytz
    Result: This is really tricky. I do not know why pytz does not converge or why it is weird around the transition times.
- [x] create Calendar.add_missing_timezones
- [x] check with actual transition times
    Two tests check this. When we parse a calendar, at least we get the right transition times in pytz and zoneinfo.
- [x] issue for test_deep_copies_are_equal[issue_722_timezone_transition_ambiguity.ics-zoneinfo]
- [x] remove todos


niccokunzmann:
I run a bit into an issue here.

I can generate them successfully for zoneinfo/dateutil.
However, when I use pytz, I run into a strange error (VTIMEZONE component != VTIMEZONE component)

    pytz -> VTIMEZONE -> pytz -> VTIMEZONE

changes the transition times.

Here VTIMEZONE == VTIMEZONE

    zoneinfo -> VTIMEZONE -> dateutil -> VTIMEZONE

Since these yield the same result (VTIMEZONE == VTIMEZONE):

    pytz -> VTIMEZONE
    zoneinfo -> VTIMEZONE

I conclude that `VTIMEZONE -> pytz` is actually a bit problematic. I will check this. In this case, `icalendar` had a bug around the transition day that the time was moved a little. e.g.

- Europe/Berlin transitions at 3:00am/2:00am and our pytz implementation turns it into 4:00/0:00
- Amerika/New_York transitions at 3:00am/1:00am and our pytz implementation turns it into 4:00/0:00
- Asia/Singapore transitions at 19820101T000000 and our pytz implementation turns it into 19820101T003000 = +30min

Next steps:

- [ ] isolate the cause of the difference
- [ ] fix the mistake if it is ours

So, in case I can  this and you depend on pytz, I still recommend that you upgrade

niccokunzmann:
I think, the interface I would like to use is this one:

    Calendar().add_missing_timezones()

This will then add all the missing timezones automatically to the calendar.

src/icalendar/cal.py
```diff
@@ -1057,6 +1096,149 @@ def get_transitions(self) -> Tuple[List[datetime], List[Tuple[timedelta, timedel
...
+        """Return a VTIMEZONE component from a timezone object.
+
+        This works with pytz and zoneinfo and any other timezone.
+        The offsets are calculated from the tzinfo object.
+
+        Parameters:
+
+        :param tzinfo: the timezone object
+        :param tzid: the tzid for this timezone in case we cannot determine it
+                 None for pytz and zoneinfo is fine.
```
stevepiercy: This param description doesn't make sense to me. Is it two sentences?

stevepiercy (changes requested):
LGTM, except for one param description.

niccokunzmann:
The pytz parsing seems to fail when we check two calendars for equality.

```
   def assert_equal(actual_value, expected_value):
        """Make sure both values are equal"""
>       assert actual_value == expected_value
E       AssertionError: assert VCALENDAR({},...TE-TIME'}))})) == VCALENDAR({},...TE-TIME'}))}))
E         
E         Use -v to get more diff

src/icalendar/tests/test_equality.py:33: AssertionError
---------------------------- Captured stdout setup -----------------------------
example file: issue_722_timezone_transition_ambiguity.ics
```

src/icalendar/cal.py
niccokunzmann: I changed the description.
https://github.com/collective/icalendar/pull/741/files#diff-209244e93cf08598138afbd0965d624109c7a013acabe550af7322aa36078f30R1131

src/icalendar/tests/test_issue_722_generate_vtimezone.py
```diff
@@ -0,0 +1,408 @@
...
+    assert generated_dt.tzname() == expected_dt.tzname() == tzname, message
+    assert generated_dt.dst() == expected_dt.dst(), message
+    assert generated_dt.utcoffset() == expected_dt.utcoffset(), message
+
+
+@pytest.mark.parametrize(
+    "uid", [0, 1, 2, 3]
+)
+def test_dateutil_timezone_when_time_is_going_backwards(calendars, tzp, uid):
+    """Sometimes, times can be ambiguous: We go back from 3am to 2 am.
```
stevepiercy: Although those who work with datetime and DST shifts have the experience of being bitten by this, we should explain it in detail. The explanation might also be helpful in the narrative docs, but I'm not sure where. There are also some really good PyCon videos and talks that discuss working with datetime.

```suggestion
    """When going from Daylight Saving Time to Standard Time, times can be ambiguous.
    For example, at 3:00 AM, the time falls back to 2:00 AM, repeating a full hour of times from 2:00 AM to 3:00 AM on the same date.
```

src/icalendar/tests/test_issue_722_generate_vtimezone.py
```diff
@@ -0,0 +1,408 @@
...
+
+    By the RFC 5545, we cannot accommodate this case. All datetimes should
+    be BEFORE the transition if ambiguous. However, dateutil can
+    create a timezone that allows being after of cause.
```
stevepiercy: I'm not sure what this sentence was trying to say, so this is my best guess. Please alter as needed.

```suggestion
    create a timezone that allows the event to be after this ambiguous time span, of course.
```

src/icalendar/cal.py
```diff
@@ -1116,10 +1345,126 @@ class Calendar(Component):
...
+        for name, value in self.property_items(sorted=False):
+            print(name, value)
+            if hasattr(value, "params"):
+                result.add(value.params.get("TZID"))
+        return result - {None}
+
+    def get_missing_tzids(self) -> set[str]:
+        """The set of missing timezone component tzids.
+
+        In order to createa :rfc:`5545` compatible calendar,
```
```suggestion
        To create a :rfc:`5545` compatible calendar,
```

stevepiercy (changes requested):
A few grammar fixes, and comments.

niccokunzmann:
Thank you for your reviews! I think this is ready now for the last review. The most important functionality has been added which allows us to add missing VTIMEZONE components to the calendar.

stevepiercy (approved):

PR #746, opened 2024-11-21
Tidy up the docstring for ``icalendar.prop.vPeriod``

stevepiercy:
See preview:

https://icalendar--746.org.readthedocs.build/en/746/api.html#icalendar.prop.vPeriod

----
📚 Documentation preview 📚: https://icalendar--746.org.readthedocs.build/


stevepiercy:
I just reformatted all the docstrings. It was a total mess. Now it is legible as HTML.

However, we should consider adopting a docstring standard, and document that adoption. I recommend Google docstring, as describe in this post:

https://stackoverflow.com/a/24385103/2214933

niccokunzmann:
Hi that looks good!

Issue #749, opened 2024-11-23
[BUG] vWeekday does not make `sign`, `weekday` available


david-venhoff:
## Describe the bug

icalendar currently does not make it possible to parse recurrence rules containing something like `BYDAY=-1SU`. To get the frequency of the recurrence from this value, one would have to manually parse `-1SU`. 
That is error prone, and it looks like icalendar [already does this](https://github.com/collective/icalendar/blob/c59966c7d9cdb58b930662870e9f938f83ba7bf4/src/icalendar/prop.py#L928-L931), but then throws away the result

## To Reproduce

- parse an icalendar with a recurrence rule containing `BYDAY=-1SU`
- See that there is no way to access `sign` and `weekday` in the resulting `vWeekday`. `relative` is made available, however without its sign.

## Expected behavior

`vWeekday` should make these data public

## Environment

 - [x] OS:                               os independent
 - [x] Python version:            python independent
 - [x] `icalendar` version: 6.1.0

niccokunzmann:
Thanks for reporting this! I estimate this to be a small change to fix. Could you create a pull request with a test for this?

david-venhoff:
Yes, I can do that :)

PR #750, opened 2024-11-25
Add `sign` and `weekday` attributes to `vWeekday` components

david-venhoff:
This pr adds the `sign` and `weekday` properties to `vWeekday` components, so that users of this library don't have to manually parse these. 

I decided to just make the existing variables available, but I could also combine `sign` and `relative` into one variable, if that is wanted.

Fixes: #749 

----
📚 Documentation preview 📚: https://icalendar--750.org.readthedocs.build/

### DIFF
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -931,6 +931,8 @@ class vWeekday(str):
         relative = match['relative']
         if weekday not in vWeekday.week_days or sign not in '+-':
             raise ValueError(f'Expected weekday abbrevation, got: {self}')
+        self.sign = sign or None
+        self.weekday = weekday or None
         self.relative = relative and int(relative) or None
         self.params = Parameters()
         return self
--- a/src/icalendar/prop.py
+++ b/src/icalendar/prop.py
@@ -913,7 +913,32 @@ class vPeriod(TimeBase):
 
 
 class vWeekday(str):
-    """This returns an unquoted weekday abbrevation.
+    """Either a ``weekday`` or a ``weekdaynum``
+
+    .. code-block:: pycon
+
+        >>> from icalendar import vWeekday
+        >>> vWeekday("MO") # Simple weekday
+        'MO'
+        >>> vWeekday("2FR").relative # Second friday
+        2
+        >>> vWeekday("2FR").weekday
+        'FR'
+        >>> vWeekday("-1SU").relative # Last Sunday
+        -1
+
+    Definition from RFC:
+
+    .. code-block:: text
+
+        weekdaynum = [[plus / minus] ordwk] weekday
+        plus        = "+"
+        minus       = "-"
+        ordwk       = 1*2DIGIT       ;1 to 53
+        weekday     = "SU" / "MO" / "TU" / "WE" / "TH" / "FR" / "SA"
+        ;Corresponding to SUNDAY, MONDAY, TUESDAY, WEDNESDAY, THURSDAY,
+        ;FRIDAY, and SATURDAY days of the week.
+
     """
     week_days = CaselessDict({
         "SU": 0, "MO": 1, "TU": 2, "WE": 3, "TH": 4, "FR": 5, "SA": 6,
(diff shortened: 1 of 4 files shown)


src/icalendar/prop.py
```diff
@@ -931,6 +931,8 @@ def __new__(cls, value, encoding=DEFAULT_ENCODING):
         relative = match['relative']
         if weekday not in vWeekday.week_days or sign not in '+-':
             raise ValueError(f'Expected weekday abbrevation, got: {self}')
+        self.sign = sign or None
+        self.weekday = weekday or None
         self.relative = relative and int(relative) or None
```
niccokunzmann, line 960: Hm. I wonder: -1 is also relative. When I use this, I think it could be good to make self.relative negative or positive.

src/icalendar/tests/prop/test_vWeekday.py
```diff
@@ -0,0 +1,29 @@
...
+    assert weekday.weekday == "SU"
+    assert weekday.relative is None
+
+
+def test_relative():
+    weekday = vWeekday("-1MO")
+    assert weekday.to_ical() == b"-1MO"
+    assert weekday.sign == "-"
+    assert weekday.weekday == "MO"
+    assert weekday.relative == 1
```
niccokunzmann: Like this could be -1. I think, I would expect this.

```python
TH - relative is None
+1TH - relative is +1
-2MO - relative is -2
```

src/icalendar/tests/prop/test_vWeekday.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from icalendar.prop import vWeekday
+
+
+def test_simple():
+    weekday = vWeekday("SU")
```
niccokunzmann, line 7: I like these examples from the tests.
We also have doctests.
If you add some examples to class definition, this will help people understand your fix and how to use this class.

Like:
```
>>> su = vWeekday("SU") # Sunday
>>> su
>>> su.weekday
>>> su.relative
>>> su.to_ical()

>>> th = vWeekday("2TH")  # second Thursday of the month?
>>> th
>>> th.weekday
>>> th.relative
>>> th.to_ical()

>>> fr = vWeekday("-1FR")  # last Friday of the month?
>>> fr
>>> fr.weekday
>>> fr.relative
```

I am writing this here because I would like your input: What does -1 mean in your view? Is it the last weekday of the month?

niccokunzmann (changes requested):
Hi, I like your PR! I left a bot of feedback and I am wondering if the sign should be integrated into the relative variable. What are your thoughts on this?

Also, I am wondering what the sign means and since you care about the sign, I would hope you know and we can document it so others know how to use the fix.

Let me know what you think!
Thanks for the tests!

src/icalendar/prop.py
david-venhoff: Yes, I agree. I just did not do it because I was not sure if it was okay to introduce a backwards incompatible change for this.

src/icalendar/tests/prop/test_vWeekday.py
david-venhoff: Yes. For example, this rrule would repeat monthly on every last friday: `FREQ=MONTHLY;BYDAY=-1FR`. I think using a relative BYDAY value is also allowed with a yearly frequency ([rfc](https://www.rfc-editor.org/rfc/rfc5545.txt), page 42)

niccokunzmann: It might be nice to extend the rrule documentation, too. I would say though that I open another issue for that.

david-venhoff:
I have one more question: According to rfc 5545, the `BYDAY` attribute for recurrence rules can be followed by a `bywdaylist`. A `bywdaylist` can contain `weekdaynum`, and a `weekdaynum` is defined like this: `weekdaynum  = [[plus / minus] ordwk] weekday`. 
However, some other attributes (e.g. `WKST`) can only be followed by a simple `weekday` (without number prefix).

Wouldn't it make sense to have two different classes for weekdays then? One that only stores a weekday and one that stores a weekday and a number.

src/icalendar/tests/prop/test_vWeekday.py
david-venhoff: I have updated the docstring and added some doctests now :)

niccokunzmann:
I think, this is good. One can always create invalid files. this package does not check validity.

Issue #743, opened 2024-11-11
Create an example for a component (Python)


niccokunzmann:
To enhance the documentation, it is nice to have examples!

- Todo
    - [x] add src/icalendar/tests/todos/example.ics (or a different name)
    - [x] add example() method with default
    - [x] show how to use the example in the class docstring
    - [x] test in src/icalendar/tests/test_examples.py
- Journal
    - [ ] add src/icalendar/tests/journals/example.ics (or a different name)
    - [ ] add example() method with default
    - [ ] show how to use the example in the class docstring
    - [ ] test in src/icalendar/tests/test_examples.py
- FreeBusy
    - [ ] add src/icalendar/tests/freebusy/example.ics (or a different name)
    - [ ] add example() method with default
    - [ ] show how to use the example in the class docstring
    - [ ] test in src/icalendar/tests/test_examples.py
- Alarm
    - [x] add src/icalendar/tests/alarms/example.ics (or a different name)
    - [x] add example() method with default
    - [x] show how to use the example in the class docstring
    - [x] test in src/icalendar/tests/test_examples.py
- Event
    - [x] add src/icalendar/tests/events/example.ics (or a different name)
    - [x] add example() method with default
    - [ ] show how to use the example in the class docstring
    - [x] test in src/icalendar/tests/test_examples.py
- Calendar
    - [x] add src/icalendar/tests/calendars/example.ics (or a different name)
    - [x] add example() method with default
    - [ ] show how to use the example in the class docstring
    - [x] test in src/icalendar/tests/test_examples.py
- Timezone
    - [x] add src/icalendar/tests/timezones/example.ics (or a different name)
    - [x] add example() method with default
    - [ ] show how to use the example in the class docstring
    - [x] test in src/icalendar/tests/test_examples.py

This is an issue for beginners.
Please open a Pull Request to let us know you are working on it or ask below if you need any help!
A pull request should address at least one item in the list above. No need to do all at once.

Please also have a look at:

- https://icalendar.readthedocs.io/en/latest/
- https://github.com/collective/icalendar/tree/main/src/icalendar/tests
- https://www.rfc-editor.org/info/rfc5545 has the specification and examples

niccokunzmann:
@prakrit55 thanks, yes you are very welcome to! Let me know if you need anything. There is no need to solve all of this at once.

prakrit55:
@niccokunzmann, do we need to create the class Docstring

niccokunzmann:
[...]

PR #753, opened 2024-11-30
docs: Create an example for a component

prakrit55:
fixing #743


prakrit55:
@niccokunzmann, we can have a discussion here.

niccokunzmann:
This is the right place. Now, could you copy in an example from the rfc?

prakrit55:
> This is the right place. Now, could you copy in an example from the rfc?

is there any way to generate rfc files, could you mention any way to do this

stevepiercy:
@prakrit55 RFC stands for [Request for Comments](https://en.wikipedia.org/wiki/Request_for_Comments). The issue to be resolved would be to copy an example from the specific [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545.html) for one of the given components, and write a Python doctest for it in its method.

I'm not sure why the issue todo for each of the three checked `add src/icalendar/tests/todos/example.ics` is checked, because I don't see the [`todo` folder in the repo](https://github.com/collective/icalendar/tree/main/src/icalendar/tests) where it says it should be.

@niccokunzmann perhaps you could update the issue description with some more details? I'm not clear myself.

prakrit55:
> @prakrit55 RFC stands for [Request for Comments](https://en.wikipedia.org/wiki/Request_for_Comments). The issue to be resolved would be to copy an example from the specific [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545.html) for one of the given components, and write a Python doctest for it in its method.
> 
> I'm not sure why the issue todo for each of the three checked `add src/icalendar/tests/todos/example.ics` is checked, because I don't see the [`todo` folder in the repo](https://github.com/collective/icalendar/tree/main/src/icalendar/tests) where it says it should be.
> 
> @niccokunzmann perhaps you could update the issue description with some more details? I'm not clear myself.

hello @stevepiercy, thanks for the reply, does the repo have any examples you can point me to

niccokunzmann:
@prakrit55, could you have a look at the documentation and the file structure and the issue - the issue points towards

- [x] what has been done already

Then, please ask again.

@stevepiercy, I will write more into the issue now.

niccokunzmann:
Thanks for your effort. I will close this now. Feel free to open an new one.

PR #755, opened 2024-12-02
Add Security Policy

niccokunzmann:
This adds a security policy as reguired by tidelift

See https://icalendar--755.org.readthedocs.build/en/755/security.html

- Contributes to #754

After the merge, I will activate the policy on GitHub here: https://github.com/collective/icalendar/security

----
📚 Documentation preview 📚: https://icalendar--755.org.readthedocs.build/


docs/security.rst
stevepiercy, line 1: For a security policy, see examples at:

- https://github.com/Pylons/.github/blob/main/SECURITY.md # I've seen this in action, and the maintainers have a very professional response.
- https://github.com/plone/.github/blob/main/SECURITY.md # @mauritsvanrees can speak Plone ones.
- https://github.com/plone/Products.CMFPlone/blob/master/.github/SECURITY.md

It's really up to you what you want to include. If you have another revision after reviewing the above items, please `@` me and I'll take another look. I'd suggest incorporating some bits and pieces from the above, especially how to work with CVEs.

stevepiercy (commented):
See comment.

docs/security.rst
niccokunzmann: Thank you for the links! I changed it to link to the Plone security page.

I will also write to the security team to ask if they have any feedback on this change.

docs/security.rst
```diff
@@ -0,0 +1,34 @@
+Security Policy
+===============
+
+This documents the security policy and actions to take to secure the security of the package, its deployment and use.
```
stevepiercy: Too many "security"s.

```suggestion
This documents the security policy and actions to take to secure the package and its deployment and use.
```

docs/security.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Supported Versions
+------------------
+
+Security vulnerabilities are fixed only for the latest version of `icalendar`.
```
stevepiercy: Use reStructuredText syntax for inline literal.

```suggestion
Security vulnerabilities are fixed only for the latest version of ``icalendar``.
```

docs/security.rst
```diff
@@ -0,0 +1,34 @@
...
+     - ❌
+   * - < 4.*
+     - ❌
+
+
+Reporting a Vulnerability
+-------------------------
+
+Please `report vulnerabilities of icalendar to Plone
+<https://github.com/plone/.github/blob/main/SECURITY.md#readme>`_.
```
stevepiercy: Do not refer to links that refer to links, especially if you cannot honor the policies to which they refer. You should instead grab the relevant bits that you want to follow and incorporate them into your custom security policy.

As this states, eventually people may report the security issue to security@plone.org. Are you a member of that email group? If not, you would need to discuss with them how to handle reports for icalendar sent to that address. It may be better to use another address.

docs/security.rst
```diff
@@ -0,0 +1,34 @@
...
+If you cannot do this, please contact one of the
+:ref:`maintainers`
+directly or open an issue.
```
stevepiercy: You should not ask people to open a security issue in a public arena, such as an issue tracker. This also conflicts with the link above, which links to https://plone.org/security/report.

If you want to create an email distribution group, and add members to it to review security, I'd suggest following Pylon Project's example as in https://groups.google.com/g/pylons-project-security/ or other free email group distribution list service that is less Google-y.

docs/security.rst
```diff
@@ -0,0 +1,34 @@
...
+   * - Version
+     - Supported
+   * - 6.*
+     - ✅
+   * - 5.*
+     - ❌
+   * - 4.*
+     - ❌
+   * - < 4.*
+     - ❌
```
stevepiercy: For accessibility, I'm not a fan of emoji. Better to use "yes" and "no" for screen reader happiness.

stevepiercy (changes requested):

docs/security.rst
niccokunzmann: We can activate GitHub's security reporting for this project.

I wonder: If a vulnerability gets reported on GitHub, then the other Plone packages do not know it exists.
If a vulnerability is reported to Plone, then I do not know it was reported.

Thus, I wonder what to do. We have two different groups here that are affected. I asked the security team to enlighten us here.

Also, nobody reported anything like this for years and it is unlikely to happen. So. I wonder if this is hypothetical. But a report should also not go into the void after a few years.

stevepiercy: > We can activate GitHub's security reporting for this project.

To enable GitHub security, you create a `SECURITY.md` file at the root of your repo. GitHub should automatically detect it. It's nothing magical. It's mere documentation of how to report security issues, and how you will respond to them.

> I wonder: If a vulnerability gets reported on GitHub, then the other Plone packages do not know it exists. If a vulnerability is reported to Plone, then I do not know it was reported.
> 
> Thus, I wonder what to do. We have two different groups here that are affected. I asked the security team to enlighten us here.

I would follow Pylons Project's workflow, modified to your preference.

https://github.com/Pylons/.github/blob/main/SECURITY.md

To summarize:

1. Users report security issues responsibly and in a manner that allows maintainers to respond in a timely and effective manner.
2. Maintainers verify and fix.
3. All the bullet points at that link.

I have no insight into the Plone Security Team's internal process. I'm not a member, and it's not transparent. I never received a report back of what they did for an issue that I recently reported to them.

> Also, nobody reported anything like this for years and it is unlikely to happen. So. I wonder if this is hypothetical. But a report should also not go into the void after a few years.

It's only hypothetical until it's not, and then it's too late.

niccokunzmann: Thanks for all the replies.

GitHub also allows security reporting. This is how it looks from the outside from the Open Web Calendar, my project. There is a button to report.
grafik

I like the one you showed me (Pylons). This is also nice to follow for someone (me) who did not go through this process yet but is in a position of responsibility.

- I will copy that in.
- change the links
- use githubs security report feature - mainly because I do not want to maintain an independent list of who gets notified - it should be the GitHub project's maintainers.
- I would add that we also notify the Plone security team.

docs/security.rst
niccokunzmann: I would opt for the GitHub process. This way, we do not need to maintain another list.

niccokunzmann:
@stevepiercy Thanks for all that help! Please review again :heart: 

- [ ] As of now, once merged, we need to activate this green button on the project.
- [ ] Also, I will wait for at least two weeks to give Plone's security team a chance to review before merge. (until Monday 6th Jan 2025)

docs/security.rst
stevepiercy: I did not know that GitHub had that feature. Your plan sounds good to me.

docs/security.rst
```diff
@@ -0,0 +1,58 @@
...
+   * - 4.*
+     - no
+   * - before 4.*
+     - no
+
+
+Reporting a Vulnerability
+-------------------------
+
+To report security issues of ``collective/icalendar``, use the ``Report a vulnerability`` button on the project's `Security Page <https://github.com/collective/icalendar/security>`_.
```
stevepiercy, line 30: Can you not create a direct link to Report a vulnerability, instead of going first to the Security page? Hm, I guess not yet, but in the next iteration, it would be better to do this in one step, instead of two.

stevepiercy (approved):
Looks great! Only question is about the direct link, but I assume that can come after this is merged.

docs/security.rst
niccokunzmann: Ok. After merge then :)

stevepiercy:
@plone/security-team please review.

niccokunzmann:
@stevepiercy, thanks for the review. There was enough time now to comment. I updated the branch and will merge it.
I opened #771 so we can track the state of all the tasks.

> @plone/security-team

This seems not to exist directly.

PR #757, opened 2024-12-09
Add from_file and to_file method

### LINKED ISSUE #756: Allow to read calendars from files
In order to make it nicer to use the icalendar interface, we can add these two types to `icalendar.Calendar.from_ical()`:

- `pathlib.Path`
- `str`

In both cases, we assume a path is given.

## Implementation suggestion 

- [ ] translate `str` into `Path`
- [ ] use `Path.read_bytes()` to get the content that would usually be passed to the `from_ical()`
- [ ] document all the parameters, see #747

abe-101:
closes #756


src/icalendar/cal.py
```diff
@@ -587,6 +588,74 @@ def is_thunderbird(self) -> bool:
...
+
+        Raises:
+            FileNotFoundError: If the file path doesn't exist
+            ValueError: If the file contents are not valid iCalendar format
+
+        Example:
+            >>> from icalendar import Calendar
+            >>> # Read a calendar file
+            >>> cal = Calendar.from_file("src/icalendar/tests/calendars/example.ics")
+            >>> # Read multiple calendars
```
niccokunzmann, line 619: You can remove the >>> here

src/icalendar/cal.py
```diff
@@ -587,6 +588,74 @@ def is_thunderbird(self) -> bool:
...
+            file: Where to write the component. Can be:
+                - A string path to a file
+                - A Path object
+            sorted: Whether parameters and properties should be lexicographically sorted.
+
+        Example:
+            >>> from icalendar import Calendar
+            >>> from pathlib import Path
+            >>> # Read a calendar file
+            >>> cal = Calendar.from_file("src/icalendar/tests/calendars/example.ics")
```
niccokunzmann: The examples here seem to be the same as above. You can use a temporary file for this.

niccokunzmann (commented):
Hi @abe-101, I left a few comments. What are your thoughts?

niccokunzmann:
Hi @abe-101, thanks for your PR!

> Are the Examples in the methods doc strings ran as part of the test suite?

Yes, if they are behind a `>>> `

> Should i add a section in usage.rst and how to use the new method?

No, I think a proper API documentation is enough for now.

> Or better, perhaps i should change the given examples that open a file to instead use the new from_file() method? and seporatly demonstrate how to use from_ical() to parse a str?

If you feel like it, a new PR might be the right place for that. Let's finish this one...

src/icalendar/cal.py
abe-101: Are you suggestion to remove the `>>>` on the lines where there is a comment? or perhaps just remove the comments altogether?

src/icalendar/cal.py
```diff
@@ -587,6 +588,74 @@ def is_thunderbird(self) -> bool:
...
+            If multiple=True:
+                A list of Component instances
+
+        Raises:
+            FileNotFoundError: If the file path doesn't exist
+            ValueError: If the file contents are not valid iCalendar format
+
+        Example:
+            >>> from icalendar import Calendar
+            >>> # Read a calendar file
```
niccokunzmann, line 617:
```suggestion

            # Read a calendar file
```

src/icalendar/cal.py
niccokunzmann: https://github.com/collective/icalendar/pull/757/files/804ab5297092428c2ec0c440f28d50db5ac686ba#r1878247067

abe-101: if I remove the `>>>` from the comments then the test fails
This doesn't work:
```
        Example:
            >>> from pathlib import Path
            # Read a calendar file
            >>> cal = Calendar.from_file("src/icalendar/tests/calendars/example.ics")
            # Read multiple calendars
            >>> cals = Calendar.from_file("src/icalendar/tests/calendars/multiple_calendar_components.ics", multiple=True)
            # or pass a Path object
            >>> path = Path("src/icalendar/tests/calendars/example.ics")
            >>> cal = Calendar.from_file(path)
```
This does work:
```
        Example:
            >>> from pathlib import Path
            >>> # Read a calendar file
            >>> cal = Calendar.from_file("src/icalendar/tests/calendars/example.ics")
            >>> # Read multiple calendars
            >>> cals = Calendar.from_file("src/icalendar/tests/calendars/multiple_calendar_components.ics", multiple=True)
            >>> # or pass a Path object
            >>> path = Path("src/icalendar/tests/calendars/example.ics")
            >>> cal = Calendar.from_file(path)
```

src/icalendar/cal.py
abe-101: I changed the examples to use the `to_file()` method

niccokunzmann:
Hi, could you write a test for that?

abe-101:
> Hi, could you write a test for that?

Sure.
Just wondering why the inline docs str tests are not enough?

niccokunzmann:
Hi, thanks! I think that the doctests are not enough.

What needs to happen in testing is this:

- [ ] write a file with this method, check the file is there and has string content
- [ ] write a file with this method and read it with this method, check that the objects are equal

For now, not having any code in the method would also pass the doctests.

src/icalendar/cal.py
niccokunzmann: It needs a new line:


```
>>> from pathlib import Path

# Read a calendar file
```

Otherwise the comment is considered output.

niccokunzmann:
@abe-101, I will close this for now. I think, it is worth documenting how to read from files but we do not need an extra function for that.

niccokunzmann:
Closing as there is no progress and the task changed slightly.
Thank you @abe-101 for contributing!

Issue #758, opened 2024-12-19
Add rrule from a string


fluxxcode:
We have a rrule string that we want to add to an event.

At the moment this doesn't seem possible because the semicolons in the rrule strings are escaped with a backslash, invalidating the rrule.

Here is an example:
```python
import icalendar

rrule_str = "FREQ=YEARLY;INTERVAL=10;BYMONTH=12"

event = Event()

# Both result in the same output
event.add("rrule", rrule_str)
event["rrule"] = rrule_str
```

Calling `.to_ical` escapes the rrule: `RRULE:FREQ=YEARLY\;INTERVAL=10\;BYMONTH=12`, making it invalid.

I think something like `event.add_unescaped` would be helpful.

niccokunzmann:
Hi, thanks! It should somehow be possible to do that.

I would look for something like this:

```python
from icalendar import vRecur

event["rrule"] = vRecur.from_ical("FREQ=YEARLY;INTERVAL=10;BYMONTH=12")
```

If you find a working example, could you add it to the documentation?
https://icalendar.readthedocs.io/en/latest/api.html#icalendar.prop.vRecur

fluxxcode:
Thank you for the answer!

```python
event["rrule"] = vRecur.from_ical("FREQ=YEARLY;INTERVAL=10;BYMONTH=12")
````

works perfectly. Will create a PR for an example in the documentation.

Issue #742, opened 2024-11-10
Document Properties with Text from RFC 5545


niccokunzmann:
The RFC 5545 describes all the properties we have in icalendar.prop.

Please add a `__doc__` string documentation for all of them.

Examples:

- [vMonth](https://icalendar.readthedocs.io/en/latest/api.html#icalendar.prop.vMonth)

```python
class vGeo:
    """Geographic Position

    Property Name:  GEO

    Purpose:  This property specifies information related to the global
        position for the activity specified by a calendar component.

    Value Type:  FLOAT.  The value MUST be two SEMICOLON-separated FLOAT
        values.

    Property Parameters:  IANA and non-standard property parameters can
        be specified on this property.

    Conformance:  This property can be specified in "VEVENT" or "VTODO"
        calendar components.

    Description:  This property value specifies latitude and longitude,
        in that order (i.e., "LAT LON" ordering).  The longitude
        represents the location east or west of the prime meridian as a
        positive or negative real number, respectively.  The longitude and
        latitude values MAY be specified up to six decimal places, which
        will allow for accuracy to within one meter of geographical
        position.  Receiving applications MUST accept values of this
        precision and MAY truncate values of greater precision.

    Example:
        GEO:37.386013;-122.082932

        >>> geo = vGeo.from_ical('37.386013;-122.082932')
        >>> geo.latitude
        37.386013
        >>> geo.longitude
        -122.082932
    """
```

This includes:

- heading
- description that gives a rough overview
- example from the RFC
- example code for icalendar

Please add one of these in a PR:

- [ ] vBinary
- [ ] vBoolean
- [ ] vCalAddress
- [ ] vDate
- [ ] vDatetime
- [ ] vDuration
- [ ] vFloat
- [ ] vFrequency
- [ ] vGeo
- [ ] vInt
- [ ] vPeriod
- [ ] vRecur
- [ ] vText
- [ ] vTime
- [ ] vUTCOffset
- [ ] vUri
- [ ] vWeekday

The following ones are special because they consist of different properties. The examples should be extended because of that:

- [ ] vDDDLists
- [ ] vDDDTypes

See also:
- #726

SerifOZ:
Can I work this issue?

niccokunzmann:
@stevepiercy If you can have a look at this issue and make the example docstring "perfect" or create a PR for a perfect docstring in e.g. vGeo, that would help the other classes. Then, we can open another issue after this is solved to "unify the format" according to the one example.

[...]

PR #759, opened 2024-12-21
Document vRecur property

fluxxcode:
This PR adds an example to the usage documentation on how to add an `RRULE` from a string.

Closes https://github.com/collective/icalendar/issues/758

----
📚 Documentation preview 📚: https://icalendar--759.org.readthedocs.build/


niccokunzmann:
Hi, this is nice... I wonder why it fails. I will have a look.

In the mean time, could you also add these examples to the vRecur class documentation?

docs/usage.rst
```diff
@@ -387,11 +387,16 @@ By extending the event with subcomponents, you can create multiple alarms.
...
+Note that if you want to add the reccurence rule from a string, you must use the vRecur property. Otherwise the rule will be escaped, making it invalid.
 
 .. code-block:: pycon
 
-    >>> event.add('rrule', {'freq': 'daily'})
+    >>> from icalendar import vRecur
+
+    >>> event.add('rrule', {'freq': 'daily', 'interval': 10})
+
+    >>> event.add('rrule', vRecur.from_ical('FREQ=DAILY;INTERVAL=10'))
```
niccokunzmann: Oh, I found it... We actually run the documentation code to make sure it works.
You add another RRULE entry and that makes the output different a few lines further down.

I think, a good place to add the documentation for this is actually the class string:

https://github.com/collective/icalendar/blob/4725c1bd57ca3afe61e7d45805a2b337842fbd29/src/icalendar/prop.py#L1102

There is almost nothing and more is appreciated! Could you add these two examples to the class string? Could you remove this line causing duplicate entry?


----

This is another issue to make this class nicer to understand:
https://github.com/collective/icalendar/issues/742
That is - if you like this class and making it pretty :heart:

docs/usage.rst
```diff
@@ -387,11 +387,16 @@ By extending the event with subcomponents, you can create multiple alarms.
     >>> alarm_24h_before.add('description', 'Reminder: Event in 24 hours')
     >>> event.add_component(alarm_24h_before)
 
-Or even recurrence.
+Or even recurrence, either from a dictionary or a string.
+Note that if you want to add the reccurence rule from a string, you must use the vRecur property. Otherwise the rule will be escaped, making it invalid.
```
stevepiercy: Add some markup for inline literal.

Use one line per sentence for easier diffs in the future.

```suggestion
Note that if you want to add the reccurence rule from a string, you must use the ``vRecur`` property.
Otherwise the rule will be escaped, making it invalid.
```

stevepiercy (changes requested):

docs/usage.rst
jannistpl: Done. I copied most of the docstring from the rfc. Let me know if you want me to adjust anything.

niccokunzmann:
@fluxxcode Have a look, can you see the tests?

https://github.com/collective/icalendar/actions/runs/12449350281/job/34754863315#step:5:192

src/icalendar/prop.py
```diff
@@ -1100,6 +1100,68 @@ def __reduce_ex__(self, _p):
...
+    .. code-block:: text
+
+        RRULE:FREQ=DAILY;COUNT=10
+
+    .. code-block:: pycon
+
+        >>> from icalendar.prop import vRecur
+        >>> rrule = vRecur.from_ical('FREQ=DAILY;COUNT=10')
+        >>> rrule
+        {'FREQ': ['DAILY'], 'COUNT': [10]}
```
niccokunzmann:
```suggestion
        vRecur({'FREQ': ['DAILY'], 'COUNT': [10]})
```

niccokunzmann:
Nice!

src/icalendar/prop.py
jannistpl: oh lol, I just saw it myself and fixed it xD

fluxxcode:
My code editor removed trailing whitespaces from other docstrings. Let me know if I should revert the diffs

stevepiercy:
Direct links to preview changes:

- https://icalendar--759.org.readthedocs.build/en/759/usage.html#example
- https://icalendar--759.org.readthedocs.build/en/759/api.html#icalendar.prop.vRecur None of the attributes are correctly formatted, except for ``Example`` because it is on its own line whereas the others are not. Can you adjust the other parameters to match the whitespace layout?

src/icalendar/prop.py
```diff
@@ -1100,6 +1100,68 @@ def __reduce_ex__(self, _p):
 
 class vRecur(CaselessDict):
     """Recurrence definition.
+
+    Property Name: RRULE
```
stevepiercy: All parameter names must appear on their own line, followed by an optional description beginning on an indented new line. Please adjust the others accordingly. ``Except`` is OK. See current rendering at https://icalendar--759.org.readthedocs.build/en/759/api.html#icalendar.prop.vRecur

```suggestion
    Property Name:
        RRULE
```

stevepiercy (changes requested):

src/icalendar/prop.py
jannistpl: The rendering now looks good

niccokunzmann:
Thanks! This is definitely a huge improvement!

PR #767, opened 2025-01-16
Shorten first example

niccokunzmann:
The first example is very useful. It should be as concise as possible.

This is tested by the tests.

----
📚 Documentation preview 📚: https://icalendar--767.org.readthedocs.build/


niccokunzmann:
Merging for progress

Issue #763, opened 2025-01-08
Test failure on Fedora with 6.1.0


limburgher:
Sample:

__________________ test_can_identify_dateutil[America/Merida] __________________

tzid = 'America/Merida'

    @with_tzid
    def test_can_identify_dateutil(tzid):
        """Check that all those dateutil timezones can be identified."""
        from dateutil.tz import gettz
>       assert tzid in tzids_from_tzinfo(gettz(tzid))
E       AssertionError: assert 'America/Merida' in ('America/Mexico_City', 'Mexico/General')
E        +  where ('America/Mexico_City', 'Mexico/General') = tzids_from_tzinfo(tzfile('/usr/share/zoneinfo/America/Merida'))
E        +    where tzfile('/usr/share/zoneinfo/America/Merida') = <dateutil.tz.tz.__get_gettz.<locals>.GettzFunc object at 0x7f543477e270>('America/Merida')

src/icalendar/tests/test_timezone_identification.py:29: AssertionError

niccokunzmann:
Thank you! Could you tell me which system it is so it can be reproduced as well as pip list?

limburgher:
Fedora 42, rawhide mock.

niccokunzmann:
Thank you! I think that the tests probably need asdjusting:

- [ ] new timezones should not be tested
- [ ] timezones that are in the list but are not supported should also not be tested

That would create consistency across platforms, I hope.

limburgher:
Still present in 6.1.1

niccokunzmann:
This is now also failing in our CI.
https://github.com/collective/icalendar/actions/runs/13164502893/job/36741212615?pr=774#step:5:309

limburgher:
6.1.2:
=================================== FAILURES ===================================
_____________ test_we_can_identify_dateutil_timezones[Asia/Manila] _____________

tzid = 'Asia/Manila'

    @pytest.mark.parametrize("tzid", available_timezones() - {"Factory", "localtime"})
    def test_we_can_identify_dateutil_timezones(tzid):
        """dateutil and others were badly supported.
    
        But if we know their shortcodes, we should be able to identify them.
        """
        tz = gettz(tzid)
>       assert tzid in tzids_from_tzinfo(tz)
E       AssertionError: assert 'Asia/Manila' in ('Asia/Brunei', 'Asia/Kuching', 'Asia/Makassar', 'Asia/Ujung_Pandang', 'Etc/GMT-8')
[...]

PR #775, opened 2025-02-05
Restrict timezone names for tests

niccokunzmann:
This is intended to be a contribution to #763.


niccokunzmann:
icalendar.timezone.tzid.tinfo2tzids (which I added) would provide us with a timezone name for a timezone even if this timezone is not easily recognized as having one, so we can generate timezones if needed.

Now, the code assumes that timezones in the past would not change.
But they do! Even between 1970 and 2000.

Here is an example of how the timezone definitions are different between each other, using EET which is +2 hours away from UTC.

```python
>>> 60*60*2  # 2 hours offset
7200
>>> from dateutil.tz import gettz
>>> from pytz import timezone
>>> import datetime
>>> d = datetime.datetime(1983, 3, 27, 3, 0)
>>> tz1 = gettz("EET")
>>> tz1.utcoffset(d)
datetime.timedelta(seconds=10800)  # wrong offset
>>> tz2 = ZoneInfo("EET")
>>> tz2.utcoffset(d)   # correct offset
datetime.timedelta(seconds=7200)
>>> tz3 = timezone("EET")
>>> tz3.utcoffset(d)  # no offset
  File "/home/nicco/icalendar/.tox/py39/lib/python3.9/site-packages/pytz/tzinfo.py", line 425, in utcoffset
    dt = self.localize(dt, is_dst)
  File "/home/nicco/icalendar/.tox/py39/lib/python3.9/site-packages/pytz/tzinfo.py", line 344, in localize
    raise NonExistentTimeError(dt)
pytz.exceptions.NonExistentTimeError: 1983-03-27 03:00:00
```

niccokunzmann:
[10bd46d](https://github.com/collective/icalendar/pull/775/commits/10bd46d14e19f1b39ca770428eef3d0539a52d02) addresses the problem that around a change of a timezone we get different results for different implementations.

niccokunzmann:
See also #780

niccokunzmann:
I am merging this to get the main branch green again, probably temporarily.

In order to really solve this problem, we need to have a look at
- #763
- #780

PR #783, opened 2025-03-12
Add name and email attribute to vCalAddress

niccokunzmann:
This adds a few convenience attributes.

- name
- email
- params (type)

----
📚 Documentation preview 📚: https://icalendar--783.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Minor changes:
 
 - Link to related package.
 - Shorten first example in documentation.
+- Add ``name`` and ``email`` properties to ``vCalAddress``.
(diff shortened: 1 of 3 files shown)


glibersat:
just fell in the pitfall of the missing "CN" from ATTENDEE, would be wonderful to merge that, thanks!

niccokunzmann:
@glibersat I will take this as a review :) Thanks!

Also: I am really happy to have the convenience methods in here. The reason is that it makes all code around the world using this library more reliable. I.e. some people might forget about edge cases and we can gather them here. Once a feature is in here, it can only get better with feedback - for everyone!

So, here is my invitation and ask: Could you have a look in your code and add an issue/pr about the ones you use?

Would you like this released?

PR #785, opened 2025-03-17
added params argument to class constructors

Horisyre:
Added `params={}` to class constructors in `icalendar.prop`.
Initialized `params` member in classes with `params` argument.

----
📚 Documentation preview 📚: https://icalendar--785.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,25 @@
 Changelog
 =========
+6.1.2 (unreleased 2025-03-19)
+------------------
+
+Minor changes:
+
+- Set default value for ``params`` as ``params={}`` in mulitple constructors in ``icalendar.prop`` to improve usability.
+- Improve object initialization performance in ``icalendar.prop``.
+- Add type hint for ``params`` in multiple constructors in ``icalendar.prop``.
+
+Breaking changes:
+
+- ...
+
+New features:
+
+- ...
+
+Bug fixes:
+
+- ...
 
 6.1.2 (unreleased)
 ------------------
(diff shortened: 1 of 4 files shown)


niccokunzmann:
@Horisyre Thank you very much! That looks great!

I would like to merge it and have a question: I see that not all lines are covered in the test file. 

For tests, it is ok if errors occur and they will count as failed then.
Could you rewrite the tests like I show below and we see if we get to 100% coverage?

old test:

```python
def test_param_vWeekday():
    try:
        obj = vWeekday("2FR",params={"SOME_PARAAM":"VALUE"}) 
        assert isinstance(obj, vWeekday)  
    except Exception as e:
        assert False, f"Error creating vWeekday object: {e}"  # line probably not used
```

new test:

```python
def test_param_vWeekday():
    obj = vWeekday("2FR",params={"SOME_PARAAM":"VALUE"}) 
    assert isinstance(obj, vWeekday)  # if this errors, that is alright.
    assert obj.params["SOME_PARAAM"] == "VALUE" # add this line, too
```

How would that be? Do you have time for this?

niccokunzmann:
Nice!

> 
    100 of 100 (100.0%) changed or added relevant lines in 2 files are covered.
    No unchanged relevant lines lost coverage.
    Overall coverage increased (+0.04%) to 96.447%

niccokunzmann:
Thank you very much!

Horisyre:
Gladly, happy to help!

> Thanks for your contribution. This is included in v6.1.2

PR #788, opened 2025-03-19
Reapply "Create stability around timezone identification"

### LINKED ISSUE #780: [BUG] main branch failing, timezone name identification way too unstable
## Describe the bug

The main branch is failing. The tests are those that try to identify the timezone id from dateutil timezones.

## To Reproduce

See 
- #779 
- #776 
- #763.

## Expected behavior

It would be nice to be able to identify timezones but as #776 shows, this is not trivial.
I think, that this should not be part of this package. We can generate timezones from actual timezone objects.
For now, the changes and assumptions make the tests fail and cause confusion.

I am wondering how to go on with this.

## Possible fixes
 
- Disable the dateutil tests to find out which timezone id is used.
- Change the timezone identification to be more dynamic, based on the local installation (maybe still problematic)
- Remove the complex timezone identification fully - we can only claim easy compatibility to zoneinfo and pytz because they have a name
- create a separate package that handles timezone identification and link to it in the documentation
- allow passing an optional timezone name
- generate a timezone with a custom name if required.
- change the way timezone names for unknown timezones are generated and use the timezone objects to generate the timezones instead of the indirect way of using the timezone id.

## Complexity

The timezone database differs on operating systems. Updates to timezones in the past are still issued. We have this complexity:

- OS
- timezone database/source: OLSON, iana, icalendar.Timezone
- timezone implementation/source pytz, zoneinfo, dateutil
- timezone database version

With this complexity, I suggest that identifying the name of the timezone is out of scope of this package. It is way too complex, an edge case (automatic generation of VTIMEZONE components). 

## Suggested Fix

I would suggest to generate the timezones from the timezone objects found in datetimes in the icalendar.Calendar. There is no need to go the indirect way of identifying the name first. However, should we know the name, we should use it (pytz, zoneinfo). I think, I added that feature because it was easy to test. I hope, removing this will help keep the package small and more stable.

I would be grateful to hear your thoughts on this.

niccokunzmann:
I accidentally worked on the `main` branch. Here is the pull request.

- use existing lookup table only to identify equivalent timezones
- return a tuple of possible timezone ids
- first in this tuple is the timezoneid passed
- test dateutil, pytz, zoneinfo
- add support for dateutil.tz.tzutc

This fixes #780

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,7 @@ New features:
 
 Bug fixes:
 
-- ...
+- Stabelize timezone id lookup, see `Issue 780 <https://github.com/collective/icalendar/issues/780>`_.
 
 6.1.2 (2025-03-19)
 ------------------
--- a/src/icalendar/timezone/tzid.py
+++ b/src/icalendar/timezone/tzid.py
@@ -6,36 +6,56 @@ datetime.tzinfo object.
 """
 from __future__ import annotations
 
+from collections import defaultdict
+from pathlib import Path
 from typing import TYPE_CHECKING, Optional
-import dateutil.tz.tz as tz
+
+from dateutil.tz import tz
+
+from icalendar.timezone import equivalent_timezone_ids_result
 
 if TYPE_CHECKING:
     from datetime import datetime, tzinfo
 
+DATEUTIL_UTC = tz.gettz("UTC")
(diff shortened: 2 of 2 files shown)


niccokunzmann:
I had feedback twice that this might indeed solve the problem.

niccokunzmann:
As this seems to fix the issues, I merge this without a review after 1 week.

PR #790, opened 2025-03-19
Add developer certificate of origin to pull requests

### LINKED ISSUE #778: Clearer licensing agreement
As part of NLNet, I wanted to clarify the licensing of this project. It is a BSD license. So, I asked:

> How can we make sure contributions are compliant with the license?

The answer was that they recommend a **Developer Certificate or Origin**. The full answer is below.

## Possible Implementation

- [ ] There is a bot that can do that for us: https://github.com/apps/dco
- [ ] And there is a pre-commit that can automate that when committing: https://github.com/mattlqx/pre-commit-sign
- [ ] we can also add a checkbox in the pull request template and add it on merge

## Questions

How are you with adding this to icalendar?
Would you like to try it out? How is it with adding extra requirements given that some people find it harder to contribute then?

## Full Answer

> You can use a CLA for contributions, however many contributors prefer it when projects use Developer Certificates of Origin (DCOs) instead (https://developercertificate.org/). This is because CLAs are frequently more complicated legal documents that require a review of terms from the developers, who may not have the legal expertise to accurately determine or protect their interests. Additionally, CLAs also require projects communicate a list of authorized contributors, which can generate some additional work for you as a project runner. Generally, CLAs are used by projects who have a vested interest in tracking in detail each individual contribution and contributor, and in carving out special terms and conditions for contributions.
>
> Conversely, a DCO is more simplistic: it is a statement that each contributor has to agree to in order to make a contribution, stating that they are allowed to do so and that the project has the right to distribute their contribution under the project's chosen license. This would simplify a lot of work for you; you can simply include the DCO text in your project repo(s) and indicate explicitly in your README or contributor guidelines that contributions indicate agreement to the DCO.
>
> I would suggest you take a look at the DCO text in the link I've provided above and consider your options, although as a general rule of thumb we at the FSFE recommend DCOs by default over CLAs. If you do feel that a CLA is preferable, you can also use the following online CLA tool (https://contributoragreements.org/ca-cla-chooser/) to generate a CLA that is in line with your needs. 

The main difference for readers that I see:
- DCO says that the code is compatible in license
- CLA assigns the copyright to someone else so they then hold the license

niccokunzmann:
This adds a Developer Certificate of Origin to all the pull requests.

See this example:

https://github.com/niccokunzmann/developer-certificate-of-origin-pr/pull/1

You can open and close it and agree to the certificate.

Fixes #778.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,48 @@
+<!--
+  1. Please mention issues.
+  2. Agree to be the original author, see below.
+--->
+
+
+
+<details><summary>
+
+- [ ] I agree to the Developer Certificate of Origin Version 1.1
+
+</summary>
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
+660 York Street, Suite 102,
+San Francisco, CA 94110 USA
+
+Everyone is permitted to copy and distribute verbatim copies of this
+license document, but changing it is not allowed.
+
+**Developer's Certificate of Origin 1.1**
+
+By making a contribution to this project, I certify that:
+
+1. The contribution was created in whole or in part by me and I
+    have the right to submit it under the open source license
+    indicated in the file; or
+
+2. The contribution is based upon previous work that, to the best
+    of my knowledge, is covered under an appropriate open source
+    license and I have the right under that license to submit that
+    work with modifications, whether created in whole or in part
+    by me, under the same open source license (unless I am
+    permitted to submit under a different license), as indicated
+    in the file; or
+
+3. The contribution was provided directly to me by some other
+    person who certified (a), (b) or (c) and I have not modified
+    it.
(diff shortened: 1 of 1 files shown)


stevepiercy (changes requested):
As a first-timer, I have no idea what its purpose is, or what to do. It needs an explanation and instructions.

As a grumpy developer, what happens if I refuse to check the box? Can I just delete the statement entirely from my pull request?

niccokunzmann:
Hm, thanks, true! And it might be said, after some looking into this, that the commits are written as 'signed off'.
That would allow tracking the DCO in the commits and not on a proprietary platform. I guess, this PR is not a solution at all!

I am hesitating to say that all commits needs signing off as that will discourage new developers? I wonder... Maybe, there is a pre-commit available or a bot that would take care of this....

stevepiercy:
What is the problem you are trying to solve? I still don't understand.

PR #792, opened 2025-03-22
Fix for TZID forward references to VTIMEZONE

dm9:
RFC 5545 allows events to rely on TZIDs referring to VTIMEZONEs that are defined after the point of reference, but icalendar doesn't currently support this. To verify, just move the VTIMEZONE block in america_new_york.ics to the end of the file and the test will fail, complaining that the detected timezone (None) does not match the TZID referred to.
	
I have an ical feed concerning an old flight that includes a forward-referenced VTIMZEONE so this an actual problem in the field.
	
The fix I'm proposing is a little clunky but works and tests OK with tox py312 anyway. Namely, if Calendar.from_ical sees a ICALENDAR with a VTIMEZONE that follows any other component (fearing that component might have included a forward reference), it reparses the whole string one more time. This solves the problem because the first time through, all the VTIMEZONEs get put into a cache by existing code, and the second time through, the TZID forward references to VTIMEZONEs are found in the cache.
	
This is my first PR ever so please be gentle :-) 

----
📚 Documentation preview 📚: https://icalendar--792.org.readthedocs.build/


CHANGES.rst
```diff
@@ -137,9 +138,7 @@ New features:
 Bug fixes:
 
 - Fix link to stable release of tox in documentation.
-- Fix a bad ``bytes`` replace in ``unescape_char``.
```
niccokunzmann: Hi, I do not know how this got changed. The old changelog could stay, I think.

src/icalendar/cal.py
```diff
@@ -1638,6 +1638,38 @@ def example(cls, name: str="example") -> Calendar:
         """Return the calendar example with the given name."""
         return cls.from_ical(get_example("calendars", name))
 
+    @classmethod
+    def from_ical(cls, st, multiple=False):
+        comps = Component.from_ical(st, multiple=True)
+        all_timezones_so_far = True
+        for comp in comps:
+            for component in comp.walk():
```
niccokunzmann: I think, `comp.subcomponents` would be enough here as it is not recursing.

I take your preference on this one!

src/icalendar/cal.py
```diff
@@ -1638,6 +1638,38 @@ def example(cls, name: str="example") -> Calendar:
...
+                if type(component) == Timezone:
```
niccokunzmann: I would test `component.name == "VTIMEZONE"` instead because of the object-oriented principle of polymorphy. However, this works.. so it is ok.

Generally for me (perfectionist object-oriented sometimes), I would choose the first possible in this order:


1. ask the object itself: `obj.can_you_do_this()`?
2. ask an attribute `obj.attr == xy`
3. `isinstance(obj, ty)`
4. `type(obj) == xy`

src/icalendar/cal.py
```diff
@@ -1638,6 +1638,38 @@ def example(cls, name: str="example") -> Calendar:
...
+                    if all_timezones_so_far:
+                        pass
+                    else:
+                        # If a preceding component refers to a VTIMEZONE defined later in the source st
+                        # (forward references are allowed by RFC 5545), then the earlier component may have
+                        # the wrong timezone attached.
+                        # However, during computation of comps, all VTIMEZONEs observed do end up in
+                        # the timezone cache. So simply re-running from_ical will rely on the cache
+                        # for those forward references to produce the correct result.
```
niccokunzmann, line 1656: Thanks for the comment!

niccokunzmann (changes requested):
Hi, thanks for this PR!
I think, that this is indeed solving the problem... :partying_face: 

You can find some comments that I made.

If you only change the changelog a little, I would already merge it and not have to edit after merge :)
The other comments are feedback, not required to change but nice to have from my point of view...

Thanks for taking the time to dig into this so far. I appreciate the test for this, too.

I wonder: code coverage says it went down. Do you know which lines are missing?

niccokunzmann:
Also, feel free to ask any questions that you like and instead of trying out things for a long time come back for conversation. I only see a GitHub website but not how you receive my comments.

Thanks again for this fix and I value it and want to merge it :)

CHANGES.rst
dm9: Not sure how that happened there, agreed!

src/icalendar/cal.py
dm9: I actually prefer the component.name test too, it's just that I already found a use of type(component)==Timezone in Timezone.get_transitions and so I thought maybe that style was preferred here for some reason. Will adopt your suggestion.

src/icalendar/cal.py
dm9: Agreed!

dm9:
I pushed a commit that addresses your comments (thanks).

I have no idea why/how the changes could have reduced test coverage. Any ideas?? 

I also have no idea why tests/nopytz is failing on this commit whereas it seemed to work fine on the first commit. Any ideas??

niccokunzmann:
Hi,

the coverage link suggests that this line is not tested: (`#L1649` in the end`)

https://github.com/collective/icalendar/blob/d3d60c61571698d23b01e78dfe1d1673120c9ad4/src/icalendar/cal.py#L1649

That can be changed through changing the if-statement.

I think, that the test failure is not your fault but because of #780 for which #788 should provide a fix. I am waiting for feedback on that till tomorrow and if I am not getting it, merge it any way for progress. Then, I can update your branch and likely, all tests run!

dm9:
Thanks Nicco. 

I'm mystified by the L1649 issue. When I set a breakpoint on that line and pytest test_timezoned.py, then the breakpoint hits. Also when I run coverage manually in my py312 environment, L1649 (along with all other new code in that area) is shown as green/covered. So I don't understand what the coverage report means about the code in this area.

Regarding L487 coverage, would you like an additional test for non-Calendar components that would cover L487? Previous code never tested that condition for non-Calendars, but I guess testing it would be a slight net improvement.

niccokunzmann:
Hi, yes if you could test that, that would be good as it was tested before and others depend on it.

dm9:
Ok I think those lines are covered now with a couple of new tests. As far as I know the only issue remaining is the nopytz thing and hopefully that will just work upon your pending push for #788.

niccokunzmann:
The test is failing... I will create a PR to your branch from my fork to fix this... This should again not be a problem for this to work. But the timezone stuff is tricky to get right, I find.

niccokunzmann:
The problem here seems to be that dateutil cannot find the timezone named America/Coyhaique

niccokunzmann:
@dm9, thanks again and sorry that you create a PR at a stage where the main branch is failing.
Could you have a look and if you do not object, merge https://github.com/dm9/icalendar/pull/1 so that the tests possibly pass?

niccokunzmann:
Thank you! I merged it.

dm9:
Nice! Thanks!

niccokunzmann:
This is available from version 6.1.3 on.

PR #797, opened 2025-04-12
Use license expression for project metadata

cdce8p:
Hatchling `1.27.0` added support for [PEP 639](https://peps.python.org/pep-0639/) license expressions. Update the project metadata for it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,13 +6,13 @@
 #
 
 [build-system]
-requires = ["hatchling", "hatch-vcs"]
+requires = ["hatchling>=1.27.0", "hatch-vcs"]
 build-backend = "hatchling.build"
 
 [project]
 name = "icalendar"
-license = { file = "LICENSE.rst", name = "BSD-2-Clause" }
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -6,13 +6,13 @@
...
 
 [build-system]
-requires = ["hatchling", "hatch-vcs"]
+requires = ["hatchling>=1.27.0", "hatch-vcs"]
 build-backend = "hatchling.build"
 
 [project]
 name = "icalendar"
-license = { file = "LICENSE.rst", name = "BSD-2-Clause" }
-#    name = "BSD-2-Clause", # TODO: is this the right short key
```
cdce8p, line 15: I believe it is since the license text matches that one here: https://spdx.org/licenses/BSD-2-Clause.html

niccokunzmann: Thanks for checking!

niccokunzmann:
Thank you for the contributions! I learned something :)

Issue #655, opened 2024-06-24
RFC 7986 - additional parameters and values


niccokunzmann:
RFC 7986 specifies a few new parameters that we should make sure we support:

- [ ] [5.1](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.1). NAME Property
- [ ] [5.2](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.2). DESCRIPTION Property
- [ ] [5.3](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.3). UID Property
- [ ] [5.4](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.4). LAST-MODIFIED Property
- [ ] [5.5](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.5). URL Property
- [ ] [5.6](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.6). CATEGORIES Property
- [ ] [5.7](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.7). REFRESH-INTERVAL Property
- [ ] [5.8](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.8). SOURCE Property
- [ ] [5.9](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.9). COLOR Property
- [ ] [5.10](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.10). IMAGE Property
- [ ] [5.11](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.11). CONFERENCE Property
- [ ] [6](https://www.rfc-editor.org/rfc/rfc7986.html#section-6). Property Parameters
- [ ] [6.1](https://www.rfc-editor.org/rfc/rfc7986.html#section-6.1). DISPLAY Property Parameter
- [ ] [6.2](https://www.rfc-editor.org/rfc/rfc7986.html#section-6.2). EMAIL Property Parameter
- [ ] [6.3](https://www.rfc-editor.org/rfc/rfc7986.html#section-6.3). FEATURE Property Parameter
- [ ] [6.4](https://www.rfc-editor.org/rfc/rfc7986.html#section-6.4). LABEL Property Parameter

These properties should have the right type in the right object, be serializable

- [ ] there is one example file with all properties
- [ ] ICS -> obj -> ICS 

See also:
- https://www.rfc-editor.org/rfc/rfc7986.html
- https://github.com/collective/icalendar/issues/589

[...]

PR #801, opened 2025-04-18
Add VCALENDAR properties NAME, DESCRIPTION and COLOR as attributes

niccokunzmann:
This is a start on #655 and adds these attributes the the calendar component:

- NAME with compatibility to X-WR-CALNAME
- DESCRIPTION with compatibility to X-WR-CALDESC
- COLOR with compatibility to X-APPLE-CALENDAR-COLOR

----
📚 Documentation preview 📚: https://icalendar--801.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ Breaking changes:
 
 New features:
 
-- ...
+- Add attributes to the calendar for properties NAME, DESCRIPTION and COLOR, see `Issue 655 <https://github.com/collective/icalendar/issues/655>`_.
 
 Bug fixes:
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ Breaking changes:
 
 New features:
 
-- Add attributes to the calendar for properties NAME, DESCRIPTION and COLOR, see `Issue 655 <https://github.com/collective/icalendar/issues/655>`_.
+- Add attributes to the calendar for properties ``NAME``, ``DESCRIPTION``, and ``COLOR``. See `Issue 655 <https://github.com/collective/icalendar/issues/655>`_.
 
 Bug fixes:
 
--- a/src/icalendar/attr.py
+++ b/src/icalendar/attr.py
@@ -7,7 +7,7 @@ from typing import Optional
 def multi_language_text_property(main_prop:str, compatibility_prop:str, doc:str) -> property:
     """This creates a text property.
 
-    This property can be defined several times with different LANGAUGE parameters.
+    This property can be defined several times with different ``LANGUAGE`` parameters.
 
     Args:
 
--- a/src/icalendar/attr.py
+++ b/src/icalendar/attr.py
@@ -11,8 +11,8 @@ def multi_language_text_property(main_prop:str, compatibility_prop:str, doc:str)
 
     Args:
 
-        main_prop: The property to set and get, e.g. NAME
-        compatibility_prop: An old property used before, e.g. X-WR-CALNAME
+        main_prop: The property to set and get, such as ``NAME``
+        compatibility_prop: An old property used before, such as ``X-WR-CALNAME``
         doc: The documentation string
     """
     def fget(self) -> Optional[str]:
--- a/src/icalendar/attr.py
+++ b/src/icalendar/attr.py
@@ -1,0 +1,40 @@
+"""Attributes of Components and properties."""
+from __future__ import annotations
+
+from typing import Optional
+
+
+def multi_language_text_property(main_prop:str, compatibility_prop:str, doc:str) -> property:
+    """This creates a text property.
+
+    This property can be defined several times with different LANGAUGE parameters.
+
+    Args:
+
+        main_prop: The property to set and get, e.g. NAME
+        compatibility_prop: An old property used before, e.g. X-WR-CALNAME
+        doc: The documentation string
+    """
+    def fget(self) -> Optional[str]:
+        """Get the property"""
+        result = self.get(main_prop, self.get(compatibility_prop))
+        if isinstance(result, list):
+            for item in result:
+                if "LANGUAGE" not in item.params:
+                    return item
+        return result
+
+    def fset(self, value:str):
+        """Set the property."""
+        fdel(self)
+        self.add(main_prop, value)
+
+    def fdel(self):
+        """Delete the property."""
+        self.pop(main_prop, None)
+        self.pop(compatibility_prop, None)
+
+    return property(fget, fset, fdel, doc)
+
+
+__all__ = ["multi_language_text_property"]
--- a/src/icalendar/cal.py
+++ b/src/icalendar/cal.py
@@ -2140,7 +2140,7 @@ class Calendar(Component):
     Conformance:
 
         This property can be specified once in an iCalendar
-        object or in "VEVENT", "VTODO", or "VJOURNAL" calendar components.
+        object or in ``VEVENT``, ``VTODO``, or ``VJOURNAL`` calendar components.
 
     Description:
 
--- a/src/icalendar/cal.py
+++ b/src/icalendar/cal.py
@@ -2062,7 +2062,7 @@ class Calendar(Component):
 
     calendar_name = multi_language_text_property(
         "NAME", "X-WR-CALNAME",
-        """This property specifies the name of the calendar
+        """This property specifies the name of the calendar.
 
     This takes care of :rfc:`7986` ``NAME`` and ``X-WR-CALNAME``.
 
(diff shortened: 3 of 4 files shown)


niccokunzmann:
@stevepiercy what I am wondering about is this:

- the `name` attribute is kind-of misleading, so I use `calendar_name`
- the other attributes are doubled with and without `calendar_`. I wonder: What do you prefer?

CHANGES.rst
```diff
@@ -15,7 +15,7 @@ Breaking changes:
 
 New features:
 
-- ...
+- Add attributes to the calendar for properties NAME, DESCRIPTION and COLOR, see `Issue 655 <https://github.com/collective/icalendar/issues/655>`_.
```
```suggestion
- Add attributes to the calendar for properties ``NAME``, ``DESCRIPTION``, and ``COLOR``. See `Issue 655 <https://github.com/collective/icalendar/issues/655>`_.
```

src/icalendar/attr.py
```diff
@@ -0,0 +1,40 @@
+"""Attributes of Components and properties."""
+from __future__ import annotations
+
+from typing import Optional
+
+
+def multi_language_text_property(main_prop:str, compatibility_prop:str, doc:str) -> property:
+    """This creates a text property.
+
+    This property can be defined several times with different LANGAUGE parameters.
```
```suggestion
    This property can be defined several times with different ``LANGUAGE`` parameters.
```

src/icalendar/attr.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Args:
+
+        main_prop: The property to set and get, e.g. NAME
+        compatibility_prop: An old property used before, e.g. X-WR-CALNAME
```
```suggestion
        main_prop: The property to set and get, such as ``NAME``
        compatibility_prop: An old property used before, such as ``X-WR-CALNAME``
```

src/icalendar/cal.py
```diff
@@ -2055,6 +2060,107 @@ def add_missing_timezones(
                 continue
             self.add_component(timezone)
 
+    calendar_name = multi_language_text_property(
+        "NAME", "X-WR-CALNAME",
+        """This property specifies the name of the calendar
```
```suggestion
        """This property specifies the name of the calendar.
```

src/icalendar/cal.py
```diff
@@ -2055,6 +2060,107 @@ def add_missing_timezones(
...
+    BEGIN:VCALENDAR
+    DESCRIPTION:This is a calendar
+    END:VCALENDAR
+    """)
+
+    color = calendar_color = multi_language_text_property(
+        "COLOR", "X-APPLE-CALENDAR-COLOR",
+        """This property specifies a color used for displaying the calendar.
+
+    This takes care of :rfc:`7986` ``COLOR`` and ``X-APPLE-CALENDAR-COLOR``.
```
```suggestion
    This implements :rfc:`7986` ``COLOR`` and ``X-APPLE-CALENDAR-COLOR``.
```

src/icalendar/cal.py
```diff
@@ -2055,6 +2060,107 @@ def add_missing_timezones(
...
+
+    Property Parameters:
+
+        IANA and non-standard property parameters can
+        be specified on this property.
+
+    Conformance:
+
+        This property can be specified once in an iCalendar
+        object or in "VEVENT", "VTODO", or "VJOURNAL" calendar components.
```
```suggestion
        object or in ``VEVENT``, ``VTODO``, or ``VJOURNAL`` calendar components.
```

src/icalendar/cal.py
```diff
@@ -2055,6 +2060,107 @@ def add_missing_timezones(
...
+
+    Description:
+
+        This property specifies a color that clients MAY use
+        when presenting the relevant data to a user.  Typically, this
+        would appear as the "background" color of events or tasks.  The
+        value is a case-insensitive color name taken from the CSS3 set of
+        names, defined in Section 4.3 of [W3C.REC-css3-color-20110607].
```
```suggestion
        names, defined in Section 4.3 of `W3C.REC-css3-color-20110607 <https://www.w3.org/TR/css-color-3/>`_.
```

src/icalendar/cal.py
```diff
@@ -2055,6 +2060,107 @@ def add_missing_timezones(
                 continue
             self.add_component(timezone)
 
+    calendar_name = multi_language_text_property(
+        "NAME", "X-WR-CALNAME",
+        """This property specifies the name of the calendar
+
+    This takes care of :rfc:`7986` ``NAME`` and ``X-WR-CALNAME``.
```
```suggestion
    This implements :rfc:`7986` ``NAME`` and ``X-WR-CALNAME``.
```

src/icalendar/cal.py
```diff
@@ -2055,6 +2060,107 @@ def add_missing_timezones(
...
+    BEGIN:VCALENDAR
+    NAME:My Calendar
+    END:VCALENDAR
+    """)
+
+    description = calendar_description = multi_language_text_property(
+        "DESCRIPTION", "X-WR-CALDESC",
+        """This property specifies the description of the calendar.
+
+    This takes care of :rfc:`7986` ``DESCRIPTION`` and ``X-WR-CALDESC``.
```
```suggestion
    This implements :rfc:`7986` ``DESCRIPTION`` and ``X-WR-CALDESC``.
```

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -0,0 +1,114 @@
+"""This tests additional attributes from RFC 7986.
```
stevepiercy:
```suggestion
"""This tests additional attributes from :rfc:`7986`.
```

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -0,0 +1,114 @@
...
+
+Some attributes are also available as X-... attributes.
```
stevepiercy:
```suggestion
Some attributes are also available as ``X-*`` attributes.
```

stevepiercy:
> @stevepiercy what I am wondering about is this:
> 
>     * the `name` attribute is kind-of misleading, so I use `calendar_name`
> 
>     * the other attributes are doubled with and without `calendar_`. I wonder: What do you prefer?

When the context of the property is obvious, such as a calendar, then I don't see the need to prefix it with `calendar_`. That would be consistent with the RFCs.

For example, the following test, before and after refactoring.

```python
def test_get_calendar_name(prop, name, calendar):
    """Get the name of the calendar."""
    calendar.add(prop, name)
    assert calendar.calendar_name == name
```

```python
def test_get_calendar_name(prop, name, calendar):
    """Get the name of the calendar."""
    calendar.add(prop, name)
    assert calendar.name == name
```

The test name is clear and should remain the same. The `name` attribute clearly belongs to the `calendar` object.

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -0,0 +1,114 @@
...
+def test_name_is_preferred(calendar, name):
+    """NAME is more important that X-WR-CALNAME"""
+    calendar.add("NAME", name)
+    calendar.add("X-WR-CALNAME", "asd")
+    assert calendar.calendar_name == name
+
+
+
+# For description, we would use the same tests as name but we also use the
+# same code, so it is alright.
```
stevepiercy:
```suggestion
# For description, we would use the same tests as name, but we also use the
# same code, so it is all right.
```

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -0,0 +1,114 @@
...
+
+@param_color
+@param_name
+def test_description(calendar, desc, name):
+    """Get the value"""
+    calendar.add(desc, name)
+    assert calendar.calendar_description == name
+
+# For color, we would use the same tests as name but we also use the
+# same code, so it is alright.
```
stevepiercy:
```suggestion
# For color, we would use the same tests as name, but we also use the
# same code, so it is all right.
```

stevepiercy (changes requested):

niccokunzmann:
> The test name is clear and should remain the same. The name attribute clearly belongs to the calendar object.

There is this problem that does not apply to any other attributes than `name`

```
class Calendar(...):
    name = "VCALENDAR" # used to identify the component
```

So, I would then choose `description` and `color` and maybe use `calendar_name`. I am thinking of this a bit more as I do not want to change the API but also be consistent with the rest that is possible. There is likely no `event_name` but `Event` has a description:

> Purpose:  This property specifies a color used for displaying the
      calendar, event, todo, or journal data.

vs.

>    Purpose:  This property specifies the name of the calendar.

That would be my proposal.

stevepiercy:
I'm sorry, I don't follow you. Where is the conflict with `name`? It would be a method of `Calendar`, if refactored from `calendar_name` to `name`.

Usage:

```pycon
    >>> calendar = Calendar()
    >>> calendar.name = "My Calendar"
    >>> print(calendar.to_ical())
    BEGIN:VCALENDAR
    NAME:My Calendar
    END:VCALENDAR
```

niccokunzmann:
This is the line:

https://github.com/collective/icalendar/blob/8add6bfe64132555e9da5d5b376ae1e4d03e8008/src/icalendar/cal.py#L1895

And this interface is so old that we cannot change it without breaking numerous downstream packages.

niccokunzmann:
Would you merge it? Should I merge it?

stevepiercy:
Yes.

PR #803, opened 2025-04-21
Add sequence property to event, todo and journal

### LINKED ISSUE #802: expose sequence property
The sequence property is one int property for events and other subcomponents.
By default, it should be 0.

https://www.rfc-editor.org/rfc/rfc5545#section-3.8.7.4

We can add a property for that. That would make some source code look much nicer.

- [ ] add SEQUENCE to event, todo, others
- [ ] join `int` property implementation with REPEAT of alarm

niccokunzmann:
This fixes #802

### DIFF
--- a/src/icalendar/__init__.py
+++ b/src/icalendar/__init__.py
@@ -1,10 +1,6 @@
 from icalendar.alarms import (
     Alarms,
     AlarmTime,
-    ComponentEndMissing,
-    ComponentStartMissing,
-    IncompleteAlarmInformation,
-    LocalTimezoneMissing,
 )
 from icalendar.cal import (
     Alarm,
@@ -13,13 +9,19 @@ from icalendar.cal import (
     ComponentFactory,
     Event,
     FreeBusy,
-    IncompleteComponent,
-    InvalidCalendar,
     Journal,
     Timezone,
     TimezoneDaylight,
     TimezoneStandard,
     Todo,
+)
+from icalendar.error import (
+    ComponentEndMissing,
+    ComponentStartMissing,
+    IncompleteAlarmInformation,
+    IncompleteComponent,
+    InvalidCalendar,
+    LocalTimezoneMissing,
 )
 
 # Parameters and helper methods for splitting and joining string with escaped
--- a/src/icalendar/alarms.py
+++ b/src/icalendar/alarms.py
@@ -17,6 +17,12 @@ from datetime import date, timedelta, tzinfo
 from typing import TYPE_CHECKING, Generator, Optional, Union
 
 from icalendar.cal import Alarm, Event, Todo
+from icalendar.error import (
+    ComponentEndMissing,
+    ComponentStartMissing,
+    IncompleteAlarmInformation,
+    LocalTimezoneMissing,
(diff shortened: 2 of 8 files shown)


niccokunzmann:
Documentation: https://icalendar--803.org.readthedocs.build/en/803/api.html#icalendar.cal.Event.sequence

src/icalendar/error.py
stevepiercy, line 1: This file does not appear in the API documentation. I think you need to add it to `docs/api.rst`.

Outside the scope of this PR, I really hate the RTD theme because I can't tell where I am in the page. For example, am I looking that the sequence property of an Event or Todo? Can you see any difference between these two links?

- https://icalendar--803.org.readthedocs.build/en/803/api.html#icalendar.cal.Event.sequence
- https://icalendar--803.org.readthedocs.build/en/803/api.html#icalendar.cal.Todo.sequence

I would recommend switching to the [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/). I use PyData Sphinx Theme for Plone Sphinx Theme.

Secondarily, I think its time to break up `api.rst` into its modules. `api` would become a folder that contains `index.rst` and one file per module. This makes it much easier to navigate and narrow down a search, too. There might be a valid use case for keeping the API on a single page, but I don't know what that would be.

stevepiercy (changes requested):
Looks good, but there is an omission.

I added a suggestion for another PR, too.

src/icalendar/error.py
niccokunzmann: > This file does not appear in the API documentation. I think you need to add it to docs/api.rst.

Should be added!

> I would recommend switching to the [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/). I use PyData Sphinx Theme for Plone Sphinx Theme.

Yes, please open an issue for that.

> There might be a valid use case for keeping the API on a single page,

The use-case is that we can shuffle things around and the links stay valid.

>  Can you see any difference between these two links?

No, they are exactly the same by source, too.

> Secondarily, I think its time to break up api.rst into its modules.

Yes, and an issue for that is good, too. There is autodoc and I wonder if we can make it easy to include all the files automatically. We will have many more attributes coming in the next year, so we should be able to cope with the onslaught of documentation :)

niccokunzmann: >     This file does not appear in the API documentation. I think you need to add it to docs/api.rst.

I added it, hence I would say this is resolved form my side.

Would you open the issues as you have a clear vision? That would allow for good closing criteria.

stevepiercy:
Created issues:

- https://github.com/collective/icalendar/issues/804
- https://github.com/collective/icalendar/issues/805

src/icalendar/error.py
stevepiercy: We can try autodoc to see if it generates a desired result. We might have to tweak the result, and if that happens we should use it once only to set up the structure, then update the docs manually going forward. We'll see.

stevepiercy (approved):

niccokunzmann:
Thanks @stevepiercy for the review!

Issue #805, opened 2025-04-22
Break up api.rst into its modules


stevepiercy:
I think its time to break up `api.rst` into its modules. `api` would become a folder that contains `index.rst` and one file per module. This makes it much easier to navigate and narrow down a search, too. There might be a valid use case for keeping the API on a single page, but I don't know what that would be.

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/803#discussion_r2053585333_

niccokunzmann:
On that note, it might be worth looking at how to better structure the documentation:

- Should each Component receive its own page?
- Should each Property (prop.py) receive its own page?

Related:

- #626

stevepiercy:
This issue depends on #804, as its layout will change navigability significantly. FTR, I find that organizing documentation according to the [Diátaxis](https://diataxis.fr/) framework makes structuring, writing, and organizing my thoughts into user-approachable documentation much clearer. I'd suggest adopting Diátaxis, but only if @niccokunzmann thinks it would be helpful. We can discuss in real time, too, if you have any concerns.

niccokunzmann:
[Diátaxis](https://diataxis.fr/) looks really cool! Thanks for sharing this! I think, making this documentation nice is worth it, too. Within this year, I would like to add well-defined access to the main properties/params so that icalendar takes care of edge cases when it comes to parsing/generating.

How much time do you have to spend on it? Would you like to give the docs a bit of a good start and I review, as so I can understand your ideas better? It feels like when I try solving something while learning it, it might get changed a lot and not quite hit the target.

stevepiercy:
@niccokunzmann started in #807.

niccokunzmann:
I would suggest the following if we would like to transition to better documentation:

- [ ] create an issue for [Diátaxis](https://diataxis.fr/)
- [ ] create sub-points of what needs changing and moved where
- [ ] describe new files and their purpose
- [ ] already consider in that the open documentation issues
- [ ] see if we can move some items from #626 over to the Diátaxis issue

From there on issues like this are basically sub-issues to restructure the documentation.

This way, we have a clearly structured communication about this.

stevepiercy:
@niccokunzmann agreed. Excellent plan. This issue (#805) would become a subissue of the Diátaxis primary issue. Let's call it "Reorganize documentation following the Diátaxis framework".

This issue (#805) would be the "API reference guide", with one page per module and an index page, falling under the reference category of Diátaxis.

[...]

PR #807, opened 2025-04-22
Switch to PyData Sphinx Theme for documentation

### LINKED ISSUE #804: Switch to PyData Sphinx Theme
I can't tell where I am in the page with the RTD theme. For example, am I looking that the sequence property of an Event or Todo? Can you see any difference between these two links?

- https://icalendar--803.org.readthedocs.build/en/803/api.html#icalendar.cal.Event.sequence
- https://icalendar--803.org.readthedocs.build/en/803/api.html#icalendar.cal.Todo.sequence

I would recommend switching to the [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/). I use PyData Sphinx Theme for Plone Sphinx Theme.

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/803#discussion_r2053585333_

stevepiercy:
Fixes #804.

----
📚 Documentation preview 📚: https://icalendar--807.org.readthedocs.build/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -190,3 +190,10 @@ norecursedirs = [
     "build",
 ]
 
+[dependency-groups]
+dev = [
+    "pydata-sphinx-theme>=0.14.4",
+    "sphinx-autobuild>=2021.3.14",
+    "sphinx-copybutton>=0.5.2",
+]
+
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Minor changes:
 
 - Use ``ruff`` to format the source code.
 - Update project metadata to use License-Expression.
+- Switch to PyData Sphinx Theme for documentation. See `Issue 803 <https://github.com/collective/icalendar/issues/804>`_.
 
 Breaking changes:
 
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -7,7 +7,6 @@ iCalendar package for Python. The inspiration has come from the email package
 in the standard lib, which he thinks is pretty simple, yet efficient and
 powerful.
 
-The ``icalendar`` package is an `RFC 5545 <https://tools.ietf.org/html/rfc5545>`
-compatible parser/generator for iCalendar files.
+The ``icalendar`` package is an :rfc:`5545` compatible parser/generator for iCalendar files.
 
 .. _`Max M`: http://www.mxm.dk
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,22 +3,6 @@ import importlib.metadata
 import datetime
 import os
 
-on_rtd = os.environ.get("READTHEDOCS", None) == "True"
-
-try:
-    import sphinx_rtd_theme
-
-    html_theme = "sphinx_rtd_theme"
-    html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
-except ImportError:
-    html_theme = "default"
-    if not on_rtd:
-        print("-" * 74)
-        print(
-            "Warning: sphinx-rtd-theme not installed, building with default " "theme."
-        )
-        print("-" * 74)
-
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.coverage",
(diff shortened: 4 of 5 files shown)


pyproject.toml
```diff
@@ -190,3 +190,10 @@ norecursedirs = [
     "build",
 ]
 
+[dependency-groups]
+dev = [
+    "pydata-sphinx-theme>=0.14.4",
+    "sphinx-autobuild>=2021.3.14",
+    "sphinx-copybutton>=0.5.2",
+]
```
stevepiercy, line 198: @niccokunzmann I use uv for managing dependencies. I added these with `uv add pydata-sphinx-theme sphinx-autobuild sphinx-copybutton`, and it automatically ensures that the dependencies are compatible. I highly recommend using uv. See https://docs.astral.sh/uv/concepts/projects/dependencies/.

You also might consider managing test dependencies the same way, and merging them into this `dev` group. Developers should not have to keep separate groups, and it encourages them to write both tests and docs with a contribution.

niccokunzmann: Hm. It is a bit duplicated at the moment.

If UV works, it would be nice to create a PR that documents how to use it.

niccokunzmann:
The theme looks nice.

I have these questions:

- [ ] (1) Dependency duplication - how do we avoid this? E.g. we could modify the read-the-docs build & tox - that can be another issue.
- [ ] (2) The idea was to distinct where you are on the API page in #807. However, when I scroll the page, I cannot see where we are. I like the nice overview on the right. I would merge it anyway because you like the theme but I also want to say that strictly, this does not solve the issue from my point of view. (only blocker)
- [x] (3) The previous links all pointed to https://icalendar.readthedocs.io/en/latest/ I assume that link will still be valid after merge as it is only a theme change.
- [ ] (4) I would like you to create an issue to move to UV if you like that. I hope it does not contradict hatchling as a build tool. To be discussed in a new issue.

grafik

I like the theme and would merge it. What do you think to 1-4?

pyproject.toml
stevepiercy: How do you currently manage dependencies? IOW, what tasks would you want to perform?

I could steal from the project I maintain.

https://plone-sphinx-theme.readthedocs.io/guides/contribute.html#update-requirements

I also use GNU Make to invoke uv and friends to make it easy to contribute.

stevepiercy:
1. Dependency management and tools should be a separate issue.
2. That looks like a bug. It should use scrollspy like on the home page, but it has a weird iframe or div getting injected. I'll look into that.
3. Correct. I would not move around files in this PR. That should be handled separately and with clear thought to set up redirects. RTD supports that, and I also use a Sphinx extension [`sphinx-reredirects`](https://documatt.com/sphinx-reredirects/).
4. I don't know if there are conflicts with hatchling, as I have not used it. I know that many folks in Plone use hatch with uv, so perhaps they also use hatchling?

I want to fix that bug in item 2 first before merge.

Also to give you an idea of the structure and layout of the theme, see https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html. This will guide decisions for #805 and other reorganization.

pyproject.toml
niccokunzmann: I do not manage dependencies for icalendar for package installs. It would be nice to e.g. fix the test and docs dependencies for more reproducibility. Compatibility assurance seems to be downstream responsibility with that. I do not know what best practices are here.

stevepiercy:
Well, I couldn't figure out what's going on. Scrollspy has some open issues in the theme https://github.com/pydata/pydata-sphinx-theme/issues?q=is%3Aissue%20state%3Aopen%20scrollspy%20. None of them appear relevant.

We could merge this PR, and continue with reorganizing the documentation, including breaking up api.rst. That might fix the issue. I think that's the way forward. @niccokunzmann what do you think?

niccokunzmann:
Yes, I wanted to be in agreement before hitting the green button :)

PR #822, opened 2025-05-04
Restructure docs

### LINKED ISSUE #820: Restructure documentation, following Diátaxis framework
The [Diátaxis](https://diataxis.fr/) framework makes structuring, writing, and organizing thoughts into user-approachable documentation much clearer. The following structure should be adopted.

- [x] Home page should reflect the new structure and point readers to the proper location. It currently is ad-hoc. Needs further discussion.
- [x] [Installing iCalendar](https://icalendar.readthedocs.io/en/latest/install.html) is more than just install. There is a difference between install for use and install for contributing. I think we should have a how-to guide for each purpose, Install at `install` and Contribute at `contribute`.
- [x] [iCalendar package](https://icalendar.readthedocs.io/en/latest/usage.html) appears to be about usage as a how-to guide and reference material, and has a shitty name. The usage parts should be in a how-to guide, and the reference parts should be broken out into the Reference guide.
- [x] [iCalendar utility](https://icalendar.readthedocs.io/en/latest/cli.html) this is a how-to guide, and should fall under that structure.
- [x] https://github.com/collective/icalendar/issues/805. This should be in a directory `api` with one file per module. It's a reference guide.
- [x] Merge [About](https://icalendar.readthedocs.io/en/latest/about.html) into [home page](https://icalendar.readthedocs.io/en/latest/index.html).
- [x] [icalendar contributors](https://icalendar.readthedocs.io/en/latest/credits.html) should go into the Contribute guide.
- [x] [Maintenance](https://icalendar.readthedocs.io/en/latest/maintenance.html) should go into the Contribute guide.
- [x] [License](https://icalendar.readthedocs.io/en/latest/license.html) needs to update the copyright. I don't know whether the Plone Foundation owns the copyright, as this project is under the collective organization. This may have other legal implications. In any case, it should not be a documentation file, but a link to the license in the root of the repo.
- [x] [Changelog](https://icalendar.readthedocs.io/en/latest/changelog.html) should go into the Reference guide section.
- [x] [Contributing](https://icalendar.readthedocs.io/en/latest/contributing.html) should go into the Contributing guide.
- [x] [Security Policy](https://icalendar.readthedocs.io/en/latest/security.html) should go into the Reference guide section.
- [x] fix PyPI icon, see https://github.com/collective/icalendar/pull/849

See also:

- https://github.com/collective/icalendar/issues/805
- https://github.com/collective/icalendar/issues/626

@niccokunzmann please let me know your thoughts about these items.

stevepiercy:
Closes #820

WIP! DO NOT MERGE!


CONTRIBUTING.rst
```diff
@@ -10,17 +10,17 @@ These are some contribution examples
...
-- Sponsor a Sprint (https://plone.org/events/sprints/whatis).
+- Sponsor a `sprint <https://plone.org/news-and-events/events/sprints>`_.
 
 
 For pull requests, keep this in mind
 ------------------------------------
 
 - Add a test which proves your fix and make it pass.
 
-- Describe your change in ``CHANGES.rst``
```
niccokunzmann: I wonder if we could instead or additionally link. We have links to headings included.
What is your preference?

docs/Makefile
```diff
@@ -1,10 +1,13 @@
...
+SHELL           = bash
 
 # You can set these variables from the command line.
-SPHINXOPTS      =
-SPHINXBUILD     = sphinx-build
-SPHINXAUTOBUILD = sphinx-autobuild
+SPHINXOPTS      ?=
+
+# Internal variables.
+SPHINXBUILD     = "$(realpath ../.venv/bin/sphinx-build)"
```
niccokunzmann: check:
- [x] This needs a venv to be set up

docs/conf.py
```diff
@@ -2,17 +2,21 @@
 import importlib.metadata
 import datetime
 import os
+import sys
+
+sys.path.insert(0, os.path.abspath("../src/"))
```
niccokunzmann: Another way is to specify `.` (project root) as a dependency.

docs/index.rst
```diff
@@ -1,24 +1,48 @@
 .. include:: ../README.rst
 
+
+.. _about-icalendar:
+
+About
```
niccokunzmann: For me, this is more `History` as it does not describe its current stage.

docs/index.rst
```diff
@@ -1,24 +1,48 @@
...
+=====
+
+`Max M`_ had often needed to parse and generate iCalendar files. Finally, he got
+tired of writing ad-hoc tools. This package is his attempt at making an
+iCalendar package for Python. The inspiration has come from the email package
+in the standard lib, which he thinks is pretty simple, yet efficient and
+powerful.
+
+The ``icalendar`` package is an :rfc:`5545` compatible parser/generator for iCalendar files.
```
niccokunzmann: This sentence has a duplicate in README.rst

docs/reference/api/icalendar.alarms.rst
```diff
@@ -0,0 +1,7 @@
+icalendar.alarms module
```
niccokunzmann, line 1: My thought: Open an issue to create tests that check that all .py files have a corresponding .rst file.

docs/reference/api/icalendar.rst
```diff
@@ -0,0 +1,35 @@
...
+   :show-inheritance:
+   :undoc-members:
+
+Subpackages
+-----------
+
+.. toctree::
+   :maxdepth: 4
+
+   icalendar.timezone
```
niccokunzmann, line 17: I personally would not distinct between modules and sub packages as they should not be treated differently. As a rule for icalendar it seems: Everything that is offered publicly can be imported with `from icalendar import ...`.
The sub-packages are just to have smaller files.
In the future, I might split up the `prop.py` and the `cal.py` file. I wonder if you would like to move them in the docs then.

What benefit do you see in splitting package/module?

docs/reference/api/icalendar.timezone.rst
```diff
@@ -0,0 +1,22 @@
+icalendar.timezone package
+==========================
+
+.. automodule:: icalendar.timezone
+   :members:
+   :show-inheritance:
+   :undoc-members:
+
+Submodules
```
niccokunzmann, line 9: To be clearer about the last comment:

We have a technical structure and we have a usability structure. Technically, we structure icalendar like this. But really, how do we want people to read the documentation?

My ideas

- icalendar
  - ... 
  - icalendar.prop
    - includes all attributes needed?
  - icalendar.timezone
    - includes all attributes needed?
    
 OR
 
 - icalendar
   - ...
   - icalendar.prop
     - ...
     - icalendar.vCalAddress in its own file

OR

- icalendar
  - ...
  - icalendar.prop.rst
  - icalendar.prop.vCalAddress.rst

I think, I would like it like No. 2/3 as each class has its own file.

Let's discuss this now.

My reasons for 1 file/class:
- If we give each class its own file, they are very well documented and the URL does not change. We can even keep the link and deprecate them.
- We can list all functions in the module
- The classes become quite big with all the additional attributes and their extensive documentation. A big file with all of them is not so nice. I rather to have a Related section in each class for what is really related.

What do you think? What is a good practice?

niccokunzmann (changes requested):
Thanks for this work!

I added comments and also a bigger wondering about how to structure the .rst files in a way that helps to not expose implementation details of the internal structure to viewers who should not be concerned with this.

Let's have a chat and use this PR to bring about a good structure from the start!

docs/reference/api/icalendar.timezone.rst
niccokunzmann: Ah. Ok. I am more clear about it.

What I want users to do is this and not use the sub-packages:

```python
```

So, I want users to view the Calendar class like this:

https://icalendar.readthedocs.io/en/latest/api/Calendar.html

So I can test something like this:

```python
import icalendar
for name in dir(icalendar):
    assert (API_PATH / (name + ".rst")).is_file()
    # also test HEADER is consistent
    ## additional documentation can be in between
    # also test FOOTER is consistent
```

Is this the right place to discuss this?

niccokunzmann:
@stevepiercy as this is a blocker for contributing new documentation, I would like this to go in soon.

stevepiercy:
@niccokunzmann I am booked up through the week. This is my top non-PR review item for icalendar.

docs/reference/api/icalendar.alarms.rst
stevepiercy: This would be useful. However, I have no idea how to create such a test. Here's what DuckDuckGo's AI spat out:

```python
import os

def check_rst_files(src_dir, docs_dir):
    py_files = [f for f in os.listdir(src_dir) if f.endswith('.py')]
    rst_files = [f for f in os.listdir(docs_dir) if f.endswith('.rst')]

    for py_file in py_files:
        corresponding_rst = py_file.replace('.py', '.rst')
        if corresponding_rst not in rst_files:
            print(f"Missing RST file for: {py_file}")

src_directory = 'src'
docs_directory = 'docs'
check_rst_files(src_directory, docs_directory)
```

docs/reference/api/icalendar.rst
stevepiercy: This is the default output from running `../.venv/bin/sphinx-apidoc -e -f -M --remove-old -o reference ../src ../src/icalendar/tests`. There are options that I can issue according to https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html. If we can find the options that do what you want automatically, then we should use those. Otherwise, we'll need to do some manual adjustment of the API .rst files.

docs/reference/api/icalendar.timezone.rst
stevepiercy: The output is automatically generated. We can structure the docs however we like, but then we diverge from automatic generation. That might be OK. API documentation is purely reference to code.

CONTRIBUTING.rst
stevepiercy: I don't understand your question. Also this file should have all its content removed and point to `docs/contributing.rst`, which is now the authoritative source. I'll resolve this comment, but please reopen or create a new issue if it hasn't already been resolved.

docs/reference/api/icalendar.rst
stevepiercy: I should also say that there are other options for documenting the API, including https://sphinx-autoapi.readthedocs.io/en/latest/, but I don't know if it does what you want any better than sphinx-apidoc.

I don't do much development with icalendar. I honestly don't know or have an opinion about what is the best way to present the API docs, and I'm looking for guidance.

docs/Makefile
stevepiercy: I'd like to use uv and GNU Make for setting up, building, and testing docs, so that the developer only needs to issue `make html` and it will install a Python, create a virtual environment, install dependencies, and build the HTML docs.

niccokunzmann: That is good! If this diverges from what is documented or we do not have that documented at all, let's open a new issue after merge.

docs/reference/api/icalendar.alarms.rst
niccokunzmann: I will create a PR with a test...

niccokunzmann: #904. Feel free to comment or merge.

docs/reference/api/icalendar.rst
niccokunzmann: I would say, any way you decide is alright. In the end, people will look at the API documentation and that has code examples and that is what they copy anyway. So, the structure seems alright to me now, looking at it. Uncoupling reference structure and module structure in my mind. They serve different purposes.

stevepiercy: In the latest version of the API docs, I used the following command:

```shell
../.venv/bin/sphinx-apidoc -f -M -e -a --remove-old -o reference ../src ../src/icalendar/tests
```

`../src/icalendar/tests` is excluded. @niccokunzmann do you want to include them? I don't know if they're useful as an API reference, and a contributor would look for examples in code anyway, which is why I excluded them.

For additional options, see https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html#options.

niccokunzmann: To exclude would be:

- icalendar/tests
- icalendar/fuzzing
- icalendar/_version

I added something like this here:

https://github.com/collective/icalendar/pull/904/files#diff-d88f7202746c071dc661364a2e1f2af3dbe1a2d29d31f1aeddb5a725efad8167R16

That is more for excluding a specific test failure for absent documentation.

niccokunzmann: I have no preference on which API doc (apidoc / autoapi) to use. I think, once it is set up, we can open an issue if we notice that something is not quite as it should.

stevepiercy:
```shell
../.venv/bin/sphinx-apidoc -f -M -e -a --remove-old -o reference ../src ../src/icalendar/tests                                                   
```

Does what we want, and it excludes fuzzing and _version.

docs/Makefile
stevepiercy: Done in 243a55ee1ffad34daad42994b42ee535dd0b0499.

docs/conf.py
stevepiercy: You took another path with setting vars. I don't have an opinion.

docs/index.rst
stevepiercy: Done in a4fda48d24d565ddfdd334ae18a39a7a089afb1a

docs/index.rst
stevepiercy: Moved to contribute/credits.rst in e9644c71152d615303d2abc52e46e33adff89f5e

SashankBhamidi (approved):
Let's wait for @niccokunzmann's review before merging.

Approving.

CONTRIBUTING.rst
niccokunzmann: - [ ] TODO: Issue?

We need to check, if this does not turn up, it should be a .md file.

niccokunzmann (approved):
Thanks, this is a blocker, so I create a follow-up issue to deal with minor adjustments that come up. Merging this big contribution ! ...

niccokunzmann:
@stevepiercy, thank you very much for this beautiful documentation site! It is a joy to read it.

PR #829, opened 2025-05-09
Fix equality of components and properties based on datetime value

### LINKED ISSUE #828: [BUG] Events not equal although times are equal (only timezones differ, not actual time)
## Describe the bug

If event times are equal but their timezones differ, then the events are not equal although they happen to have the same info in it.

## To Reproduce

```
>>> from icalendar import Event
>>> e1 = Event()
>>> e2 = Event()
>>> from datetime import datetime
>>> d1 = datetime(2025, 5, 8, 12, 0, tzinfo=ZoneInfo("UTC"))
>>> d2 = d1.astimezone(ZoneInfo("Europe/Zurich"))
>>> d2 == d1
True
>>> e1 == e2
True
>>> e1.start = d1
>>> e2.start = d2
>>> e1 == e2
False
```

## Expected behavior

```
>>> e1 == e2
True
```

## Environment

 - [x] OS: Ubuntu                              
 - [x] Python version:  3.9          
 - [x] `icalendar` version: 6.2.0     

## Additional context

- [x] I tested it with the latest version `pip3 install https://github.com/collective/icalendar.git`

niccokunzmann:
This fixes #828 

- we make the properties comparable with each other
- also, a list of one datetime will be equal to one datetime to aid in adding parameters in different ways.


niccokunzmann:
After this, I would like to create a pull request for code formatting.

src/icalendar/tests/test_issue_828.py
```diff
@@ -0,0 +1,162 @@
...
+    """The values are not equal."""
+    a = transform(d1)
+    b = transform(d2)
+    assert_not_eq(a, b)
+
+
+@pytest.mark.parametrize("ddd_type", [to_vDD, to_vDDDTypes])
+@param_equal_dts
+def test_not_equal_if_parameters_differ(d1, d2, ddd_type):
+    """If the items are equal but the parameters differ, we they should not equal."""
```
stevepiercy:
```suggestion
    """If the items are equal but the parameters differ, they should not be equal."""
```

stevepiercy (changes requested):
Just a grammar issue, otherwise LGTM!

niccokunzmann:
Thanks!

PR #831, opened 2025-05-13
Test that issue 336 can be closed

### LINKED ISSUE #336: Datetimes with timezones from dateutil.tz badly supported
See also #333.

Here is some test code to illustrate the issue:

```
import datetime
import pytz
from dateutil import tz
import icalendar

utc1 = pytz.utc
utc2 = tz.UTC
oslo1 = pytz.timezone('Europe/Oslo')
oslo2 = tz.gettz('Europe/Oslo')
remote1 = pytz.timezone('Brazil/DeNoronha')
remote2 = tz.gettz('Brazil/DeNoronha')

for tz in (utc1, utc2, oslo1, oslo2, remote1, remote2):
    myevent = icalendar.Event()
    myevent.add('dtstart', datetime.datetime.now().astimezone(tz))
    print(myevent.to_ical().decode('ASCII'))
```

Particularly the 'remote2' output line is quite unacceptable ...

```
DTSTART;TZID=-02;VALUE=DATE-TIME:20211117T133359
```

There seems to be no easy fix for this, since there seems to be no obvious way to get from `tz.gettz('Brazil/DeNoronha')` to `Brazil/DeNoronha`.  (It's not just me being eurocentric - DeNoronha is really one of the most remote time zones for most of the world population, just check the map).  I see that the tzinfo object from dateutil has a file name - perhaps it's possible to reverse-engineer something from the filename - but it will be hard to make a platform-independent solution.

I could perhaps try to work out something, but only if someone (preferably someone with the authority to merge pull requests) says it's a good idea.

niccokunzmann:
This closes #336

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,7 +24,7 @@ New features:
 
 Bug fixes:
 
-- ...
+- Test that we support dateutil timezones as outlined in `Issue 336 <https://github.com/collective/icalendar/issues/336>`_.
 
 
 6.2.0 (2025-05-07)
--- a/src/icalendar/tests/test_issue_336_dateutil_timezone.py
+++ b/src/icalendar/tests/test_issue_336_dateutil_timezone.py
@@ -1,0 +1,27 @@
+"""We would like to be able to get the right timezone names.
+
+See https://github.com/collective/icalendar/issues/336
+
+It appears that the timezone Brazil/DeNoronha is actually America/Noronha.
+"""
+
+from datetime import datetime
+from dateutil import tz
+
+from icalendar import Event
+from icalendar.timezone import tzid_from_tzinfo
+
+
+def test_timezone_name_directly():
+    """Try to get the name directly."""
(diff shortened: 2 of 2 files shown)


niccokunzmann:
@tobixen have a look! You opened #336 and might be interested to know that it works now :)

tobixen:
Awesome! :-)

This does indeed solve the issue I raised on #336, but I'm not sure it is sufficient to solve interoperability with some of the caldav servers, see also https://github.com/home-assistant/core/issues/142555

The easiest way seems to send timestamps to the server in UTC, but then again that will cause problems with recurrences.

tobixen:
If the icalendar library now also supports to generate the VTIMEZONE component, then that's perfect.  Next step is to send Olson identifier to servers that support it, and VTIMEZONE to servers that don't.  Late 2026, probably :-)

niccokunzmann:
Yes, it does support it and it is slow though because it will work with any timezone object. There is a Python package that includes pre-made VTIMEZONE components, I think.

PR #834, opened 2025-05-13
Deprecate the UID Generator

### LINKED ISSUE #816: Use Python's UID generator
Currently, we have our own implementation for UID: [icalendar.tools.UIDGenerator](https://icalendar.readthedocs.io/en/latest/api.html#icalendar.tools.UIDGenerator). This is not really needed any more.

- [x] remove the use of this class internally
- [x] adjust the documentation
- [x] use the Python's built-in uuid module for what happens inside
- [x] add deprecation notice for future removal if used

See also:

- #740

niccokunzmann:
This fixes #816

----
📚 Documentation preview 📚: https://icalendar--834.org.readthedocs.build/

### DIFF
--- a/src/icalendar/tools.py
+++ b/src/icalendar/tools.py
@@ -17,7 +17,7 @@ class UIDGenerator:
     
     .. deprecated:: 6.3.0
 
-        Use the Python standard library's :function:`uuid4` instead.
+        Use the Python standard library's :func:`uuid.uuid4` instead.
     """
 
     chars = list(ascii_letters + digits)
@@ -28,7 +28,7 @@ class UIDGenerator:
 
          .. deprecated:: 6.3.0
 
-             Use the Python standard library's :function:`uuid4` instead.
+             Use the Python standard library's :func:`uuid.uuid4` instead.
         """
         warn(
             "Use https://docs.python.org/3/library/uuid.html#uuid.uuid4 instead.",
@@ -39,20 +39,15 @@ class UIDGenerator:
 
     @staticmethod
     def uid(host_name="example.com", unique=""):
-        """Generates a unique id consisting of:
-            datetime-uniquevalue@host.
-        Like:
+        """Generates a unique ID consisting of ``datetime-uniquevalue@host``.
+
+        For example:
+            
             20050105T225746Z-HKtJMqUgdO0jDUwm@example.com
+        
+        .. deprecated:: 6.3.0
 
-"""Generates a unique ID consisting of ``datetime-uniquevalue@host``.
-   
-   For example:
-       
-       20050105T225746Z-HKtJMqUgdO0jDUwm@example.com
-   
-   .. deprecated:: 6.3.0
-
-        Use the Python standard library's :function:`uuid4` instead.
+            Use the Python standard library's :func:`uuid.uuid5` instead.
         """
         from icalendar.prop import vDatetime, vText
         warn(
--- a/src/icalendar/tools.py
+++ b/src/icalendar/tools.py
@@ -22,9 +22,12 @@ class UIDGenerator:
 
     @staticmethod
     def rnd_string(length=16) -> str:
-        """Generates a string with random characters of length."""
+        """Generates a string with random characters of length.
+
+        Use https://docs.python.org/3/library/uuid.html#uuid.uuid4 instead.
+        """
         warn(
-            "Use https://docs.python.org/3/library/uuid.html#uuid.uuid4",
+            "Use https://docs.python.org/3/library/uuid.html#uuid.uuid4 instead.",
             WillBeRemovedInVersion7,
             stacklevel=1
         )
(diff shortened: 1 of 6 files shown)


src/icalendar/tools.py
```diff
@@ -1,19 +1,36 @@
...
 from icalendar.parser_tools import to_unicode
 
+from .error import WillBeRemovedInVersion7
+
 
 class UIDGenerator:
-    """If you are too lazy to create real uid's."""
+    """If you are too lazy to create real uid's.
+
+    Use https://docs.python.org/3/library/uuid.html#uuid.uuid4
```
stevepiercy: I assume this will be released in the next minor version. Also let's make sure the reference to Python docs renders properly.

```suggestion
    """Use this only if you're too lazy to create real UUIDs.
    
    .. deprecated:: 6.3.0

        Use the Python standard library's :function:`uuid4` instead.
```

src/icalendar/tools.py
```diff
@@ -1,19 +1,36 @@
...
 
     chars = list(ascii_letters + digits)
 
     @staticmethod
-    def rnd_string(length=16):
-        """Generates a string with random characters of length."""
+    def rnd_string(length=16) -> str:
+        """Generates a string with random characters of length.
+
+        Use https://docs.python.org/3/library/uuid.html#uuid.uuid4 instead.
```
stevepiercy:
```suggestion
         .. deprecated:: 6.3.0

             Use the Python standard library's :function:`uuid4` instead.
```

src/icalendar/tools.py
```diff
@@ -22,8 +39,15 @@ def uid(host_name="example.com", unique=""):
             datetime-uniquevalue@host.
         Like:
             20050105T225746Z-HKtJMqUgdO0jDUwm@example.com
+
+        Use https://docs.python.org/3/library/uuid.html#uuid.uuid5 instead.
```
stevepiercy: I can't make a suggestion for the entire docstring in the GitHub UI, so this one will need to be applied manually.

```suggestion
"""Generates a unique ID consisting of ``datetime-uniquevalue@host``.
   
   For example:
       
       20050105T225746Z-HKtJMqUgdO0jDUwm@example.com
   
   .. deprecated:: 6.3.0

        Use the Python standard library's :function:`uuid4` instead.
```

stevepiercy (changes requested):

niccokunzmann:
@stevepiercy have a look again.

Thanks for the suggestions. I did not know the syntax.
Also: There is a change in the changelog file but it is not recognized. I wonder what happened to Plone's check.

niccokunzmann:
@mauritsvanrees, We have a change in the Changelog.rst file but the check is still red. It seems to me that it only checks the first commit of a PR.

mauritsvanrees:
@niccokunzmann That check has been flaky of late.  See https://github.com/plone/mr.roboto/issues/168
You can probably ignore it.

src/icalendar/tools.py
```diff
@@ -1,29 +1,60 @@
...
     @staticmethod
     def uid(host_name="example.com", unique=""):
-        """Generates a unique id consisting of:
-            datetime-uniquevalue@host.
-        Like:
+        """Generates a unique ID consisting of ``datetime-uniquevalue@host``.
+
+        For example:
+            
             20050105T225746Z-HKtJMqUgdO0jDUwm@example.com
```
stevepiercy: This should not render as a clickable mailto link, but as a code block. Verify at https://icalendar--834.org.readthedocs.build/en/834/api.html#icalendar.tools.UIDGenerator.

```suggestion
            .. code-block:: text

                20050105T225746Z-HKtJMqUgdO0jDUwm@example.com
```

src/icalendar/tools.py
```diff
@@ -1,29 +1,60 @@
...
+        
+        .. deprecated:: 6.2.1
+
+            Use the Python standard library's :func:`uuid.uuid5` instead.
```
stevepiercy, line 52: Why do we use uuid5 here and below instead of uuid4, as we did in the earlier docstrings?

stevepiercy (changes requested):
A syntax fix, and a question to check which UUID function to use.

stevepiercy:
Check this file.

- https://icalendar--834.org.readthedocs.build/en/834/api.html#icalendar.tools.UIDGenerator

src/icalendar/tools.py
niccokunzmann: uuid5 is closer to the parameters. It gets influenced by the domain, too.

stevepiercy: Should we change uuid4 to uuid5? Just want to make sure that the difference is deliberate and not a typo.

niccokunzmann: This is not a version thing of what is better/later: uuid4 is random, uuid5 uses certain information to generate the random value with a hash so you can see if they are from the same 'source'/domain.

stevepiercy: I understand the differences of the two functions and other uuid functions as well. My question is whether the difference here was a deliberate choice or a typo. It sounds like the former.

stevepiercy (approved):
It renders correctly now.

If the UUID thing was a deliberate choice, and not a typo, then OK to merge.

niccokunzmann:
Thanks!

PR #835, opened 2025-05-13
Add UID property

### LINKED ISSUE #740: Create the UID property
Please create the UID property in icalendar.cal.Component.

- get is always str, empty if no value is given
- set
- create a docstring with text from the rfc https://www.rfc-editor.org/rfc/rfc7986.html#section-5.3
- reference the rfc
- add an example in the docstring of how to set a properly generated uid
- log changes
- test the code

- [ ] for the alarm, also cosider `X-ALARMUID`

see also:
- https://github.com/collective/icalendar/issues/69#issuecomment-2128256014
- https://github.com/collective/icalendar/issues/655
- https://github.com/collective/icalendar/discussions/662
- #816

niccokunzmann:
This adds unified access to UID.
Fix #740 

----
📚 Documentation preview 📚: https://icalendar--835.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,7 +20,7 @@ Breaking changes:
 
 New features:
 
-- ...
+- Add the ``uid`` property to ``Alarm``, ``Event``, ``Calendar``, ``Todo``, and ``Journal`` components. See `Issue 740 <https://github.com/collective/icalendar/issues/740>`_.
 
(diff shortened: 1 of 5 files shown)


stevepiercy (approved):
LGTM: https://icalendar--832.org.readthedocs.build/en/832/api.html#icalendar.prop.vRecur

Issue #826, opened 2025-05-07
[BUG] Documentation Version is invalid


niccokunzmann:
## Describe the bug

When visiting the [stable documentation](https://icalendar.readthedocs.io/en/stable/), I cannot see the current version. Instead, I see this:

![Image](https://github.com/user-attachments/assets/224f3b5c-e1a5-4f38-be6d-b00478392d50)

The problem here is that we build the package for the documentation without pulling the tags first and being on a tag.
With that, hatch-vcs will create this version information.

## Expected behavior

I would like to see the latest stable tag in the title. In this case v6.2.0.

## Environment

readthe docs

## Additional context

- https://pypi.org/project/hatch-vcs/
- #823

niccokunzmann:
Solved by #837

niccokunzmann:
https://icalendar.readthedocs.io/ now redirects to the `stable` build of the documentation. So: Only if we create a new release, users of icalendar will see the changes we made on the recommended documentation site.

PR #837, opened 2025-05-14
get tags and set version

niccokunzmann:
By building the package and pulling the tags, we should be able to build the documentation so it shows the version of the package.

@stevepiercy What are your thoughts on the commands? Would it be better to put them into make?

Contributes to #826

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,7 +26,7 @@ Bug fixes:
 
 - Test that we can add an RRULE as a string. See `Issue 301 <https://github.com/collective/icalendar/issues/301>`_.
 - Test that we support dateutil timezones as outlined in `Issue 336 <https://github.com/collective/icalendar/issues/336>`_.
-- Build documentation on readthedocs with the version identifier. See `Issue 826 <https://github.com/collective/icalendar/issues/826>`_.
+- Build documentation on Read the Docs with the version identifier. See `Issue 826 <https://github.com/collective/icalendar/issues/826>`_.
 
 
 6.2.0 (2025-05-07)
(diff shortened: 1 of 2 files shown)


niccokunzmann:
image
[efc5f62](https://github.com/collective/icalendar/pull/837/commits/efc5f62a1b835f82e5819d64b18509c2c5147e8f)

CHANGES.rst
```diff
@@ -26,6 +26,7 @@ Bug fixes:
 
 - Test that we can add an RRULE as a string. See `Issue 301 <https://github.com/collective/icalendar/issues/301>`_.
 - Test that we support dateutil timezones as outlined in `Issue 336 <https://github.com/collective/icalendar/issues/336>`_.
+- Build documentation on readthedocs with the version identifier. See `Issue 826 <https://github.com/collective/icalendar/issues/826>`_.
```
```suggestion
- Build documentation on Read the Docs with the version identifier. See `Issue 826 <https://github.com/collective/icalendar/issues/826>`_.
```

stevepiercy (changes requested):
Use proper brand name.

stevepiercy:
> By building the package and pulling the tags, we should be able to build the documentation so it shows the version of the package.

That's creative. I do this in the `conf.py` of one package I maintain.

```python
from my_package import __version__
from packaging.version import Version

# ...

# The version info for the project you're documenting, acts as replacement for
# |version| and |release|, also used in various other places throughout the
# built documents.
#
# The full version, including alpha/beta/rc tags.
release = __version__
# The short X.Y version.
version = Version(str(release)).base_version
```

niccokunzmann:
Yeah. This is a hatchling issue...

stevepiercy:
So... what would you like to do? Wait for a fix in hatchling or is it possible to use my `conf.py` method?

niccokunzmann:
What happened before this PR:

1. git clone
2. there is no src/icalendar/_version.py file
3. build docs
4. default version

What happens after this PR:

1. git clone
2. get tags
3. use python -m build to generate src/icalendar/_version.py with hatchling-vcs
4. build docs
5. correct version

---

Using your conf.py file modification, the version could still be the default one - it does not generate the _version.py file.

But instead of discussing this, we could just merge it and see if it works on a tag. I would like to create a release today so that we can check.

niccokunzmann:
So, I am merging it now, so I can try it out with a new release..

PR #842, opened 2025-05-15
Introduce new() to Calendar class

### LINKED ISSUE #596: Mandatory properties are neither enforced nor set to a default
According to RFC 5545, every VCAL needs a `PRODID` and `VERSION`. If those are not explicitly added, `Calendar.to_ical()` produces non-conforming output because they are missing.

It could be a deliberate design choice to leave this open. But I wonder, how often do people want to generate invalid output, and shouldn't the library enforce a valid calendar object somehow?

Just some suggestions here.
### VERSION
As this library claims to be implementing RFC 5545, I think this should just be `2.0` by default. If you add it again with a different value, that's fine, it will be replaced.

### PRODID
What about setting this to something like `-//python.collective.icalendar//NONSGML python.collective.icalendar.app.event//EN`? (I'm not sure how exactly these strings should be constructed, I just wrote something based on a test case I saw in this repo). Same here, if you add it again with a different value, it will be replaced.

Both of above improvements are fully backwards-compatible, AFAIK. Please let me know your thoughts.

niccokunzmann:
`icalendar.Calendar` now has a `new()` classmethod which allows the creation of valid calendars.

Fixes #596.
Contributes to #843

----
📚 Documentation preview 📚: https://icalendar--842.org.readthedocs.build/


niccokunzmann:
Closing in favour of #845

Issue #843, opened 2025-05-15
Aid the creation of valid calendars.


niccokunzmann:
Some properties are required, others optional, some are strictly typed.

In order to start creating valid calendar times with many other features included, this issue proposes adding a `new()` method to the `Component` classes.

Example:

```python
>>> event = Event.new(start=date(2025, 10, 11))
>>> print(event.to_ical().decode())
BEGIN:VEVENT
DTSTART;VALUE=DATE:20251011
UID:<randomly-generated-uuid4>
DTSTAMP:<now>
END:VEVENT
>>> calendar = Calendar.new(components=[event])
```

Benefits:
- The calendar can create a valid calendar from the start
- E.g. DTSTART is required IF the calendar has no method -> we can run checks on the whole object structure
- We can use getters and setters to set the values, providing additional verification without code duplication
- Generation and setting of default values is still possible
- We do not deprecate or interfere with the minimal way of `event = Event()`

See also:
- https://github.com/collective/icalendar/issues/596
- https://github.com/collective/icalendar/issues/315
- https://github.com/collective/icalendar/discussions/662
- https://github.com/collective/icalendar/discussions/536
- https://github.com/collective/icalendar/discussions/548
- #817

PR #845, opened 2025-05-16
Add new() method to component classes

niccokunzmann:
This creates a new method for the other classes.

Fixes https://github.com/collective/icalendar/issues/596.
Contributes to https://github.com/collective/icalendar/issues/843

See also
- #842 

Changes:
- add a new() classmethod
- use the ruff formatter
- add some new attributes to the new method
- split up cal.py

----
📚 Documentation preview 📚: https://icalendar--845.org.readthedocs.build/


niccokunzmann:
@stevepiercy, I could use some feedback here:

- how is it with the splitting of cal.py
- I would like to extend the new() interface in the future without breaking changes - that is why I included the error message. What are your thoughts?

niccokunzmann:
@jacadzaca, this is a bigger change and if you have time, I would value your input/a read through for having another pair of eyes on possible things that I might not consider/forget.

src/icalendar/cal/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    "Event",
+    "FreeBusy",
+    "IncompleteComponent",
+    "Journal",
+    "Timezone",
+    "TimezoneDaylight",
+    "TimezoneStandard",
+    "Todo",
+    "get_example",
+    "types_factory",
```
niccokunzmann: TODO: 

- [ ] make types factory singleton, too
- [ ] document move to singleton
- [ ] remove export from module (cleaner)

niccokunzmann:
This PR has a huge diff.

Essentially:

- it splits up cal.py in to a module
- makes the components_factory a singleton
- uses the ruff pre-commit to also format the code in hundreds of places to make it better in style.


I would rather avoid doing this work again and it blocks the rest of my work as most of it would touch the cal.py module.
If you could give this a review and we merge it soon, then that would be much appreciated.

stevepiercy:
I have no interest in reviewing changes from automatic formatting. I would strongly recommend not to apply formatting in this PR, and instead do it in a subsequent PR. It makes it difficult to review the substantive changes.

niccokunzmann:
Hi @stevepiercy, I understand that! I crisp PR would be good.

I would like to share from my side that I did this PR twice because of merging: Twice going through the code base and also twice going through 100+ places manually to have the pre-commit accept my commits.

It is ok for me if you do not review the whole code. I trust that I can copy the code quite well - the tests are running.

For me, we have two threads of conversation:
1. `new()` (important)
2. pre-commit (annoying)

## new()

The real important question here for me is: How to handle `new()` and  if you have ideas about this.

- Does it feel good and extensible?
- `new()` will include checks and these will get more thorough over time. That means: Code created with icalendar 7 might break with icalendar 7+ because we introduce more checks. How do we properly document this fact? I would not want to raise the major number each time the checks change... Do you have some ideas from other projects?

## pre-commit

Yeah. Reviewing autoformatting is annoying.
I am wondering: We have not formatted the whole code base. There was a PR before for the whole code base. However, ruff has evolved since then and e.g. sorts the names in `__all__` exports of modules.
So, when we use the pre-commit, we will always get formatting into our PR when a file is committed.

1. Sometimes, ruff requires manual intervention to resolve the requirement. I think, this can be a bit much for other contributors.
2. A formatting PR is a bit like this one: touches lots of code and then other PRs might have merge conflicts. So, it is good to put that in there another time. I would like the ruff checks to be mandatory green. This PR does quite a bit towards that.

## summary

`new()` is important. The rest is refactoring, formatting, copy and small clean-up changes.

niccokunzmann:
I am merging this now to progress as this blocks further nice developments.

PR #847, opened 2025-05-18
Update README.rst

niccokunzmann:
Update the README to use new getter/setter properties

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,14 +81,14 @@ Such a calendar can then be edited and saved again.
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


stevepiercy (approved):

niccokunzmann:
@stevepiercy, I would like to put you down as the documentation lead in the maintenance section of the documentation. 
With that, I want to clarify the roles of who does what.
What do you think?

stevepiercy:
> @stevepiercy, I would like to put you down as the documentation lead in the maintenance section of the documentation. With that, I want to clarify the roles of who does what. What do you think?

Sounds good to me. Something like "Tag @stevepiercy for any documentation matters." to start.

PR #849, opened 2025-05-19
make pypi link visible

niccokunzmann:
This makes the link visible.


stevepiercy:
This is the way https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/header-links.html#svg-image-icons. This PR is a workaround that doesn't accurately represent the link to PyPI.

niccokunzmann:
Ah right. I will close this and add it as an item to the docs issue.

PR #850, opened 2025-05-19
Remove quotes from safe characters

### LINKED ISSUE #836: [BUG] icalendar unnecessarily quotes TZID for custom timezones in DTSTART/DTEND
`icalendar` adds quotes to the `TZID` parameter in `DTSTART`/`DTEND` properties when re-serializing an ICS file, if the `TZID` refers to a custom timezone defined in a `VTIMEZONE` component (e.g., "Eastern Standard Time" from Outlook). This can cause validation/interoperability issues.

**The Issue:**

If an input ICS has:
```
BEGIN:VTIMEZONE
TZID:Eastern Standard Time
...
END:VTIMEZONE
...
DTSTART;TZID=Eastern Standard Time:20241028T170000
```

`icalendar` re-serializes it as:
```
DTSTART;TZID="Eastern Standard Time":20241028T170000
```
The `TZID` in the `VTIMEZONE` definition remains unquoted, but the quoting in event properties is the problem.

**Steps to Reproduce:**

1.  **Input `.ics` (minimal):**
    ```vcf
    BEGIN:VCALENDAR
    METHOD:PUBLISH
    PRODID:Microsoft Exchange Server 2010
    VERSION:2.0
    BEGIN:VTIMEZONE
    TZID:Eastern Standard Time
    BEGIN:STANDARD
    DTSTART:16010101T020000
    TZOFFSETFROM:-0400
    TZOFFSETTO:-0500
    RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=1SU;BYMONTH=11
    END:STANDARD
    BEGIN:DAYLIGHT
    DTSTART:16010101T020000
    TZOFFSETFROM:-0500
    TZOFFSETTO:-0400
    RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=2SU;BYMONTH=3
    END:DAYLIGHT
    END:VTIMEZONE
    BEGIN:VEVENT
    UID:minimal-demo-event-est-20241028@example.com
    SUMMARY:Anonymous Test Event for TZID
    DTSTART;TZID=Eastern Standard Time:20241028T170000
    DTEND;TZID=Eastern Standard Time:20241028T180000
    DTSTAMP:20250514T023916Z
    END:VEVENT
    END:VCALENDAR
    ```

2.  **Python code:**
    ```python

    ics_content = b"""... (paste above ICS content here) ..."""
    cal = Calendar.from_ical(ics_content)
    print(cal.to_ical().decode())
    ```

3.  **Actual Output (snippet):**
    ```vcf
    DTSTART;TZID="Eastern Standard Time":20241028T170000
    DTEND;TZID="Eastern Standard Time":20241028T180000
    ```

**Expected Behavior:**

`DTSTART`/`DTEND` `TZID` should remain unquoted if it refers to a defined `VTIMEZONE` and the `TZID` string itself doesn't require quoting per RFC 5545.
Example: `DTSTART;TZID=Eastern Standard Time:20241028T170000`

**Impact:**

This quoting can cause validation failures and interoperability problems with systems expecting the original unquoted format for such `VTIMEZONE`-defined TZIDs.

**Suggestion:**

`icalendar` could avoid quoting a `TZID` in event properties if it matches a `TZID` in a `VTIMEZONE` (and the string doesn't contain special characters requiring quotes).

Context on Outlook's non-standard TZIDs: [Google Support Thread](https://support.google.com/calendar/thread/152611780/importing-calendar-events-via-ics-files-get-wrong-timezone?hl=en)

Thanks!

Related issue https://github.com/collective/icalendar/issues/44

niccokunzmann:
This fixes #836.

These characters were escaped but do not need escaping: white space, single quote


niccokunzmann:
I will do a bit more work to check why the quoting is expected.

niccokunzmann:
I added escaping for a few parameters in there, e.g. CN which is a name and the RFC quotes it although it is not required to be quoted.

abe-101:
lgtm
I did some testing against mergecalweb and it indeed does resolve the issue
Thank you very much for the quick turn around

PR #856, opened 2025-05-21
Make sure we do not loose information

### LINKED ISSUE #187: Property values ignoring property parameters

niccokunzmann:
Make sure we do not loose information because the VALUE parameter has a wrong type for datetime values

This attempts to fix #187.

Also:

- add proper type to from_ical
- add VALUE enum
- add value getter for Parameters
- convert date to tzinfo if we would loose the TZID information in the value

To merge after #845

### DIFF
--- a/src/icalendar/parser.py
+++ b/src/icalendar/parser.py
@@ -7,8 +7,11 @@ conversion is attempted.
 """
 from __future__ import annotations
 
+import functools
 import os
+from typing import Optional
 from icalendar.caselessdict import CaselessDict
+from icalendar.enums import VALUE
 from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE
 from icalendar.parser_tools import SEQUENCE_TYPES
 from icalendar.parser_tools import to_unicode
@@ -171,6 +174,29 @@ def q_join(lst, sep=",", always_quote=False):
     return sep.join(dquote(itm, always_quote=always_quote) for itm in lst)
 
 
+def single_string_parameter(func):
+    """Create a parameter getter/setter for a single string parameter."""
+
+    name = func.__name__
+
+    @functools.wraps(func)
+    def fget(self: Parameters):
+        """Get the value."""
+        return self.get(name)
+
+    def fset(self: Parameters, value: str|None):
+        """Set the value"""
+        if value is None:
+            fdel(self)
+        else:
+            self[name] = value
+
+    def fdel(self: Parameters):
+        """Delete the value."""
+        self.pop(name, None)
+
+    return property(fget, fset, fdel, doc=func.__doc__)
+
 class Parameters(CaselessDict):
     """Parser and generator of Property parameter strings. It knows nothing of
     datatypes. Its main concern is textual structure.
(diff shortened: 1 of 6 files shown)


src/icalendar/tests/test_issue_187_type_and_value_do_not_match.py
```diff
@@ -0,0 +1,46 @@
...
+)
+def test_wrong_dates_are_converted(string, expected_params, expected_property_value):
+    """We convert these dates to the correct value if they turn up.
+
+    We want to make sure that these are working, have values and do not throw exceptions.
+    Information should not get lost except if it is contradictory.
+    So, the tests can be changed if a different resolution is needed.
+
+    The problem: The VALUE parameter might not correspond to the actual value.
+    However, we correctly parse the value and do not loose information.
```
stevepiercy:
```suggestion
    However, we correctly parse the value and do not lose information.
```

src/icalendar/prop.py
```diff
@@ -571,12 +572,15 @@ def from_ical(cls, ical, timezone=None):
...
+                tzinfo = tzp.timezone(timezone)
+                if tzinfo is not None:
+                    return to_datetime(vDate.from_ical(ical)).replace(tzinfo=tzinfo)
             return vDate.from_ical(ical)
-        elif len(ical) in (6, 7):
+        if len(ical) in (6, 7):
             return vTime.from_ical(ical)
-        else:
-            raise ValueError(f"Expected datetime, date, or time, got: '{ical}'")
+        raise ValueError(f"Expected datetime, date, or time, got: '{ical}'")
```
stevepiercy:
```suggestion
        raise ValueError(f"Expected datetime, date, or time. Got: '{ical}'")
```

stevepiercy (changes requested):
Minor grammar and spelling fixes. Otherwise LGTM and can be merged.

src/icalendar/parser.py
```diff
@@ -171,6 +174,29 @@ def q_join(lst, sep=",", always_quote=False):
     return sep.join(dquote(itm, always_quote=always_quote) for itm in lst)
 
 
+def single_string_parameter(func):
```
stevepiercy, line 184: This did not show up in the docs PR preview. Not sure why.

stevepiercy: https://icalendar--856.org.readthedocs.build/en/856/search.html?q=single_string_parameter

stevepiercy (changes requested):

stevepiercy:
Needs change log.

Docs to review:

- https://icalendar--856.org.readthedocs.build/en/856/api.html#icalendar.enums.VALUE

src/icalendar/parser.py
niccokunzmann: This is ok. It is used internally only and might be moved. I would not advertise it to people and so it is missing in `__all__`.

stevepiercy: Thanks for the explanation.

Issue #859, opened 2025-05-30
Add vCard support


michaelgrahamevans:
Consider adding a `Card` type for parsing and creating [RFC 6350](https://www.rfc-editor.org/rfc/rfc6350) vCard objects.

This might be out of scope of this library, however I see there is some [soft support](https://github.com/collective/icalendar/pull/143) for the format and there was [previously interest in adding full support](https://github.com/collective/icalendar/issues/186).

niccokunzmann:
Thank you for reporting this!
That is a good idea.
Would you like to work on this? With a bit of help?

tobixen:
(The vobject library is a "competitor" to the icalendar library.  It was unmaintained and orphaned for a longer while, but it looks like there is an actively maintained official fork of the project with presence on pypi and own web-page)

niccokunzmann:
@tobixen what is the fork that works? I think, we should mention both in the related projects.

tobixen:
The one that is currently on pypi has github project page https://github.com/py-vobject/vobject and home page http://py-vobject.github.io/ ... though it only shows history up until 2018 and is also reachable by https://vobject.sameenkarim.com/ ... and the py-vobject seems to be a fork of Sameen Karim's repository, https://github.com/skarim/vobject.  

There is also http://vobject.skyhouseconsulting.com/history.html and https://github.com/skarim/vobject

My memories are vague, but we had serious issues as the caldav library was very much dependent on vobject.  Freely from my memory, the dominant fork was at some point https://github.com/tBaxter/vobject - it is forked from https://github.com/adieu/vobject ... then again, if the adieu-version is the original vobject, I'd expect there to be more open pull requests.

Of course I can do more archaeological work on this, I should have quite some information on the history of vObject in my mailboxes and there should also be quite much information about it out there on the open Internet ... but I don't think it serves any purpose, I think we can trust pypi on what is the currently actively maintained version of vObject :-)

tobixen:
Of course, the full story is condensed down in https://github.com/py-vobject/vobject#history :-)

PR #861, opened 2025-06-02
Add VAVAILABILITY + AVAIABLE for RFC 7953

### LINKED ISSUE #654: RFC 7953 - VAVAILABILITY component
In order to claim compatibility to RFC 7953, we should add an `Availability` component equivalent to `Event` to represent `VAVAILABILITY`

Add
- [ ] class Availability
- [ ] dtstamp REQUIRED
- [ ] uid REQUIRED
- [ ] busytype
- [ ] class
- [ ] created ONCE
- [ ] description ONCE
- [ ] dtstart REQUIRED
- [ ] last-mod ONCE
- [ ] location ONCE
- [ ] organizer
- [ ] priority
- [ ] seq
- [ ] summary ONCE
- [ ] url
- [ ] dtend XOR duration
- [ ] categories
- [ ] comment
- [ ] contact
- [ ] recurid ONCE
- [ ] rrule ONCE
- [ ] Examples from RFC

See also:
- https://www.rfc-editor.org/rfc/rfc7953.html
- https://github.com/collective/icalendar/issues/589
- #596

niccokunzmann:
- add classes
- add new()
- add setters and getters
- check docs to reflect changes
- docs now use the 

TODO:

- [x] fully check new()
- [x] add check that new() required start
- [x] add changelog
- [x] Test other examples
- [x] add start and end, attributed for scheduling
- [x] make start a must

Fixes #654

----
📚 Documentation preview 📚: https://icalendar--861.org.readthedocs.build/


docs/conf.py
```diff
@@ -1,7 +1,12 @@
...
-import importlib.metadata
 import datetime
-import os
+import importlib.metadata
+import sys
+from pathlib import Path
+
+HERE = Path(__file__).parent
+SRC = HERE.parent / "src"
+sys.path.insert(0, str(SRC))  # update docs from icalendar source for livehtml
```
stevepiercy, line 9: These seems over complicated, and I'm not clear what problem it is supposed to solve. Can you explain?

stevepiercy (commented):
@niccokunzmann I started a review, then realized you have TODO items in the PR description. Do you intend to do them in this PR or another? If this one, then I'll hold off on a review, else I'll resume my review. Please let me know. Thank you!

docs/conf.py
niccokunzmann: When I did recurring-ical-events, this put the icalendar import to the front of the paths to check. So that the source is used and not the installed library. With this, sources are renewed and autodoc loads realoaded modules.

niccokunzmann:
@stevepiercy, I will do the TODO items first and then notify you. Thank you!

niccokunzmann:
@tobixen, this might be of interest to you as VAVAILABILITY and RFC 7953 is used in conjunction with CALDAV. Would you like to give this a review from this perspective?

tobixen:
> src/icalendar/cal/avaiable.py

Shouldn't it be available with a lower case L?

I'm growing old enough that I ought to use glasses while sitting by the computer, yet I rarely use them ... and I also have problems with the l-key on my keyboard at the moment, so this looks like a mistake I could do :-)

tobixen:
I've had a brief look, but it's not exactly a code review.  It's a big changeset, and I would have to start reading the RFC - I don't have time for that right now at least - and it seems like the rest of the week will be very busy.

An issue should be filed in the caldav project to implement support for this and the relevant RFCs.

niccokunzmann:
After this, I think, I could start with #859

Issue #219, opened 2017-03-16
Can't parse iOS location


hobarrera:
Parsing this [apparently] valid file fails/crashes:

```
BEGIN:VCALENDAR
CALSCALE:GREGORIAN
PRODID:-//Apple Inc.//iOS 10.2.1//EN
VERSION:2.0
BEGIN:VTODO
COMPLETED:20170312T122019Z
CREATED:20170312T051318Z
DTSTAMP:20170312T122023Z
LAST-MODIFIED:20170312T122021Z
PERCENT-COMPLETE:100
PRIORITY:1
SEQUENCE:0
STATUS:COMPLETED
SUMMARY:Cosas en el freezer
UID:5ACD2904-D0CE-4097-BC75-C1EB95C0BB22
X-APPLE-SORT-ORDER:510988398
BEGIN:VALARM
ACKNOWLEDGED:20170312T122015Z
ACTION:DISPLAY
DESCRIPTION:Reminder
TRIGGER;VALUE=DATE-TIME:19760401T005545Z
UID:E13840A6-C30C-4BF0-A3E1-73EF27E34BD9
X-APPLE-PROXIMITY:ARRIVE
X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS=Ayacucho 983\\nPlata Baja 
 \"A\"\\nC1111AAC CABA\\nArgentina\\nArgentina;X-APPLE-ABUID="ab://Home";
 X-APPLE-RADIUS=0;X-APPLE-REFERENCEFRAME=1;X-TITLE=Home:geo:
 -34.597613,-58.395858
X-WR-ALARMUID:E13840A6-C30C-4BF0-A3E1-73EF27E34BD9
END:VALARM
END:VTODO
END:VCALENDAR
```

Traceback:

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/icalendar/parser.py", line 241, in from_ical
    validate_param_value(v, quoted=False)
  File "/usr/lib/python3.6/site-packages/icalendar/parser.py", line 132, in validate_param_value
    raise ValueError(value)
ValueError: Ayacucho 983%5CnPlata Baja \"A\"%5CnC1111AAC CABA%5CnArgentina%5CnArgentina

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/icalendar/parser.py", line 336, in parts
    strict=self.strict)
[...]

PR #866, opened 2025-06-10
Fix Issue 219

niccokunzmann:
This makes a start on Issue #219

TODO:

- [ ] just disabling verification does the trick. This seems to be not intended. We should test verification. https://github.com/collective/icalendar/pull/866/files#r2137725206
- [ ] we should check that in case the validation fails, we unescape `\"`.

----
📚 Documentation preview 📚: https://icalendar--866.org.readthedocs.build/


src/icalendar/parser.py
```diff
@@ -245,12 +243,11 @@ def from_ical(cls, st, strict=False):
                         v2 = v.strip('"')
                         validate_param_value(v2, quoted=True)
                         vals.append(rfc_6868_unescape(v2))
+                    # validate_param_value(v, quoted=False)
```
niccokunzmann: This makes the tests pass.

SashankBhamidi:
Hi @niccokunzmann, I'm going to start getting back to reviews to close/merge PRs starting from oldest to new. 

Good to see we have so many new contributions! 

What do you want to do with this PR? Lmk.

niccokunzmann:
I close it. It is linked. If anyone wants to pick up work on this. They are welcome or can just have a look and copy what they need. Thanks!

PR #872, opened 2025-07-09
unique identifier must be included in VEVENT, VTODO, and VJOURNAL #315

### LINKED ISSUE #315: unique identifier must be included in VEVENT, VTODO, and VJOURNAL

niccokunzmann:
This fixes #315.

The work has been done before. This closes and tests this explicitly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,7 @@ New features:
 - Add ``stamp``, ``last_modified``, ``created``, ``CREATED``, ``busy_type``, ``class``, ``comments``, ``contacts``, ``location``, ``organizer``, ``priority``, and ``url`` properties to components that use them.
 - Add ``availabilities`` attribtue to ``Calendar``.
 - Add ``status``, ``transparency``, and ``attendees`` properties. See `Issue 841 <https://github.com/collective/icalendar/issues/841>`_.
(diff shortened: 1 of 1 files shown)


joooeey:
I [was asked](https://github.com/collective/icalendar/issues/315#issuecomment-3052290812) to provide feedback on this PR. Let me say that I have not worked with this library in almost five years. The test in this PR seems relatively good. It checks that the UID field is included and that it is not empty. There is no check that the uuid is unique. Writing a uniqueness test is hard; it would require a known case from real code where duplicate IDs show up.

The problem back in 2020 was that the UID was not included at all (if I interpreted my notes correctly).

niccokunzmann:
Thanks for your feedback...
Merging this for progress.

PR #877, opened 2025-07-10
Add tests for more properties

### LINKED ISSUE #655: RFC 7986 - additional parameters and values

niccokunzmann:
This PR fixes #655, making icalendar compatible to RFC 7986.

todo:

- all properties
  - [x] LAST-MODIFIED Property
  - [x] [5.5](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.5). URL Property
  - [x] [5.7](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.7). REFRESH-INTERVAL Property
  - [x] [5.8](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.8). SOURCE Property
  - [x] [5.10](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.10). IMAGE Property
  - [x] [5.11](https://www.rfc-editor.org/rfc/rfc7986.html#section-5.11). CONFERENCE Property
- [x] proper value types
- [x] add label of compatibility to docs

https://icalendar--877.org.readthedocs.build/en/877/

### DIFF
--- a/src/icalendar/cal/calendar.py
+++ b/src/icalendar/cal/calendar.py
@@ -493,7 +493,8 @@ Description:
         """Set the REFRESH-INTERVAL."""
         if not isinstance(value, timedelta) and value is not None:
             raise TypeError(
-                "REFRESH-INTERVAL must be either a positive timedelta, or None to delete it."
+                "REFRESH-INTERVAL must be either a positive timedelta,"
+                " or None to delete it."
             )
         if value is not None and value.total_seconds() <= 0:
             raise ValueError("REFRESH-INTERVAL must be a positive timedelta.")
@@ -546,6 +547,7 @@ Description:
             organization: The organization name. Used to generate `prodid` if not provided.
             prodid: The :attr:`prodid` of the component. If None and organization is provided,
                 generates a `prodid` in format "-//organization//name//language".
+            refresh_interval: The :attr:`refresh_interval` of the calendar.
             source: The :attr:`source` of the calendar.
             uid: The :attr:`uid` of the calendar.
                 If None, this is set to a new :func:`uuid.uuid4`.


src/icalendar/cal/todo.py
```diff
@@ -314,6 +320,7 @@ def new(
             classification: The :attr:`classification` of the todo.
             color: The :attr:`color` of the todo.
             comments: The :attr:`Component.comments` of the todo.
+            conferences: The :attr:`conferences` if the todo.
```
SashankBhamidi: Just found one tiny typo:

```suggestion
            conferences: The :attr:`conferences` of the todo.
```

src/icalendar/prop/conference.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @classmethod
+    def from_uri(cls, uri: vUri):
+        """Create a Conference from a URI."""
+        return cls(
+            uri,
+            feature=uri.params.get("feature"),
+            label=uri.params.get("label"),
+            language=uri.params.get("language"),
+        )
```
SashankBhamidi: This method crashes when `uri` is a string instead of `vUri` object.

Reproduction:
```python
event = Event()
event.add('CONFERENCE', 'tel:+1-555-1234')  # Raw string
conferences = event.conferences  # ❌ AttributeError: 'str' object has no attribute 'params'
```

  Fix: Add type checking:
```python
@classmethod
def from_uri(cls, uri):
    if isinstance(uri, str):
        uri = vUri(uri)
    return cls(...)```

File: src/icalendar/prop/image.py [hunk 3]
```diff
@@ -0,0 +1,90 @@
...
+        altrep: str | None = None,
+        display: str | None = None,
+    ):
+        """Create a new image according to :rfc:`7986`."""
+        self.uri = uri
+        self.b64data = b64data
+        self.fmttype = fmttype
+        self.altrep = altrep
+        self.display = display
+
```
SashankBhamidi, line 85: Image allows both URI and binary data simultaneously.

Reproduction:
```python
img = Image(uri='http://example.com/img.jpg', b64data='base64data')  # Should be invalid
```

Fix: Add validation after line 75:
```python
if uri is not None and b64data is not None:
    raise ValueError("Image cannot have both URI and binary data (RFC 7986)")        
if uri is None and b64data is None:
    raise ValueError("Image must have either URI or binary data")
```

src/icalendar/cal/calendar.py
```diff
@@ -458,6 +461,42 @@ def add_missing_timezones(
...
+        """
+        refresh_interval = self.get("REFRESH-INTERVAL")
+        return refresh_interval.dt if refresh_interval else None
+
+    @refresh_interval.setter
+    def refresh_interval(self, value: timedelta | None):
+        """Set the REFRESH-INTERVAL."""
+        del self.refresh_interval
+        if value is not None:
+            self.add("REFRESH-INTERVAL", value)
```
SashankBhamidi, line 503: Accepts negative refresh intervals (RFC requires positive).

Reproduction:
```python
cal.refresh_interval = timedelta(hours=-1)  # Should be rejected
```

Fix: Add validation after line 490:
```python
if value is not None and value.total_seconds() <= 0:
    raise ValueError("Refresh interval must be positive (RFC 7986)")
```

src/icalendar/attr.py
```diff
@@ -1763,19 +1810,159 @@ def set_end_with_locking(
...
+        iCalendar object or in "VEVENT", "VTODO", or "VJOURNAL" calendar
+        components.
+    """
+    images = self.get("IMAGE", [])
+    if not isinstance(images, SEQUENCE_TYPES):
+        images = [images]
+    return [Image.from_property_value(img) for img in images]
+
+
+images_property = property(_get_images)
```
SashankBhamidi, line 1866: `images` property is read-only while `conferences` has getter/setter/deleter.

Users can't set images like they can set conferences

```python
event.conferences = [Conference(...)]  # ✅ Works
event.images = [Image(...)]           # ❌ AttributeError: no setter
```

Fix: Either add setter/deleter like conferences, or document as read-only.

Or document the read-only behavior if intentional.

src/icalendar/tests/test_conference.py
```diff
@@ -0,0 +1,144 @@
...
+    return request.param
+
+
+@pytest.fixture
+def component(component_class):
+    """Create a component."""
+    return component_class()
+
+
+def test_no_conferenes(component):
```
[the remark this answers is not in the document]
SashankBhamidi: Valid and should be implemented.

src/icalendar/prop/__init__.py
```diff
@@ -77,10 +76,13 @@ class vBinary:
     """Binary property values are base 64 encoded."""
 
     params: Parameters
+    obj: str
 
-    def __init__(self, obj):
+    def __init__(self, obj, params: dict[str:str] | None = None):
```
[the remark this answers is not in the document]
SashankBhamidi: Valid and should be implemented.

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -212,3 +211,56 @@ def test_set_if_multiple_colors(
     color_component.add("COLOR", "green")
     color_component.color = color
     assert color_component.color == color
+
+
+def test_refresh_interval_default(calendar: Calendar):
+    """REFRESH-INTERVAL default."""
+    assert calendar.get("REFESH-INTERVAL") is None
```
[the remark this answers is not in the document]
SashankBhamidi: Valid and should be implemented.

src/icalendar/prop/conference.py
niccokunzmann: 51bcca7

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -212,3 +211,56 @@ def test_set_if_multiple_colors(
...
+    calendar.refresh_interval = timedelta(hours=1)
+    assert calendar.refresh_interval == timedelta(hours=1)
+    assert calendar["REFRESH-INTERVAL"].td == timedelta(hours=1)
+
+
+def test_refresh_interval_set_to_value_by_dict(calendar: Calendar):
+    """REFRESH-INTERVAL setting."""
+    calendar.add("REFRESH-INTERVAL", timedelta(days=1, hours=2))
+    assert calendar.refresh_interval == timedelta(days=1, hours=2)
+    assert calendar["REFRESH-INTERVAL"].dt == timedelta(days=1, hours=2)
```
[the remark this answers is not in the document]
niccokunzmann: no, this is a test for the compatibility.

niccokunzmann:
@SashankBhamidi Thanks for your review! I addressed the issues, I believe. Let me know if there is anything else to do and feel free to merge it if you are satisfied.

src/icalendar/cal/calendar.py
```diff
@@ -458,6 +462,51 @@ def add_missing_timezones(
...
+        return refresh_interval.dt if refresh_interval else None
+
+    @refresh_interval.setter
+    def refresh_interval(self, value: timedelta | None):
+        """Set the REFRESH-INTERVAL."""
+        if not isinstance(value, timedelta) and value is not None:
+            raise TypeError(
+                "REFRESH-INTERVAL must be a positive timedelta or None (to delete it)."
+            )
+        if value is not None and value.total_seconds() < 0:
```
SashankBhamidi:
```suggestion
        if value is not None and value.total_seconds() <= 0:
```

RFC 7986 Section 5.7 requires positive duration. Zero isn't positive, so change the comparison to reject zero values for proper compliance.

src/icalendar/cal/calendar.py
```diff
@@ -458,6 +462,51 @@ def add_missing_timezones(
...
+            raise ValueError("REFRESH-INTERVAL must be a positive timedelta.")
+        del self.refresh_interval
+        if value is not None:
+            self.add("REFRESH-INTERVAL", value)
```
SashankBhamidi:
```suggestion
        if value is not None:
            del self.refresh_interval
            self.add("REFRESH-INTERVAL", value)
        else:
            del self.refresh_interval
```

Validate first, then delete. Current logic removes existing value before validation completes, which breaks the principle of atomic operations.

src/icalendar/attr.py
```diff
@@ -1763,13 +1829,157 @@ def set_end_with_locking(
...
+            ... ]
+            >>> print(event.to_ical())
+            BEGIN:VEVENT
+            CONFERENCE;FEATURE="PHONE,MODERATOR";LABEL=Moderator dial-in;LANGUAGE=EN:t
+             el:+1-412-555-0123,,,654321
+            END:VEVENT
+
+    """
+    conferences = self.get("CONFERENCE", [])
+    if not isinstance(conferences, Sequence):
```
SashankBhamidi:
```suggestion
    if not isinstance(conferences, SEQUENCE_TYPES):
        conferences = [conferences]
```

Use SEQUENCE_TYPES for consistency with images property at line 1861. Both properties should follow the same pattern for maintainable code.

src/icalendar/tests/test_rfc_7986.py
```diff
@@ -212,3 +211,84 @@ def test_set_if_multiple_colors(
...
+def test_invalid_refresh_interval(calendar: Calendar):
+    with pytest.raises(ValueError):
+        calendar.refresh_interval = timedelta(seconds=-1)
+
+
+def test_0_refresh_interval(calendar: Calendar):
+    """REFRESH-INTERVAL zero."""
+    calendar.refresh_interval = timedelta(0)
+    assert calendar.refresh_interval == timedelta(0)
+    assert calendar["REFRESH-INTERVAL"].dt == timedelta(0)
```
SashankBhamidi:
```suggestion
def test_0_refresh_interval(calendar: Calendar):
    """REFRESH-INTERVAL zero should be invalid per RFC 7986."""
    with pytest.raises(ValueError, match="positive timedelta"):
        calendar.refresh_interval = timedelta(0)
```

Update test to expect ValueError for zero timedelta, matching the RFC requirement for positive duration after implementing the validation fix.

src/icalendar/cal/calendar.py
```diff
@@ -458,6 +462,51 @@ def add_missing_timezones(
...
+        """
+        refresh_interval = self.get("REFRESH-INTERVAL")
+        return refresh_interval.dt if refresh_interval else None
+
+    @refresh_interval.setter
+    def refresh_interval(self, value: timedelta | None):
+        """Set the REFRESH-INTERVAL."""
+        if not isinstance(value, timedelta) and value is not None:
+            raise TypeError(
+                "REFRESH-INTERVAL must be a positive timedelta or None (to delete it)."
```
```suggestion
                "REFRESH-INTERVAL must be either a positive timedelta, or None to delete it."
```

src/icalendar/prop/__init__.py
```diff
@@ -197,7 +199,7 @@ def from_ical(cls, ical: ICAL_TYPE):
 
 
 class vCalAddress(str):
-    """Calendar User Address
+    r"""Calendar User Address
```
stevepiercy: This looks like a stray keystroke snuck in.

```suggestion
    """Calendar User Address
```

src/icalendar/prop/__init__.py
```diff
@@ -717,6 +719,17 @@ def from_ical(cls, ical, timezone=None):
             return vTime.from_ical(ical)
         raise ValueError(f"Expected datetime, date, or time. Got: '{ical}'")
 
+    @property
+    def td(self) -> timedelta:
+        """Compatibility property returning self.dt.
+
+        This class is used to replace different time components.
+        Some of them contain a datetime or date (.dt).
+        Some of them contain a timedelta (.td).
```
stevepiercy: Use inline literals.

```suggestion
        """Compatibility property returning ``self.dt``.

        This class is used to replace different time components.
        Some of them contain a datetime or date (``.dt``).
        Some of them contain a timedelta (``.td``).
```

src/icalendar/prop/conference.py
```diff
@@ -0,0 +1,98 @@
...
+        which.
+
+        The "LANGUAGE" property parameter can be used to specify the
+        language used for text values used with this property (as per
+        Section 3.2.10 of :rfc:`5545`).
+
+    Example:
+        The following are examples of this property:
+
+        .. code-block:: ical
```
stevepiercy: There's no ical lexer in Pygments

```suggestion
        .. code-block:: text
```

src/icalendar/attr.py
```diff
@@ -1312,6 +1321,44 @@ def fset(self: Component, value: str | StrEnum | None) -> None:
...
+    .. code-block:: text
+
+        URL:http://example.com/pub/calendars/jsmith/mytime.ics
+
+""",
+)
+
+source_property = single_string_property(
+    "SOURCE",
+    """A URI where calendar data can be refreshed from.
```
```suggestion
    """A URI from where calendar data can be refreshed.
```

src/icalendar/attr.py
```diff
@@ -1763,13 +1829,157 @@ def set_end_with_locking(
         raise ValueError(msg)
 
 
+def _get_images(self: Component) -> list[Image]:
```
stevepiercy, line 1832: Do you want to display this private method? It currently doesn't display in the docs.

src/icalendar/attr.py
```diff
@@ -1763,13 +1829,157 @@ def set_end_with_locking(
...
+
+    Conformance:
+        This property can be specified multiple times in an
+        iCalendar object or in "VEVENT", "VTODO", or "VJOURNAL" calendar
+        components.
+
+    .. note::
+
+        At the present moment, this property is read-only. If you require a setter,
+        please open an issue or a pull request.
```
stevepiercy, line 1858: Based on this note, I think you expect this method and its docstring to be published. However when I alter the docs locally, it still doesn't display. Does it need to be added to `__all__` or something else?

```rst
.. automodule:: icalendar.attr
   :members:
   :private-members:
```

Perhaps these will help?

- https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-private-members
- https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoproperty

Alternatively you could follow the pattern for `icalendar.attr.property_del_duration`, `icalendar.attr.property_get_duration`, and `icalendar.attr.property_set_duration`.

src/icalendar/attr.py
```diff
@@ -1763,13 +1829,157 @@ def set_end_with_locking(
...
+    images = self.get("IMAGE", [])
+    if not isinstance(images, SEQUENCE_TYPES):
+        images = [images]
+    return [Image.from_property_value(img) for img in images]
+
+
+images_property = property(_get_images)
+
+
+def _get_conferences(self: Component) -> list[Conference]:
```
stevepiercy, line 1869: Also private member, not displayed in docs. Do you want it to display?

stevepiercy (changes requested):
@niccokunzmann a couple of suggestions and questions.

I like @SashankBhamidi suggestions, too.

src/icalendar/cal/calendar.py
niccokunzmann: I will test this now..

src/icalendar/tests/test_rfc_7986.py
niccokunzmann: deleted the test, this is now covered by test_invalid_refresh_interval

src/icalendar/prop/conference.py
niccokunzmann: Ok! Hm... AI hallucination :) DONE.

SashankBhamidi: > Ok! Hm... AI hallucination :)

What do you use?

src/icalendar/attr.py
niccokunzmann: Yes, because it is used for the property only. This way, we are free to refactor our API in the future without breaking compatibility.

src/icalendar/attr.py
niccokunzmann: See https://github.com/collective/icalendar/pull/877#discussion_r2387799765

src/icalendar/attr.py
niccokunzmann: It is here, in the property: https://icalendar--877.org.readthedocs.build/en/877/api.html#icalendar.cal.Calendar.images

Also see: https://github.com/collective/icalendar/pull/877#discussion_r2387799765

src/icalendar/prop/conference.py
niccokunzmann: Windsurf AI and then probably Visual Studio Code does its thing, too.

src/icalendar/attr.py
stevepiercy: @niccokunzmann sorry, I'm not clear on whether you're saying "yes" to my first or second sentence. Display it or not?

As a side note, in my work on the API docs, I currently use [`-P` to display private members](https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html#cmdoption-sphinx-apidoc-P). In general, what would be your preferred default, show or hide private members in the API docs?

I can also override the default for specific modules, classes, and exceptions with [`:private-members:`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-private-members).

src/icalendar/attr.py
stevepiercy: See https://github.com/collective/icalendar/pull/877#discussion_r2389258252

src/icalendar/attr.py
stevepiercy: Ah, OK, I expected to see its private name of `_get_images`. I didn't understand where it should appear. Noted. Thank you!

src/icalendar/attr.py
niccokunzmann: I would say: Hide private members. I use private functions for better modularity & code re-use and we have a nice public interface. If we expose private functions, we have less flexibility because people might depend on them. But I would like to refactor without thinking about that.

Please do not display this method. It is internal and might be renamed. Nobody should use it outside of icalendar.

stevepiercy: Removed display of private methods.

src/icalendar/cal/calendar.py
```diff
@@ -468,31 +519,38 @@ def new(
...
-            method: The :attr:`method` of the component.
-            name: The :attr:`calendar_name` of the component.
+            last_modified: The :attr:`Component.last_modified` of the calendar.
+            method: The :attr:`method` of the calendar.
+            name: The :attr:`calendar_name` of the calendar.
             organization: The organization name. Used to generate `prodid` if not provided.
             prodid: The :attr:`prodid` of the component. If None and organization is provided,
                 generates a `prodid` in format "-//organization//name//language".
-            uid: The :attr:`uid` of the component.
+            source: The :attr:`source` of the calendar.
```
SashankBhamidi, line 551:
```suggestion
            prodid: The :attr:`prodid` of the component. If None and organization is provided,
                generates a `prodid` in format "-//organization//name//language".
            refresh_interval: The :attr:`refresh_interval` of the calendar.
            source: The :attr:`source` of the calendar.
```

Missing documentation for `refresh_interval` parameter in Calendar.new() method. Parameter exists in signature (line 527) but missing from docstring.

SashankBhamidi:
Great work! Feel free to merge after the doc fix.

src/icalendar/attr.py
```diff
@@ -1763,13 +1829,158 @@ def set_end_with_locking(
...
+            CONFERENCE;FEATURE="PHONE,MODERATOR";LABEL=Moderator dial-in;LANGUAGE=EN:t
+             el:+1-412-555-0123,,,654321
+            END:VEVENT
+
+    """
+    conferences = self.get("CONFERENCE", [])
+    if not isinstance(conferences, SEQUENCE_TYPES):
+        conferences = [conferences]
+        conferences = [conferences]
+    return [Conference.from_uri(conference) for conference in conferences]
```
SashankBhamidi, line 1958:
```suggestion
    if not isinstance(conferences, SEQUENCE_TYPES):
        conferences = [conferences]
    return [Conference.from_uri(conference) for conference in conferences]
```

Remove duplicate line that creates nested list `[[conference]]` instead of `[conference]`. This breaks Conference.from_uri() when it tries to access `.params` on a list object instead of a vUri object.

src/icalendar/attr.py
```diff
@@ -1763,13 +1829,158 @@ def set_end_with_locking(
...
+
+    .. note::
+
+        At the present moment, this property is read-only. If you require a setter,
+        please open an issue or a pull request.
+    """
+    images = self.get("IMAGE", [])
+    if not isinstance(images, SEQUENCE_TYPES):
+        images = [images]
+    return [Image.from_property_value(img) for img in images]
```
SashankBhamidi, line 1863:
```python
    # Cache could be added, but this is complex due to mutability
    # For now, document the behavior or consider lazy evaluation
    return [Image.from_property_value(img) for img in images]
```

The current implementation recreates Image objects on each property access, which is inefficient for repeated access. Caching could be added but is complex due to mutability concerns.

Consider documenting this behavior or implementing lazy evaluation if performance becomes an issue.

src/icalendar/cal/calendar.py
```diff
@@ -521,6 +579,10 @@ def new(
         calendar.calscale = calscale
         calendar.categories = categories
         calendar.uid = uid
+        calendar.url = url
+        calendar.refresh_interval = refresh_interval
+        calendar.source = source
```
SashankBhamidi, line 586:
```suggestion
    if refresh_interval is not None:
        calendar.refresh_interval = refresh_interval
    if source is not None:
        calendar.source = source
```

Minor optimization to skip setter validation when values are None. Current approach is safer but slightly less efficient during object construction.

src/icalendar/prop/conference.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def to_uri(self) -> vUri:
+        """Convert the Conference to a vUri."""
+        params = {}
+        if self.feature:
+            params["FEATURE"] = self.feature
+        if self.label:
+            params["LABEL"] = self.label
+        if self.language:
+            params["LANGUAGE"] = self.language
```
SashankBhamidi, line 94:
```suggestion
    if self.feature:
        params["FEATURE"] = self.feature if isinstance(self.feature, str) else ",".join(self.feature)
    if self.label:
        params["LABEL"] = self.label if isinstance(self.label, str) else ",".join(self.label)
    if self.language:
        params["LANGUAGE"] = self.language if isinstance(self.language, str) else ",".join(self.language)
```

Normalize list values to comma-separated strings for consistent iCalendar serialization.

Current implementation may pass lists directly to vUri params, causing inconsistent output.

src/icalendar/prop/image.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    @classmethod
+    def from_property_value(cls, value: vUri | vBinary | vText):
+        """Create an Image from a property value."""
+        params: dict[str, str] = {}
+        if not hasattr(value, "params"):
+            raise TypeError("Value must be URI or BINARY.")
+        value_type = value.params.get("VALUE", "").upper()
```
SashankBhamidi, line 47:
```suggestion
        if not hasattr(value, "params"):
            raise TypeError("Value must be URI or BINARY.")
        try:
            value_type = value.params.get("VALUE", "").upper()
        except (AttributeError, TypeError):
            raise TypeError("Value must have valid params attribute.")
```

Add try-catch around params.get() to handle cases where params exists but isn't dict-like. Provides better error handling for malformed property values.

src/icalendar/attr.py
SashankBhamidi: Creates nested lists, causing Conference.from_uri() to fail.

SashankBhamidi: Completely breaks conferences property getter

src/icalendar/attr.py
niccokunzmann: The image class has properties which are lazily evaluated and only copying takes place. That is alright.

I would like to merge this soon. It is more a perfectionist strive than of any value. That RFC might be rarely used in the next years. It is ok to do it like this for now and not spend too much time. When value comes, change comes and contributors will come to do it.

This is more about showing that we support the RFC and so people can actually use it and come back with feedback.

src/icalendar/cal/calendar.py
niccokunzmann: I do not want to apply this change it because I want setting to None to delete it and also use all parameters in the same way inside of new and not make exceptions for some to be treated differently. That is why None is considered in the setters at all, because of the new() method. So, `if ?? is not None` should not not be used in `new()`. The ease of reading it matters, too.

src/icalendar/prop/conference.py
niccokunzmann: This goes into a new issue because it needs to be tested. Could you open this after merge?

src/icalendar/prop/image.py
niccokunzmann: This requires a test, please open an issue after merge.

PR #883, opened 2025-07-29
ci: add CodeQL, Safety CVE scan, and Bandit rules to CI

raajheshkannaa:
### Why
* **Static analysis missing** – the project doesn’t yet run CodeQL.
* **Supply-chain CVE checks missing** – Safety scan adds an automated alert if any dependency has a reported CVE.
* **Security lint** – Ruff is already in use, but Bandit rules (`--select B`) were not enabled.

### What’s in this PR
| Added / Changed | Purpose |
|-----------------|---------|
| `.github/workflows/codeql.yml` | Runs GitHub CodeQL (Python) on push / PR. |
| Updated `.github/workflows/ruff.yml` | *Adds two steps*<br>• **Bandit rules** via `ruff --select B`<br>• **Safety** CVE scan job (uploads report artifact) |
| _(no other files touched)_ | No functional code changes. |

### Impact
* Immediate SAST coverage (CodeQL) and CVE reporting (Safety) for every PR.
* Zero runtime impact; CI-only additions.
* Bandit rules catch insecure code patterns early.

Happy to maintain these security workflows going forward—feedback welcome!

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,31 @@
+name: codeql-analysis
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  analyze:
+    name: CodeQL Scan
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


stevepiercy:
I don't see the point in adding these duplicative features, and it increases CI run times, consuming more electricity, without need. I could be wrong, and if so, there needs to be a good explanation for why this is needed, especially since it is a product from the contributor's employer. It's irresponsible not to disclose that up front, too.

PR #894, opened 2025-08-31
Workaround broken ICS files with empty RDATE

rudis:
Hi,

from time to time I stumble over these broken ICS files with an empty RDATE which are currently rejected by icalendar (manually stripping these lines yield a successful parse). This is invalid according to https://www.rfc-editor.org/rfc/rfc5545#section-3.8.5.2 but it would be nice if such files could be successfully parsed anyway.

Best,
Simon

----
📚 Documentation preview 📚: https://icalendar--894.org.readthedocs.build/


niccokunzmann (approved):
This code change is tested and solves the issue.

Issue #762, opened 2024-12-31
Documentation: Add link to types-icalendar


niccokunzmann:
This project adds the typing information for icalendar:
https://pypi.org/project/types-icalendar/

I think, we should document this in the installation section or related work if people are interested.

- [ ] link to project
- [ ] describe its purpose

stevepiercy:
This option of supporting types has the most overhead. How do you test the types? How do you keep them in sync with the `icalendar` repo? How do you synch releases between the the two repos?

Did you consider shipping the stub files with the `icalendar` distribution?

Refs:

- https://typing.readthedocs.io/en/latest/index.html
- https://peps.python.org/pep-0526/ (history)

niccokunzmann:
Your idea is nice. However it takes work, so the proposed change helps for now and if someone wants to take on your idea, it can be implemented! :heart:

stevepiercy:
Oops, I omitted an important clause. I meant to say "This option of supporting types _in a separate repo_ has the most overhead.

Anyway, there is a good discussion about the various options in the Python Typing docs. If you choose to use stub files, I don't see how keeping them in the same repo takes more work than in a separate repo, but sounds like more work to me.

- [How to provide type annotations?](https://typing.readthedocs.io/en/latest/guides/libraries.html#providing-type-annotations)
- In [Distributing type information](https://typing.readthedocs.io/en/latest/spec/distributing.html):
  - [Stub files](https://typing.readthedocs.io/en/latest/spec/distributing.html#stub-files)
  - [Type information in libraries](https://typing.readthedocs.io/en/latest/spec/distributing.html#type-information-in-libraries)

PredictiveManish:
@niccokunzmann Does this issue required to be worked?

niccokunzmann:
@PredictiveManish Thanks for asking, yes. @stevepiercy and I have proposed different ways on how to go about typing. Please choose one of them that you see fit and help people to use icalendar with types by changing the documentation.

stevepiercy:
@niccokunzmann I think this issue can be closed. Nothing needs to be documented. A new issue requesting inline typing could be created, if you think that would be useful, maybe for first-timers?

PR #911, opened 2025-10-10
Add Python typing support infrastructure

### LINKED ISSUE #395: Support Python typing
I get:

```
Skipping analyzing "icalendar": module is installed, but missing library stubs or py.typed marker - mypy
```

SashankBhamidi:
## Summary

Adds PEP 561 type checking infrastructure to enable mypy and other type checkers to recognize inline type annotations in the icalendar package.

## Changes

- Add `py.typed` marker file per PEP 561
- Configure `pyproject.toml` to distribute the marker in wheel packages  
- Add `mypy` and type stub dependencies to test extras
- Update changelog

## Implementation Notes

This PR establishes the foundation for type checking without attempting to fix all existing type errors. The codebase already has inline type annotations in 31+ files, and this change makes them usable by external type checkers.

Type annotations will continue to be added incrementally as new code is written, following the maintainer's gradual approach.

## References

- Fixes #395
- Related to #762 (external types-icalendar documentation)
- Implements PEP 561: https://peps.python.org/pep-0561/

----
📚 Documentation preview 📚: https://icalendar--911.org.readthedocs.build/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -28,7 +28,7 @@ jobs:
         path: ./out/artifacts
     - name: Upload Sarif
       if: always() && steps.build.outcome == 'success'
-      uses: github/codeql-action/upload-sarif@v3
+      uses: github/codeql-action/upload-sarif@v4
       with:
         # Path to SARIF file relative to the root of the repository
         sarif_file: cifuzz-sarif/results.sarif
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -86,6 +86,9 @@ test = [
     "coverage",
     "hypothesis",
     "pytz",
+    "mypy>=1.0",
```
niccokunzmann, line 89: That is within `test =`, however, I do not see it used.

SashankBhamidi: @niccokunzmann I added it thinking developers would use it locally for type checking. Remove it or add a CI step to run mypy? (Though running mypy now would show ~400 errors since types are being added incrementally)

niccokunzmann: I would say leave it but run mypy and see how many errors there are - can you create a `good first issue` to fix one or more type errors?

PR #331, opened 2021-10-13
Strong types, not ignoring the VALUE parameter, better error handling

thet:
Continuation of https://github.com/collective/icalendar/pull/196
See old PR for comments.

PR #915, opened 2025-10-16
Fix VALUE parameter handling for date/time properties

### LINKED ISSUE #349: Adding an EXDATE with type date does not set the element's VALUE as DATE
If you set an EXDATE with an argument of type datetime.date, iCalendar does not set the VALUE of the element as DATE.

This violates RFC5545 section 3.8.5.1, which states that:

> Value Type:  The default value type for this property is DATE-TIME.
>       The value type can be set to DATE.

and (in the format definition)

> ;Value MUST match value type

Minimal code to reproduce:
```
import icalendar
import datetime

# Create an event with necessary elements
ev = icalendar.Event()
ev.add('UID', 'ical-test-20220129')
ev.add('DTSTAMP', datetime.datetime.now(datetime.timezone.utc))
ev.add('SUMMARY', 'EXDATE test')

# Add a date as start date
date = datetime.date(year=2022, month=1, day=1)
ev.add('DTSTART', date)

# Add weekly repeat rule
ev.add('RRULE', {'FREQ':'WEEKLY'})

# Add a date as exception date
exdate = datetime.date(year=2022, month=1, day=8)
ev.add('EXDATE', exdate)

print(ev.to_ical().decode())
```

With Python 3.9.10, icalendar 4.0.9, this gives output:
```
BEGIN:VEVENT
SUMMARY:EXDATE test
DTSTART;VALUE=DATE:20220101
DTSTAMP;VALUE=DATE-TIME:20220129T154609Z
UID:ical-test-20220129
RRULE:FREQ=WEEKLY
EXDATE:20220108
END:VEVENT
```
The error is that the EXDATE line should read:
```
EXDATE;VALUE=DATE:20220108
```
Note that in the above code, setting the DTSTART and DTSTAMP elements correctly sets their VALUE parameter from the given argument.

Workaround: The add('EXDATE',...) call can be changed to explicitly set the VALUE:
```
ev.add('EXDATE', exdate, parameters={'VALUE':'DATE'})
```

SashankBhamidi:
## What this fixes

When you add a property with a `datetime.date` object, the library wasn't setting `VALUE=DATE` in the iCalendar output. This caused problems with other calendar applications that need this parameter to correctly interpret the date format.

## What changed

- Python `date` objects now automatically get `VALUE=DATE`
- Python `time` objects now automatically get `VALUE=TIME`  
- The VALUE parameter is properly respected when parsing iCalendar data
- RDATE/EXDATE always use vDDDLists to handle lists correctly

## How it works

Added a helper function `_infer_value_type()` that looks at the Python type and figures out what VALUE parameter should be set. The encoding path checks this and auto-sets the parameter. The decoding path extracts VALUE from the parameters and uses it to pick the right parser.

## Attribution

This builds on earlier work by @thet (PR #331) and @stlaz (PR #196), completing the fix they started.

Fixes #349
Related to #187

## Example

Before:
```python
ev.add('EXDATE', datetime.date(2022, 1, 8))
# Generated: EXDATE:20220108  (missing VALUE=DATE)
```

After:
```python
ev.add('EXDATE', datetime.date(2022, 1, 8))
# Generated: EXDATE;VALUE=DATE:20220108  (correct!)
```

----
📚 Documentation preview 📚: https://icalendar--915.org.readthedocs.build/

### DIFF
--- a/src/icalendar/cal/component.py
+++ b/src/icalendar/cal/component.py
@@ -13,40 +13,12 @@ from icalendar.parser import Contentline, Contentlines, Parameters, q_join, q_sp
 from icalendar.parser_tools import DEFAULT_ENCODING
 from icalendar.prop import TypesFactory, vDDDLists, vText
 from icalendar.timezone import tzp
+from icalendar.tools import is_date, is_datetime
 
 if TYPE_CHECKING:
     from icalendar.compatibility import Self
 
 _marker = []
-
-
-def _infer_value_type(value: date | datetime | timedelta | time | tuple | list) -> str | None:
-    """Infer the ``VALUE`` parameter from a Python type.
-
-    Args:
-        value: Python native type, one of :py:class:`date`, :py:mod:`datetime`, :py:class:`timedelta`, :py:mod:`time`, :py:class:`tuple`, or :py:class:`list`.
-
-    Returns:
-        str or None: The ``VALUE`` parameter string, for example, "DATE", "TIME", or other string, or ``None``
-            if no specific ``VALUE`` is needed.
-    """
-    if isinstance(value, list):
-        if not value:
-            return None
-        # Check if ALL items are date (but not datetime)
-        if all(isinstance(item, date) and not isinstance(item, datetime) for item in value):
-            return "DATE"
-        # Check if ALL items are time
-        if all(isinstance(item, time) for item in value):
-            return "TIME"
-        # Mixed types or other types - don't infer
-        return None
-    if isinstance(value, date) and not isinstance(value, datetime):
-        return "DATE"
-    if isinstance(value, time):
-        return "TIME"
-    # Don't infer PERIOD - it's too risky and vPeriod already handles it
-    return None
 
 
 class Component(CaselessDict):
@@ -97,6 +69,35 @@ class Component(CaselessDict):
         if cls._components_factory is None:
             cls._components_factory = ComponentFactory()
         return cls._components_factory.get(name, Component)
+
+    @staticmethod
+    def _infer_value_type(value: date | datetime | timedelta | time | tuple | list) -> str | None:
+        """Infer the ``VALUE`` parameter from a Python type.
+
+        Args:
+            value: Python native type, one of :py:class:`date`, :py:mod:`datetime`, :py:class:`timedelta`, :py:mod:`time`, :py:class:`tuple`, or :py:class:`list`.
+
+        Returns:
+            str or None: The ``VALUE`` parameter string, for example, "DATE", "TIME", or other string, or ``None``
+                if no specific ``VALUE`` is needed.
+        """
+        if isinstance(value, list):
+            if not value:
+                return None
+            # Check if ALL items are date (but not datetime)
+            if all(is_date(item) for item in value):
+                return "DATE"
+            # Check if ALL items are time
+            if all(isinstance(item, time) for item in value):
+                return "TIME"
+            # Mixed types or other types - don't infer
+            return None
+        if is_date(value):
+            return "DATE"
+        if isinstance(value, time):
+            return "TIME"
+        # Don't infer PERIOD - it's too risky and vPeriod already handles it
+        return None
 
     def __init__(self, *args, **kwargs):
         """Set keys to upper for initial dict."""
@@ -150,7 +151,7 @@ class Component(CaselessDict):
             if parameters and "VALUE" in parameters:
                 value_param = parameters["VALUE"]
             elif not isinstance(value, cls.types_factory.all_types):
-                inferred = _infer_value_type(value)
+                inferred = cls._infer_value_type(value)
                 if inferred:
                     value_param = inferred
                     # Auto-set the VALUE parameter


CHANGES.rst
```diff
@@ -63,6 +63,7 @@ Bug fixes:
 - Fix invalid calendar: Parsing a date with TZID results in a datetime to not loose the timezone. See `Issue 187 <https://github.com/collective/icalendar/issues/187>`_.
 - Fix timezone placement in ``add_missing_timezones()``: ``VTIMEZONE`` components now appear before ``VEVENT`` and other components that reference them. See `Issue 844 <https://github.com/collective/icalendar/issues/844>`_.
 - Fixed ``Todo.duration`` and ``Event.duration`` to return ``DURATION`` property when set, even without ``DTSTART``. See `Issue 867 <https://github.com/collective/icalendar/issues/867>`_.
+- Fix VALUE parameter handling: ``datetime.date`` objects now correctly set ``VALUE=DATE`` parameter when added to properties like ``EXDATE``, ``RDATE``, and ``DTSTART``. The VALUE parameter is also properly used when parsing iCalendar data. See `Issue 349 <https://github.com/collective/icalendar/issues/349>`_.
```
```suggestion
- Fix ``VALUE`` parameter handling: ``datetime.date`` objects now correctly set ``VALUE=DATE`` parameter when added to properties like ``EXDATE``, ``RDATE``, and ``DTSTART``. The ``VALUE`` parameter is also properly used when parsing iCalendar data. See `Issue 349 <https://github.com/collective/icalendar/issues/349>`_.
```

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
 _marker = []
 
 
+def _infer_value_type(value):
+    """Infer the VALUE parameter from a Python type.
```
```suggestion
    """Infer the ``VALUE`` parameter from a Python type.
```

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
...
+
+    Args:
+        value: Python native type (date, datetime, timedelta, time, tuple, list)
```
stevepiercy: The object might need to be tweaked to render as a link to the Python docs, but this is my guess.

```suggestion
        value: Python native type, one of :py:class:`date`, :py:mod:`datetime`, :py:class:`timedelta`, :py:mod:`time`, :py:class:`tuple`, or :py:class:`list`.
```

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
...
+
+    Returns:
+        str or None: The VALUE parameter string (e.g., "DATE", "TIME") or None
```
```suggestion
        str or None: The ``VALUE`` parameter string, for example, "DATE", "TIME", or other string, or ``None``
```

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
...
+            if no specific VALUE is needed
```
```suggestion
            if no specific ``VALUE`` is needed.
```

src/icalendar/prop/__init__.py
```diff
@@ -2189,8 +2189,28 @@ def __init__(self, *args, **kwargs):
...
     )
 
-    def for_property(self, name):
-        """Returns a the default type for a property or parameter"""
+    def for_property(self, name, value_param=None):
+        """Returns the type class for a property or parameter.
+
+        Args:
+            name: Property or parameter name
+            value_param: Optional VALUE parameter (e.g., "DATE", "DATE-TIME")
```
stevepiercy:
```suggestion
            value_param: Optional ``VALUE`` parameter, for example, "DATE", "DATE-TIME", or other string.
```

stevepiercy (changes requested):
I reviewed only docstrings, which look pretty good to me, with a few enhancements.

SashankBhamidi:
Roger, I always request your review for docs 

_I certainly don't learn from Steve's doc's suggestions (every PR I've worked on)_

stevepiercy:
> Roger, I always request your review for docs
> 
> _I certainly don't learn from Steve's doc's suggestions (every PR I've worked on)_

Well, don't take everything I do as perfect. I just realized that I reviewed a private method, which doesn't render in the docs. That's what I get for going on auto-pilot. :wink: It's all good anyway.

src/icalendar/prop/__init__.py
```diff
@@ -2189,8 +2189,28 @@ def __init__(self, *args, **kwargs):
...
+        # Special case: RDATE and EXDATE always use vDDDLists to support list values
+        # regardless of the VALUE parameter
+        if name.lower() in ("rdate", "exdate"):
+            return self["date-time-list"]
+
+        # Only use VALUE parameter for known properties that support multiple value types
+        # (like DTSTART, DTEND, etc. which can be DATE or DATE-TIME)
+        # For unknown/custom properties, always use the default type from types_map
+        if value_param and name.lower() in self.types_map:
+            value_param_lower = value_param.lower()
```
niccokunzmann: Any `.lower()` calls are not required because this is a CaselessDict (uppercase/lowercase does not matter)
Please remove them.

src/icalendar/prop/__init__.py
```diff
@@ -2189,8 +2189,28 @@ def __init__(self, *args, **kwargs):
         }
     )
 
-    def for_property(self, name):
-        """Returns a the default type for a property or parameter"""
+    def for_property(self, name, value_param=None):
```
niccokunzmann: Please add a type hint.

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
 _marker = []
 
 
+def _infer_value_type(value):
```
niccokunzmann: Please add a type hint here.
```suggestion
def _infer_value_type(value:date|datetime|timedelta|time|tuple|list) -> str|None:
```

niccokunzmann (changes requested):
Thank you for this PR! I really like the tests. They show that this can close the issue.

I added a few hints on what I would like changed because it is project specific stuff.

On another note: We use pytest which makes some tests smaller than unittest. If you like, have a look - no need to change that. Both test styles are welcome.

SashankBhamidi:
Hmm.

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
...
+
+    Returns:
+        str or None: The ``VALUE`` parameter string, for example, "DATE", "TIME", or other string, or ``None``
+            if no specific ``VALUE`` is needed.
+    """
+    if isinstance(value, list):
+        if not value:
+            return None
+        # Check if ALL items are date (but not datetime)
+        if all(isinstance(item, date) and not isinstance(item, datetime) for item in value):
```
```suggestion
        if all(is_date(item) for item in value):
```

adding

```
from icalendar.tools import is_date, is_datetime
```

src/icalendar/cal/component.py
```diff
@@ -20,6 +20,35 @@
...
+            return "DATE"
+        # Check if ALL items are time
+        if all(isinstance(item, time) for item in value):
+            return "TIME"
+        # Mixed types or other types - don't infer
+        return None
+    if isinstance(value, date) and not isinstance(value, datetime):
```
```suggestion
    if is_date(value, date):
```

src/icalendar/cal/component.py
```diff
@@ -116,7 +145,21 @@ def _encode(cls, name, value, parameters=None, encode=1):
             # Don't encode already encoded values.
             obj = value
         else:
-            klass = cls.types_factory.for_property(name)
+            # Extract VALUE parameter if present, or infer it from the Python type
+            value_param = None
+            if parameters and "VALUE" in parameters:
+                value_param = parameters["VALUE"]
+            elif not isinstance(value, cls.types_factory.all_types):
+                inferred = _infer_value_type(value)
```
niccokunzmann: If _infer_value_type is used only once and that is here, it should be located in the class. That would increase [cohesion](https://ducmanhphan.github.io/2019-03-23-Coupling-and-Cohension-in-OOP/).

SashankBhamidi:
> @SashankBhamidi You might want to register for https://hacktoberfest.com/ if you have not done so, before merge :)

No clue what the fest is about, but created an account. Merging.

PR #916, opened 2025-10-16
added ics_value property to vPeriod, vDuration,vDate,vText

### LINKED ISSUE #876: Properly decode every value type (polymorphy, beginner)
At the moment, each value type is different.
We have a method `Component.decoded()` which puts an incomplete if/else dispatch into place to get the decoded value.

It would be easier if all value types had a property such as `ics_value` which would result in the object used.
This would e.g. be

- `.td` for timedelta
- `.dt` for date
- `str(self)` for vText

- [ ] Please implement such a property on all value types in `icalendar.prop`
- [ ] make `Component.decode()` use this kind of value.

TejasR002:
## Closes issue

- [ ] Closes #876 

## Description

added ics_value property to vPeriod, vDuration,vDate,vText.

## Checklist

- [ ] I've run and ensured all tests pass locally by following [Running Tests](https://icalendar.readthedocs.io/en/latest/install.html#running-tests).
- [ ] I've added myself to `docs/credits.rst` as a contributor in this pull request or have done so previously.

## Additional information

Upload screenshots, videos, links to documentation, or any other relevant information.

----
📚 Documentation preview 📚: https://icalendar--916.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,8 @@ New features
 - Created a separate path to contribute to documentation. See `Issue 929 <https://github.com/collective/icalendar/issues/929>`_.
 - Added a code of conduct. See `Issue 907 <https://github.com/collective/icalendar/issues/907>`_.
 - Added redirects of moved files and 404 not found page to catch any other obsolete URLs. See `Issue 959 <https://github.com/collective/icalendar/issues/959>`_.
+- Added artificial intelligence (AI) policy. See `Issue 933 <https://github.com/collective/icalendar/issues/933>`_.
+
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -145,6 +145,7 @@ linkcheck_ignore = [
     r"https://groups.google.com/g/icalendar-coc/",
     r"https://pypi.org/manage/project/icalendar/collaboration/",
     # Ignore specific anchors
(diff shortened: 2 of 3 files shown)


SashankBhamidi:
@TejasR002 Tests are failing. Let me know when this is ready for a review. 

**Edit:** It looks like this branch is behind `main`. We recommend updating and testing against the latest version.

src/icalendar/prop/__init__.py
```diff
@@ -1153,8 +1167,14 @@ def __repr__(self):
     def dt(self):
         """Make this cooperate with the other vDDDTypes."""
         return (self.start, (self.duration if self.by_duration else self.end))
+    @property
+    def ics_value(self):
+        if self.by_duration:
+            return (self.start, self.duration)
```
niccokunzmann: This duplicates `.dt`. Please just return `self.dt`.

Also, they all need a type hint in the end. I updated the issue to be clearer on the scope.

src/icalendar/cal/component.py
niccokunzmann, line 1: Hi, I think for now, it would be ok to remove the changes in this file here. Then, the tests should pass. There are a lot of classes that will receive the `.ics_value` property. Adding some is a good first contribution. Embedding that behaviour into icalendar might be too hard for a first-time contribution. If you remove the changes in this file and add tests for the existing new `.ics_value` properties, that is a really great contribution!

PR #935, opened 2025-10-22
Fix #355: Preserve URL-encoded characters in content line parsing

### LINKED ISSUE #355: Unwanted unescaping of http url strings

SashankBhamidi:
The parser was corrupting URL-encoded characters like `%3A` (colon) in DESCRIPTION fields. Facebook event URLs with complex query parameters would get mangled during parsing.

The root cause: `escape_string()` and `unescape_string()` were being used as a workaround to hide backslash-escaped characters during parsing. This approach couldn't distinguish between RFC 5545 backslash escaping (like `\,`) and actual URL encoding (like `%3A`), so everything got incorrectly unescaped.

The fix rewrites `Contentline.parts()` to properly track:
- Double-quote delimited sections (case-sensitive)
- Backslash escaping outside quotes
- Delimiter detection only when not quoted or escaped

Added a new `unescape_backslash()` function that only handles RFC 5545 backslash sequences without touching URL encoding. This preserves `%XX` values while still correctly unescaping `\,` `\;` `\:` `\` and `\n`.

Tests verify both URL preservation and backslash escaping still work correctly.

Based on original work by @NicoHood in #356. Tests from @niccokunzmann.

Closes #355
Closes #356
Closes #426

----
📚 Documentation preview 📚: https://icalendar--935.org.readthedocs.build/


src/icalendar/parser.py
```diff
@@ -428,37 +449,64 @@ def from_parts(
...
+                escaped = False
+
+            # Validate parsing results
             if not value_split:
-                value_split = i + 1
+                # No colon found - value is empty, use end of string
+                value_split = len(self)
+
+            # Extract name - if no delimiter, take whole string for validate_token to reject
+            name = self[:name_split] if name_split else self
```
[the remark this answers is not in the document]
SashankBhamidi: Rejecting, this change would break backward compatibility with error handling.

CHANGES.rst
```diff
@@ -71,6 +71,7 @@ Bug fixes:
...
 - Fixed ``Todo.duration`` and ``Event.duration`` to return ``DURATION`` property when set, even without ``DTSTART``. See `Issue 867 <https://github.com/collective/icalendar/issues/867>`_.
 - Fix ``VALUE`` parameter handling: ``datetime.date`` objects now correctly set ``VALUE=DATE`` parameter when added to properties like ``EXDATE``, ``RDATE``, and ``DTSTART``. The ``VALUE`` parameter is also properly used when parsing iCalendar data. See `Issue 349 <https://github.com/collective/icalendar/issues/349>`_.
+- Fix URL-encoded characters being incorrectly unescaped during content line parsing. The parser now properly handles backslash escaping and double-quoted sections without corrupting URL-encoded values like ``%3A`` (colon) in DESCRIPTION fields. Added ``unescape_backslash()`` function to separate RFC 5545 backslash escaping from URL encoding. Added type hints to ``Contentline.parts()`` method. See `Issue 355 <https://github.com/collective/icalendar/issues/355>`_.
```
```suggestion
- Fix URL-encoded characters being incorrectly unescaped during content line parsing. The parser now properly handles backslash escaping and double-quoted sections without corrupting URL-encoded values like ``%3A`` (colon) in DESCRIPTION fields. Added ``unescape_backslash()`` function to separate :rfc:`5545` backslash escaping from URL encoding. Added type hints to ``Contentline.parts()`` method. See `Issue 355 <https://github.com/collective/icalendar/issues/355>`_.
```

src/icalendar/parser.py
```diff
@@ -428,37 +450,64 @@ def from_parts(
             return cls(f"{name};{params}:{values}")
         return cls(f"{name}:{values}")
 
-    def parts(self):
-        """Split the content line up into (name, parameters, values) parts."""
+    def parts(self) -> tuple[str, Parameters, str]:
+        """Split the content line up into (name, parameters, values) parts.
```
stevepiercy:
```suggestion
        """Split the content line into ``name``, ``parameters``, and ``values`` parts.
```

src/icalendar/parser.py
```diff
@@ -428,37 +450,64 @@ def from_parts(
...
+
+        Properly handles escaping with backslashes and double-quote sections
+        to avoid corrupting URL-encoded characters in values.
+
+        Example with parameter::
+
+            DESCRIPTION;ALTREP="cid:part1.0001@example.org":The Fall'98 Wild
```
stevepiercy: Use explicit syntax highlighter.

```suggestion
        Example with parameter:
        
        .. code-block:: text

            DESCRIPTION;ALTREP="cid:part1.0001@example.org":The Fall'98 Wild
```

src/icalendar/parser.py
```diff
@@ -428,37 +450,64 @@ def from_parts(
...
+
+        Example without parameters::
+
+            DESCRIPTION:The Fall'98 Wild
```
stevepiercy:
```suggestion
        Example without parameters:
        
        .. code-block:: text

            DESCRIPTION:The Fall'98 Wild
```

src/icalendar/parser.py
```diff
@@ -331,6 +330,29 @@ def unescape_string(val):
     )
 
 
+def unescape_backslash(val):
+    """Unescape backslash sequences in iCalendar text.
+
+    Unlike unescape_string(), this only handles actual backslash escapes
```
stevepiercy: Let's make sure this generates a correct link in the PR preview.

```suggestion
    Unlike :py:meth:`unescape_string`, this only handles actual backslash escapes
```

src/icalendar/parser.py
```diff
@@ -331,6 +330,29 @@ def unescape_string(val):
...
+    per RFC 5545, not URL encoding. This preserves URL-encoded values
```
stevepiercy:
```suggestion
    per :rfc:`5545`, not URL encoding. This preserves URL-encoded values
```

src/icalendar/parser.py
```diff
@@ -331,6 +330,29 @@ def unescape_string(val):
...
+    like %3A in URLs.
```
stevepiercy:
```suggestion
    like ``%3A`` in URLs.
```

src/icalendar/parser.py
```diff
@@ -331,6 +330,29 @@ def unescape_string(val):
...
+
+    Uses a placeholder for escaped backslashes (\\\\) to prevent them from
```
stevepiercy:
```suggestion
    Uses a placeholder for escaped backslashes (``\\\\``) to prevent them from
```

stevepiercy (changes requested):
Let's make sure that the reST markup renders correctly in the PR preview. Also the docstrings might need to be rewrapped.

stevepiercy:
@SashankBhamidi for some unknown reason `unescape_backslash` does not appear in the docs. Can you have a look? https://icalendar--935.org.readthedocs.build/en/935/reference/api/icalendar.parser.html?readthedocs-diff=true&readthedocs-diff-chunk=2. Does it appear when you build locally?

stevepiercy:
> Can default to squash merging in repo settings. This is what I personally have setup in my profile:

That allows it, but doesn't make it the default. I think you have to select it once on the repo, and then GitHub remembers it. Of course, I'm old and forgetful. 👴

stevepiercy (approved):
Docs look good now. https://icalendar--935.org.readthedocs.build/en/935/reference/api/icalendar.parser.html#icalendar.parser.unescape_backslash

Let's wait for @niccokunzmann's technical review.

src/icalendar/parser.py
```diff
@@ -331,6 +330,29 @@ def unescape_string(val):
...
+    Unlike :py:meth:`unescape_string`, this only handles actual backslash escapes
+    per :rfc:`5545`, not URL encoding. This preserves URL-encoded values
+    like ``%3A`` in URLs.
+
+    Uses a placeholder for escaped backslashes (``\\\\``) to prevent them from
+    being processed during other escape sequence replacements.
+    """
+    # Use a placeholder to avoid conflicts
+    backslash_placeholder = "\x00BACKSLASH\x00"
+    return (
```
niccokunzmann: A more efficient way here would be to use `re.replace`. The reason: Each replace performs a copy operation on the whole string. Also, it protects us from using a `backslash_placeholder`.

I will create a PR.

Also, the tests should make sure that all values are properly escaped.

I think, I would like to have tests on this method specifically. That will allow easier changes in the future.

src/icalendar/parser.py
```diff
@@ -428,37 +450,68 @@ def from_parts(
...
-            for i, ch in enumerate(st):
-                if not in_quotes:
-                    if ch in ":;" and not name_split:
+            name_split: int | None = None
+            value_split: int | None = None
+            in_quotes: bool = False
+            escaped: bool = False
+
+            for i, ch in enumerate(self):
+                if ch == '"' and not escaped:
```
niccokunzmann, line 472: At the moment, I do not have the head space for understanding this. But there are lots of tests, so it should be correct.

SashankBhamidi:
@niccokunzmann Makes sense on the regex approach, 6 string copies is inefficient. For tests: should I add unit tests for `unescape_backslash()` covering all escape sequences (`\,` `\;` `\:` `\\` `\n` `\N`) now, or will you handle that in your PR? The parsing state machine handles quoted sections and escaped characters correctly per the test cases, but direct tests would make refactoring safer.

niccokunzmann:
@SashankBhamidi have a look:
https://github.com/SashankBhamidi/icalendar/pull/3

You can merge it into this PR.

src/icalendar/parser.py
niccokunzmann: https://github.com/SashankBhamidi/icalendar/pull/3

SashankBhamidi:
> @SashankBhamidi have a look: [SashankBhamidi#3](https://github.com/SashankBhamidi/icalendar/pull/3)
> 
> You can merge it into this PR.

LGTM, two issues I'm going to nitpick however

Issue #938, opened 2025-10-23
Add type hints [beginner, Python, typing]


niccokunzmann:
icalendar exists since Python 2. Nowadays, type hints are a huge support for developers!

If you are a newcomer to icalendar, please add type hints to untyped functions. Here is your checklist:

- [ ] choose a function that does not have a return type or argument types in `src/icalendar` (the tests)
- [ ] find all places where it is used (easiest for functions that are not used much) 
- [ ] from its content and usage, add type hints
- [ ] create a pull request with your changes, mentioning this issue.

---

This is a rolling issue: When closed, copy and open again or re-open.

niccokunzmann:
Thanks for your contributions. @owl-tech-sui @lillianpenner @jkseppan They are included in release v7.0.0!

xxiaoxiong:
may i ask is this problem still here now?  can i work on it?

niccokunzmann:
@xxiaoxiong Yes, please. This is always there until we close it. Feel free to just start where you like.

cybs-joe:
Hi! I'd like to work on this as my first open source contribution , same checklist right ? and is it still available?

stevepiercy:
@cybs-joe yes, and yes, and please follow https://icalendar.readthedocs.io/en/stable/contribute/index.html. Thank you!

Priyanshu-pulak:
Hi! I'd like to contribute by updating iCalender to use Python type hinting features. Is this still open for contributions & accepting PRs?

stevepiercy:
@Priyanshu-pulak yes, and please follow https://icalendar.readthedocs.io/en/stable/contribute/index.html. Thank you!

goAuD:
Hi, I would like to add the missing return type annotation to unescape_backslash in scr/icalendar/parser/property.py. I checked its implementation, usages, and existing tests, and confirmed that it returns *str*. Is this scope suitable as a small contribution to #938?

stevepiercy:
@goAuD Please go ahead!

PR #940, opened 2025-10-26
Add type hints in icalendar/tools.py

jkseppan:
## Closes issue

See #938 (doesn't close it as there are many more type hints to add)

## Description

The union of date and datetime is actually date, since datetime inherits from date, but I think the type hints are still useful documentation. Annotating the return value of is_datetime as TypeIs[datetime] should help type checkers.

## Checklist

- [N/A] I've added or updated tests if applicable.
- [x] I've run and ensured all tests pass locally by following [Running Tests](https://icalendar.readthedocs.io/en/latest/install.html#running-tests).
- [x] I've added myself to `docs/credits.rst` as a contributor in this pull request or have done so previously.

## Additional information

https://docs.python.org/3/library/typing.html#typing.TypeIs

----
📚 Documentation preview 📚: https://icalendar--940.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,7 @@ Minor changes:
 - Update LICENSE file heading formatting.
 - Document ``icalendar.version`` module.
 - Use ``CONTRIBUTING.md`` in favour of ``.rst`` file.
+- Add type hints to ``icalendar.tools`` module.
 
 Breaking changes:
 
--- a/docs/contribute/credits.rst
+++ b/docs/contribute/credits.rst
@@ -61,6 +61,7 @@ Contributors
 - Jochen Sprickerhof <icalendar@jochen.sprickerhof.de>
 - Johannes Raggam <johannes@raggam.co.at>
 - Jordan Kiang <jordan@cozi.com>
+- `Jouni K. Seppänen <https://github.com/jkseppan>`_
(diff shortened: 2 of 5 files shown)


jkseppan:
Thanks! I assume that the `(?i)` on most lines means case-insensitive comparison and should not be specified for proper names that should always start with a capital letter.

Documentation here: https://icalendar--940.org.readthedocs.build/en/940/reference/api/icalendar.tools.html

niccokunzmann (approved):

niccokunzmann:
Thanks!

CHANGES.rst
```diff
@@ -39,6 +39,7 @@ Minor changes:
 - Update LICENSE file heading formatting.
 - Document ``icalendar.version`` module.
 - Use ``CONTRIBUTING.md`` in favour of ``.rst`` file.
+- Add type hints to ``icalendar.tools`` module.
```
stevepiercy, line 42: FYI, this is fine, but you can also use reStructuredText to link to the module.

```rest
:py:mod:`icalendar.tools`
```

This will render as an inline literal at the root of the repo, but as a link where it gets included in the docs at https://icalendar.readthedocs.io/en/latest/reference/changelog.html.

stevepiercy (commented):
Follow up remarks.

PR #943, opened 2025-10-26
Fix #898: Todo.end returns start for date DTSTART

### LINKED ISSUE #898: TODO.end = TODO.start if no DURATION/DUE is given
This should be checked in a test: TODO.end = TODO.start if no DURATION/DUE is given

- [ ] create a test that checks this case
- [ ] change code if needed

Context: The RFC makes a statement about the Event but there is nothing about the TODO. I doubt though that the TODO is endless but I do not know.

@tobixen - you work more with TODOs. It is a case that TODOs have only a DTSTART. What is the expected end of such a TODO? The same as the start or is it endless? (None)

_Originally posted by @niccokunzmann in https://github.com/collective/icalendar/pull/896#discussion_r2355068476_

SashankBhamidi:
Fixes #898

Todo with date-only DTSTART was returning start + 1 day instead of start.

Changed get_end_property() to check component type:
- Event: date → start + 1 day, datetime → start (unchanged)
- Todo: both date and datetime → start (fixed)

----
📚 Documentation preview 📚: https://icalendar--943.org.readthedocs.build/


src/icalendar/tests/test_issue_898_todo_end.py
```diff
@@ -0,0 +1,53 @@
...
+        """Todo with date DTSTART should have end = start (NOT start + 1 day)."""
+        todo = Todo()
+        todo.add("UID", "test-todo-date")
+        todo.add("DTSTART", date(2026, 1, 1))
+
+        # This should be start, not start + 1 day
+        assert todo.end == date(2026, 1, 1)
+
+    def test_todo_with_datetime_dtstart_only(self):
+        """Todo with datetime DTSTART should have end = start."""
```
niccokunzmann, line 52:
```suggestion
        """We assume that the Todo with datetime DTSTART should have end = start.
        
        It is not clear from the RFC.
        end could also be None but this works best at the moment.
        """
```

I would add the assumption.

niccokunzmann (changes requested):
I am not definite on this and would add this comment with a hint that allows people to change this test in the future.

src/icalendar/tests/test_issue_898_todo_end.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_event_with_datetime_dtstart_only(self):
+        """Event with datetime DTSTART should have end = start."""
+        event = Event()
+        event.add("UID", "test-event-datetime")
+        event.add("DTSTART", datetime(2026, 1, 1, 12, 0))
+
+        assert event.end == datetime(2026, 1, 1, 12, 0)
+
+    def test_todo_with_date_dtstart_only(self):
+        """Todo with date DTSTART should have end = start (NOT start + 1 day)."""
```
niccokunzmann:
```suggestion
        """Todo with date DTSTART should have end = start (NOT start + 1 day).
        
        It is not clear from the RFC.
        end could also be None but this works best at the moment.
        """
```

src/icalendar/tests/test_issue_898_todo_end.py
niccokunzmann: Wrong test..

PR #990, opened 2025-11-13
Update check of which files trigger a RTD build.

### LINKED ISSUE #848: Check RTD build skip requirements
In the .readthedocs.yml file, we skip the build under certain conditions.
These conditions will not fit any more, especially after we merge #845.
So, they need revisiting:

- [ ] Which Python files are excluded, which are not?
- [ ] Going a level deeper is not considered
- [ ] Documentation restructuring will maybe require revisiting the scope of the .rst check.

stevepiercy:
Closes #848

- Use a glob of `**` to grab all `.py` file in `src/icalendar/`, subject to the three exclusions at the end of the list
- Add two docs `.github/workflows/` files
- Add .vale.ini
- Specify `CHANGES.rst` as it is the only remaining .rst file outside of the `docs` directory that gets included by Sphinx at build time.
- change log

----
📚 Documentation preview 📚: https://icalendar--990.org.readthedocs.build/


niccokunzmann (approved):
Nice!

PR #1002, opened 2025-11-21
Revise wording of issue and pull request templates

stevepiercy:
## Description

- Fix checklist item wording in pull request template.
- Increase spacing and improve formatting and wording to make the bug report and blank issue templates easier to read.

A change log isn't needed, since this is for internal use only and an update to the initial versions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/empty-issue.md
+++ b/.github/ISSUE_TEMPLATE/empty-issue.md
@@ -7,11 +7,15 @@ assignees: ''
 
 ---
 
-<!-- This is an unstructured issue template to use if you do not need guidance.
-Here are some hints though:
-- add ICS example files if you can
-- show code if possible
-- show error output or output that differs from what you expect
-- it is nice to connect to the value of this issue
+<!--
+This is an unstructured issue template to use if you don't need guidance.
+
+Here are some hints, though.
+
+- Add example `.ics` files, if relevant.
+- Show code.
+- Show logs, stack traces, error output, or other output that differs from what you expect.
+- Add links to related issues or pull requests.
+
 Thanks for taking your time to report this!
- -->
+-->


.github/ISSUE_TEMPLATE/empty-issue.md
```diff
@@ -7,11 +7,15 @@ assignees: ''
...
-<!-- This is an unstructured issue template to use if you do not need guidance.
-Here are some hints though:
-- add ICS example files if you can
-- show code if possible
-- show error output or output that differs from what you expect
-- it is nice to connect to the value of this issue
+<!--
+This is an unstructured issue template to use if you don't need guidance.
+
+Here are some hints, though.
```
SashankBhamidi, line 13:
```suggestion
Here are some hints, though:
```
Pardon my English but I've always used colon while listing, is that an actual thing? If not, suggestion may be ignored.

SashankBhamidi (approved):

.github/ISSUE_TEMPLATE/empty-issue.md
stevepiercy: Lists don't have a hard and fast English grammar rule. It's completely style. The Microsoft Writing Style Guide, which we've adopted, suggests three different acceptable list forms. The [Introductory text](https://learn.microsoft.com/en-us/style-guide/scannable-content/lists#introductory-text) section is really good at explaining usage.

I prefer to write complete sentences ending with a period, instead of sentence fragments ending with a colon or headings, because of accessibility. It helps people either with vision impairments using screen readers or who use translation tools, because there is more context. It takes more time to write, but I want docs to be open to the broadest audience possible.

Two of the problems that authors blip over when writing introductory sentence fragments is:

- writers can't count
- grammar and consistent style falls apart
- see what I did there?

And AI is horrendous for this because it has so many bad examples, such as the foregoing list, from which to copy.

SashankBhamidi: Ah, makes sense. Thanks for explaining!

PR #1038, opened 2025-12-19
Release 7.0.0a3

niccokunzmann:
This follows the release process for a new release.

----
📚 Documentation preview 📚: https://icalendar--1038.org.readthedocs.build/

### DIFF
--- a/docs/contribute/maintenance.rst
+++ b/docs/contribute/maintenance.rst
@@ -91,6 +91,8 @@ However, only people with ``Environments/Configure PyPI`` access can approve an 
 
 #.  Check that the file :file:`CHANGES.rst` is up to date with the `latest merged pull requests <https://github.com/collective/icalendar/pulls?q=is%3Apr+is%3Amerged>`_, and the version you want to release is correctly named.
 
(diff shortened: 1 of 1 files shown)


docs/contribute/maintenance.rst
```diff
@@ -103,16 +103,16 @@ However, only people with ``Environments/Configure PyPI`` access can approve an
 
         git checkout main
         git pull
-        git checkout -b release main
-        git add CHANGES.rst
+        git checkout -b release-$VERSION main
```
stevepiercy, line 108: Oh, now I understand why you have that code block in Step 1!

Would you please add a sentence to Step 1 (I can't make a suggestion on that line in GitHub):

```rst
    Set an environment variable to use in subsequent commands during the release process.
```

stevepiercy (changes requested):
One tweak, and I approve. Thank you!

docs/contribute/maintenance.rst
niccokunzmann: 1916368 - how is that?

stevepiercy: Perfect!

PR #1039, opened 2025-12-19
Update version to '6.x' in version-switcher.json

niccokunzmann:
This is to see if a general version in the version switcher would work.

## Closes issue

I would like changes to the version switcher to be minimal.

## Description

This tries out a 6.x version like for the others.

## Checklist

- [ ] I've added myself to `docs/credits.rst` as a contributor in this pull request or have done so previously.

----
📚 Documentation preview 📚: https://icalendar--1039.org.readthedocs.build/

### DIFF
--- a/docs/_static/version-switcher.json
+++ b/docs/_static/version-switcher.json
@@ -5,7 +5,7 @@
     },
     {
         "name": "6.x (stable)",
-        "version": "v6.3.2",
+        "version": "6.x",
         "url": "https://icalendar.readthedocs.io/en/stable/",
         "preferred": "true"
     },


niccokunzmann:
@stevepiercy What do you think of this change? It seems to work but I am not sure where the data is used. In the end, I would like to avoid changing this file.
Are all versions using the version switcher of stable or of latest?

stevepiercy (commented):
So far, so good.

You'll need to update the RTD configuration at https://app.readthedocs.org/dashboard/icalendar/rules/

stevepiercy:
`docs/_static/version-switcher.json` exists only in `main`, or "latest" in the URL or "dev" in the switcher on RTD.

All versions have the following code in their `docs/conf.py`.

```python
# Define the json_url for our version switcher.
json_url = "https://icalendar.readthedocs.io/en/latest/_static/version-switcher.json"

# Define the version we use for matching in the version switcher.
version_match = os.environ.get("READTHEDOCS_VERSION")
release = icalendar.__version__

# If READTHEDOCS_VERSION doesn't exist, we're not on RTD
# If it's an integer, we're in a PR build and the version isn't correct.
# If it's "latest", change to "dev" in the version switcher.
if not version_match or version_match.isdigit() or version_match == "latest":
    # For local development, infer the version to match from the package.
    if "a" in release or "b" in release or "rc" in release or "dev" in release:
        # Override the fully qualified URL in dev mode.
        json_url = "_static/version-switcher.json"
        version_match = "dev"
    else:
        version_match = f"v{release}"
elif version_match == "stable":
    version_match = f"v{release}"
```

Does that help? It's tricky stuff!

niccokunzmann:
I see this:

grafik

So, I wonder what to update...

niccokunzmann:
@stevepiercy Would it be ok to merge this and then have a look if it all works? If it does not, we can revert the commit.

niccokunzmann:
Ah, I did not see the comment! Yes, that sounds good with setting new rules.

I would like to avoid the 7.x branch in some ways. It seems like an extra step. But we can try it out for a bit. I guess it

- will be protected
- and always the same a main OR as the latest tag?

readthedocs also allows clicking a button to build branches. So, I would just tinker with it. Maybe, `stable` on RTD can match tags always. Then, we do not need the branch. I think, that is what it is doing anyway. We can change the rules. I should have a look and report back.

stevepiercy:
> I would like to avoid the 7.x branch in some ways. It seems like an extra step. But we can try it out for a bit.

Yes, it is, and it's a common practice, although somewhat new to icalendar.

> I guess it
> 
>     * will be protected

We already have a rule for `*.x`, so `7.x` would inherit that protection.

>     * and always the same a main OR as the latest tag?

Not necessarily. When a breaking change enters `main`, then you wouldn't want to backport it to `7.x`. Then further along, you'd consider when to do an `8.x` release and a new branch upon release. A `latest` tag is not appropriate, as its name must be unique on a specific commit.

> Maybe, `stable` on RTD can match tags always. Then, we do not need the branch. I think, that is what it is doing anyway. We can change the rules. I should have a look and report back.

I thought more about this, and I think my suggestion to try to override how RTD automatically determines the default (stable) _version_ was a bad idea. Let's just use what they offer for free. See https://docs.readthedocs.com/platform/stable/versions.html#versions-are-git-tags-and-branches:

> If your project has any tags or branches with a name following [semantic versioning](https://semver.org/) (with or without a `v` prefix), we also create a `stable` version tracking your most recent release. If you want a custom `stable` version, create either a tag or branch in your project with that name.

So that means that the tag with the highest version number that follows semver would be the stable version. I expect when you create a `7.0.0` tag, then RTD will automatically make it the `stable` version. We wouldn't have to do anything.

Meanwhile, RTD would not set the commit for a non-stable tag, such as `7.0.0a`, as its `stable` version.

> This matches tags and should set the last pushed tag as stable. Maybe, we should set it to version 7 only? In case we push a version 6 one? `(?:\d\d+|7|8|9)\.\d+\.\d+` ?

I think that would match any backports to older versions, so if there was a bugfix in 6, then that tagged version might become the default (stable) version, even though `7.0.0` would be the actual stable release and version. I'd prefer to see what happens on RTD with the `7.0.0` tag and release.

niccokunzmann:
Alright. Can you create a list of steps of what we need to do to get this merged? I am a bit confused on what to do next.

I assume, we will document the 7.x branch and do the releases from there. That needs documenting then. I will do that on release and create an issue once we have that branch.

Also, when we add a 7x. branch, then we need to document it:

- possibly here: https://github.com/collective/icalendar/pull/1083#discussion_r2702467926
- and update all the JSON files in all the \d.x branches?
- update this JSON config

What are the next steps?

stevepiercy:
Closing to avoid merge conflict resolution.

PR #1042, opened 2025-12-20
Link timedelta to datetime.timedelta in the Python standard library documentation

### LINKED ISSUE #951: timedelta not linked in documentation
When I check out the locally built documentation, I can see that timedelta is not linked to the Python std lib documentation.

Image

stevepiercy:
## Closes issue

- [X] Closes #951

## Description

Link timedelta to datetime.timedelta in the Python standard library documentation.

## Checklist

- [X] I've added or updated tests if applicable.

----
📚 Documentation preview 📚:

- https://icalendar--1042.org.readthedocs.build/en/1042/reference/changelog.html
- https://icalendar--1042.org.readthedocs.build/en/1042/reference/api/icalendar.prop.html#icalendar.prop.vText.GAP
- https://icalendar--1042.org.readthedocs.build/en/1042/reference/api/icalendar.cal.event.html#icalendar.cal.event.Event.duration
- https://icalendar--1042.org.readthedocs.build/en/1042/reference/api/icalendar.cal.todo.html#icalendar.cal.todo.Todo.duration

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- Simplify contributors and add supporters in credits. See `pull request 1035 <https://github.com/collective/icalendar/pull/1041>`_.
+- ...
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
@@ -43,6 +43,13 @@ Bug fixes
 ~~~~~~~~~
 
 - Link ``timedelta`` to :py:class:`datetime.timedelta` in the Python standard library documentation. See `Issue 951 <https://github.com/collective/icalendar/issues/951>`_.
+
+Documentation
+~~~~~~~~~~~~~
+
+- Simplify contributors and add supporters in credits. See `pull request 1035 <https://github.com/collective/icalendar/pull/1041>`_.
+- Add a section in the change log for Documentation. See `Issue 1043 <https://github.com/collective/icalendar/issues/1043>`_.
+
 
 7.0.0a3 (2025-12-19)
 --------------------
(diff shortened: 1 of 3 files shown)


src/icalendar/cal/component.py
```diff
@@ -84,7 +84,7 @@ def get_component_class(cls, name: str) -> type[Component]:
 
     @staticmethod
     def _infer_value_type(
-        value: date | datetime | timedelta | time | tuple | list,
+        value: date | datetime | datetime.timedelta | time | tuple | list,
```
```
import datetime
```
or
```
from datetime import datetime, timedelta
```

in the second case, datetime.timedelta does not work.

src/icalendar/cal/event.py
```diff
@@ -296,7 +296,7 @@ def _get_start_end_duration(self):
     )
 
     @property
-    def duration(self) -> timedelta:
+    def duration(self) -> datetime.timedelta:
```
niccokunzmann: https://github.com/collective/icalendar/blob/eb9c4d4475c4982a8d0570b99d9239081bd62ed5/src/icalendar/cal/event.py#L6

I don't think this is supposed to work. The class `datetime` has no attribute `timedelta`..

```
>>> datetime.datetime.timedelta
AttributeError: type object 'datetime.datetime' has no attribute 'timedelta'

```

stevepiercy: @niccokunzmann what tool do you use to get that traceback?

I use pytest directly without tox, and it didn't catch this. There's nothing obvious in our docs either. https://icalendar.readthedocs.io/en/latest/contribute/development.html

src/icalendar/cal/component.py
stevepiercy: Fixed in 51c3a6a412b3b37c119aef501a45c1345a189b43

src/icalendar/cal/event.py
stevepiercy: Fixed in 51c3a6a412b3b37c119aef501a45c1345a189b43

Issue #158, opened 2015-03-24
Lazy parsing


untitaker:
Icalendar currently raises exceptions on minor things like invalid datetimes as described in https://github.com/collective/icalendar/issues/155. I think iCalendar could parse property values lazily, i.e. when `Component.__getitem__` is called, since some broken properties shouldn't stop the user from accessing the rest.

untitaker:
It also would help with the performance problems I have when manipulating iCalendar files in a way that doesn't involve properties at all. E.g. in [vdirsyncer](https://github.com/untitaker/vdirsyncer) I had to write my own parser, both due to performance and tolerance of broken items.

thet:
+1, sounds like a good enhancement.

niccokunzmann:
This is included in v7.1.0.

Issue #104, opened 2013-07-31
Content line could not be parsed into parts Exception


rchrd2:
Is there any way to ignore offending entries and parse the rest of the file? 

Without any icalendar repair tools, it is difficult to read a broken file. 

I tried repairing the file with this regex, but it does not work still:

```
    asciidata = re.sub(r'^(?!([A-Z]+):)', r'     ', asciidata, count=0, flags=re.MULTILINE)

    try:
        cal = icalendar.Calendar.from_ical(asciidata)
    except Exception as detail:
        print >>sys.stderr, detail
```

When the exception is thrown, it prevents the rest of the file from being parsed.

Thank you.

potuz:
I am hitting the same error on a valid V4.0 vcard (using khal to extract bdays)

```
debug: Updating 0a9acecd6abeb4db248c933231cecfc8e4939c315cbe38bd4e7c1a952df690eb.vcf because 1498507884569668096.000000000 != None
debug: Updating b7vf5c4ge3a33le3rr8o4ubor7dl3wzs7byn.vcf because 1498518463768040960.000000000 != None
Unknown exception happened.
Traceback (most recent call last):
  File "/usr/lib64/python3.4/site-packages/icalendar/parser.py", line 230, in from_ical
    key, val = q_split(param, '=')
ValueError: need more than 1 value to unpack

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.4/site-packages/icalendar/parser.py", line 336, in parts
    strict=self.strict)
  File "/usr/lib64/python3.4/site-packages/icalendar/parser.py", line 255, in from_ical
    % (param, exc))
ValueError: 'data' is not a valid parameter string: need more than 1 value to unpack

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.4/site-packages/khal/khalendar/khalendar.py", line 303, in _update_vevent
    update(event.raw, href=href, etag=etag, calendar=calendar)
  File "/usr/lib64/python3.4/site-packages/khal/khalendar/backend.py", line 276, in update_birthday
    ical = icalendar.Event.from_ical(vevent)
  File "/usr/lib64/python3.4/site-packages/icalendar/cal.py", line 324, in from_ical
    name, params, vals = line.parts()
  File "/usr/lib64/python3.4/site-packages/icalendar/parser.py", line 346, in parts
    % (self, exc)
[...]

PR #1044, opened 2025-12-21
Error-tolerant property parsing for Events (vBrokenProperty fallback)

SashankBhamidi:
Implements error-tolerant property parsing for Event components with vBrokenProperty fallback.

## What this does

- **vBrokenProperty(vText)**: Simple vText subclass that preserves raw values and stores parse metadata
- **Eager parsing with fallback**: Properties parsed immediately during from_ical(), broken ones become vBrokenProperty
- **Error recording**: Parse errors recorded immediately in `component.errors`
- **Event-only**: Only Events have `ignore_exceptions=True` (Calendar/Timezone stay strict per #104)
## Background

Resurrects @stlaz's 97c2ce3 (2016) vText fallback concept. Implements Nicco's requested architecture from review - simple vText subclass instead of LazyProperty wrapper.

## Why Event-only?

Initially tried error tolerance for all components, but:
- `test_cal_strict_parsing` broke - VTIMEZONE needs errors raised during `from_ical()`
- Issue #104 intentionally made Calendar strict while Events are lenient
- `test_dont_ignore_exceptions_on_broken_calendars_issue_104` validates this design

Extending to Calendar/Timezone would be a breaking change. Tracked in #1051 for separate discussion.

## What this doesn't do

This PR focuses on **error tolerance** for Event properties. It does **not** address performance for large calendars - all Event objects are still created eagerly during `from_ical()`.

Lazy **subcomponents** (the real performance win) tracked in #1050.

## Related

Addresses #158 (error tolerance aspect - broken properties don't block access to valid ones), #1050 (lazy subcomponents - performance aspect of #158), #1051 (Calendar/Timezone error tolerance discussion)

----
📚 Documentation preview 📚: https://icalendar--1044.org.readthedocs.build/en/1044/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,7 +42,7 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
-- Event components now have error-tolerant property parsing. Properties with parsing errors fall back to vText, preserving the raw value and allowing access to other valid properties. Errors are recorded in ``component.errors``. Partially addresses `#158 <https://github.com/collective/icalendar/issues/158>`_.
+- Event components now have error-tolerant property parsing. Properties with parsing errors fall back to :class:`~icalendar.prop.vBrokenProperty`, preserving the raw value and allowing access to other valid properties. Errors are recorded in ``component.errors``. Partially addresses `#158 <https://github.com/collective/icalendar/issues/158>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/reference/api/icalendar.cal.rst
+++ b/docs/reference/api/icalendar.cal.rst
@@ -23,6 +23,5 @@ Submodules
    icalendar.cal.examples
    icalendar.cal.free_busy
    icalendar.cal.journal
-   icalendar.cal.lazy
    icalendar.cal.timezone
    icalendar.cal.todo
--- a/src/icalendar/cal/component.py
+++ b/src/icalendar/cal/component.py
@@ -647,15 +647,6 @@ class Component(CaselessDict):
         if len(self.subcomponents) != len(other.subcomponents):
             return False
 
-        # Unwrap lazy properties before comparison
-        # LazyProperty instances won't compare equal to parsed values,
-        # so we must access all properties to trigger parsing first.
-        # Performance tradeoff: correctness over speed for large calendars.
-        for key in self.keys():
-            _ = self[key]
-        for key in other:
-            _ = other[key]
-
         properties_equal = super().__eq__(other)
         if not properties_equal:
             return False
--- a/src/icalendar/prop/__init__.py
+++ b/src/icalendar/prop/__init__.py
@@ -3547,7 +3547,7 @@ class vUnknown(vText):
 
 
 class vBrokenProperty(vText):
-    """Property that failed to parse - preserves raw value as text.
+    """Property that failed to parse, preserving raw value as text.
 
     Represents property values that failed to parse with their expected
     type. The raw iCalendar string is preserved for round-trip serialization.
--- a/src/icalendar/tests/test_components_break_on_bad_ics.py
+++ b/src/icalendar/tests/test_components_break_on_bad_ics.py
@@ -25,9 +25,7 @@ def test_rdate_dosent_become_none_on_invalid_input_issue_464(events):
     """Issue #464 - [BUG] RDATE can become None if value is invalid
     https://github.com/collective/icalendar/issues/464
     """
-    # With lazy parsing, we need to access RDATE to trigger parsing and error recording
     event = events.issue_464_invalid_rdate
-    _ = event.get("RDATE")  # Trigger lazy parsing
 
     # After VALUE parameter fix, the assertion checks for the full value string
     # in the error message, making the test more flexible
--- a/src/icalendar/tests/test_equality.py
+++ b/src/icalendar/tests/test_equality.py
@@ -60,7 +60,7 @@ def test_parsed_calendars_are_equal_if_parsed_again_jcal(source_file, tzp):
     source -> calendar -> jcal -> same calendar
     """
     if source_file.source_file in ("issue_464_invalid_rdate.ics",):
-        pytest.skip("Invalid RDATE falls back to vText, which can't round-trip to jCal.")
+        pytest.skip("Invalid RDATE falls back to vBrokenProperty, which can't round-trip to jCal.")
     copy_of_calendar = Component.from_jcal(source_file.to_jcal())
     assert_equal(copy_of_calendar, source_file)
 
--- a/src/icalendar/tests/test_error_tolerant_parsing.py
+++ b/src/icalendar/tests/test_error_tolerant_parsing.py
@@ -113,7 +113,7 @@ END:VCALENDAR"""
 
 
 def test_property_equality():
-    """Verify equality comparisons work with lazy properties."""
+    """Verify equality comparisons work with properties."""
     ical_str = b"""BEGIN:VCALENDAR
 VERSION:2.0
 PRODID:test
@@ -134,8 +134,8 @@ END:VCALENDAR"""
     assert event1["SUMMARY"] == "Test Event"
 
 
-def test_lazy_property_to_ical():
-    """Verify to_ical() works correctly with lazy properties."""
+def test_property_to_ical():
+    """Verify to_ical() works correctly with properties."""
     ical_str = b"""BEGIN:VCALENDAR
 VERSION:2.0
 PRODID:test
@@ -158,8 +158,8 @@ END:VCALENDAR"""
     assert b"SUMMARY:Test Event" in output
 
 
-def test_multiple_lazy_properties_in_list():
-    """Verify lists of lazy properties work correctly."""
+def test_multiple_properties_in_list():
+    """Verify lists of properties work correctly."""
     ical_str = b"""BEGIN:VCALENDAR
 VERSION:2.0
 PRODID:test
@@ -184,7 +184,7 @@ END:VCALENDAR"""
     assert all(isinstance(dt, vDDDLists) for dt in exdates)
 
 
-def test_lazy_property_with_tzid():
+def test_property_with_tzid():
     """Verify TZID parameter is handled correctly."""
     ical_str = b"""BEGIN:VCALENDAR
 VERSION:2.0
@@ -210,7 +210,7 @@ END:VCALENDAR"""
 
 
 def test_freebusy_comma_separated_values():
-    """Verify FREEBUSY comma-separated values work with lazy parsing."""
+    """Verify FREEBUSY comma-separated values work with error-tolerant parsing."""
     ical_str = b"""BEGIN:VCALENDAR
 VERSION:2.0
 PRODID:test
--- a/src/icalendar/tests/test_unit_cal.py
+++ b/src/icalendar/tests/test_unit_cal.py
@@ -430,14 +430,7 @@ def test_calendar_with_parsing_errors_has_an_error_in_one_event(calendars):
     attribute. The error in the following is the third EXDATE: it has an
     empty DATE.
     """
-    # Access EXDATE properties to trigger lazy parsing and error recording.
-    # With lazy parsing, errors are only recorded when properties are accessed,
-    # not during from_ical(). This is intentional for performance.
-    events = calendars.parsing_error.walk("VEVENT")
-    for e in events:
-        _ = e.get("EXDATE", [])
-
-    errors = [e.errors for e in events]
+    errors = [e.errors for e in calendars.parsing_error.walk("VEVENT")]
     assert errors == [[], [("EXDATE", "Expected datetime, date, or time. Got: ''")]]
 
 


stevepiercy (changes requested):
With the new module, you'll need to run `make apidoc`, then `make livehtml`, to ensure that the new module gets documented.

It should appear somewhere under https://icalendar--1044.org.readthedocs.build/en/1044/reference/api/icalendar.cal.html

stevepiercy:
@SashankBhamidi fyi, when you open a branch from your own fork instead of this repo, RTD won't automatically insert a link to the pull request preview in the description, and we'll need to hunt it down as maintainers in https://app.readthedocs.org/projects/icalendar/builds/.

SashankBhamidi:
I'm crashing out if this fails again

SashankBhamidi:
In f7d8c5db I added AttributeError handling to the fuzzer harness. The crash was happening because malformed VTIMEZONE components with nested VEVENTs would cause an AttributeError when the timezone code tried to access .dt on a vText object.

Now CIFuzz is reporting a timeout on this input: `QkVHSU46CkRJUjoKRElSOgpESVI6` (base64)

Which decodes to:
```
BEGIN:
DIR:
DIR:
DIR:
```

I can't reproduce this locally at all. The input completes in under 1ms and raises `ValueError: Found no components where exactly one is required` as expected. I've tested it with the fuzzer harness directly and tried all the parameter combinations, everything passes instantly.

I've looked at the changes I made for lazy parsing, I catch Exception instead of specific types, have nested try/except for vText fallback, and manually create vText as a last resort. But this timeout input fails during component parsing before it ever reaches the lazy parsing code.

src/icalendar/cal/component.py
```diff
@@ -128,6 +129,36 @@ def __bool__(self):
...
+
+        # Handle single property
+        if isinstance(value, LazyProperty):
+            parsed = value.get_parsed_value()
+            super().__setitem__(key, parsed)
+            return parsed
+
+        return value
+
+    def get(self, key, default=None):
```
niccokunzmann, line 195: TODO: Check if we need this.

src/icalendar/cal/component.py
```diff
@@ -434,35 +465,54 @@ def from_ical(cls, st, multiple: bool = False) -> Self | list[Self]:  # noqa: FB
...
+                        component.add(name, lazy_prop, encode=0)
+                else:
+                    # Parse eagerly for strict components (maintain current behavior)
+                    for val in vals_list:
+                        try:
+                            if tzid:
+                                parsed_val = factory.from_ical(val, tzid)
+                            else:
+                                parsed_val = factory.from_ical(val)
+                        except ValueError:
```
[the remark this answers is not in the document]
niccokunzmann: TODO: Dependabot's comment should be tested, not just fixed.

src/icalendar/cal/component.py
```diff
@@ -434,35 +465,53 @@ def from_ical(cls, st, multiple: bool = False) -> Self | list[Self]:  # noqa: FB
...
-                        ]
-                    # Workaround broken ICS files with empty RDATE
-                    elif name == "RDATE" and vals == "":
-                        parsed_components = []
-                    else:
-                        parsed_components = [factory(factory.from_ical(vals))]
-                except ValueError as e:
-                    if not component.ignore_exceptions:
-                        raise
-                    component.errors.append((uname, str(e)))
```
niccokunzmann, line 551: component.errors seems to be empty after this change whereas it was filled before. I suggest it still is filled as that is not required to be changed.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,163 @@
...
+        self._value_param = value_param
+        self._tzid = tzid
+        self._component = component
+        self._parsed_value = None
+        self._parse_error = None
+
+    @property
+    def params(self):
+        """Get parameters without triggering parse."""
+        return object.__getattribute__(self, "_params")
```
niccokunzmann: TODO: check if `__getattribute__` is needed as `__getattribute__` is not defined in this class.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,163 @@
+"""Lazy parsing wrapper for iCalendar property values."""
```
niccokunzmann: TODO: As this is a property, it should be in `icalendar.prop.lazy`

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,163 @@
...
+                parsed_value = vText(raw_value)
+                parsed_value.params = params
+
+            object.__setattr__(self, "_parsed_value", parsed_value)
+            object.__setattr__(self, "_parse_error", parse_error)
+
+            if component is not None:
+                component.errors.append((property_name, parse_error))
+
+    def get_parsed_value(self):
```
niccokunzmann: Return type can be VPROPERTY

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,163 @@
...
+        """Get the parsed value, parsing if necessary."""
+        self._ensure_parsed()
+        return object.__getattribute__(self, "_parsed_value")
+
+    def __getattr__(self, name):
```
niccokunzmann: `__getattr__` is only called when the whole hierarchy is traversed. This is generally very slow.

src/icalendar/tests/fuzzed/__init__.py
```diff
@@ -52,7 +53,7 @@ def fuzz_v1_calendar(
                     event.to_ical()
             else:
                 c.to_ical()
-    except (ValueError, TypeError) as e:
+    except (ValueError, TypeError, AttributeError) as e:
```
niccokunzmann: I would not want to add this in some way as we want to catch those and the 'attribute' is too general. The fuzzer is there to detect the edge cases that developers cannot find easily. This blankets every spelling mistake we make in untested code. It would be good to catch them.

src/icalendar/tests/test_components_break_on_bad_ics.py
```diff
@@ -25,13 +25,17 @@ def test_rdate_dosent_become_none_on_invalid_input_issue_464(events):
     """Issue #464 - [BUG] RDATE can become None if value is invalid
     https://github.com/collective/icalendar/issues/464
     """
+    # With lazy parsing, we need to access RDATE to trigger parsing and error recording
```
niccokunzmann: Hm. This is confusing me:

1. errors are not useful any more. I would expect to be able to go though them to see if there is anything. How are they used? Is there documentation on this? This is a change in what is expected to work for a long time.
2. This is not discoverable. Creating a property called `errors` that then goes through all those and does the work would be the nicer approach.
3. What if I am interested in errors - iterate over everything twice?

src/icalendar/tests/test_equality.py
```diff
@@ -59,6 +59,8 @@ def test_parsed_calendars_are_equal_if_parsed_again_jcal(source_file, tzp):
 
     source -> calendar -> jcal -> same calendar
     """
+    if source_file.source_file in ("issue_464_invalid_rdate.ics",):
```
niccokunzmann, line 62: TODO: That should be able to to round trip. This is an edge case that should be recorded in an issue or solved, I think.

src/icalendar/tests/test_lazy_parsing.py
```diff
@@ -0,0 +1,378 @@
...
+    event = cal.walk("VEVENT")[0]
+
+    # Access valid property should work
+    dtend = event["DTEND"]
+    assert isinstance(dtend, vDDDTypes)
+    assert dtend.dt.year == 2025
+
+    # Access broken property should work (as vText)
+    dtstart = event["DTSTART"]
+    assert isinstance(dtstart, vText)
```
niccokunzmann: TODO: I think that it would be nice to subclass vText and create a vBroken that can have a few more methods or properties to it. This is a hunch of how I would implement it to delegate more to the parsed property.

src/icalendar/tests/test_error_tolerant_parsing.py
```diff
@@ -0,0 +1,378 @@
...
+    assert isinstance(dtstart, vText)
+
+    # Other properties accessible
+    summary = event["SUMMARY"]
+    assert str(summary) == "Test Event"
+
+
+def test_lazy_property_equality():
+    """Verify equality comparisons work with lazy properties."""
+    ical_str = b"""BEGIN:VCALENDAR
```
niccokunzmann, line 117: Hint: For these strings, we have the calendars directory. This makes the tests smaller.

niccokunzmann (changes requested):
Hm, essentially, this changes the way we work with the icalendar.prop internally. I am not convinced that it is needed or required to put all values into the LazyProperty. This is confusing because of the issue which suggests a solution to several use cases, one of them being fast parsing and one of them error tolerance. For error tolerance, I am not convinced that a lazy approach is required.

To address part of the issue, this is sufficient:

- Create a `class vBrokenProperty(vText)`
- Fallback to that when there is a parsing error.
- One can then also record which property type would have been expected to parse the value.
- still record all errors as was done before.
- remove LazyProperty

Personally, I am resistant to touch the core when it is not needed as I am here to maintain the package with minimal required changes for downstream use.

src/icalendar/cal/component.py
SashankBhamidi: Kept it. Now just delegates to `CaselessDict.get()` - standard dict interface. Simplified from the lazy unwrapping version.

SashankBhamidi: Kept it. Now just delegates to CaselessDict.get() - standard dict interface. Simplified from the lazy unwrapping version.

src/icalendar/cal/component.py
SashankBhamidi: Fixed. Errors now appended immediately during parsing in the exception handler, not deferred to property access.

src/icalendar/tests/fuzzed/__init__.py
SashankBhamidi: Removed. vBrokenProperty is a plain vText subclass, won't raise AttributeError. Fuzzer only catches ValueError and TypeError now.

src/icalendar/tests/test_components_break_on_bad_ics.py
SashankBhamidi: Fixed with eager parsing. Errors are now in component.errors immediately after from_ical() completes, no property access needed. 

The comment about "need to access RDATE" is outdated and removed. Test now verifies errors exist without accessing properties.

src/icalendar/tests/test_equality.py
SashankBhamidi: Yeha, verified working. vBrokenProperty.to_ical() inherits from vText and returns the raw escaped value. 

Round-trip works: parse → serialize → parse preserves broken properties. 

Added test_vbroken_property_metadata() that validates this.

src/icalendar/tests/test_lazy_parsing.py
SashankBhamidi: Done. Stores what type was expected and why parsing failed.

SashankBhamidi: Created vBrokenProperty(vText) with metadata properties: expected_type, property_name, parse_error.

src/icalendar/tests/test_error_tolerant_parsing.py
SashankBhamidi: Good point. Will use existing test files where applicable in future updates 👍

src/icalendar/cal/component.py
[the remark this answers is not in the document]
SashankBhamidi: The current code catches `Exception` at component.py:574, which handles both `ValueError` and `TypeError` (and any other exceptions that might occur during parsing). This is actually more robust than the Copilot suggestion.

SashankBhamidi:
> TODO: Dependabot's comment should be tested, not just fixed.

@niccokunzmann See https://github.com/collective/icalendar/pull/1044#discussion_r2676323303

SashankBhamidi:
The timeout is fuzzer instrumentation overhead in CI, not a performance issue.

Eager parsing is more predictable: all parsing happens upfront in `from_ical()` instead of unpredictably when properties are accessed. File parses successfully and error handling works correctly.

src/icalendar/prop/__init__.py
```diff
@@ -3546,6 +3546,70 @@ def examples(cls) -> list[vUnknown]:
     from icalendar.param import VALUE
 
 
+class vBrokenProperty(vText):
+    """Property that failed to parse - preserves raw value as text.
```
stevepiercy:
```suggestion
    """Property that failed to parse, preserving raw value as text.
```

docs/reference/api/icalendar.cal.rst
```diff
@@ -23,5 +23,6 @@ Submodules
    icalendar.cal.examples
    icalendar.cal.free_busy
    icalendar.cal.journal
+   icalendar.cal.lazy
```
stevepiercy: Where's `src/icalendar/cal/lazy.py`? See https://app.readthedocs.org/projects/icalendar/builds/30946111/#301159577--160. It should show up in the navigation once pushed.

stevepiercy (changes requested):

src/icalendar/cal/component.py
```diff
@@ -614,6 +647,15 @@ def __eq__(self, other):
         if len(self.subcomponents) != len(other.subcomponents):
             return False
 
+        # Unwrap lazy properties before comparison
```
niccokunzmann: This is about lazy parsing. Could you remove all changes that are not needed?

src/icalendar/tests/test_components_break_on_bad_ics.py
niccokunzmann: I still see this in the PR:

```
    _ = event.get("RDATE")  # Trigger lazy parsing
```

src/icalendar/tests/test_equality.py
niccokunzmann: Can this line be removed?

src/icalendar/tests/test_error_tolerant_parsing.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def test_broken_property_doesnt_block_others():
+    """Verify one broken property doesn't prevent accessing others."""
+    ical_str = b"""BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:test
+BEGIN:VEVENT
+UID:test-123
+DTSTART:INVALID-DATE
```
niccokunzmann, line 92: This calendar and the calendar above are the same. When you use AI to generate test cases, then duplication happens. Duplication makes it harder to maintain code. Test code is code, too.
This code breaks the pattern that is there for the other tests: The icalendar code is contained in the `.ics` files and accessed with `calendars.calendar_file_name`. That cleans up the test cases, see
- https://docs.pytest.org/en/stable/explanation/anatomy.html#test-anatomy
- https://docs.pytest.org/en/stable/explanation/fixtures.html

How do you feel about refactoring?

src/icalendar/tests/test_error_tolerant_parsing.py
niccokunzmann: This whole test could be skipped if this calendar was in a test file. Could you move it to a test file?

src/icalendar/tests/test_lazy_parsing.py
```diff
@@ -0,0 +1,394 @@
...
+    event1 = cal1.walk("VEVENT")[0]
+    event2 = cal2.walk("VEVENT")[0]
+
+    # Equality should work
+    assert event1["SUMMARY"] == event2["SUMMARY"]
+    assert event1["SUMMARY"] == "Test Event"
+
+
+def test_lazy_property_to_ical():
+    """Verify to_ical() works correctly with lazy properties."""
```
niccokunzmann: This is a test for lazy parsing. I though lazy parsing was removed. Is it removed?

src/icalendar/cal/component.py
SashankBhamidi: Removed this lazy parsing code block.

src/icalendar/tests/test_lazy_parsing.py
SashankBhamidi: Yes, lazy parsing is removed. The implementation uses eager parsing with vBrokenProperty fallback. Renamed the test file to test_error_tolerant_parsing.py and updated all function names/docstrings to remove 'lazy' terminology

SashankBhamidi:
> Can this line be removed?

Tested the jCal round-trip, it fails because vBrokenProperty preserves raw text that can't be properly parsed back from jCal format. Updated the comment from 'vText' to 'vBrokenProperty' to be accurate. The skip is still needed.

SashankBhamidi:
Once this PR is merged, I'll open a follow-up PR to refactor the inline test calendars to use .ics fixture files as tracked in #1081.

src/icalendar/tests/test_error_tolerant_parsing.py
SashankBhamidi: I'll open a follow-up PR for that once this is merged, to keep the scope focused here

src/icalendar/cal/component.py
niccokunzmann: It is probably nice if it is improved. Because that allows us to document how Components act specifically.

src/icalendar/cal/component.py
```diff
@@ -510,49 +521,71 @@ def from_ical(cls, st, multiple: bool = False) -> Self | list[Self]:  # noqa: FB
...
-                            raw_value = line_str[colon_idx + 1:]
+
+                # Determine TZID for datetime properties
+                tzid = params.get("TZID") if params and name in datetime_names else None
+
+                # Handle special cases for value list preparation
+                if name.upper() == "CATEGORIES":
+                    # Special handling for CATEGORIES - need raw value
+                    # before unescaping to properly split on unescaped commas
+                    from icalendar.parser import split_on_unescaped_comma
```
niccokunzmann, line 532: The import CAN move to the top. Is it in a loop?

src/icalendar/prop/__init__.py
```diff
@@ -3546,6 +3546,70 @@ def examples(cls) -> list[vUnknown]:
...
+        value,
+        encoding=DEFAULT_ENCODING,
+        /,
+        params: dict[str, Any] | None = None,
+        expected_type: str | None = None,
+        property_name: str | None = None,
+        parse_error: str | None = None,
+    ):
+        self = super().__new__(cls, value, encoding, params=params)
+        object.__setattr__(self, "expected_type", expected_type)
```
niccokunzmann, line 3570: This works. However, meta programming is a code smell. Please just set the attributes. I wonder if the `__slots__` need to be extended to include the `__slots__` of vText? What do you think?

src/icalendar/prop/__init__.py
```diff
@@ -3546,6 +3546,70 @@ def examples(cls) -> list[vUnknown]:
...
+        object.__setattr__(self, "property_name", property_name)
+        object.__setattr__(self, "parse_error", parse_error)
+        return self
+
+    def __repr__(self) -> str:
+        return (
+            f"vBrokenProperty({str(self)!r}, "
```
niccokunzmann, line 3577: That is nice!

src/icalendar/tests/test_equality.py
niccokunzmann: This is a TODO and should be recorded as an issue: 
 
- [ ] vBrokenProperty needs to be serializable to jCal
- [ ] parsing jCal with such a broken property should work and result in the same value (vText/vUnkown/vBrokenProperty)

niccokunzmann (approved):
Thanks! I left a few comments but I would merge it, too.

src/icalendar/cal/component.py
SashankBhamidi: It's inside the parsing loop to avoid a circular import (parser imports prop, prop would import parser). Could refactor the split function to a shared module if preferred.

src/icalendar/prop/__init__.py
SashankBhamidi: Used `object.__setattr__` because vText inherits from `str` which is immutable. Will look into extending `__slots__` to include vText's slots.

src/icalendar/tests/test_equality.py
SashankBhamidi: Tracked in #1084

CHANGES.rst
```diff
@@ -42,7 +42,7 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
-- ...
+- Event components now have error-tolerant property parsing. Properties with parsing errors fall back to :class:`~icalendar.prop.vBrokenProperty`, preserving the raw value and allowing access to other valid properties. Errors are recorded in ``component.errors``. Partially addresses `#158 <https://github.com/collective/icalendar/issues/158>`_.
```
stevepiercy, line 45: I'm not clear what I can do with ``component.errors``, if anything. Does this thing need to be documented?

SashankBhamidi: `component.errors` is a list of `(property_name, error_message)` tuples. This attribute already exists but isn't documented. This PR ensures errors are populated immediately during `from_ical()`.

Yes, it should be documented. I'll create an issue to track adding documentation for error handling.

src/icalendar/cal/component.py
niccokunzmann: I would move it to the top of the function, then.

Issue #433, opened 2022-10-04
Split tests for custom components and calendar parsing


angatha:
This issue arises from #417. See that PR for more context.

In #417, a testcase was refactored that uses `icalendar.Calendar.from_ical(...)` to parse a [custom component](https://github.com/jacadzaca/icalendar/blob/4661b3b618fb368dd56b1e133ea014428e9a6c7a/src/icalendar/tests/calendars/issue_178_custom_component_contains_other.ics).

This test case should be split into two. One for Calendar and one for custom components.

This issue relates to #432 which documents how to parse custom components.

PR #1055, opened 2025-12-26
Clarify custom component parsing

### LINKED ISSUE #432: Clarify how custom components should be parsed
This issue arises from #417. See that PR for more context.

The test case from #417 uses `icalendar.Calendar.from_ical(...)` to parse a custom component. As for now (#417), that method is an effectively static class method. (It is a class method but doesn't use the `cls` argument.) It uses an instance of `icalendar.cal.ComponentFactory` to determine the class that represents the component, defaulting to `icalendar.cal.Component`.

## What is the correct way of parsing a custom component?

Doing the analog of the current behavior, is one option. `Calendar` doesn't redefine `from_ical(...)` so actually the call is equivalent to `icalendar.cal.Component.from_ical(...)`. A second option would be to model custom components with two new types. (Be welcome to add other solutions)

- Use `icalendar.cal.Component.from_ical(...)` directly.
- Introduce an `XComponent` and an `IanaComponent` (See [RFC 5545](https://www.rfc-editor.org/rfc/rfc5545#section-3.6))

Since currently there is no difference between `icalendar.Calendar.from_ical(...)` and `icalendar.cal.Component.from_ical(...)` when parsing an VCALENDAR (and other components), is it a feature, that they could be used interchangeable, is it a bug, that they don't validate the type or is it just undefined behavior when relying on the behavior of `icalendar.Calendar.from_ical(...)`? That should be clarified when documenting the new components (if that solution is chosen) but also it may be something to add for other components.

## Steps to take

1. clarify the questions from above
2. implement chosen solution
3. potentially adjust tests extracted from #433

SashankBhamidi:
Closes #432 and #433.

Clarifies how icalendar handles custom components (X-components, IANA-components).

**Changes:**
- New comprehensive how-to guide: docs/how-to/custom-components.rst
- Enhanced docstrings in ComponentFactory, Component.from_ical(), Calendar.from_ical()
- New test_custom_components.py with 25 tests covering all aspects
- Reorganized tests from test_parsing.py
- All documentation examples tested via doctests

**Note:**
No code changes to library. Documents existing behavior and validates RFC 5545 compliance.

----
📚 Documentation preview 📚: https://icalendar--1055.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,7 +28,7 @@ Minor changes
 ~~~~~~~~~~~~~
 
 - Clarified custom component (X-* and IANA-registered) parsing behavior through enhanced documentation and comprehensive how-to guide. Custom components are automatically handled by the library with no special configuration required. See `Issue #432 <https://github.com/collective/icalendar/issues/432>`_.
-- Reorganized custom component tests into dedicated ``test_custom_components.py`` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
+- Reorganized custom component tests into a dedicated :file:`test_custom_components.py` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
--- a/docs/how-to/custom-components.rst
+++ b/docs/how-to/custom-components.rst
@@ -307,16 +307,16 @@ The following example shows how to add a feature for a calendar, where an applic
 Related content
 ===============
 
-:class:`icalendar.Component`
+:class:`~icalendar.cal.component.Component`
     Base component class
 
 :class:`icalendar.cal.component_factory.ComponentFactory`
     Component factory
 
-:meth:`icalendar.Component.from_ical`
+:meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`
     Parse components
 
-:meth:`icalendar.Calendar.from_ical`
+:meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`
     Parse calendars
 
 :rfc:`5545`
--- a/src/icalendar/cal/component.py
+++ b/src/icalendar/cal/component.py
@@ -105,7 +105,7 @@ class Component(CaselessDict):
             component_class: Component subclass to register. Must have a ``name`` attribute.
 
         Raises:
-            ValueError: If component_class has no ``name`` attribute.
+            ValueError: If ``component_class`` has no ``name`` attribute.
             ValueError: If a component with this name is already registered.
 
         Example:
@@ -116,7 +116,7 @@ class Component(CaselessDict):
             ...         return "custom"
             >>> Component.register(XExample)
         """
-        if not hasattr(component_class, 'name') or component_class.name is None:
+        if not hasattr(component_class, "name") or component_class.name is None:
             raise ValueError(f"{component_class} must have a 'name' attribute")
 
         if cls._components_factory is None:
--- a/src/icalendar/cal/component_factory.py
+++ b/src/icalendar/cal/component_factory.py
@@ -34,8 +34,7 @@ class ComponentFactory(CaselessDict):
         >>> custom_class()
         X-VENDOR({})
 
-    If a component class is not yet supported, it can be either created using :meth:`get_component_class` or added manually as a subclass of :class:`~icalendar.Component`.
-
+    If a component class is not yet supported, it can be either created using :meth:`get_component_class` or added manually as a subclass of :class:`~icalendar.cal.component.Component`.
     See :doc:`/how-to/custom-components` for details.
     """
 


.vale.ini
```diff
@@ -11,3 +11,7 @@ BasedOnStyles = Vale, Microsoft
 Microsoft.Contractions = suggestion
 Microsoft.Units = suggestion
 Microsoft.GeneralURL = No
+
+# Ignore Sphinx cross-reference directives (:py:class:, :py:meth:, etc.)
+# to prevent false positives from Microsoft.Spacing on "y.C"
+TokenIgnores = (:[a-z]+:[a-z]+:`[^`]+`)
```
SashankBhamidi: Added `TokenIgnores` to prevent Microsoft.Spacing false positives on Sphinx directives (`:py:class:`, `:py:meth:`, etc.). 

@niccokunzmann @stevepiercy Could you confirm this is the right approach? I noticed `r.C` in the vocabulary for `:rfc:` directives. Is there a preferred way to handle Sphinx directives in Vale for this project?

SashankBhamidi: Also, see [74db40d](https://github.com/collective/icalendar/pull/1055/commits/74db40da4b777cb06b650f01894f12d138b98b03)

stevepiercy: In general, it's all in the configuration of Vale.

There are the accept.txt and reject.txt files in https://github.com/collective/icalendar/tree/main/docs/styles/config/vocabularies/icalendar. These are vocabularies https://vale.sh/docs/keys/vocab. One thing to note is that when you add an entry to accept.txt, then you don't need to add a complementary entry to reject.txt. I contributed the last bullet point in the list near the top of the page to their docs.

There's also the possibility of creating custom styles https://vale.sh/docs/styles.

I often configure `.vale.ini` to ignore or change the severity level to "suggestion". I haven't tried `TokenIgnores` yet.

---

For this specific issue, I'd check the specific vocabulary file, then figure out why it matches with https://regex101.com/ and the Vale docs. When I understand why it matches, then I work on an override that doesn't break other stuff. The example you tried when I use https://regex101.com/r/7DkSBA/3 implies it won't do exactly what you want.

Don't change the vocabulary files from third parties, but override them, because changes to those third-party files will get overwritten by the upstream provider.

This process can be challenging, fun, and often frustrating, such as trying to enforce the correct spelling of Node.js while allowing the word "node". I still haven't figured out this specific puzzle.

---

Side note, this filter on the CLI would be helpful to standardize headings https://vale.sh/docs/filters.

CHANGES.rst
```diff
@@ -27,7 +27,14 @@ We still recommend checking out the new features and giving feedback in the repo
...
 ~~~~~~~~~~~~~
 
-- ...
+- Clarified custom component (X-* and IANA-registered) parsing behavior through enhanced
+  documentation and comprehensive how-to guide. Custom components are automatically handled
+  by the library with no special configuration required.
+  See `Issue #432 <https://github.com/collective/icalendar/issues/432>`_.
+- Reorganized custom component tests into dedicated ``test_custom_components.py`` file
+  with expanded test coverage for ``Component.from_ical()``, ``Calendar.from_ical()``,
+  and ``ComponentFactory`` usage.
```
stevepiercy: Check my syntax to ensure the objects render and link to the API reference.

```suggestion
  with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`,
  and :class:`~icalendar.cal.component_factory.ComponentFactory` usage.
```

src/icalendar/cal/calendar.py
```diff
@@ -85,6 +85,18 @@ def example(cls, name: str = "example") -> Calendar:
 
     @classmethod
     def from_ical(cls, st, multiple=False):
+        """Parse iCalendar data into Calendar instances.
+
+        Wraps :meth:`Component.from_ical() <icalendar.Component.from_ical>` with
```
stevepiercy: I think this should link to the Component module, instead of the icalendar alias. Compare:

- https://icalendar--1055.org.readthedocs.build/en/1055/reference/api/icalendar.html#icalendar.Component.from_ical
- https://icalendar--1055.org.readthedocs.build/en/1055/reference/api/icalendar.cal.component.html#icalendar.cal.component.Component.from_ical

```suggestion
        Wraps :meth:`Component.from_ical() <icalendar.cal.component.Component.from_ical>` with
```

src/icalendar/cal/calendar.py
```diff
@@ -85,6 +85,18 @@ def example(cls, name: str = "example") -> Calendar:
...
+        timezone forward-reference resolution and VTIMEZONE caching.
+
+        Args:
+            st: iCalendar data as bytes or string
+            multiple: If True, returns list. If False, returns single calendar.
```
```suggestion
            multiple: If ``True``, returns list. If ``False``, returns single calendar.
```

src/icalendar/cal/component.py
```diff
@@ -363,7 +363,20 @@ def property_items(
 
     @classmethod
     def from_ical(cls, st, multiple: bool = False) -> Self | list[Self]:  # noqa: FBT001
-        """Populates the component recursively from a string."""
+        """Parse iCalendar data into component instances.
+
+        Handles standard and custom components (X-*, IANA-registered).
```
stevepiercy: Sphinx renders the * as a link to a non-existent id. Making it an inline literal avoids that bug. I've never seen that before!

https://icalendar--1055.org.readthedocs.build/en/1055/reference/api/icalendar.cal.component.html#icalendar.cal.component.Component.from_ical

```suggestion
        Handles standard and custom components (``X-*``, IANA-registered).
```

src/icalendar/cal/component.py
```diff
@@ -363,7 +363,20 @@ def property_items(
...
+
+        Args:
+            st: iCalendar data as bytes or string
+            multiple: If True, returns list. If False, returns single component.
```
```suggestion
            multiple: If ``True``, returns list. If ``False``, returns single component.
```

src/icalendar/cal/component.py
```diff
@@ -363,7 +363,20 @@ def property_items(
...
+
+        Returns:
+            Component or list of components
+
+        See Also:
```
stevepiercy, line 426: Oh, cool! This is Napolean black magic that supports admonitions in this format instead of the clumsy `.. seealso::` syntax.

src/icalendar/cal/component_factory.py
```diff
@@ -24,9 +24,23 @@ class ComponentFactory(CaselessDict):
...
+
+        >>> factory = ComponentFactory()
+        >>> custom_class = factory.get_component_class('X-VENDOR')
+        >>> custom_class()
+        X-VENDOR({})
+
     If a component class is not yet supported, it can be either created
     using :meth:`get_component_class` or added manually as a subclass of
     :class:`~icalendar.Component`.
+
```
stevepiercy: Merge into one paragraph.

```suggestion
```

src/icalendar/cal/component_factory.py
```diff
@@ -24,9 +24,23 @@ class ComponentFactory(CaselessDict):
         >>> event_class()
         VEVENT({})
 
+    **Custom Components:**
+
+    Automatically creates component classes for unknown names (X-components,
+    IANA-components). Custom components are never dropped (RFC 5545).
```
stevepiercy:
```suggestion
    IANA-components). Custom components are never dropped per :rfc:`5545`.
```

src/icalendar/cal/component_factory.py
```diff
@@ -24,9 +24,23 @@ class ComponentFactory(CaselessDict):
         >>> event_class()
         VEVENT({})
 
+    **Custom Components:**
+
```
stevepiercy: This is not good semantic markup because it renders as `<strong>Custom Components:</strong>`. It should be a heading or removed. However, headings in docstrings are limited to the options listed in https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections.

```suggestion
```

src/icalendar/cal/component_factory.py
```diff
@@ -24,9 +24,23 @@ class ComponentFactory(CaselessDict):
...
+    Automatically creates component classes for unknown names (X-components,
```
stevepiercy:
```suggestion
    Automatically creates custom component classes for unknown names (X-components,
```

docs/how-to/custom-components.rst
stevepiercy, line 1: I'm going to do a direct edit to this file for compliance with our style guide, soon to be published from another PR in draft status.

src/icalendar/cal/component.py
SashankBhamidi: 😂😂😂

docs/how-to/custom-components.rst
SashankBhamidi: Could you do this after the merge? Otherwise the draft will run into conflicts, which just adds unnecessary work.

src/icalendar/cal/component.py
SashankBhamidi: I've been giggling for 5 minutes now

docs/how-to/custom-components.rst
stevepiercy: Sorry, didn't see this note until I had already jumped in.

stevepiercy: But there shouldn't be any merge conflicts, because I edited only this one file. The others are suggestions on other files.

SashankBhamidi: I thought you were going to open a new PR instead of working on this one, this works too, looks good.

SashankBhamidi:
@stevepiercy Thanks for the editorial polish and the Vale fix! All your review suggestions have been addressed and pushed. 

Regarding the documentation examples, I tested all 7 examples in the file and they all work correctly with their imports. The `NameError` you encountered was likely from copy-pasting just the component parsing line without the `from icalendar import Component` line above it. 

Your TokenIgnores solution works perfectly. I agree it's a sledgehammer approach, but given the variety of characters that can appear in reStructuredText roles, it's the right call. 

I will wait for a technical review now.

docs/how-to/custom-components.rst
stevepiercy: Oh, yeah, that's right. Sorry, my brain was in editorial mode, not developer mode.

stevepiercy (approved):
Docs and change log looks good, all links resolve to proper targets. Thank you!

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,303 @@
...
+                    raise ValueError(f"Missing required property: {prop}")
+
+        def get_vendor_id(self):
+            """Convenience method for vendor ID."""
+            return self.get("X-VENDOR-ID")
+
+    # Register with the singleton factory
+    # First call to get_component_class() initializes the factory
+    Component.get_component_class("VEVENT")  # Ensure factory exists
+    Component._components_factory.add_component_class(XVendorComponent)
```
niccokunzmann: This is private and should not be accessed. There are methods specifically so that you can override them. Use those.

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,303 @@
...
+                if prop not in self:
+                    raise ValueError(f"Missing required property: {prop}")
+
+        def get_vendor_id(self):
+            """Convenience method for vendor ID."""
+            return self.get("X-VENDOR-ID")
+
+    # Register with the singleton factory
+    # First call to get_component_class() initializes the factory
+    Component.get_component_class("VEVENT")  # Ensure factory exists
```
niccokunzmann: You can subclass and override this.
Then, use your component class to parse this or create your own ComponentFactory inside that.

SashankBhamidi:
I didn't realize `_components_factory` was private, I saw it worked and documented it. My bad.

Thanks for pointing this out!

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,321 @@
...
+    >>> cal = Calendar.from_ical(calendar_data)
+    >>> len(cal.subcomponents)
+    1
+
+Create custom component subclasses
+==================================
+
+While the dynamic component creation works for most cases, you can create explicit component subclasses for custom components that need special behavior.
+
+Subclass Component and override ``get_component_class()``:
```
niccokunzmann: You do not do that.

Here, you have a chance to do the following:

1. Imagine the loveliest interface you would like to use.
2. Use it in the code
3. implement it
4. add tests for edge cases

For that, you MUST ignore any knowledge about implementation.

E.g. I would like to use icalendar:

```
from icalendar import Component

class XMyComponent(Component):
    # ...

Component.register(XMyComponent)  # check for name, document duplication
```

That will allow you to use the component class everywhere in your program.

Discussion: Is this needed? Maybe, but rare.

The problem with using any `_` attributes, functions and methods: I added these because it is a quick fix. For providing the best API, it is bad to expose them.

Also in use here and generally should be avoided if possible: https://en.wikipedia.org/wiki/Singleton_pattern

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,321 @@
...
+                    raise ValueError(f"Missing required property: {prop}")
+
+        def get_vendor_id(self):
+            """Convenience method for vendor ID."""
+            return self.get("X-VENDOR-ID")
+
+    class CustomComponent(Component):
+        """Component subclass with custom factory."""
+
+        _components_factory = ComponentFactory()
```
niccokunzmann: You must not document to use private attributes - my opinion.

Please extend the interface to serve the purpose most exquisitely.

niccokunzmann:
From my perspective, this can be merged on a purely functional base. It adds a bit of conflict but also there is a chance to create a really nice interface to use. I will wait for what you would like to do about it. It is better than what was there before.

CHANGES.rst
```diff
@@ -27,7 +27,8 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- ...
+- Clarified custom component (X-* and IANA-registered) parsing behavior through enhanced documentation and comprehensive how-to guide. Custom components are automatically handled by the library with no special configuration required. See `Issue #432 <https://github.com/collective/icalendar/issues/432>`_.
+- Reorganized custom component tests into dedicated ``test_custom_components.py`` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
```
stevepiercy: I missed this on my previous pass.

```suggestion
- Reorganized custom component tests into a dedicated :file:`test_custom_components.py` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
```

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,323 @@
...
+    >>> custom.add("x-theme", "dark")
+    >>> cal.add_component(custom)
+    >>> b"BEGIN:X-MYAPP-SETTINGS" in cal.to_ical()
+    True
+
+
+Related content
+===============
+
+:class:`icalendar.Component`
```
stevepiercy: I forgot to verify all these links, and found a few to resolve.

```suggestion
:class:`~icalendar.cal.component.Component`
```

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,323 @@
...
+    Base component class
+
+:class:`icalendar.cal.component_factory.ComponentFactory`
+    Component factory
+
+:meth:`icalendar.Component.from_ical`
```
stevepiercy: Also broken link.

```suggestion
:meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`
```

docs/how-to/custom-components.rst
```diff
@@ -0,0 +1,323 @@
...
+    Parse components
+
+:meth:`icalendar.Calendar.from_ical`
```
stevepiercy:
```suggestion
:meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`
```

src/icalendar/cal/component_factory.py
stevepiercy: This really should be one paragraph, not two. It's weird to have two separate paragraphs for one subject.

src/icalendar/cal/component.py
```diff
@@ -97,6 +97,40 @@ def get_component_class(cls, name: str) -> type[Component]:
...
 
+    @classmethod
+    def register(cls, component_class: type[Component]) -> None:
+        """Register a custom component class.
+
+        Args:
+            component_class: Component subclass to register. Must have a ``name`` attribute.
+
+        Raises:
+            ValueError: If component_class has no ``name`` attribute.
```
```suggestion
            ValueError: If ``component_class`` has no ``name`` attribute.
```

src/icalendar/cal/component.py
```diff
@@ -97,6 +97,40 @@ def get_component_class(cls, name: str) -> type[Component]:
...
+
+        Example:
+            >>> from icalendar import Component
+            >>> class XExample(Component):
+            ...     name = "X-EXAMPLE"
+            ...     def custom_method(self):
+            ...         return "custom"
+            >>> Component.register(XExample)
+        """
+        if not hasattr(component_class, 'name') or component_class.name is None:
```
```suggestion
        if not hasattr(component_class, "name") or component_class.name is None:
```

src/icalendar/cal/component_factory.py
```diff
@@ -24,9 +24,19 @@ class ComponentFactory(CaselessDict):
...
+    IANA-components). Custom components are never dropped per :rfc:`5545`.
+
+    .. code-block:: pycon
+
+        >>> factory = ComponentFactory()
+        >>> custom_class = factory.get_component_class('X-VENDOR')
+        >>> custom_class()
+        X-VENDOR({})
+
+    If a component class is not yet supported, it can be either created using :meth:`get_component_class` or added manually as a subclass of :class:`~icalendar.Component`.
```
stevepiercy: Make it link.

```suggestion
    If a component class is not yet supported, it can be either created using :meth:`get_component_class` or added manually as a subclass of :class:`~icalendar.cal.component.Component`.
```

stevepiercy (changes requested):
I double-checked the links and found several broken ones. I also found some minor style and grammar stuff.

SashankBhamidi:
Yes merge this, does not show me an option to merge from my phone

stevepiercy:
Yeah, when you enable auto-merge, it hides the checkbox "Merge without waiting for requirements to be met (bypass rules)". Also since @niccokunzmann already approved in writing, but not by selecting "Approve" in a review, automerge would never work. Anyway, I disabled automerge, checked the box, and merged.

SashankBhamidi:
I dont have permissions to bypass and merge anyway so only thing I can do is click auto-merge :) 

Github mobile app needs updates

PR #1057, opened 2025-12-27
Fix CaselessDict.__eq__ crash with non-dict types

### LINKED ISSUE #1016: CaselessDict.__eq__ crashes when comparing with non-dict types
## Describe the bug

`CaselessDict.__eq__` crashes when comparing with non-dict types.

## To reproduce

```python
from icalendar.caselessdict import CaselessDict

d = CaselessDict()
d['test'] = 1
d == "string"  # AttributeError: 'str' object has no attribute 'items'
```

## Expected behavior

Should return `False` instead of crashing.

## Root cause

```python
def __eq__(self, other: object) -> bool:
    return self is other or dict(self.items()) == dict(other.items())
```

## Proposed fix

```python
def __eq__(self, other: object) -> bool:
    if not isinstance(other, dict):
        return NotImplemented
    return self is other or dict(self.items()) == dict(other.items())
```

---

Found during review of PR #1015.

SashankBhamidi:
Fixes #1016

Returns NotImplemented when comparing CaselessDict with non-dict types instead of raising AttributeError.

Changes:
- Modified CaselessDict.__eq__ to check isinstance(other, dict)
- Added comprehensive tests for equality/inequality with non-dict types
- Updated CHANGES.rst

----
📚 Documentation preview 📚: https://icalendar--1057.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,7 +42,7 @@ New features
 Bug fixes
 ~~~~~~~~~
 
-- Fixed :py:class:`CaselessDict.__eq__ <icalendar.caselessdict.CaselessDict>` to return ``NotImplemented`` when comparing with non-dict types instead of raising ``AttributeError``. See `Issue #1016 <https://github.com/collective/icalendar/issues/1016>`_.
+- Fixed :meth:`icalendar.caselessdict.CaselessDict.__eq__` to return ``NotImplemented`` when comparing with non-dict types instead of raising ``AttributeError``. See `Issue #1016 <https://github.com/collective/icalendar/issues/1016>`_.
 - Link ``timedelta`` to :py:class:`datetime.timedelta` in the Python standard library documentation. See `Issue 951 <https://github.com/collective/icalendar/issues/951>`_.
 
 Documentation


src/icalendar/tests/test_unit_caselessdict.py
```diff
@@ -125,3 +125,40 @@ def test_CaselessDict(self):
...
+
+    def test_eq_with_non_dict_types(self):
+        """Test that CaselessDict.__eq__ handles non-dict comparisons correctly."""
+        CaselessDict = icalendar.caselessdict.CaselessDict
+
+        d = CaselessDict()
+        d['test'] = 1
+
+        # Should return False for non-dict types, not crash
+        self.assertFalse(d == "string")
```
niccokunzmann, line 137: We use pytest. Why do you use the unittest API? Pytest will generate nice messages for an assert statement.

SashankBhamidi: I followed the existing `unittest.TestCase` pattern in this file. I can convert to pytest assertions if you'd like?

niccokunzmann: No need! I just wondered.

PR #1058, opened 2025-12-27
Fix CATEGORIES comma escaping (#127)

### LINKED ISSUE #127: string conversion of summary field converts "," to "\,"
when the calendar event summary field contains a string with a comma, for example
"key=value1,value2"
the to_ical() function converts the above string to:  "key=value1\,value2"

Is this a bug or a feature?

thanks - bob s.

SashankBhamidi:
Fixes #127

Categories with commas now round-trip correctly.

- Add split_on_unescaped_comma() helper
- Special handling in Component.from_ical()
- Update vCategory to accept pre-split lists

----
📚 Documentation preview 📚: https://icalendar--1058.org.readthedocs.build/


CHANGES.rst
```diff
@@ -43,6 +43,7 @@ Bug fixes
 ~~~~~~~~~
 
 - Link ``timedelta`` to :py:class:`datetime.timedelta` in the Python standard library documentation. See `Issue 951 <https://github.com/collective/icalendar/issues/951>`_.
+- Fix round-trip parsing of :class:`~icalendar.prop.vCategory` (CATEGORIES property) when category values contain commas. Categories like ``'Meeting, John'`` now correctly survive to_ical() → from_ical() round-trips instead of being split into multiple categories. Added :func:`~icalendar.parser.split_on_unescaped_comma` helper function. See `Issue #127 <https://github.com/collective/icalendar/issues/127>`_.
```
stevepiercy: Avoid symbols, preferring narrative style. Link to methods, and ensure that I got the syntax correct by looking at the PR preview.

```suggestion
- Fix round-trip parsing of :class:`~icalendar.prop.vCategory` (CATEGORIES property) when category values contain commas. Categories like ``'Meeting, John'`` now correctly survive round trips between :meth:`~icalendar.Component.to_ical` and :meth:`~icalendar.Component.from_ical` instead of being split into multiple categories. Added :func:`~icalendar.parser.split_on_unescaped_comma` helper function. See `Issue #127 <https://github.com/collective/icalendar/issues/127>`_.
```

src/icalendar/parser.py
```diff
@@ -547,6 +547,57 @@ def unescape_backslash(val: str):
     )
 
 
+def split_on_unescaped_comma(text: str) -> list[str]:
+    r"""Split text on unescaped commas and unescape each part.
```
stevepiercy, line 551: Typo?

```suggestion
    """Split text on unescaped commas and unescape each part.
```

src/icalendar/parser.py
```diff
@@ -547,6 +547,57 @@ def unescape_backslash(val: str):
...
+    After splitting, unescapes backslash sequences in each part.
+
+    Args:
+        text: Text with potential escaped commas (e.g., "foo\\, bar,baz")
+
+    Returns:
+        List of unescaped category strings
+
+    Examples:
+        >>> from icalendar.parser import split_on_unescaped_comma
```
stevepiercy: Explicitly use the pycon lexer, and indent all other subsequent lines in the Python console examples.

```suggestion
    .. code-block:: pycon

            >>> from icalendar.parser import split_on_unescaped_comma
```

stevepiercy (changes requested):
Minor syntax tweaks. Still needs a technical review. Thank you!

src/icalendar/parser.py
SashankBhamidi: The `r"""` is a raw string literal in Python, which is the correct way to write docstrings that contain backslashes (like `\\,`). Without the r, we'd have to double-escape everything.

src/icalendar/parser.py
SashankBhamidi: Done in 57f06d0

CHANGES.rst
```diff
@@ -43,6 +43,7 @@ Bug fixes
 ~~~~~~~~~
 
 - Link ``timedelta`` to :py:class:`datetime.timedelta` in the Python standard library documentation. See `Issue 951 <https://github.com/collective/icalendar/issues/951>`_.
+- Fix round-trip parsing of :class:`~icalendar.prop.vCategory` (CATEGORIES property) when category values contain commas. Categories like ``'Meeting, John'`` now correctly survive round trips between :meth:`~icalendar.Component.to_ical` and :meth:`~icalendar.Component.from_ical` instead of being split into multiple categories. Added :func:`~icalendar.parser.split_on_unescaped_comma` helper function. See `Issue #127 <https://github.com/collective/icalendar/issues/127>`_.
```
stevepiercy: Nope, I was off. Even though it links, it goes to the long list for the icalendar alias, instead of `Component`. Let's try this one.

```suggestion
- Fix round-trip parsing of :class:`~icalendar.prop.vCategory` (CATEGORIES property) when category values contain commas. Categories like ``'Meeting, John'`` now correctly survive round trips between :meth:`Component.to_ical <icalendar.cal.component.Component.to_ical>` and :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>` instead of being split into multiple categories. Added :func:`~icalendar.parser.split_on_unescaped_comma` helper function. See `Issue #127 <https://github.com/collective/icalendar/issues/127>`_.
```

stevepiercy (changes requested):
Better link target. Otherwise docs look good. Thank you!

src/icalendar/parser.py
```diff
@@ -547,6 +547,58 @@ def unescape_backslash(val: str):
...
+
+        >>> from icalendar.parser import split_on_unescaped_comma
+        >>> split_on_unescaped_comma(r"foo\, bar,baz")
+        ['foo, bar', 'baz']
+        >>> split_on_unescaped_comma("a,b,c")
+        ['a', 'b', 'c']
+        >>> split_on_unescaped_comma(r"a\,b\,c")
+        ['a,b,c']
+        >>> split_on_unescaped_comma(r"Work,Personal\,Urgent")
+        ['Work', 'Personal,Urgent']
```
stevepiercy: The `Examples:` heading should remain. Sorry, I wasn't clear.

```suggestion
    Examples:

        .. code-block:: pycon
    
            >>> from icalendar.parser import split_on_unescaped_comma
            >>> split_on_unescaped_comma(r"foo\, bar,baz")
            ['foo, bar', 'baz']
            >>> split_on_unescaped_comma("a,b,c")
            ['a', 'b', 'c']
            >>> split_on_unescaped_comma(r"a\,b\,c")
            ['a,b,c']
            >>> split_on_unescaped_comma(r"Work,Personal\,Urgent")
            ['Work', 'Personal,Urgent']

```

stevepiercy (commented):
One more miscommunication to handle.

niccokunzmann:
Is there an issue for this?

src/icalendar/prop/__init__.py
```diff
@@ -854,7 +854,7 @@ def __init__(
         self.params = Parameters(params)
 
     def __iter__(self):
-        return iter(vCategory.from_ical(self.to_ical()))
+        return iter(self.cats)
```
niccokunzmann, line 857: That looks good!

src/icalendar/tests/test_parsing.py
```diff
@@ -260,6 +260,38 @@ def test_unescape_char():
...
+    from icalendar.parser import split_on_unescaped_comma
+
+    # Simple case
+    assert split_on_unescaped_comma("a,b,c") == ["a", "b", "c"]
+
+    # Escaped comma
+    assert split_on_unescaped_comma("a\\,b,c") == ["a,b", "c"]
+
+    # Multiple escaped commas
+    assert split_on_unescaped_comma("a\\,b\\,c") == ["a,b,c"]
```
niccokunzmann, line 253: A test could be added for the categories `"\,"` and `"\\,"` to make sure that escaped escape sequences work.

niccokunzmann (commented):
I am unsure about this. This adds a feature or a bug or changes ambiguous behaviour.

Which application does allow commas in the categories?

niccokunzmann:
The issue related to this is: https://github.com/collective/icalendar/issues/127

How to justify this:

- Can you show that this is in line with the RFC? [SPECIFICATION]
- Which applications allow commas in categories? [COMPATIBILITY]
- Who benefits from this? Can you show code that uses icalendar out there that requires this change? [USABILITY]

`\` escaping is used in programming languages but is not part of the RFC as far as I remember. `python-icalendar` being able to round-trip is not a justification to change what people expect or what other libraries do.

This is a widely used library and this changes behaviour. Why is this change correct?

niccokunzmann:
Hm it seems old issues are sometimes open because they are not straightforward to address, because there is ambiguity or because there was not enough discussion. You can create pull requests for those but they are more like a contribution for the discussion and I have to think about merging them because I would maybe even revert them if they do not address a problem but introduce new problems.

SashankBhamidi:
No longer solid huh, thanks Nicco

Your merge to main lowkey was made with ♥️

Issue #994, opened 2025-11-14
[BUG] Cross-component links are broken sometimes


niccokunzmann:
## Describe the bug

When I go to [Availability.new()](https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.cal.availability.html#icalendar.cal.availability.Availability.new), several links across classes are broken: [Edit: this example as since been fixed in #1064, and is an example of what to do for the remaining broken links.]

<img width="496" height="115" alt="Image" src="https://github.com/user-attachments/assets/0e40351f-9d1d-4d70-bd10-b8513f956606" />

## Expected behavior

I would like all links to attributes and functions in the documented classes to be clickable.

Sphinx has been configured with the following setting. See https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-autoclass-inherited-members.

```python
autodoc_default_options = {
    ...
    "inherited-members": "CaselessDict",
```

This brings in dozens of attributes in the documentation in the right-hand sidebar.

## To fix

1. [Build the documentation locally](https://icalendar.readthedocs.io/en/latest/contribute/documentation/build-check.html).

2. Search the built HTML documentation to find attributes and functions that are inline literals but not hyperlinked, using a regular expression such as the following for attributes or functions, respectively.

   ```regex
   [^>]<code class="xref py py-attr docutils literal notranslate"><span class="pre">
   ```

   ```regex
   [^>]<code class="xref py py-func docutils literal notranslate"><span class="pre">
   ```

   If the attribute is hyperlinked, then it will be preceded by an opening `<a>` tag and its attributes. We want to find anything _except_ `>`, followed by the HTML markup.

3. Use a short form of the object reference to link to the attribute or function. Example from `icalendar.cal.availability.Availability.new()` (PR #1064):

   ```python
   comments: The :attr:`comments` of the availability.
   ```

stevepiercy:
While working on #1716, I stumbled across Sphinx's nitpicky mode. To run it:

```shell
make clean livehtml SPHINXOPTS="-n"
```

You'll see a list of all the broken cross-references, and a few more.

I suggest that we run Sphinx in nitpicky mode in CI after this issue is resolved.

PR #1064, opened 2025-12-31
Fix link of `uid` via icalendar.attr.uid_property import

stevepiercy:
- Partially addresses #994.

## Description

Resolve broken cross-reference links in `new()` methods.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -121,6 +121,7 @@ Documentation
 - Fixed broken links in ``docs/how-to/usage.rst`` documentation. Part of :issue:`1158`
 - Changed Sphinx configuration to inherit from ``CaselessDict``, displaying all of its class members in the in-page navigation.
   Fixed broken cross-component links in ``Availability.new()``. :issue:`994`
+- Configured Sphinx to append the docstring from the ``__init__`` method into its class docstring. :issue:`1156`
 
 
 7.0.0a3 (2025-12-19)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -166,7 +166,7 @@ apidoc_modules = [
         },
     }
 ]
-
+autoclass_content = "both"
 
 # -- sphinx.ext.autodoc options -------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
@@ -175,7 +175,6 @@ autodoc_default_options = {
     "inherited-members": "CaselessDict",
     "members": True,
     "show-inheritance": True,
-    "special-members": "__init__",
     "undoc-members": True,
 }
 
--- a/docs/contribute/documentation/style-guide.rst
+++ b/docs/contribute/documentation/style-guide.rst
@@ -255,20 +255,27 @@ All items should terminate with a period.
 
 The following docstrings items are the most frequently used in icalendar, although they may be others.
 
+.. important::
+
+    When writing docstrings for both a class and its ``__init__`` method, consider the following points.
+
+    -   A class's docstring must not contain ``Attributes``, ``Parameters``,  ``Returns``, or ``Raises`` sections.
+    -   ``Parameters``, ``Returns``, and ``Raises`` sections may be in the ``__init__`` method's docstring.
+    -   The ``__init__`` method's docstring gets automatically appended to its class's docstring by Sphinx.
+        See :pull:`1164` for an example.
+    -   The one-line summary in the ``__init__`` method's docstring should serve as a narrative bridge from the ending of its class docstring.
+
 Summary
     Docstrings must begin with a one-line summary of the Python object, terminated by a period.
 
 Description
     When the one-line summary is insufficient to describe the Python object, then write an overall description of what it does, without going into details of how it does it.
     Leave implementation details to the code, and optionally inline code comments.
+
     Separate the summary and description with a blank line.
 
 ``Attributes``
     Each attribute should consist of its name, type, and a brief description.
-    
-    .. note::
-
-        If you know how to avoid manually entering the type for an attribute, please see :issue:`1156`.
 
 ``Parameters``
     Each parameter should consist of its name and a brief description.
--- a/src/icalendar/alarms.py
+++ b/src/icalendar/alarms.py
@@ -42,16 +42,6 @@ class AlarmTime:
     An AlarmTime instance combines an alarm component with its resolved
     trigger time and additional state information, such as acknowledgment
     and snoozing.
-
-    Attributes:
-        alarm (Alarm): The underlying VALARM component.
-        trigger (datetime): The computed trigger time.
-        acknowledged_until (datetime | None): Time in UTC until which the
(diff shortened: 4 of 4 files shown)


niccokunzmann:
Looks like a fix. Is uid part of the superclass Component?

stevepiercy:
I found this, but I don't know whether that makes it 🦸 . hehe.

https://github.com/collective/icalendar/blob/6a8e9743170370f215fd02574f1e922a6a7850d7/src/icalendar/cal/component.py#L9-L20

https://github.com/collective/icalendar/blob/6a8e9743170370f215fd02574f1e922a6a7850d7/src/icalendar/cal/component.py#L688-L688

For the other attrs that don't link, should we do a similar treatment?

- comments
- components
- concepts
- created
- last_modified
- links
- refids
- related_to
- stamp

Also this issue turns out to be less about documentation and more about code and feature compliance with the RFCs. I get concerned about tests as well. I'd prefer to focus on documentation, and let a first-timer pick this up by giving enough direction for them to run with it.

stevepiercy:
It looks like we had crossed GitHub comments, and I just noticed you opened https://github.com/collective/icalendar/issues/1066.

The syntax you suggested doesn't work. The target must always resolve to a fully qualified dotted Python path.

For this one, I'd go with markup that indicates in which class `uid` resides to disambiguate the context. If it displays only `uid`, then it can mislead the reader to think it's in the class `Availability`.

```python
:attr:`icalendar.cal.component.Component.uid`
```

or

```python
:attr:`Component.uid <icalendar.cal.component.Component.uid>`
```

At this point, I'm not sure how to proceed. I pushed a commit so you can see the change.

niccokunzmann:
Hm. I don't mind either way. One looks nicer and the other one is compact in code. They are both functionally valid. So, please choose how you like the docs.

I start thinking that it is more compact then to rename these properties if they have multiple imports so they can be imported in the class like `from icalendar.cal.properties import uid`.
If all these attributes are actually on the page of the respective component that would be quite nice for the reader, right? Docs is important and that might be quite good. What do you think?

What about consistency and other methods like to_ical? A style can be tried out here and then an issue created for first timers to make it consistent across all classes.

Is there a way some inherited functions and atrributes can be displayed? Is that desirable?

stevepiercy:
> Hm. I don't mind either way. One looks nicer and the other one is compact in code. They are both functionally valid. So, please choose how you like the docs.

Just to be clear, I'm saying that these two hyperlinks are not functionally valid _as hyperlinks_ after Sphinx renders it to HTML. As Python code, they may be functionally valid, either with or without the two lines of the import and the variable assignment.

- "Without import" goes to current development docs without any `Availability.uid` in the right-hand navigation or in the page.
 
  Screenshot 2026-01-01 at 2 01 13 PM

- "With import" goes to pull request preview docs with the `Availability.uid` in the right-hand navigation _and_ in the page.

  Screenshot 2026-01-01 at 2 02 29 PM
  Screenshot 2026-01-01 at 2 03 31 PM

When there's a change in code, I don't think this is a personal preference, but an intentional design decision. That goes beyond mere appearance. I don't feel it's right for me to make a unilateral choice here.

> I start thinking that it is more compact then to rename these properties if they have multiple imports so they can be imported in the class like `from icalendar.cal.properties import uid`. If all these attributes are actually on the page of the respective component that would be quite nice for the reader, right? Docs is important and that might be quite good. What do you think?

I think it's confusing to write reStructuredText `role`s for Python objects in icalendar.

To back up a bit, I first noticed this when I resolved all the "duplicate object" warnings emitted by Sphinx in a separate recent PR. Even Sphinx was confused, ha! That has been cleared up, specifically by always using a full dotted Python path in the target. That leaves us with the question of what should be the actual target, either the superclass or subclass?

I don't feel qualified to answer that question by myself. I don't actually use or develop in icalendar, although I've absorbed a basic understanding from just writing docs. That's where your experience, and that of other users and developers, is extremely valuable.

I think it would be reasonable to say that if I were writing code, and I couldn't find `Availability.uid` in the docs, I would assume it's not possible to use. I'd first look for it in the in-page navigation, and find nothing. Then I'd search the page, and see no separate entry. I would find `uid` in the page, buried in the `Availability.new` method.

Now imagine doing the same thing for an attribute that does appear, such as `busy_type`. My brain notices there's a difference, but doesn't know why, then shrugs it off as something to come back to later.

This difference causes a small dissonance for me as a documentarian, but this may be acceptable to users and developers.

> What about consistency and other methods like to_ical? A style can be tried out here and then an issue created for first timers to make it consistent across all classes.

Do you mean where `to_ical` in https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.prop.html is applied across all relevant classes? That's really clear to me and doesn't short-circuit my brain.
 
> Is there a way some inherited functions and atrributes can be displayed? Is that desirable?

I'm not sure what you have in mind. Do you envision something like a list of inherited objects in a separate section, some way of marking them inline where they appear in the page, or something else? Here are some options.

1. Try [`:inherited-members:`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-autoclass-inherited-members) and configure it in `make apidoc`.
2. Try [Sphinx AutoAPI](https://github.com/readthedocs/sphinx-autoapi) instead of apidoc and autodoc.
3. Add pretty pictures with https://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html.
4. Customize the source .rst files to display whatever we want, but the trade-off would be the loss of automatic generation of the .rst files from Python source files via `make apidoc`.
5. Ask in the Python documentation Discord server, but I'd like to better understand what you seek when I present the question.

stevepiercy:
@niccokunzmann this PR is ready for review.

I also updated the description in issue #994 to direct first-timers to pick it up and run with it.

CHANGES.rst
```diff
@@ -118,6 +118,9 @@ Documentation
   See `smartquotes_action <https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes_action>`_.
 - Fix incorrect return type annotation in :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`. :issue:`1141`
 - Fixed broken links in ``docs/how-to/usage.rst`` documentation. Part of :issue:`1158`
+- Changed Sphinx configuration to inherit from ``CaselessDict`` and display all of its members in the in-page navigation.
```
```suggestion
- Changed Sphinx configuration to inherit from ``CaselessDict``, displaying all of its class members in the in-page navigation.
```

I don't really mind if this is changed.

niccokunzmann (dismissed):
This looks nice. I had a look at the docs. https://icalendar--1064.org.readthedocs.build/en/1064/reference/api/icalendar.cal.availability.html#icalendar.cal.availability.Availability.created

CHANGES.rst
stevepiercy: The suggestion you made didn't make sense grammatically, so I altered it slightly.

niccokunzmann (approved):

PR #1067, opened 2026-01-03
Added upgrade guide

stevepiercy:
## Closes issue

- [X] Closes #997

## Description

- Added upgrade guide.
- Add anchor that doesn't resolve on GitHub to ``linkcheck_ignore``
- Fix ``<unknown>:1057: SyntaxWarning: invalid escape sequence '\,'`` warning that Sphinx emitted. What a stinker!

## Checklist

- [X] I've added a change log entry to `CHANGES.rst`.
- [X] I've added or updated tests if applicable.
- [X] I've run and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [X] I've added or edited documentation, both as docstrings to be rendered in the API documentation and narrative documentation, as necessary.

Please carefully review the examples and the suggested deprecations, and make inline suggestions for improvement.

----
📚 Documentation preview 📚: https://icalendar--1067.org.readthedocs.build/

PR #1071, opened 2026-01-06
Make typing-extensions optional and fix is_pytz_dt type hint

jkseppan:
Change is_pytz_dt return type from TypeIs to TypeGuard: if it returns
false, the input could still be a datetime so we shouldn't narrow.

Move typing-extensions from core dependencies to test extra.

Import TypeIs and TypeGuard from typing_extensions depending on
Python version, as recommended by

https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module

Importing from typing_extensions upon ImportError did not work in
my testing, presumably because mypy doesn't really import the name
but has special handling.

----
📚 Documentation preview 📚: https://icalendar--1071.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,13 +31,13 @@ Minor changes
 - Created an :meth:`~cal.alarm.Alarm.example` method for the :class:`~cal.alarm.Alarm` component. See `Issue 743`_.
 - Clarified custom component (X-* and IANA-registered) parsing behavior through enhanced documentation and comprehensive how-to guide. Custom components are automatically handled by the library with no special configuration required. See `Issue #432 <https://github.com/collective/icalendar/issues/432>`_.
 - Reorganized custom component tests into a dedicated :file:`test_custom_components.py` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
-- The `typing-extensions` dependency on Python < 3.13 is now optional, part of the `test` extra.
+- The ``typing-extensions`` dependency on Python < 3.13 is now optional, part of the ``test`` extra.
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
 - :meth:`Component.decoded` now returns a string instead of bytes for text properties.
-- The :func:`icalendar.tools.is_pytz_dt` return value is now hinted as `TypeGuard[datetime]`, not `TypeIs[datetime]`, since it returning false should not allow narrowing as non-datetime.
+- The :func:`icalendar.tools.is_pytz_dt` return value is now hinted as ``TypeGuard[datetime]``, not ``TypeIs[datetime]``, since returning ``False`` should not allow narrowing it as non-datetime.
 
 New features
 ~~~~~~~~~~~~


src/icalendar/compatibility.py
```diff
@@ -16,8 +18,24 @@
...
+if TYPE_CHECKING:
+    if sys.version_info >= (3, 10):
+        from typing import TypeGuard
+    else:
+        from typing_extensions import TypeGuard
+
+    if sys.version_info >= (3, 13):
+        from typing import TypeIs
+    else:
+        from typing_extensions import TypeIs
```
stevepiercy, line 30: This doesn't look right to me. Let's say you have Python 3.12.

Line 23 yields `from typing import TypeGuard`.

But line 30 overrides that with `from typing_extensions import TypeIs`.

I'm not clear what you intended, but I think you intend to have upper bounds on the Python version in the `else` statements.

CHANGES.rst
```diff
@@ -31,11 +31,13 @@ Minor changes
 - Created an :meth:`~cal.alarm.Alarm.example` method for the :class:`~cal.alarm.Alarm` component. See `Issue 743`_.
 - Clarified custom component (X-* and IANA-registered) parsing behavior through enhanced documentation and comprehensive how-to guide. Custom components are automatically handled by the library with no special configuration required. See `Issue #432 <https://github.com/collective/icalendar/issues/432>`_.
 - Reorganized custom component tests into a dedicated :file:`test_custom_components.py` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
+- The `typing-extensions` dependency on Python < 3.13 is now optional, part of the `test` extra.
```
stevepiercy: rST uses double backticks for inline literals, whereas md uses single.

```suggestion
- The ``typing-extensions`` dependency on Python < 3.13 is now optional, part of the ``test`` extra.
```

CHANGES.rst
```diff
@@ -31,11 +31,13 @@ Minor changes
...
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
 - :meth:`Component.decoded` now returns a string instead of bytes for text properties.
+- The :func:`icalendar.tools.is_pytz_dt` return value is now hinted as `TypeGuard[datetime]`, not `TypeIs[datetime]`, since it returning false should not allow narrowing as non-datetime.
```
stevepiercy: Some grammar and rST markup.

```suggestion
- The :func:`icalendar.tools.is_pytz_dt` return value is now hinted as ``TypeGuard[datetime]``, not ``TypeIs[datetime]``, since returning ``False`` should not allow narrowing it as non-datetime.
```

src/icalendar/compatibility.py
stevepiercy: Oh, nm. This is OK. I had to look up that:

- TypeIs is new in 3.13 https://docs.python.org/3/library/typing.html#typing.TypeIs
- TypeGuard is new in 3.10 https://docs.python.org/3/library/typing.html#typing.TypeGuard

jkseppan:
Thanks for sticking with it! I agree that it's a bit dense on first read. Applied the grammar and markup improvements.

CHANGES.rst
```diff
@@ -31,13 +31,13 @@ Minor changes
...
 - Reorganized custom component tests into a dedicated :file:`test_custom_components.py` file with expanded test coverage for :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`, :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, and :class:`~icalendar.cal.component_factory.ComponentFactory` usage. See `Issue #433 <https://github.com/collective/icalendar/issues/433>`_.
-- The `typing-extensions` dependency on Python < 3.13 is now optional, part of the `test` extra.
+- The ``typing-extensions`` dependency on Python < 3.13 is now optional, part of the ``test`` extra.
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
 - :meth:`Component.decoded` now returns a string instead of bytes for text properties.
-- The :func:`icalendar.tools.is_pytz_dt` return value is now hinted as `TypeGuard[datetime]`, not `TypeIs[datetime]`, since it returning false should not allow narrowing as non-datetime.
+- The :func:`icalendar.tools.is_pytz_dt` return value is now hinted as ``TypeGuard[datetime]``, not ``TypeIs[datetime]``, since returning ``False`` should not allow narrowing it as non-datetime.
```
stevepiercy, line 40: Actually, I wonder if this is a "breaking change", because v7 is still in alpha and it's OK to break stuff in alphas. IMO, "breaking" means that it breaks a previously released major version, which is consistent with SemVer.

In v6, we didn't have type hints, as shown in this permalink:

https://github.com/collective/icalendar/blob/51f017e6918e360d4b8437f50599c6ae85343f1e/src/icalendar/tools.py#L89-L91

I think moving this to "Minor changes" would be OK.

Also, when we release icalendar major versions, beginning with v7, we'll have an [Upgrade guide](https://github.com/collective/icalendar/pull/1067), where we can highlight breaking stuff by pulling items from these sections.

stevepiercy (commented):
Question to consider, but otherwise I approve. Thank you!

CHANGES.rst
niccokunzmann: I would never consider type hints as breaking changes. The code works with the hints and without.

jkseppan: Sure, the code works, but the previous type hint promised slightly too much, so some users may now see new type errors. But I'll move this to minor changes. EDIT: oh, it got merged already. Thanks!

Issue #1072, opened 2026-01-06
apply Google Style to a function documentation [beginner, Python]


niccokunzmann:
icalendar has a lot of old functions with docstrings that do not follow a consistent format. Please improve one or more docstrings in a module to conform to our [style guide].

## Your Contribution

You do not need to [set up icalendar for development](https://icalendar.readthedocs.io/en/latest/contribute/documentation/build-check.html). You can [use the GitHub web interface](https://icalendar.readthedocs.io/en/latest/contribute/documentation/small-edits.html) for this contribution if you like.

1. Search for a function in the [src/icalendar/](https://github.com/collective/icalendar/tree/main/src/icalendar) directory. Good candidates include classes with no docstring, `__init__` methods of classes that have no docstring, and docstrings with any of the following patterns:
   - `:param`
   - `:raises`
   - `:returns`
   - `:rtype`
   - `:type`
2. Improve the documentation string according to the [style guide] (better summary, example, spacing, headings, parameters, ...). You can also add type hints, as described in issue #938.
3. Create a Pull Request (PR). When creating it write `Contributes to #1072`, not `Closes`.
4. The documentation preview will be built by Read the Docs for the PR, and a link added to the bottom of the PR description. Please find the function in the API documentation and link it in the PR description. If Read the Docs doesn't insert a link after a few seconds, then visit [icalendar PR preview builds](https://app.readthedocs.org/projects/icalendar/builds/) to find, copy, and paste the URL for your PR preview into the description.
5. Wait for feedback and get it merged :)

## Example

[Pull Request 1070](https://github.com/collective/icalendar/pull/1070/changes) changes the docstrings. Below, you can see the old doctring:

<img width="790" height="539" alt="Screenshot 2026-01-05 at 3 05 21 PM" src="https://github.com/user-attachments/assets/11689eed-769f-4852-aa88-4711c4c42556" />

After using the style guide, there is a one-line summary, an optional description and the parameters and result are clearly highlighted and described: 

<img width="790" height="963" alt="Screenshot 2026-01-05 at 3 05 56 PM" src="https://github.com/user-attachments/assets/6cad836d-7c1e-4160-980c-369d6475a313" />

Location: https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.parser_tools.html

## Related:

- https://github.com/collective/icalendar/pull/1054#issuecomment-3712403073
- [style guide]
- #938 

[...]

PR #1077, opened 2026-01-15
docs: apply Google Style to parser.py docstrings

Eshaan-byte:
Updates 11 function docstrings in src/icalendar/parser.py to follow the Google Style guide as outlined in the project Updates 11 function docstrings in `src/icalendar/parser.py` to follow the Google Style guide as outlined in the [project documentation](https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html).

## Functions Updated
- `escape_char()` - Added Parameters, Returns, and Note sections explaining TEXT escaping rules
- `unescape_char()` - Added comprehensive documentation with replacement order details
- `param_value()` - Documented parameter value conversion with RFC references
- `validate_token()` - Added Parameters and Raises sections for token validation
- `validate_param_value()` - Documented validation for quoted and unquoted values
- `dquote()` - Explained quoting rules with Parameters and Returns sections
- `q_split()` - Added detailed documentation with examples showing quote-aware splitting
- `q_join()` - Documented joining with examples
- `escape_string()` - Added conversion table for backslash to percent encoding
- `unescape_string()` - Added reverse conversion documentation
- `unescape_list_or_string()` - Documented handling of both string and list inputs

## What Was Improved
All docstrings now include:
- One-line summaries
- Proper  Parameters  and  Returns  sections (following project standard, not "Args")
- RFC references using `:rfc:` directive (RFC 5545, RFC 6868)
- Cross-references using `:py:func:` for linking to related functions

Contributes to #1072

----
📚 Documentation preview 📚: https://icalendar--1077.org.readthedocs.build/


niccokunzmann (dismissed):
Thanks! This looks good to me! @stevepiercy might also want to look.

CHANGES.rst
```diff
@@ -27,6 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
+- Updated 11 function docstrings in :mod:`icalendar.parser` to follow the Google Style guide, improving API documentation clarity and consistency. See `Issue #1072 <https://github.com/collective/icalendar/issues/1072>`_.
```
stevepiercy: Would you please move this under the Documentation subheading?

src/icalendar/parser.py
```diff
@@ -30,7 +30,27 @@
 
 
 def escape_char(text):
-    """Format value according to iCalendar TEXT escaping rules."""
+    """Format value according to iCalendar TEXT escaping rules.
```
stevepiercy: First, I think I need to add a new rule to our docstring style guide.

```rst
Escape docstrings
-----------------

Avoid double-escaping in docstrings.
Use the raw ``r`` indicator immediately before the leading docstring delimiter ``"""``.
Thus the docstring and the rendered content will have the same level of escapes.
It will also be less confusing for readers of the source code.

.. seealso::

    ``parser.split_on_unescaped_semicolon`` example.
```

What do you think?

Continuing, would you please make the following change, as well as update this docstring to have only single escapes?

```suggestion
    r"""Format value according to iCalendar TEXT escaping rules.
```

src/icalendar/parser.py
```diff
@@ -30,7 +30,27 @@
...
+
+    Escapes special characters in text values according to :rfc:`5545` rules.
```
stevepiercy: Is there a specific section about escaping? I couldn't find one explicitly. Would this be it?

https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.11

src/icalendar/parser.py
```diff
@@ -30,7 +30,27 @@
...
+
+    Note:
+        The replacement order is critical:
+
+        1. ``\\N`` -> newline (to handle existing escapes)
+        2. ``\\`` -> ``\\\\`` (escape backslashes)
+        3. ``;`` -> ``\\;`` (escape semicolons)
+        4. ``,`` -> ``\\,`` (escape commas)
+        5. ``\\r\\n`` -> ``\\n`` (normalize line endings)
+        6. ``\\n`` -> ``\\n`` (escape newlines)
```
stevepiercy: Adjust to single escapes per above comment.

src/icalendar/parser.py
```diff
@@ -30,7 +30,27 @@
...
+    Parameters:
+        text: The text to escape (str or bytes).
+
+    Returns:
+        The escaped text with special characters escaped.
+
+    Note:
+        The replacement order is critical:
+
+        1. ``\\N`` -> newline (to handle existing escapes)
```
stevepiercy:
```suggestion
        1. ``\N`` -> ``\n`` (to normalize newlines to lowercase)
```

src/icalendar/parser.py
stevepiercy: This one looks weird if there's no transformation. I think this is more accurate. What do you think?

```suggestion
        6. ``\n`` -> ``\n`` (escape newlines with raw string)
```

src/icalendar/parser.py
```diff
@@ -44,6 +64,27 @@ def escape_char(text):
 
 
 def unescape_char(text):
+    """Unescape iCalendar TEXT values.
+
+    Reverses the escaping applied by :py:func:`escape_char` according to
+    :rfc:`5545` TEXT escaping rules.
+
+    Parameters:
+        text: The escaped text (str or bytes).
```
stevepiercy: If you use type hinting, then we don't need to specify str or bytes. Would you like to try that on the method? Otherwise, we have to manually write it out, as shown.

```suggestion
        text: The escaped text (``str`` or ``bytes``).
```

src/icalendar/parser.py
```diff
@@ -44,6 +64,27 @@ def escape_char(text):
...
+
+    Returns:
+        The unescaped text, or None if text is neither str nor bytes.
```
stevepiercy:
```suggestion
        The unescaped text, or ``None`` if text is neither ``str`` nor ``bytes``.
```

src/icalendar/parser.py
```diff
@@ -44,6 +64,27 @@ def escape_char(text):
...
+
+    Note:
+        The replacement order is critical to avoid double-unescaping:
+
+        1. ``\\N`` -> ``\\n`` (intermediate step)
+        2. ``\\r\\n`` -> ``\\n`` (normalize line endings)
+        3. ``\\n`` -> newline (unescape newlines)
+        4. ``\\,`` -> ``,`` (unescape commas)
+        5. ``\\;`` -> ``;`` (unescape semicolons)
+        6. ``\\\\`` -> ``\\`` (unescape backslashes last)
```
stevepiercy: See previous comment about double-escapes.

src/icalendar/parser.py
```diff
@@ -107,7 +148,20 @@ def foldline(line, limit=75, fold_sep="\r\n "):
...
+    """Convert a parameter value to its iCalendar representation.
+
+    Applies :rfc:`6868` escaping and optionally quotes the value according
+    to :rfc:`5545` parameter value formatting rules.
+
+    Parameters:
+        value: The parameter value to convert. Can be a sequence, string, or
+            object with a ``to_ical()`` method.
+        always_quote: If True, always enclose the value in double quotes.
+            Defaults to False (only quote when necessary).
```
stevepiercy:
```suggestion
        always_quote: If ``True``, always enclose the value in double quotes.
            Defaults to ``False`` (only quote when necessary).
```

src/icalendar/parser.py
```diff
@@ -129,13 +183,37 @@ def param_value(value, always_quote=False):
...
 def validate_param_value(value, quoted=True):
+    """Validate a parameter value for unsafe characters.
+
+    Checks parameter values for characters that are not allowed according to
+    :rfc:`5545`. Uses different validation rules for quoted and unquoted values.
+
+    Parameters:
+        value: The parameter value to validate.
+        quoted: If True, validate as a quoted value (allows more characters).
+            If False, validate as an unquoted value (stricter). Defaults to True.
```
stevepiercy:
```suggestion
        quoted: If ``True``, validate as a quoted value (allows more characters).
            If ``False``, validate as an unquoted value (stricter). Defaults to ``True``.
```

src/icalendar/parser.py
```diff
@@ -147,7 +225,21 @@ def validate_param_value(value, quoted=True):
 
 
 def dquote(val, always_quote=False):
-    """Enclose parameter values containing [,;:] in double quotes."""
+    """Enclose parameter values in double quotes when needed.
+
+    Parameter values containing special characters ``[,;:']`` must be enclosed
```
stevepiercy: Are square brackets "special characters"? I couldn't find them as such in the RFC.

src/icalendar/parser.py
```diff
@@ -147,7 +225,21 @@ def validate_param_value(value, quoted=True):
...
+    in double quotes according to :rfc:`5545`. Double-quote characters in the
+    value are replaced with single quotes since they're forbidden in parameter
+    values.
+
+    Parameters:
+        val: The parameter value to quote.
+        always_quote: If True, always enclose in quotes regardless of content.
+            Defaults to False (only quote when necessary).
```
stevepiercy:
```suggestion
        always_quote: If ``True``, always enclose in quotes regardless of content.
            Defaults to ``False`` (only quote when necessary).
```

src/icalendar/parser.py
```diff
@@ -158,7 +250,32 @@ def dquote(val, always_quote=False):
 
 # parsing helper
 def q_split(st, sep=",", maxsplit=-1):
-    """Splits a string on char, taking double (q)uotes into considderation."""
+    """Split a string on separator, respecting double quotes.
```
stevepiercy:
```suggestion
    """Split a string on a separator, respecting double quotes.
```

src/icalendar/parser.py
```diff
@@ -158,7 +250,32 @@ def dquote(val, always_quote=False):
...
+
+    Splits the string on the separator character, but ignores separators that
+    appear inside double-quoted sections. This is needed for parsing parameter
+    values that may contain quoted strings.
+
+    Parameters:
+        st: The string to split.
+        sep: The separator character. Defaults to ",".
```
stevepiercy:
```suggestion
        sep: The separator character. Defaults to ``,``.
```

src/icalendar/parser.py
```diff
@@ -158,7 +250,32 @@ def dquote(val, always_quote=False):
...
+        maxsplit: Maximum number of splits to perform. If -1 (default),
+            perform all possible splits.
```
stevepiercy:
```suggestion
        maxsplit: Maximum number of splits to perform. If ``-1`` (default),
            then perform all possible splits.
```

src/icalendar/parser.py
```diff
@@ -181,7 +298,29 @@ def q_split(st, sep=",", maxsplit=-1):
...
+    """Join a list with a separator, quoting items as needed.
+
+    Joins list items with the separator, applying :py:func:`dquote` to each item
+    to add double quotes when they contain special characters.
+
+    Parameters:
+        lst: The list of items to join.
+        sep: The separator to use. Defaults to ",".
+        always_quote: If True, always quote all items. Defaults to False
+            (only quote when necessary).
```
stevepiercy:
```suggestion
        sep: The separator to use. Defaults to ``,``.
        always_quote: If ``True``, always quote all items. Defaults to ``False``
            (only quote when necessary).
```

src/icalendar/parser.py
```diff
@@ -512,6 +651,26 @@ def from_jcal_property(cls, jcal_property: list):
 
 
 def escape_string(val):
+    """Escape backslash sequences to URL-encoded hex values.
```
stevepiercy: Also apply the newly proposed docstring escaping rule in the Note below.

```suggestion
    r"""Escape backslash sequences to URL-encoded hex values.
```

src/icalendar/parser.py
```diff
@@ -522,6 +681,25 @@ def escape_string(val):
 
 
 def unescape_string(val):
+    """Unescape URL-encoded hex values to their original characters.
```
stevepiercy: Also apply the newly proposed docstring escaping rule in the Note below.

```suggestion
    r"""Unescape URL-encoded hex values to their original characters.
```

src/icalendar/parser.py
```diff
@@ -698,6 +876,17 @@ def rfc_6868_escape(param_value: str) -> str:
...
+    """Unescape a value that may be a string or list of strings.
+
+    Applies :py:func:`unescape_string` to the value. If the value is a list,
+    unescapes each element.
+
+    Parameters:
+        val: A string or list of strings to unescape.
+
+    Returns:
+        The unescaped value(s).
```
stevepiercy: Don't use parentheses, just pluralize. Much easier.

```suggestion
        The unescaped values.
```

src/icalendar/parser.py
```diff
@@ -30,7 +30,27 @@
 
 
 def escape_char(text):
```
stevepiercy: By adding a type hint, we get a nice bonus for free. See next suggestion.

```suggestion
def escape_char(text: str | bytes):
```

src/icalendar/parser.py
```diff
@@ -30,7 +30,27 @@
...
-    """Format value according to iCalendar TEXT escaping rules."""
+    r"""Format value according to iCalendar TEXT escaping rules.
+
+    Escapes special characters in text values according to :rfc:`5545#section-3.3.11` rules.
+    The order of replacements matters to avoid double-escaping.
+
+    Parameters:
+        text: The text to escape (``str`` or ``bytes``).
```
stevepiercy: By adding a type hint (see above), the manual types aren't necessary.

```suggestion
        text: The text to escape.
```

src/icalendar/parser.py
```diff
@@ -44,6 +64,27 @@ def escape_char(text):
 
 
 def unescape_char(text):
```
stevepiercy: Same thing with type hint, but also including a return type, which also gets rendered automatically.

```suggestion
def unescape_char(text: str | bytes) -> str | bytes:
```

src/icalendar/parser.py
```diff
@@ -44,6 +64,27 @@ def escape_char(text):
...
+    r"""Unescape iCalendar TEXT values.
+
+    Reverses the escaping applied by :py:func:`escape_char` according to
+    :rfc:`5545#section-3.3.11` TEXT escaping rules.
+
+    Parameters:
+        text: The escaped text (``str`` or ``bytes``).
```
stevepiercy:
```suggestion
        text: The escaped text.
```

src/icalendar/parser.py
```diff
@@ -44,6 +64,27 @@ def escape_char(text):
...
+
+    Returns:
+        The unescaped text, or ``None`` if text is neither ``str`` nor ``bytes``.
```
stevepiercy:
```suggestion
        The unescaped text, or ``None`` if ``text`` is neither ``str`` nor ``bytes``.
```

src/icalendar/parser.py
```diff
@@ -107,7 +148,20 @@ def foldline(line, limit=75, fold_sep="\r\n "):
 
 
 def param_value(value, always_quote=False):
```
stevepiercy: It would be nice to have a type hint here, and, if necessary, update the docstring accordingly. Can you come up with a type hint? This one is beyond my familiarity with type hints. Maybe your editor has some AI feature?

Now that I think about it, all the methods for which you've added docstrings could be further improved with type hints. Would you like to take that next step, too?

src/icalendar/parser.py
```diff
@@ -129,13 +183,37 @@ def param_value(value, always_quote=False):
 
 
 def validate_token(name):
+    """Validate that a name is a valid iCalendar token.
```
stevepiercy: Let's apply `r` here and make the regex consistent between source and rendered HTML.

```suggestion
    r"""Validate that a name is a valid iCalendar token.
```

src/icalendar/parser.py
```diff
@@ -129,13 +183,37 @@ def param_value(value, always_quote=False):
...
+
+    Checks if the name matches the :rfc:`5545` token syntax using the NAME
+    regex pattern (``[\\w.-]+``).
```
stevepiercy:
```suggestion
    regex pattern (``[\w.-]+``).
```

stevepiercy (changes requested):
Thanks for the changes. While reviewing them, I realized that all the methods to which you've added docstrings could be further improved with type hints. I've added a couple that were easy for me. Would you like to try the remaining ones? It would also contribute to the open issue https://github.com/collective/icalendar/issues/938.

src/icalendar/parser.py
```diff
@@ -106,8 +147,21 @@ def foldline(line, limit=75, fold_sep="\r\n "):
 # Property parameter stuff
 
 
-def param_value(value, always_quote=False):
-    """Returns a parameter value."""
+def param_value(value: Any, always_quote: bool = False) -> str:
```
stevepiercy: Is there a more specific type than `Any`?

Is this valid?

```python
def param_value(value: list | tuple | range | str | object, always_quote: bool = False) -> str:
```

src/icalendar/parser.py
```diff
@@ -521,7 +680,26 @@ def escape_string(val):
     )
 
 
-def unescape_string(val):
+def unescape_string(val: str) -> str:
+    r"""Unescape URL-encoded hex values to their original characters.
+
+    Reverses :py:func:`escape_string` by converting percent-encoded hex values
```
stevepiercy:
```suggestion
    Reverses :func:`escape_string` by converting percent-encoded hex values
```

Eshaan-byte: Good catch! Yes, `:func:` is shorter and works just as well since we're already in the Python domain. I'll update it. Thanks!

src/icalendar/parser.py
Eshaan-byte: True, Looking at the code, the function accepts:
1. Sequences (list or tuple, from SEQUENCE_TYPES)  
2. Strings
3. Any object with a `to_ical()` method
Actually, we can avoid `Any` entirely by using a Protocol for objects with `to_ical()`:

from collections.abc import Sequence
from typing import Protocol

class HasToIcal(Protocol):
    """Protocol for objects with a to_ical method."""
    def to_ical(self) -> bytes: ...

def param_value(
    value: Sequence[str] | str | HasToIcal, 
    always_quote: bool = False
) -> str:

stevepiercy: That's even better. Would you please implement that?

src/icalendar/parser.py
SashankBhamidi: @stevepiercy Given how often this shows up in PRs, I think adding this rule would prevent ongoing confusion.

stevepiercy: @SashankBhamidi see issue opened yesterday https://github.com/collective/icalendar/issues/1080

src/icalendar/parser.py
```diff
@@ -29,6 +30,14 @@
     from icalendar.prop import VPROPERTY
 
 
+class HasToIcal(Protocol):
+    """Protocol for objects with a to_ical method."""
+
+    def to_ical(self) -> bytes:
+        """Convert to iCalendar format."""
+        ...
```
stevepiercy, line 38: I was today days old when I learned of the [Ellipsis constant in Python](https://docs.python.org/3/library/constants.html#Ellipsis) and its use in [type hinting](https://docs.python.org/3/library/typing.html#annotating-callable-objects). 🤯 All this time, I thought it was just a lazy person's notation for elided code!

src/icalendar/parser.py
```diff
@@ -29,8 +30,36 @@
...
+
+    Note:
+        The replacement order is critical:
+
+        1. ``\N`` -> ``\n`` (normalize newlines to lowercase)
+        2. ``\`` -> ``\\`` (escape backslashes)
+        3. ``;`` -> ``\;`` (escape semicolons)
+        4. ``,`` -> ``\,`` (escape commas)
+        5. ``\r\n`` -> ``\n`` (normalize line endings)
+        6. ``\n`` -> ``\n`` (escape newlines with raw string)
```
stevepiercy: I'm flip-flopping on this one. It renders in the docs as if nothing happens, but then the explanation indicates something _does_ happen, which I find confusing. What do you think about this?

```suggestion
        6. ``"\n"`` -> ``r"\n"`` (escape newlines with raw newlines)
```

With the double-quotes and `r`, it's an exception to how we describe the code, but I can't think of a cleaner and clearer, although inconsistent with the preceding list items, way of describing it. Please let me know. Thank you!

src/icalendar/parser.py
Eshaan-byte: 😄 Yes! The `...` (Ellipsis) in Protocols is pretty cool. It says "this method signature exists, but we don't care about the implementation" - perfect for structural subtyping. Python's type system has some nice surprises!

src/icalendar/parser.py
Eshaan-byte: Good catch! I see the confusion now - `\n` -> `\n` looks like a no-op.

Looking at the actual code `text.replace("\n", r"\n")`, I considered a few options:

Option 1: Your suggestion
"\n" -> r"\n" (escape newlines with raw newlines)

Option 2: Be explicit about the character
newline character -> \n (escape as literal backslash-n)

Option 3: Show it's a transformation
\n (newline) -> \n (escaped literal)

I'm leaning toward Option 2 because it removes all ambiguity - you immediately know we're converting an invisible character to visible text. But I can see the value in your approach too since it shows the raw string prefix.

What's your preference?

stevepiercy: Actually, I'm leaning toward merging Options 1 and 2 with a change from "escape" to "transform". I prefer the use of code to symbolize what's going, and the explicit narrative text in parentheses removes ambiguity and reinforces my learning. What do you think of this version?

```suggestion
        6. ``"\n"`` -> ``r"\n"`` (transform a newline character to a literal, or raw, newline character)
```

I also thought about using the verb "normalize" for this one, but it didn't ring quite as true as "transform" because the latter goes across (Latin root _trans_) forms.

Eshaan-byte: I like it! "Transform" is definitely better than "escape" here since we're changing the form, not just escaping it. And showing both the quoted versions makes it crystal clear what's happening - the reader can see exactly how the raw string differs.

I'll update it to your suggested version. Thanks for the thoughtful explanation about the Latin root too! 👍

src/icalendar/parser.py
niccokunzmann:
```python
class HasToIcal(Protocol):
```

That is cool. The first time that I see this.
Would you use that instead of this: https://github.com/collective/icalendar/blob/5043e1037572f9ce1b429f5f94007f7d8adefe15/src/icalendar/prop/__init__.py#L3890 ?

Eshaan-byte: Good question! They serve different purposes:

Protocol (like `HasToIcal`) - for structural typing:

TypeAlias Union (like `VPROPERTY`) - for nominal typing:

For `VPROPERTY`, I'd keep the Union because you want a known, finite set of property types where the type checker can catch if someone passes an invalid type.

For `param_value`, Protocol was better because we accept any object with `to_ical()`, including types we haven't defined yet.

Rule of thumb: Protocol for "anything with this interface", Union for "exactly one of these types".

stevepiercy: I've added this to https://github.com/collective/icalendar/issues/1080 for future work. Nice explanation!

PR #1089, opened 2026-01-18
feat: add NamedTuple support for vCard ADR/N properties

### LINKED ISSUE #1060: Improvements for vCard properties (PR #1059 follow-up)
Follow-up improvements suggested in #1059:

**Use NamedTuple for field access:**
- [ ] vAdr: `adr.street` instead of `adr.fields[2]`
- [ ] vN: `name.family_name` instead of `name.fields[0]`
- [ ] vOrg: `org.name` instead of `org.fields[0]`

References: [1](https://github.com/collective/icalendar/pull/1059#discussion_r2649252230), [2](https://github.com/collective/icalendar/pull/1059#discussion_r2649252667), [3](https://github.com/collective/icalendar/pull/1059#discussion_r2649252845)

**Parametrize tests and add edge cases:**
- [ ] Use pytest parametrization
- [ ] Test escaped escape sequences (\\\\;)
- [ ] Test other escape character interactions

Reference: [4](https://github.com/collective/icalendar/pull/1059#discussion_r2649254191)

**Reorganize tests:**
- Move vCard tests to `src/icalendar/tests/rfc6350/`

Reference: [5](https://github.com/collective/icalendar/pull/1059#discussion_r2649254688)

SashankBhamidi:
Add AdrFields and NFields NamedTuples for structured field access on vAdr and vN. Add name/units properties to vOrg. Reorganize vCard tests into rfc_6350_vcard directory.

Closes #1060

----
📚 Documentation preview 📚: https://icalendar--1089.org.readthedocs.build/

### DIFF
--- a/src/icalendar/prop/__init__.py
+++ b/src/icalendar/prop/__init__.py
@@ -1280,6 +1280,7 @@ class vOrg:
 
     default_value: ClassVar[str] = "TEXT"
     params: Parameters
+    fields: tuple[str, ...]
 
     def __init__(
         self,


src/icalendar/prop/__init__.py
niccokunzmann, line 1287: Could you add the type hint here, too?

niccokunzmann, line 1326: Can you use the NamedTuple here, too?

src/icalendar/prop/__init__.py
```diff
@@ -1288,6 +1332,21 @@ def __repr__(self):
...
+        """The organization name (first field)."""
+        return self.fields[0]
+
+    @property
+    def units(self) -> tuple[str, ...]:
+        """The organizational unit names (remaining fields after the name)."""
+        return self.fields[1:]
+
+    @property
+    def ical_value(self) -> tuple[str, ...]:
```
niccokunzmann, line 1347: Is it possible to use the type hint here, too?

SashankBhamidi:
@niccokunzmann Thanks for the review!

Added the `fields: tuple[str, ...]` class attribute to vOrg for consistency.

Regarding NamedTuple for `from_ical()` and `ical_value` on vOrg: ORG has variable-length fields per RFC 6350 (organization name + zero or more unit names), so a fixed NamedTuple isn't possible. ADR (7 fields) and N (5 fields) are fixed, which is why they can use NamedTuples.

SashankBhamidi:
To clarify on the type hints - they were already present:

- `__init__` parameter: `fields: tuple[str, ...] | list[str] | str`
- `from_ical` return: `-> tuple[str, ...]`
- `ical_value` return: `-> tuple[str, ...]`

The class attribute I added (`fields: tuple[str, ...]`) documents the type of `self.fields` after initialization (which is always a tuple, even though the constructor accepts list/str).

niccokunzmann (approved):
Thanks for the clarification!

Issue #1076, opened 2026-01-11
Use "Parameters" in docstrings, not "Args" or "Arguments"


stevepiercy:
"Parameters" is preferred. See https://docs.python.org/3/faq/programming.html#what-is-the-difference-between-arguments-and-parameters and https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections.

I omitted that detail in https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure.

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/1075#discussion_r2679460695_

PR #1090, opened 2026-01-19
Add behavioral tests for LazyCalendar

SashankBhamidi:
#1201 was inspired by this PR and superseded the implementation. This PR now contributes the behavioral tests on top.

- `to_ical()` works correctly before and after accessing lazy components
- 15 tests covering serialization, `.todos` and `.journals` property access, sequential and mixed access patterns, edge cases (empty calendar, timezone-only, forward timezone references, `multiple=True`), and `contains_uid` substring false-positive behavior

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,8 @@ Bug fixes
 Documentation
 ~~~~~~~~~~~~~
 
+- Run ``sphinx-build`` with ``-W`` to turn warnings into errors. :issue:`1306`
+- Added `sphinx-llms-txt <https://sphinx-llms-txt.readthedocs.io/en/stable/>`_ extension to generate :file:`llms.txt` and :file:`llms-full.txt` files for AI/LLM documentation consumption. :issue:`1302`
 - Fixed CI Vale check reporting and resolved Vale errors. :pr:`1278`
 - Include :file:`Makefile` in documentation workflow path filters so documentation CI runs when Makefile logic changes, and keep Vale failures visible in CI output. :issue:`1277`
 - Document how to install icalendar on Alpine Linux. :pr:`1290`
@@ -52,7 +54,7 @@ Documentation
 - Add Repology badge and distribution installation instructions to install documentation. :issue:`1119`
 - Updated references to :class:`~icalendar.prop.uri.vUri` and :class:`~icalendar.enums.RELTYPE` classes in :file:`attr.py`. :issue:`1158`
 - Convert docstrings in ``attr.py`` and ``cal/calendar.py`` to Google Style format. :issue:`1072`
-- Explained import shortcuts in :doc:`component-api` documentation. :issue:`1161`
+- Explained import shortcuts in :doc:`../explanation/api-design` documentation. :issue:`1161`
 - Added tutorial for creating a calendar with events with attendees. :pr:`1262`
 - Added recognition of NLnet Foundation for its funding and Open Collective for donations to the documentation footer. :issue:`1214`
 - Documented ``vText`` properties according to :rfc:`5545#section-3.3.11`. :issue:`742`
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ DOCS_DIR        = ./docs/
 BUILDDIR        = ../_build
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -W -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 VALEFILES       := $(shell find $(DOCS_DIR) -type f -name "*.rst" -print)  # Also add `src` for docstrings.
 VALEOPTS        ?=
 PYTHONVERSION   = >=3.11,<3.15
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,6 +22,7 @@ extensions = [
     "sphinx_design",
     "sphinx_icalendar",
     "sphinx_issues",
+    "sphinx_llms_txt",
     "sphinx_reredirects",
 ]
 source_suffix = {".rst": "restructuredtext"}
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -91,6 +91,13 @@ icalendar lets you create, inspect, and modify calendaring information with Pyth
     tutorials/index
 
 .. toctree::
+    :caption: Explanation
+    :maxdepth: 2
+    :hidden:
+
+    explanation/index
+
+.. toctree::
     :caption: Reference
     :hidden:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,7 @@ docs = [
     "sphinx-design>=0.5.0",
     "sphinx-icalendar>=0.1.0",
     "sphinx-issues>=5.0.1",
+    "sphinx-llms-txt>=0.7.1",
     "sphinx-notfound-page>=1.1.0",
     "sphinx-reredirects>=0.1.6",
     "vale>=3.12.0",


src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,392 @@
...
+
+        Returns:
+            The parsed Component object.
+        """
+        if index in self._parsed_indices:
+            # Already parsed - find and return the component from subcomponents
+            comp_name, _ = self._raw_components[index]
+            for comp in self.subcomponents:
+                if comp.name == comp_name:
+                    return comp
```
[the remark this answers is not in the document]
SashankBhamidi: This code path is rarely executed. Callers (`_parse_all_of_type`, `_parse_all_raw`) check `_parsed_indices` before calling `_parse_raw_component`, so the "already parsed" branch is defensive code. Premature optimization.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,392 @@
...
+                parsed_val = factory.from_ical(vals, tzid)
+            else:
+                parsed_val = factory.from_ical(vals)
+            vals_inst = factory(parsed_val)
+            vals_inst.params = params if params else Parameters()
+            cal.add(name, vals_inst, encode=0)
+        except Exception:
+            # Calendar is strict, so we should raise, but for robustness
+            # just add as text if it fails
+            pass
```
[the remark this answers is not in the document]
SashankBhamidi: This matches `Calendar.from_ical()` behavior. Calendar-level property parsing failures shouldn't crash the entire calendar parsing. The library doesn't track property-level errors for calendars.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,392 @@
...
+                continue
+            if comp_name == name:
+                self._parse_raw_component(i)
+
+    def _parse_all_raw(self) -> None:
+        """Parse all remaining raw components."""
+        for i in range(len(self._raw_components)):
+            if i not in self._parsed_indices:
+                self._parse_raw_component(i)
+
```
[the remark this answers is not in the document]
SashankBhamidi: Same as comment 3 - thread-safety is not a design goal. The parent `Calendar` class is also not thread-safe. Adding locks would introduce complexity and overhead for a feature that isn't expected.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,433 @@
...
+class TestLazyCalendarJournals:
+    """Test journal component access."""
+
+    def test_journals_property(self):
+        """Accessing journals works correctly."""
+        cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+        journals = cal.journals
+        assert len(journals) == 1
+        assert str(journals[0]["SUMMARY"]) == "Test Journal 1"
```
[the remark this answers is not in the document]
SashankBhamidi: Error handling for malformed components is delegated to `Component.from_ical()` which has its own test coverage. LazyCalendar doesn't introduce new error handling logic.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,392 @@
...
+        return super().todos
+
+    @property
+    def journals(self) -> list[Journal]:
+        """All journal components in the calendar.
+
+        Journals are parsed on first access.
+        """
+        self._parse_all_of_type("VJOURNAL")
+        return super().walk("VJOURNAL")
```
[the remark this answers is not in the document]
SashankBhamidi: `Calendar` doesn't have a `journals` property - it only has `events`, `todos`, `timezones`, `freebusy`, and `availabilities`. Using `walk("VJOURNAL")` is the correct implementation.

src/icalendar/cal/lazy.py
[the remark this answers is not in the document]
SashankBhamidi: Same as comment 1 - this code path is rarely executed. Callers check `_parsed_indices` first.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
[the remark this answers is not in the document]
SashankBhamidi: Thread-safety is not a design goal. The parent `Calendar` class is also not thread-safe. This is consistent with the library's existing design.

stevepiercy:
@SashankBhamidi would you please resolve any open conversations between you and copilot, unless you want someone else to comment, in which case, please tag them?

src/icalendar/cal/lazy.py
stevepiercy, line 1: Please run `make apidoc` to generate the `.rst` source file, and push. That should make this module appear at https://icalendar--1090.org.readthedocs.build/en/1090/reference/api/icalendar.cal.lazy.html

Once I can look at both source and rendered HTML, then I'll do a more thorough review.

CHANGES.rst
```diff
@@ -43,6 +43,7 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
+- Added :class:`~icalendar.cal.lazy.LazyCalendar` for lazy subcomponent parsing. This subclass of :class:`~icalendar.Calendar` defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY components until they are accessed via ``.events``, ``.todos``, ``.walk()``, etc. Calendar properties and VTIMEZONE components are still parsed eagerly. This is useful for accessing calendar metadata or counting events without full parsing overhead. See `Issue #1050 <https://github.com/collective/icalendar/issues/1050>`_.
```
stevepiercy: `LazyCalendar` should render as a link, once... well, see my next suggestion.

```suggestion
- Added :class:`~icalendar.cal.lazy.LazyCalendar` for lazy subcomponent parsing. This subclass of :class:`~icalendar.Calendar` defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY components until they are accessed via ``.events``, ``.todos``, ``.walk()``, or other component methods. Calendar properties and VTIMEZONE components are still parsed eagerly. This is useful for accessing calendar metadata or counting events without full parsing overhead. See `Issue #1050 <https://github.com/collective/icalendar/issues/1050>`_.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+    from icalendar.cal.event import Event
+    from icalendar.cal.free_busy import FreeBusy
+    from icalendar.cal.journal import Journal
+    from icalendar.cal.todo import Todo
+
+
+class LazyCalendar(Calendar):
+    """Calendar with lazy subcomponent parsing for large files.
+
+    Parses Calendar properties and VTIMEZONE components eagerly, but defers
```
stevepiercy: If it doesn't link, then use the longer form of `~icalendar.cal.calendar.Calendar`.

```suggestion
    """:class:`Calendar` with lazy subcomponent parsing for large files.

    Parses :class:`Calendar` properties and VTIMEZONE components eagerly, but defers
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+    parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY
+    until accessed via ``.events``, ``.todos``, ``.walk()``, etc.
```
stevepiercy:
```suggestion
    until accessed via ``.events``, ``.todos``, ``.walk()``, or other component methods.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        cal["VERSION"]  # Returns vText('2.0')
+
+        # Events are parsed on first access
+        len(cal.events)  # Returns 1500
+
+    Note:
+        Once accessed, components are fully parsed and cached.
+        Subsequent accesses return the same parsed objects.
+
+        The ``to_ical()`` method produces correct output whether
```
stevepiercy: Let's link to the method. Hopefully it uses the current context, but if not, then use the fully qualified path and leading tilde `~icalendar.cal.lazy.LazyCalendar.from_ical`.

```suggestion
        The :meth:`to_ical` method produces correct output whether
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        components have been parsed or not.
+    """
+
+    # Components that should be parsed lazily
```
stevepiercy: Move this into a docstring so that it appears in the HTML.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+    LAZY_COMPONENTS: ClassVar[frozenset[str]] = frozenset(
+        {"VEVENT", "VTODO", "VJOURNAL", "VFREEBUSY", "VAVAILABILITY"}
+    )
+
+    # Components that must be parsed eagerly (needed for timezone resolution)
```
stevepiercy: Move this into a docstring so that it appears in the HTML.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        # raw_lines is a list of Contentline objects
+        self._raw_components: list[tuple[str, list[Contentline]]] = []
+        # Track which have been parsed (by index)
+        self._parsed_indices: set[int] = set()
+
+    @classmethod
+    def from_ical(cls, st, multiple: bool = False):
+        """Parse iCalendar data with lazy subcomponent parsing.
+
+        Calendar properties and VTIMEZONE are parsed immediately.
```
stevepiercy:
```suggestion
        :class:`Calendar` properties and VTIMEZONE are parsed immediately.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        Other subcomponents (VEVENT, VTODO, etc.) are stored as raw
```
stevepiercy: Don't use Latin abbreviations per https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/etc.

```suggestion
        Other subcomponents—such as VEVENT or VTODO—are stored as raw
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        content lines and parsed on first access.
+
+        Args:
```
stevepiercy: See https://github.com/collective/icalendar/issues/1076

```suggestion
        Parameters:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+            st: iCalendar data as bytes or string
```
stevepiercy: Let's use type hints ftw!

```suggestion
            st (bytes | str): iCalendar data.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+            multiple: If ``True``, returns list. If ``False``, returns single calendar.
```
stevepiercy:
```suggestion
            multiple (bool): If ``True``, returns list. If ``False``, returns single calendar.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+
+        Returns:
+            LazyCalendar or list of LazyCalendars
```
stevepiercy:
```suggestion
            LazyCalendar or list of LazyCalendars.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Store raw lines for unparsed components: list of (name, raw_lines)
+        # raw_lines is a list of Contentline objects
+        self._raw_components: list[tuple[str, list[Contentline]]] = []
+        # Track which have been parsed (by index)
+        self._parsed_indices: set[int] = set()
+
+    @classmethod
+    def from_ical(cls, st, multiple: bool = False):
```
stevepiercy: I think this is the correct return type hint, but maybe not. When provided, you'll get the `Returns` heading automatically and don't need to specify it manually.

```suggestion
    def from_ical(cls, st, multiple: bool = False) -> LazyCalendar | list[LazyCalendar]:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+            # just add as text if it fails
+            pass
+
+    def _parse_raw_component(self, index: int) -> Component:
+        """Parse a raw component by index and add to subcomponents.
+
+        If the component at the given index has already been parsed,
+        returns the previously parsed component from subcomponents.
+
+        Args:
```
stevepiercy:
```suggestion
        Parameters:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+            index: Index into self._raw_components
```
stevepiercy:
```suggestion
            index: Index into self._raw_components.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        for i in range(len(self._raw_components)):
+            if i not in self._parsed_indices:
+                self._parse_raw_component(i)
+
+    def walk(self, name=None, select=lambda _: True):
+        """Recursively traverse component and subcomponents.
+
+        Parses lazy components as needed.
+
+        Args:
```
stevepiercy:
```suggestion
        Parameters:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+            if comp_name == name:
+                self._parse_raw_component(i)
+
+    def _parse_all_raw(self) -> None:
+        """Parse all remaining raw components."""
+        for i in range(len(self._raw_components)):
+            if i not in self._parsed_indices:
+                self._parse_raw_component(i)
+
+    def walk(self, name=None, select=lambda _: True):
```
stevepiercy: It would be nice to have a returns type hint, but I don't know how to do that for this method. Maybe we have a clue from https://github.com/collective/icalendar/pull/1077#pullrequestreview-3674481325 or https://stackoverflow.com/a/39624147/2214933 ?

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        FreeBusy components are parsed on first access.
+        """
+        self._parse_all_of_type("VFREEBUSY")
+        return super().freebusy
+
+    @property
+    def availabilities(self) -> list[Availability]:
+        """All Availability components in the calendar.
+
+        Availability components are parsed on first access.
```
stevepiercy: or `~icalendar.cal.availability.Availability`.

```suggestion
        """All :class:`Availability` components in the calendar.

        :class:`Availability` components are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        """All event components in the calendar.
+
+        Events are parsed on first access.
+        """
+        self._parse_all_of_type("VEVENT")
+        return super().events
+
+    @property
+    def todos(self) -> list[Todo]:
+        """All todo components in the calendar.
```
stevepiercy: Try short form, else `~icalendar.cal.todo.Todo`.

```suggestion
        """All :class:`Todo` components in the calendar.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+
+        Todos are parsed on first access.
```
stevepiercy: I'm going from memory, but I think you have to escape the space to get plurals for roles.

```suggestion
        :class:`Todo`\ s are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        """
+        self._parse_all_of_type("VTODO")
+        return super().todos
+
+    @property
+    def journals(self) -> list[Journal]:
+        """All journal components in the calendar.
+
+        Journals are parsed on first access.
```
stevepiercy: Same. `~icalendar.cal.journal.Journal`.

```suggestion
        """All :class:`Journal` components in the calendar.

        :class:`Journal`\ s are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        """
+        self._parse_all_of_type("VJOURNAL")
+        return super().walk("VJOURNAL")
+
+    @property
+    def freebusy(self) -> list[FreeBusy]:
+        """All FreeBusy components in the calendar.
+
+        FreeBusy components are parsed on first access.
```
stevepiercy: or `~icalendar.cal.free_busy.FreeBusy`

```suggestion
        """All :class:`FreeBusy` components in the calendar.

        :class:`FreeBusy` components are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+        self._parse_all_of_type("VAVAILABILITY")
+        return super().availabilities
+
+    def property_items(self, recursive=True, sorted: bool = True):
+        """Returns properties in this component and subcomponents.
+
+        For unparsed components, reconstructs from stored raw lines
+        to ensure correct round-trip serialization.
+
+        Args:
```
stevepiercy:
```suggestion
        Parameters:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+    @property
+    def availabilities(self) -> list[Availability]:
+        """All Availability components in the calendar.
+
+        Availability components are parsed on first access.
+        """
+        self._parse_all_of_type("VAVAILABILITY")
+        return super().availabilities
+
+    def property_items(self, recursive=True, sorted: bool = True):
```
stevepiercy: I'm not sure if this is correct syntax for the return type, maybe use a trailing comma for the tuple?

```suggestion
    def property_items(self, recursive=True, sorted: bool = True) -> list[tuple(name,value)]:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,398 @@
...
+
+        For unparsed components, reconstructs from stored raw lines
+        to ensure correct round-trip serialization.
+
+        Args:
+            recursive: Include subcomponents.
+            sorted: Sort property names.
+
+        Returns:
+            List of (name, value) tuples.
```
stevepiercy: If we have a returns type hint, this can be removed.

stevepiercy (changes requested):
This is really good. I've made suggestions to add both links to objects and type hints, as well as a few grammar and style things.

src/icalendar/cal/lazy.py
SashankBhamidi: The parent `Calendar.walk()` method returns `list[Component]`. We could add the same here, but since this just calls `super().walk()`, the inherited type hint should apply.

SashankBhamidi:
@stevepiercy Thanks for the review!

I’d like to wait for @niccokunzmann’s review before merging. Nicco, I added comments to Copilot’s suggestions. They explain the reasoning behind the approach and should help if any of the logic looks off.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+    from icalendar.cal.event import Event
+    from icalendar.cal.free_busy import FreeBusy
+    from icalendar.cal.journal import Journal
+    from icalendar.cal.todo import Todo
+
+
+class LazyCalendar(Calendar):
+    """A :class:`Calendar` with lazy subcomponent parsing for large files.
+
+    Parses :class:`Calendar` properties and VTIMEZONE components eagerly, but
```
stevepiercy: meh, it didn't link.

```suggestion
    """A :class:`~icalendar.cal.calendar.Calendar` with lazy subcomponent parsing for large files.

    Parses :class:`~icalendar.cal.calendar.Calendar` properties and VTIMEZONE components eagerly, but
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
+"""Lazy subcomponent parsing for large calendars.
+
+See :issue:`1050`.
```
stevepiercy: We don't yet support this syntax, but we should. See https://github.com/collective/icalendar/issues/1091.

Until then.

```suggestion
See `Issue #1050 <https://github.com/collective/icalendar/issues/1050>`_.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+        cal["VERSION"]  # Returns vText('2.0')
+
+        # Events are parsed on first access
+        len(cal.events)  # Returns 1500
+
+    Note:
+        Once accessed, components are fully parsed and cached.
+        Subsequent accesses return the same parsed objects.
+
+        The :meth:`to_ical` method produces correct output whether
```
stevepiercy: Didn't link. Let's try this.

```suggestion
        The :meth:`~icalendar.cal.lazy.LazyCalendar.from_ical` method produces correct output whether
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+        # Track which have been parsed (by index)
+        self._parsed_indices: set[int] = set()
+
+    @classmethod
+    def from_ical(
+        cls, st: bytes | str, multiple: bool = False
+    ) -> LazyCalendar | list[LazyCalendar]:
+        """Parse iCalendar data with lazy subcomponent parsing.
+
+        :class:`Calendar` properties and VTIMEZONE are parsed immediately.
```
stevepiercy:
```suggestion
        :class:`~icalendar.cal.calendar.Calendar` properties and VTIMEZONE are parsed immediately.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+        :class:`FreeBusy` components are parsed on first access.
+        """
+        self._parse_all_of_type("VFREEBUSY")
+        return super().freebusy
+
+    @property
+    def availabilities(self) -> list[Availability]:
+        """All :class:`Availability` components in the calendar.
+
+        :class:`Availability` components are parsed on first access.
```
stevepiercy:
```suggestion
        """All :class:`~icalendar.cal.availability.Availability` components in the calendar.

        :class:`~icalendar.cal.availability.Availability` components are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+        :class:`Event`\\ s are parsed on first access.
+        """
+        self._parse_all_of_type("VEVENT")
+        return super().events
+
+    @property
+    def todos(self) -> list[Todo]:
+        """All :class:`Todo`\\ s in the calendar.
+
+        :class:`Todo`\\ s are parsed on first access.
```
stevepiercy:
```suggestion
        """All :class:`~icalendar.cal.todo.Todo`\\ s in the calendar.

        :class:`~icalendar.cal.todo.Todo`\\ s are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+        """
+        self._parse_all_of_type("VTODO")
+        return super().todos
+
+    @property
+    def journals(self) -> list[Journal]:
+        """All :class:`Journal`\\ s in the calendar.
+
+        :class:`Journal`\\ s are parsed on first access.
```
stevepiercy:
```suggestion
        """All :class:`~icalendar.cal.journal.Journal`\\ s in the calendar.

        :class:`~icalendar.cal.journal.Journal`\\ s are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+        """
+        self._parse_all_of_type("VJOURNAL")
+        return super().walk("VJOURNAL")
+
+    @property
+    def freebusy(self) -> list[FreeBusy]:
+        """All :class:`FreeBusy` components in the calendar.
+
+        :class:`FreeBusy` components are parsed on first access.
```
stevepiercy:
```suggestion
        """All :class:`~icalendar.cal.free_busy.FreeBusy` components in the calendar.

        :class:`~icalendar.cal.free_busy.FreeBusy` components are parsed on first access.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,401 @@
...
+    ) -> list[tuple[str, Any]]:
+        """Returns properties in this component and subcomponents.
+
+        For unparsed components, reconstructs from stored raw lines
+        to ensure correct round-trip serialization.
+
+        Parameters:
+            recursive: Include subcomponents.
+            sorted: Sort property names.
+        """
```
stevepiercy: I got Returns confused with the automatic Return type. This is still useful.

```suggestion

        Returns:
            List of (name, value) tuples.
        """
```

src/icalendar/cal/lazy.py
SashankBhamidi: This is wrong, the docstring is talking about `to_ical()`, not `from_ical()`. The correct fix should be:

```suggestion
        The :meth:`~icalendar.cal.lazy.LazyCalendar.to_ical` method produces correct output whether
```

src/icalendar/cal/lazy.py
SashankBhamidi: I shall modify to your suggestion, I do not want to get back to this after 1091 is resolved :)

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+        with open("large_calendar.ics", "rb") as f:
+            cal = LazyCalendar.from_ical(f.read())
+
+        # Access metadata immediately (no event parsing)
+        cal["VERSION"]  # Returns vText('2.0')
+
+        # Events are parsed on first access
+        len(cal.events)  # Returns 1500
+
+    Note:
```
niccokunzmann:
```suggestion
    .. note::
```

Does that work?

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+                        ]
+
+                        if component_name in cls.EAGER_COMPONENTS:
+                            # Parse VTIMEZONE eagerly using existing Component.from_ical
+                            raw_str = "\r\n".join(str(ln) for ln in raw_lines)
+                            component = Component.from_ical(raw_str)
+                            cal.add_component(component)
+                            if component_name == "VTIMEZONE" and "TZID" in component:
+                                tzp.cache_timezone_component(component)
+                        elif component_name in cls.LAZY_COMPONENTS:
```
niccokunzmann: TODO: The question I have here is that of a side effect: How likely is it that the cached timezones will change in between parsing calendars lazily?

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+            )
+        return calendars[0]
+
+    @classmethod
+    def _parse_calendar_property(cls, cal, name, params, vals, line):
+        """Parse a single calendar-level property.
+
+        Reuses the property parsing logic from Component.from_ical.
+        """
+        from icalendar.parser import Parameters, split_on_unescaped_comma
```
niccokunzmann: This import should be at the top if possible.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+
+        uname = name.upper()
+        value_param = params.get("VALUE") if params else None
+        factory = cls.types_factory.for_property(name, value_param)
+
+        datetime_names = ("DTSTART", "DTEND", "RECURRENCE-ID", "DUE", "RDATE", "EXDATE")
```
niccokunzmann: The following lines look like they duplicate behaviour from other components. It is best to refactor the code into `_` methods so that it can be reused. They would reside in the `Component` class.  

Code duplication is a [code smell](https://code-smells.com/dispensables/duplicate-code). I would like to avoid that.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+
+        Parameters:
+            name: If provided, only return components with this name.
+            select: Optional filter function.
+
+        Returns:
+            List of matching components.
+        """
+        if name is not None:
+            self._parse_all_of_type(name.upper())
```
niccokunzmann: That is cool! I like the little optimization here!

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+        return super().walk(name, select)
+
+    @property
+    def events(self) -> list[Event]:
+        """All :class:`~icalendar.cal.event.Event`\\ s in the calendar.
+
+        :class:`~icalendar.cal.event.Event`\\ s are parsed on first access.
+        """
+        self._parse_all_of_type("VEVENT")
+        return super().events
```
niccokunzmann: This is not required, see the implementation of events.

```python
    @property
    def events(self) -> list[Event]:
        return self.walk("VEVENT")
```

Please remove all those attributes.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+
+        Parameters:
+            recursive: Include subcomponents.
+            sorted: Sort property names.
+
+        Returns:
+            List of (name, value) tuples.
+        """
+        v_text = self.types_factory["text"]
+        properties = [("BEGIN", v_text(self.name).to_ical())]
```
niccokunzmann: The way to avoid code duplication here is to refactor the component class, I think. Please touch the code there.

```
for subcomponent in self.subcomponents:
                properties += subcomponent.property_items(sorted=sorted)
```

Could become `for subcomponent in _iterate_components_with_properties():`. `_iterate_components_with_properties` gets overwritten in the LazyCalendar.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+                    # Output raw lines directly
+                    for raw_line in raw_lines:
+                        # Use marker that content_line will handle
+                        properties.append((self._RAW_LINE_MARKER, raw_line))
+
+        properties.append(("END", v_text(self.name).to_ical()))
+        return properties
+
+    def content_line(self, name, value, sorted: bool = True):
+        """Returns property as content line.
```
niccokunzmann: Nice! This is cool. I like it.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,433 @@
+"""Tests for LazyCalendar - lazy subcomponent parsing for large calendars.
+
+See https://github.com/collective/icalendar/issues/1050
+"""
+
+from icalendar import Calendar, LazyCalendar
+
+# Test calendar with multiple component types
+CALENDAR_WITH_EVENTS_AND_TODOS = b"""BEGIN:VCALENDAR
```
niccokunzmann: All those go into test files.

The conftest.py file needs adapting to yield a feature called 'lazy_calendar' if you like to test this specifically.

However what I recommend is adjusting the `calendars` fixture to be parametrized and then let all the tests run with the LazyCalendar AND the Calendar. This way, all tests run on both classes. That is in the end what would be nicest... It is it least worth a try and see which tests fail on the LazyCalendar.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def test_events_property_triggers_parsing(self):
+        """Accessing .events parses and returns events."""
+        cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+        # Before access
+        assert len(cal._parsed_indices) == 0
+
+        # Access events
+        events = cal.events
```
niccokunzmann: This assertion can be skipped if we use test parametrization with LazyCalendar and Calendar.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,433 @@
...
+        assert len(events) == 2
+
+        # Events should now be parsed
+        assert len(cal._parsed_indices) == 2  # Only VEVENT parsed
+
+        # Check event details
```
niccokunzmann: This assertion should also be redundant if we use test parametrization with LazyCalendar and Calendar.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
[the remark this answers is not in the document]
niccokunzmann: Concurrent here probably means not parallel.

Do all combinations (.event, .todo, ...) work to access the components? This can be done with test parametrization.

niccokunzmann (changes requested):
Hi, I like that there is a lot of coverage already. I added a few suggestions. I am happy to add some code changes, too. Especially since it might be very specific what I have in mind. Let me know what you like to do and delegate some work to me that you would rather like to review. Thanks for the PR! Really nice! That is soo needed!

src/icalendar/cal/lazy.py
SashankBhamidi: The timezone caching behavior is identical to regular `Calendar.from_ical()` - both call `tzp.cache_timezone_component()` for VTIMEZONE components. LazyCalendar parses VTIMEZONE eagerly (same timing as Calendar), so there's no difference in caching behavior between the two classes.

src/icalendar/cal/lazy.py
SashankBhamidi: The import of `Parameters` and `split_on_unescaped_comma` from `icalendar.parser` is kept inside `_parse_calendar_property()` to avoid circular import issues at module load time - the parser module imports Component-related classes.

src/icalendar/cal/lazy.py
SashankBhamidi: Hmm.

SashankBhamidi: Agreed.

SashankBhamidi: Would you prefer I:
1. Refactor the Component class in this PR to extract reusable methods like `_parse_property()`, `_parse_datetime_property()`, etc., or
2. Accept this for the initial implementation and create a follow-up issue for refactoring?

src/icalendar/cal/lazy.py
SashankBhamidi: I think there might be a misunderstanding here. The property overrides are essential for lazy parsing to work correctly.

SashankBhamidi: The parent `Calendar.events` is implemented as:
```python
@property
def events(self) -> list[Event]:
    return self.walk("VEVENT")
```

However, `walk()` without arguments parses ALL components. So if we don't override `.events`, accessing it would trigger parsing of all VEVENTs, VTODOs, VJOURNALs, etc., defeating the purpose of lazy parsing.

The overrides ensure:
- `.events` only parses VEVENTs
- `.todos` only parses VTODOs  
- `.journals` only parses VJOURNALs

This is the core optimization - you can access calendar metadata and count events without parsing 1000s of todos and journals.

Should we keep them?

src/icalendar/cal/lazy.py
SashankBhamidi: I'd like to defer this to a follow-up PR to keep this one focused on LazyCalendar

src/icalendar/tests/test_issue_1050_lazy_calendar.py
SashankBhamidi: Excellent suggestion! However, the current test file has tests specifically for lazy parsing mechanics (checking `_parsed_indices`, `_raw_components`, etc.) that only apply to LazyCalendar.

SashankBhamidi: I will create a follow-up issue for parametrizing the entire test suite?

SashankBhamidi: https://github.com/collective/icalendar/issues/1093

src/icalendar/cal/lazy.py
stevepiercy: `Note:` works as is. PyData Sphinx Theme supports turning headings into admonitions without using the klunky reStructuredText directive form.

Screenshot 2026-01-20 at 1 57 04 PM

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+        with open("large_calendar.ics", "rb") as f:
+            cal = LazyCalendar.from_ical(f.read())
+
+        # Access metadata immediately (no event parsing)
+        cal["VERSION"]  # Returns vText('2.0')
+
+        # Events are parsed on first access
+        len(cal.events)  # Returns 1500
+
+    .. note::
```
stevepiercy: Please revert. See https://github.com/collective/icalendar/pull/1090#discussion_r2710216010

```suggestion
    Note:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+    methods.
+
+    This is useful for:
+
+    - Accessing calendar metadata without parsing all events
+    - Counting events without full parsing overhead
+    - Working with very large calendars (1000+ events)
+
+    Example::
+
```
stevepiercy: Let's use the explicit lexer form. This also means that the theme will change the appearance of the Example heading from plain text to something prettier.

```suggestion
    Example:

        .. code-block:: py

```

## Current example

Screenshot 2026-01-20 at 2 08 25 PM

## Nice Example heading example

https://icalendar--1090.org.readthedocs.build/en/1090/reference/api/icalendar.cal.alarm.html#icalendar.cal.alarm.Alarm

Screenshot 2026-01-20 at 2 09 02 PM

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,404 @@
...
+        # Parse a large calendar file
+        with open("large_calendar.ics", "rb") as f:
+            cal = LazyCalendar.from_ical(f.read())
+
+        # Access metadata immediately (no event parsing)
+        cal["VERSION"]  # Returns vText('2.0')
+
+        # Events are parsed on first access
+        len(cal.events)  # Returns 1500
```
stevepiercy: Indent.

```suggestion
            from icalendar import LazyCalendar

            # Parse a large calendar file
            with open("large_calendar.ics", "rb") as f:
                cal = LazyCalendar.from_ical(f.read())

            # Access metadata immediately (no event parsing)
            cal["VERSION"]  # Returns vText('2.0')

            # Events are parsed on first access
            len(cal.events)  # Returns 1500
```

stevepiercy (changes requested):

stevepiercy (dismissed):
Ah, my OCD itch has been scratched. Look at this beautiful documentation!

https://icalendar--1090.org.readthedocs.build/en/1090/reference/api/icalendar.cal.lazy.html#icalendar.cal.lazy.LazyCalendar

Thank you!

niccokunzmann:
A thought that I had: Is it more developer friendly to call that class BigCalendar instead of LazyCalendar? The parsing is deferred but really the use-case is to parse huge calendar files with it. Could also be FastCalendar.

Then, it would be good to know when this calendar actually kicks in as being useful. A benchmark is nice for that.

src/icalendar/cal/lazy.py
niccokunzmann: I would like to avoid as many follow-up issues as possible because they are a maybe promise.
The problem that I have is if we add code that is not of good quality (smelly) then this creates maintenance effort over time and unforeseen problems for users.

I would like #1093 to be part of this PR, too. 

I took on the maintainer role of the project - that is to assure that it stays flexible in the future and adapts to requirements. So, this edges into the realm where I am not willing in my free time to make sure that duplicated code is working.

I would like these to be applied where needed:

- https://refactoring.guru/extract-method
- https://en.wikipedia.org/wiki/Template_method_pattern

It is quite basic for me and I can do it. I will use option 1 and as I said, I am happy to work on it. Could also be in `main`.

So, personally, I like clean code and good testing. So, looking at the code being copied, I get some reactions: Neither TDD nor clean code. Sorry about that. It does not make it easier for me to review this. It would be really nice for me to have this PR cleaner. I would really appreciate that. However, I am in a conflict here with progress and that is important for the project. So, I will take some time to think about that. Maybe someone else wants to give the go-ahead in the mean time.

src/icalendar/cal/lazy.py
niccokunzmann: Hm. The parser should not import icalendar.cal.lazy in this way, I think. That needs investigating.

Could you point out to me where that import takes place?

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,405 @@
...
+
+        comp_name, raw_lines = self._raw_components[index]
+        raw_str = "\r\n".join(str(ln) for ln in raw_lines)
+        component = Component.from_ical(raw_str)
+        self.subcomponents.append(component)
+        self._parsed_indices.add(index)
+        return component
+
+    def _parse_all_of_type(self, name: str) -> None:
+        """Parse all raw components of a given type."""
```
niccokunzmann:
```suggestion
        """Parse all raw components of a given type."""
        name = name.upper()
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,405 @@
...
+
+        Parameters:
+            name: If provided, only return components with this name.
+            select: Optional filter function.
+
+        Returns:
+            List of matching components.
+        """
+        if name is not None:
+            self._parse_all_of_type(name.upper())
```
niccokunzmann:
```suggestion
            self._parse_all_of_type(name)
```

This is a clearer responsibility -> move parameter requirements into the method that requires them.

src/icalendar/cal/lazy.py
niccokunzmann: > However, walk() without arguments parses ALL components. 

This statement seems to not apply to what I wrote about as walk is not used without arguments.

See the code below. Actually it should work.

```
        if name is not None:
            self._parse_all_of_type(name.upper())
```

See https://github.com/collective/icalendar/pull/1090/changes#r2712250334

src/icalendar/cal/lazy.py
niccokunzmann: If we submit this fix now, it creates more issues than it solves.
To not duplicate code is a very basic skill of a programmer. If you want to work on the core logic of icalendar, I ask you to apply basic standards to your code. Otherwise, we are getting into a mess in the future - after merge basically.

SashankBhamidi:
@niccokunzmann Thanks for the detailed review. I've addressed your feedback in the latest commit.

---

> TODO: The question I have here is that of a side effect: How likely is it that the cached timezones will change in between parsing calendars lazily?

As mentioned earlier, timezone caching is identical to `Calendar.from_ical()`. VTIMEZONE is parsed eagerly, same timing as Calendar.

---

> This import should be at the top if possible.

> Could you point out to me where that import takes place?

Moved to top. Investigated and confirmed there is no circular import - `icalendar.parser` does not import `icalendar.cal.lazy`. Imports work fine at module level.

---

> The following lines look like they duplicate behaviour from other components. It is best to refactor the code into `_` methods so that it can be reused.

> I would like these to be applied where needed:
> - https://refactoring.guru/extract-method
> - https://en.wikipedia.org/wiki/Template_method_pattern

Done. Applied both:
- **Extract Method**: `Component._parse_single_property()` for shared property parsing
- **Template Method**: `Component._iter_property_items_subcomponents()` hook that LazyCalendar overrides

---

> I would like #1093 to be part of this PR, too.

Added `calendar_class`, `lazy_calendars`, and `any_calendars` fixtures to `conftest.py`.

---

> This is not required, see the implementation of events... Please remove all those attributes.

Removed all redundant property overrides. You were right - `Calendar.events` calls `self.walk("VEVENT")` which triggers `LazyCalendar.walk()` → `_parse_all_of_type("VEVENT")`.

---

> The way to avoid code duplication here is to refactor the component class... Could become `for subcomponent in _iterate_components_with_properties():`.

Done. Added `_iter_property_items_subcomponents()` in Component, overridden in LazyCalendar.

---

> Suggested change: move `.upper()` into `_parse_all_of_type()`

Done.

---

Also added `Calendar.journals` property for API consistency (was missing).

Ready for you to take over.

SashankBhamidi:
## Summary of Changes

Addressed all PR review feedback:

1. **Moved imports to top of lazy.py** - confirmed no circular import issues
2. **Extracted `Component._parse_single_property()`** - shared property parsing logic
3. **Applied Template Method pattern** - added `Component._iter_property_items_subcomponents()` 
4. **Removed redundant property overrides** - `events`, `todos`, `journals`, `freebusy`, `availabilities` now inherit from parent
5. **Moved `.upper()` into `_parse_all_of_type()`**
6. **Added `Calendar.journals` property** - for API consistency with `events`, `todos`, etc.
7. **Added test fixtures** - `calendar_class`, `lazy_calendars`, `any_calendars` for parametrized testing

## Deferred Items

1. **`subcomponents` as property** - Per review discussion, you mentioned you'd handle converting `subcomponents` from a list to a property in a separate change.

2. **Migrating existing tests to `any_calendars`** - The fixtures are in place (#1093), but existing tests still use the `calendars` fixture. Migrating them to run with both Calendar and LazyCalendar could be done incrementally in future PRs.

## Merged with main

Branch is now up-to-date with latest `main` (merged today). All 25 LazyCalendar tests pass.

src/icalendar/cal/component.py
```diff
@@ -406,6 +406,26 @@ def walk(self, name=None, select=lambda _: True) -> list[Component]:
...
+    def _iter_property_items_subcomponents(
+        self, sorted: bool = True
+    ) -> list[tuple[str, object]]:
+        """Iterate over subcomponents for property_items().
+
+        This is a hook method that can be overridden by subclasses to customize
+        how subcomponents are serialized. For example, LazyCalendar overrides
+        this to handle unparsed raw components.
+
+        Args:
```
stevepiercy: Even though this is a private method and won't show up in the HTML docs, we ought to follow Python best practices in docstrings. Args are the values, whereas parameters are the things that accept arguments. See https://github.com/collective/icalendar/issues/1076 for details. Maybe I need to explicitly add something to the style guide at https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure? Please let me know. Thank you!

```suggestion
        Parameters:
```

src/icalendar/cal/component.py
```diff
@@ -427,12 +447,118 @@ def property_items(
...
+        line,
+        *,
+        ignore_errors: bool = False,
+    ) -> None:
+        """Parse a single property and add it to a component.
+
+        This is a helper method that handles property parsing logic shared between
+        :meth:`from_ical` and :class:`~icalendar.cal.lazy.LazyCalendar`.
+
+        Args:
```
```suggestion
        Parameters:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
+"""Lazy subcomponent parsing for large calendars.
+
+See `Issue #1050 <https://github.com/collective/icalendar/issues/1050>`_.
```
stevepiercy: Use shiny new syntax.

```suggestion
See issue :issue:`1050`.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+from icalendar.parser import Contentline, Contentlines
+from icalendar.timezone import tzp
+
+
+class LazyCalendar(Calendar):
+    """A :class:`~icalendar.cal.calendar.Calendar` with lazy subcomponent parsing for large files.
+
+    Parses :class:`~icalendar.cal.calendar.Calendar` properties and VTIMEZONE components eagerly, but
+    defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY
+    until accessed via ``.events``, ``.todos``, ``.walk()``, or other component
```
stevepiercy: Should each of these be linked to the component methods? I couldn't easily find a good target, except:

- https://icalendar--1090.org.readthedocs.build/en/1090/reference/api/icalendar.cal.component.html#icalendar.cal.component.Component.walk for `walk`

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+        Other subcomponents—such as VEVENT or VTODO—are stored as raw
+        content lines and parsed on first access.
+
+        Parameters:
+            st (bytes | str): iCalendar data as bytes or string.
+            multiple (bool): If ``True``, returns list. If ``False``, returns
+                single calendar.
+
+        Returns:
+            LazyCalendar or list of LazyCalendar instances.
```
stevepiercy:
```suggestion
            :class:`~icalendar.cal.lazy.LazyCalendar` or list of :class:`~icalendar.cal.lazy.LazyCalendar` instances.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+        raw_str = "\r\n".join(str(ln) for ln in raw_lines)
+        component = Component.from_ical(raw_str)
+        self.subcomponents.append(component)
+        self._parsed_indices.add(index)
+        return component
+
+    def _parse_all_of_type(self, name: str) -> None:
+        """Parse all raw components of a given type.
+
+        Args:
```
stevepiercy:
```suggestion
        Parameters:
```

CHANGES.rst
```diff
@@ -46,6 +46,9 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
+- Added :class:`~icalendar.cal.lazy.LazyCalendar` for lazy subcomponent parsing. This subclass of :class:`~icalendar.Calendar` defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY components until they are accessed via :attr:`~icalendar.cal.calendar.Calendar.events`, :attr:`~icalendar.cal.calendar.Calendar.todos`, :meth:`~icalendar.cal.component.Component.walk`, or other component methods. Calendar properties and VTIMEZONE components are still parsed eagerly. This is useful for accessing calendar metadata or counting events without full parsing overhead. See :issue:`1050`.
+- Added :attr:`~icalendar.cal.calendar.Calendar.journals` property to :class:`~icalendar.Calendar` for API consistency with :attr:`~icalendar.cal.calendar.Calendar.events`, :attr:`~icalendar.cal.calendar.Calendar.todos`, and other component accessors. See :issue:`1050`.
```
stevepiercy: The class `Calendar` didn't link. Maybe try this?

```suggestion
- Added :class:`~icalendar.cal.lazy.LazyCalendar` for lazy subcomponent parsing. This subclass of :class:`~icalendar.cal.calendar.Calendar` defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY components until they are accessed via :attr:`~icalendar.cal.calendar.Calendar.events`, :attr:`~icalendar.cal.calendar.Calendar.todos`, :meth:`~icalendar.cal.component.Component.walk`, or other component methods. Calendar properties and VTIMEZONE components are still parsed eagerly. This is useful for accessing calendar metadata or counting events without full parsing overhead. See :issue:`1050`.
- Added :attr:`~icalendar.cal.calendar.Calendar.journals` property to :class:`~icalendar.cal.calendar.Calendar` for API consistency with :attr:`~icalendar.cal.calendar.Calendar.events`, :attr:`~icalendar.cal.calendar.Calendar.todos`, and other component accessors. See :issue:`1050`.
```

stevepiercy (dismissed):
All good for docs now. Maybe one more pass from @niccokunzmann?

SashankBhamidi:
Thanks Steve, looks good, yes. Waiting on Nicco.

src/icalendar/cal/component.py
```diff
@@ -443,12 +463,118 @@ def property_items(
...
+        params,
+        vals: str,
+        line,
+        *,
+        ignore_errors: bool = False,
+    ) -> None:
+        """Parse a single property and add it to a component.
+
+        This is a helper method that handles property parsing logic shared between
+        :meth:`from_ical` and :class:`~icalendar.cal.lazy.LazyCalendar`.
```
stevepiercy: Let's make this explicit instead of implicit. It doesn't render because it's private, but it's still good to use the fully dotted path.

```suggestion
        :meth:`~icalendar.cal.component.from_ical` and :class:`~icalendar.cal.lazy.LazyCalendar`.
```

src/icalendar/cal/component.py
```diff
@@ -443,12 +463,118 @@ def property_items(
...
+
+        Parameters:
+            component: The component to add the property to.
+            name: Property name (e.g., "SUMMARY", "DTSTART").
```
stevepiercy: No Latin abbreviations.

```suggestion
            name: Property name, for example, "SUMMARY" or "DTSTART".
```

src/icalendar/cal/component.py
```diff
@@ -443,12 +463,118 @@ def property_items(
...
+            params: Property parameters.
+            vals: Raw property value string.
+            line: The original content line (for CATEGORIES special handling).
+            ignore_errors: If True, silently ignore parsing errors.
+                          If False, raise exceptions on parse errors.
```
```suggestion
            ignore_errors: If ``True``, silently ignore parsing errors.
                          If ``False``, raise exceptions on parse errors.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,301 @@
...
+
+            # Parse a large calendar file
+            with open("large_calendar.ics", "rb") as f:
+                cal = LazyCalendar.from_ical(f.read())
+
+            # Access metadata immediately (no event parsing)
+            cal["VERSION"]  # Returns vText('2.0')
+
+            # Events are parsed on first access
+            len(cal.events)  # Returns 1500
```
stevepiercy: This example should use the pycon lexer for Python console, and its syntax should be modified accordingly. Then the code gets automatically tested.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,301 @@
...
+        The :meth:`~icalendar.cal.component.Component.to_ical` method produces correct output whether
+        components have been parsed or not.
+    """
+
+    #: Components that should be parsed lazily.
+    LAZY_COMPONENTS: ClassVar[frozenset[str]] = frozenset(
+        {"VEVENT", "VTODO", "VJOURNAL", "VFREEBUSY", "VAVAILABILITY"}
+    )
+
+    #: Components that must be parsed eagerly (needed for timezone resolution).
```
stevepiercy:
```suggestion
    #: Components that must be parsed eagerly for timezone resolution.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,301 @@
...
+        component = Component.from_ical(raw_str)
+        self.subcomponents.append(component)
+        self._parsed_indices.add(index)
+        return component
+
+    def _parse_all_of_type(self, name: str) -> None:
+        """Parse all raw components of a given type.
+
+        Parameters:
+            name: Component type name (case-insensitive, e.g., "VEVENT" or "vevent").
```
stevepiercy:
```suggestion
            name: Component type name, case-insensitive, for example, "VEVENT" or "vevent".
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,301 @@
...
+            if i not in self._parsed_indices:
+                self._parse_raw_component(i)
+
+    def walk(self, name=None, select=lambda _: True) -> list[Component]:
+        """Recursively traverse component and subcomponents.
+
+        Parses lazy components as needed.
+
+        Parameters:
+            name: If provided, only return components with this name.
```
stevepiercy:
```suggestion
            name (str | None): If provided, only return components with this name.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,301 @@
...
+            select: Optional filter function.
```
stevepiercy: Can you add a fancy type annotation for this parameter? See https://icalendar.readthedocs.io/en/latest/contribute/development.html#type-hints

CHANGES.rst
```diff
@@ -148,6 +148,9 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
+- Added :class:`~icalendar.cal.lazy.LazyCalendar` for lazy subcomponent parsing. This subclass of :class:`~icalendar.cal.calendar.Calendar` defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY components until they are accessed via :attr:`~icalendar.cal.calendar.Calendar.events`, :attr:`~icalendar.cal.calendar.Calendar.todos`, :meth:`~icalendar.cal.component.Component.walk`, or other component methods. Calendar properties and VTIMEZONE components are still parsed eagerly. This is useful for accessing calendar metadata or counting events without full parsing overhead. See :issue:`1050`.
+- Added :attr:`~icalendar.cal.calendar.Calendar.journals` property to :class:`~icalendar.cal.calendar.Calendar` for API consistency with :attr:`~icalendar.cal.calendar.Calendar.events`, :attr:`~icalendar.cal.calendar.Calendar.todos`, and other component accessors. See :issue:`1050`.
+- Added ``calendar_class``, ``lazy_calendars``, and ``any_calendars`` pytest fixtures for parametrized testing of both :class:`~icalendar.Calendar` and :class:`~icalendar.cal.lazy.LazyCalendar`. See :issue:`1093`.
```
```suggestion
- Added ``calendar_class``, ``lazy_calendars``, and ``any_calendars`` pytest fixtures for parametrized testing of both :class:`~icalendar.cal.calendar.Calendar` and :class:`~icalendar.cal.lazy.LazyCalendar`. See :issue:`1093`.
```

stevepiercy: Update against `main` and move up to the unreleased version.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+
+from datetime import datetime
+from pathlib import Path
+
+from icalendar import Calendar, Event, LazyCalendar
+
+CALENDARS_FOLDER = Path(__file__).parent / "calendars"
+
+
+def load_calendar(filename):
```
niccokunzmann: please use the fixture provided.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    "issue_1050_calendar_with_events_and_todos.ics"
+)
+SIMPLE_CALENDAR = load_calendar("issue_1050_simple_calendar.ics")
+TIMEZONE_ONLY_CALENDAR = load_calendar("issue_1050_timezone_only_calendar.ics")
+EMPTY_CALENDAR = load_calendar("issue_1050_empty_calendar.ics")
+FORWARD_REF_CALENDAR = load_calendar("issue_1050_forward_timezone_reference.ics")
+MULTI_CAL = load_calendar("issue_1050_multiple_calendars.ics")
+
+
+def test_empty_calendar():
```
niccokunzmann: use lazy_calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """Empty calendar has no subcomponents and is not lazy."""
+    cal = LazyCalendar.from_ical(EMPTY_CALENDAR)
+
+    assert str(cal["VERSION"]) == "2.0"
+    assert len(cal.subcomponents) == 0
+    assert not cal.is_lazy()
+
+
+def test_timezone_only_calendar_stays_lazy():
```
niccokunzmann: use lazy_calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """Calendar with only VTIMEZONE stays lazy until subcomponents are accessed."""
+    cal = LazyCalendar.from_ical(TIMEZONE_ONLY_CALENDAR)
+
+    assert cal.is_lazy()
+    timezones = cal.walk("VTIMEZONE")
+    assert len(timezones) == 1
+    assert cal.is_lazy()
+
+
+def test_todos_property():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """Accessing .todos parses and returns todo components."""
+    cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+    todos = cal.todos
+    assert len(todos) == 1
+    assert str(todos[0]["SUMMARY"]) == "Test Todo 1"
+
+
+def test_journals_property():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """Accessing .journals returns journal components."""
+    cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+    journals = cal.journals
+    assert len(journals) == 1
+    assert str(journals[0]["SUMMARY"]) == "Test Journal 1"
+
+
+def test_to_ical_before_access():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+
+    assert b"BEGIN:VCALENDAR" in ical_output
+    assert b"END:VCALENDAR" in ical_output
+    assert b"BEGIN:VTIMEZONE" in ical_output
+    assert b"BEGIN:VEVENT" in ical_output
+    assert b"BEGIN:VTODO" in ical_output
+    assert b"BEGIN:VJOURNAL" in ical_output
+
+
+def test_to_ical_after_partial_access():
```
niccokunzmann: - [ ] use calendars fixture
- [ ] parametrize access to different attributes
- [ ] use the same assert as above (no code duplication) OR better: Test that the ICS data stays the same in all three cases:
    1. before access
    2. after access of one subcomponent
    3. when fully parsed

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+
+    _ = cal.events
+    ical_output = cal.to_ical()
+
+    assert b"BEGIN:VTIMEZONE" in ical_output
+    assert b"BEGIN:VEVENT" in ical_output
+    assert b"BEGIN:VTODO" in ical_output
+
+
+def test_round_trip_preserves_component_counts():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    ical_output = cal.to_ical()
+
+    cal2 = Calendar.from_ical(ical_output)
+    assert len(cal2.walk("VEVENT")) == 2
+    assert len(cal2.walk("VTODO")) == 1
+    assert len(cal2.walk("VJOURNAL")) == 1
+    assert len(cal2.walk("VTIMEZONE")) == 1
+
+
+def test_round_trip_without_access():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+    ical_output = cal.to_ical()
+
+    cal2 = Calendar.from_ical(ical_output)
+    assert len(cal2.walk("VEVENT")) == 2
+    assert len(cal2.walk("VTODO")) == 1
+
+
+def test_tzid_accessible_on_lazy_events():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """VTIMEZONE is available when lazy events are parsed."""
+    cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+    event = cal.events[0]
+    assert event["DTSTART"].params.get("TZID") == "America/New_York"
+
+
+def test_forward_timezone_reference():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """VTIMEZONE defined after events in the file is still resolved correctly."""
+    cal = LazyCalendar.from_ical(FORWARD_REF_CALENDAR)
+
+    assert len(cal.events) == 1
+    dtstart = cal.events[0]["DTSTART"].dt
+    assert dtstart.tzinfo is not None
+    assert str(dtstart.tzinfo) == "Europe/Berlin"
```
niccokunzmann: we also have tzp functions that get the tzid. This should be alright, too.

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def test_add_component_appears_in_events():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    new_event.add("UID", "new-event@example.com")
+    new_event.add("SUMMARY", "New Event")
+    new_event.add("DTSTART", datetime(2025, 2, 1, 10, 0, 0))
+    cal.add_component(new_event)
+
+    uids = sorted(str(e["UID"]) for e in cal.events)
+    assert uids == ["new-event@example.com", "simple-event@example.com"]
+
+
+def test_multiple_calendars():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """from_ical(..., multiple=True) returns all calendars."""
+    cals = LazyCalendar.from_ical(MULTI_CAL, multiple=True)
+    assert len(cals) == 2
+    assert len(cals[0].events) == 1
+    assert len(cals[1].events) == 1
+
+
+def test_sequential_property_access():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+    """Accessing different component type properties in sequence returns correct results."""
+    cal = LazyCalendar.from_ical(CALENDAR_WITH_EVENTS_AND_TODOS)
+
+    assert len(cal.journals) == 1
+    assert len(cal.events) == 2
+    assert len(cal.todos) == 1
+    assert len(cal.events) == 2
+
+
+def test_mixed_property_and_walk_access():
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+def test_contains_uid_substring_match_parses_more():
+    """with_uid() may parse components whose content contains the UID as a substring.
+
+    :class:`~icalendar.parser.ical.component.ComponentIcalParser` does a substring
+    scan of raw lines to decide whether to parse a component for ``with_uid()``.
+    A DESCRIPTION that happens to contain the UID string causes that component
+    to be parsed even though it is not the target. This is intentional: parse more,
+    never miss.
+    """
+    ics = (
```
niccokunzmann: use an ICS file

src/icalendar/tests/test_issue_1050_lazy_calendar.py
```diff
@@ -0,0 +1,194 @@
...
+        b"BEGIN:VCALENDAR\r\nVERSION:2.0\r\n"
+        b"BEGIN:VEVENT\r\nUID:abc\r\nSUMMARY:target\r\nEND:VEVENT\r\n"
+        b"BEGIN:VEVENT\r\nUID:xyz\r\nDESCRIPTION:contains abc in text\r\nEND:VEVENT\r\n"
+        b"END:VCALENDAR\r\n"
+    )
+    cal = LazyCalendar.from_ical(ics)
```
niccokunzmann: use calendars fixture

src/icalendar/tests/test_issue_1050_lazy_calendar.py
SashankBhamidi: Consolidated into a single parametrized `test_to_ical_roundtrip` with `access` in `["none", "partial", "full"]`. Used component count assertions after re-parsing rather than byte equality, since `Calendar` and `LazyCalendar` may not produce identical bytes.

niccokunzmann: Yes, right... they might produce different output but should not lose information.

PR #1105, opened 2026-01-26
Document `__init__` and `__new__` methods

### LINKED ISSUE #1079: Document `__init__` methods
> I wonder where the parameters to the `__init__` function should be declared - do they show up when you put the in `__init__`? I am wondering for my own learning.

Almost always, docstrings should be in both the class and its `__init__`. The only exception is when all members in `__init__` are private. See https://stackoverflow.com/a/54189717.

To show `__init__`'s docstrings, which is the modern way I should have used in the first place:

```python
# -- Options apidoc output -------------------------------------------------
apidoc_modules = [
    {
        "path": SRC / "icalendar",
        "destination": "reference/api",
        "exclude_patterns": [
            "*tests*",
        ],
        "separate_modules": True,
        "automodule_options": {
            "ignore-module-all",
            "members",
            "show-inheritance",
            "undoc-members",
        }
    }
]


# -- Options autodoc output -------------------------------------------------
autodoc_default_options = {
    "ignore-module-all": True,
    "members": True,
    "show-inheritance": True,
    "special-members": "__init__",
    "undoc-members": True,
}


```

Also remove the `SPHINX_APIDOC_OPTIONS` environment variable from `Makefile`, and run `make apidoc` again.

See https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_options.

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/issues/1078#issuecomment-3757259287_

stevepiercy:
Closes #1079

## Description

Document `__init__` and `__new__` methods.

## Checklist

## Additional information

I see warnings that I can't explain. Help!

```console
WARNING: error while formatting signature for icalendar.cal.alarm.Alarm.Triggers.__new__: Handler <function process_signature at 0x105f32de0> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'Triggers') [autodoc]
WARNING: error while formatting signature for icalendar.prop.AdrFields.__new__: Handler <function process_signature at 0x105f32de0> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'AdrFields') [autodoc]
WARNING: error while formatting signature for icalendar.prop.NFields.__new__: Handler <function process_signature at 0x105f32de0> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'NFields') [autodoc]
```

----
📚 Documentation preview 📚: https://icalendar--1105.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -85,7 +85,8 @@ Documentation
 - Replaced "Arguments" and "Args" with "Parameters". :issue:`1076`
 - Added documentation of how to set up a development environment with git. :issue:`906`
 - Documented how to install and use pre-commit. :issue:`996`
-- Documented ``__init__`` and ``__new__`` methods. :issue:`1079`
+- Added a new rule to our docstring style guide to escape docstrings, and added a new section for type hints in the code conventions section. :issue:`1080`
+- Documented ``__init__`` methods. :issue:`1079`
 
 
 7.0.0a3 (2025-12-19)


Makefile
```diff
@@ -127,15 +126,6 @@ test: clean linkcheckbroken  ## Clean docs build, then run vale and linkcheckbro
 
 
 # development
-.PHONY: apidoc
```
niccokunzmann, line 130: If you remove this command, then the documentation for it should also go. Is there a documentation?

stevepiercy: I had edited it in https://github.com/collective/icalendar/pull/1105/files#diff-190ebed9e5332b158fdf8ecdf5501904528b686042d8c3519099719ca867e532, but not removed it.

stevepiercy:
> > Configure Sphinx to not document **new** for the three classes that yield warnings. I'm not sure if this is possible, but I can try.
> 
> Only the classes in `icalendar.prop` need the documentation of their `__new__` methods. The other modules do not need that as they do not subclass static types like `str` and explicitly override `__new__`.

Option 3 is not possible. I could remove `__new__` from the configuration, leaving just `__init__`, to remove the 3 warnings. However that would result in the `__new__` methods not getting documented.

I also tried editing `docs/reference/api/icalendar.prop.rst`:

```rst
.. automodule:: icalendar.prop
   :members:
   :show-inheritance:
   :special-members: __init__,__new__
   :undoc-members:
```

That still yielded two warnings.

```console
WARNING: error while formatting signature for icalendar.prop.AdrFields.__new__: Handler <function process_signature at 0x106133740> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'AdrFields') [autodoc]
WARNING: error while formatting signature for icalendar.prop.NFields.__new__: Handler <function process_signature at 0x106133740> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'NFields') [autodoc]
```

Additionally, every time we run `sphinx-build`, `:special-members:` gets clobbered. I'm not sure if it's possible to insert a hook in sphinx-build to modify the .rst source. It doesn't matter because the warnings would still appear.

So we can either document `__new__` and live with the warnings (not good), not document `__new__`, or rewrite the code so that they don't return the exception.

@niccokunzmann what do you think? I have no idea how to rewrite the code.

stevepiercy:
@niccokunzmann I removed `__new__` for now to make progress. I think we should make a new issue for completing `__new__` to keep it focused on the issue you reported.

CHANGES.rst
```diff
@@ -85,7 +85,7 @@ Documentation
 - Replaced "Arguments" and "Args" with "Parameters". :issue:`1076`
 - Added documentation of how to set up a development environment with git. :issue:`906`
 - Documented how to install and use pre-commit. :issue:`996`
-- Added a new rule to our docstring style guide to escape docstrings, and added a new section for type hints in the code conventions section. :issue:`1080`
+- Documented ``__init__`` and ``__new__`` methods. :issue:`1079`
```
```suggestion
- Added a new rule to our docstring style guide to escape docstrings, and added a new section for type hints in the code conventions section. :issue:`1080`
- Documented ``__init__`` methods. :issue:`1079`
```

niccokunzmann (changes requested):
I think one change and then merge it!

CHANGES.rst
stevepiercy: Hmm. Bad merge. This is complete.

stevepiercy:
@niccokunzmann one more check please per https://github.com/collective/icalendar/pull/1105/commits/d9211a191927068c18b95b376546be66ad71e82e

niccokunzmann:
Thanks, I created https://github.com/collective/icalendar/issues/1118 and it includes the documentation effort as well as moving the docstring into `__new__` once we found a solution.

PR #1113, opened 2026-01-29
add checkmark and crosses for easier visual distinction

niccokunzmann:
## Closes issue

For me, this is makes understanding it quicker.

grafik

## Description

Add checkmarks and crosses to RFC compatibility.

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1113.org.readthedocs.build/


SashankBhamidi (approved):
LGTM.

niccokunzmann:
Let's open an issue for that or a PR.

I see this:

> If non-text content is [pure decoration](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html#dfn-pure-decoration), is used only for visual formatting, or is not presented to users, then it is implemented in a way that it can be ignored by [assistive technology](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html#dfn-assistive-technology).
> from https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html

See also:
- https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/
- https://www.w3.org/WAI/standards-guidelines/act/rules/46ca7f/

stevepiercy:
In the context of a _definition term_, it's not decoration, though. The emoji is intended to create _meaning_, similar to a legend in a chart. The definition term is read back as literally "checkmark RFC `<number>`". What does that even mean? It's also not clear whether it means "incompatible" or "compatibility not guaranteed".

It might be better to combine the two definition lists into a table. Tables aren't always helpful with screen readers, either, so I tend to avoid them.

| RFC | Description | Supported |
|---|---|---|
| [RFC 2445](https://datatracker.ietf.org/doc/html/rfc2445.html) | Obsoleted by [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545.html) | ✅ Yes |

Another option would be to create subheadings each with an optional decoration. Again, I dislike emojis that are supposed to convey meaning without ARIA labels. I lean toward this one, as it provides clear separation.

## Compatible with RFCs

icalendar is compatible with the following RFC standards.

- `<list of compatible RFCs>`

## Not guaranteed compatibility

The maintainers of icalendar do not claim compatibility with the following RFC, although it might work.

- `<list of not guaranteed compatible RFC>`

niccokunzmann:
I am in favour od the headings with a listing. The checkmarks give it a bit of structure but the text itself is super confusing when you just read it straight out. So something like

SUPPORTED RFCS

- RFC 5545 <name of RFC> is supported. Explanation in next line
- ... is suported.

PR #1116, opened 2026-01-29
Add with_uid method to component (fixes #950)

### LINKED ISSUE #950: search with Component.with_uid()
Since #658 and #657, UID is quite important to reference other components.

I would like to retrieve all subcomponents with a particular UID!

```python
component = icalendar.Component()
components = component.with_uid("uid")
```

Requirements:

- [ ] the return type is a list of components, including alarms
- [ ] they all have the uid
- [ ] use walk() to generate the result
- [ ] create a nice example
- [ ] Document this not just in the API but in a how-to "Find all components with a certain UID"

likhith1253:
## Closes issue

- [x] Closes #950

## Description

Hi, I implemented the `with_uid` method in the `Component` class to find subcomponents by their UID.

I also added a test case to verify it works. I ran the tests locally and the new test passed.

## Checklist


src/icalendar/cal/component.py
```diff
@@ -405,6 +405,10 @@ def walk(self, name=None, select=lambda _: True) -> list[Component]:
             name = name.upper()
         return self._walk(name, select)
 
+    def with_uid(self, uid):
```
```suggestion
    def with_uid(self, uid:str) -> Component:
```

This adds types.

src/icalendar/cal/component.py
```diff
@@ -405,6 +405,10 @@ def walk(self, name=None, select=lambda _: True) -> list[Component]:
...
+        """Return a list of components with the given UID."""
```
niccokunzmann: Could you apply the docstring format that we like to have?

https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#python-docstrings

tests/test_issue_950.py
```diff
@@ -0,0 +1,29 @@
+import unittest
```
niccokunzmann: Could you remove this file? It seems that everything is covered by src/icalendar/tests/test_issue_950.py

src/icalendar/cal/component.py
```diff
@@ -405,6 +405,17 @@ def walk(self, name=None, select=lambda _: True) -> list[Component]:
             name = name.upper()
         return self._walk(name, select)
 
+    def with_uid(self, uid):
```
```suggestion
    def with_uid(self, uid:str) -> list[Component]:
```

niccokunzmann (changes requested):
Thanks! It appears somehow the type hints got lost. I added the suggestion again.

Have a look at the docs: https://icalendar--1116.org.readthedocs.build/en/1116/reference/api/icalendar.cal.component.html#icalendar.cal.component.Component.with_uid

Looks ready after this change!
Could you apply it?

src/icalendar/cal/component.py
likhith1253: Applied the type hints. Thanks!

PR #1131, opened 2026-02-04
Fix/issue 1129 typing

### LINKED ISSUE #1129: typing of Component.from_ical
When I run type checks, then I see this error in my project:

```
error: Call to untyped function "from_ical" of "Calendar" in typed context  [no-untyped-call]
```

The code shows no type hints:

```
    @classmethod
    def from_ical(cls, st, multiple=False):

```

I would like this to be typed. Also that it has type overloading for `multiple=False` and `multiple=True` so that the list return type is considered if multiple=True.

likhith1253:
## Closes issue

- [ ] Closes #1129 

## Description

Added proper type hints and overloaded signatures to Component.from_ical to correctly reflect the return type based on the multiple argument.
multiple=False (default) returns single Component instance (Self).
multiple=True returns list of Component instances (list[Self]).

----
📚 Documentation preview 📚: https://icalendar--1131.org.readthedocs.build/


src/icalendar/cal/component.py
```diff
@@ -403,6 +403,17 @@ def walk(self, name=None, select=lambda _: True) -> list[Component]:
             name = name.upper()
         return self._walk(name, select)
 
+    def with_uid(self, uid: str) -> list[Component]:
```
niccokunzmann, line 406: Hi, did you do another PR before? It seems that they got mixed up into one. Could you create a new PR based on the `main` branch with the relevant commits?

You can use `git cherry-pick` to just pick the commits to put into the new PR.

src/icalendar/cal/component.py
```diff
@@ -433,8 +444,16 @@ def property_items(
         properties.append(("END", v_text(self.name).to_ical()))
         return properties
 
+    @overload
```
niccokunzmann, line 447: Thanks, this looks good.

src/icalendar/cal/component.py
niccokunzmann: No need to do this if it is already merged in #1116

Issue #987, opened 2025-11-11
[Refactoring, Python] Move icalendar.prop classes into their own files


niccokunzmann:
`icalendar/prop/__init__.py` is quite big. It would be nicer to have all the classes located in their own files.

For `icalendar.prop`:

- [x] move classes into their own files
- [x] file names: without v and lowercase
- [x] import all classes in `__init__.py`
- [x] adjust documentation in accordance with icalendar.cal
- [x] make sure all links to them still work

For `icalendar.parser`:

- [x] move classes into their own files
- [x] import all classes in `__init__.py`
- [x] adjust documentation in accordance with icalendar.cal
- [x] make sure all links to them still work

stevepiercy:
To generate API .rst files automatically, run `make apidoc`. See https://icalendar.readthedocs.io/en/latest/contribute/development.html#apidoc and https://github.com/collective/icalendar/blob/main/Makefile#L131-L138.

PR #1136, opened 2026-02-05
Issue 987 - refactor icalendar.prop

niccokunzmann:
## Closes issue

- [x] Contributes to #987

## Description

Write a description of the fixes or improvements.

## Checklist

## Additional information

Documentation: https://icalendar--1136.org.readthedocs.build/en/1136/reference/api/icalendar.prop.html

### DIFF
--- a/docs/how-to/usage.rst
+++ b/docs/how-to/usage.rst
@@ -474,7 +474,7 @@ By extending the event with subcomponents, you can create multiple alarms.
     >>> event.add_component(alarm_24h_before)
 
 You can even add a recurrence, either from a dictionary or a string.
-Note that if you want to add the recurrence rule from a string, you must use the :class:`icalendar.prop.vRecur` property.
+Note that if you want to add the recurrence rule from a string, you must use the :class:`~icalendar.prop.recur.recur.vRecur` property.
 Otherwise the rule will be escaped, making it invalid.
 
 .. code-block:: pycon


niccokunzmann:
@owl-tech-sui Thanks for the suggestions. I added entries into the change log and worked on the imports. I would leave the little bits that are out-of-scope in this PR because I would not invest time into them otherwise.

The `tox.ini` change is required as the `--group` option was making my tests fail locally.

niccokunzmann:
Could you give this another review, please?

stevepiercy:
FTR, the hack I found was to checkout `main`, copy `prop.__init__.py` into a scratch file in PyCharm, then checkout this branch, and finally compare the scratch file against each file. It shows a very nice diff.

Screenshot 2026-02-08 at 7 06 31 PM

I was able to run through all the diffs quickly this way.

It looks like all the things that @owl-tech-sui caught have been resolved. I didn't find anything major. There was a minor thing, a pattern I observed, a change in the type hint return type, from a specific type, for example, `-> vWeekday`, to `-> Self`. I assume that was deliberate.

stevepiercy (approved):

CHANGES.rst
```diff
@@ -45,6 +45,9 @@ Breaking changes
 ~~~~~~~~~~~~~~~~
 
 - Drop support for Python 3.8 and 3.9. See :issue:`977`.
+- ``DURATION_REGEX`` moved from :mod:`icalendar.prop` to :mod:`icalendar.prop.dt.duration`. See :issue:`987`.
+- ``WEEKDAY_RULE`` moved from :mod:`icalendar.prop` to :mod:`icalendar.prop.recur.weekday`. See :issue:`987`.
+- Fix: Removed ``tzid_from_dt`` and ``tzid_from_tzinfo`` from :mod:`icalendar.prop` as they exist in :mod:`icalendar.timezone`.
```
SashankBhamidi: The third breaking change entry says "Fix: Removed" which reads oddly under "Breaking changes" - it's not a fix, it's an intentional removal. Also missing the issue reference that the other two entries have.

```suggestion
- Removed ``tzid_from_dt`` and ``tzid_from_tzinfo`` from :mod:`icalendar.prop` as they exist in :mod:`icalendar.timezone`. See :issue:`987`.
```

tox.ini
```diff
@@ -8,6 +8,9 @@ envlist = py310,py311,py312,py313,py314,pypy3,docs,nopytz
 [testenv]
 package = editable
 commands =
+    # Minimal pip version
+    # See https://pip.pypa.io/en/latest/user_guide/#dependency-groups
+    pip install --upgrade pip>=25.1
```
SashankBhamidi, line 13: `pip install --upgrade pip>=25.1` inside tox runs - this is useful context but the `>=25.1` gets interpreted by the shell. Did you mean `"pip>=25.1"` with quotes, or is pip smart enough to handle this? On Windows shells especially this could behave unexpectedly. Worth a quick sanity check.

src/icalendar/prop/dt/list.py
```diff
@@ -0,0 +1,105 @@
...
+        prop = jcal_property[:3]
+        dts = []
+        for value in values:
+            dts.append(vDDDTypes.from_jcal(prop + [value]))
+        return cls(
+            dts,
+            params=Parameters.from_jcal_property(jcal_property),
+        )
+
+    __hash__ = None
```
SashankBhamidi, line 102: Putting `__hash__ = None` at the very bottom after `from_jcal` is easy to miss. Class-level declarations like this are conventionally placed near the top of the class body alongside `__slots__`, type annotations, etc. Same applies to `recur.py:294`.

src/icalendar/tests/test_issue_987.py
```diff
@@ -0,0 +1,22 @@
+"""This adds tests for refactoring."""
```
SashankBhamidi: The test file is called `test_issue_987` but the docstring just says "This adds tests for refactoring." - could mention what specifically is being tested (unhashability of mutable types).

```suggestion
"""Tests for issue #987: property class modularization.

Verifies that mutable property types (vDDDLists, vRecur) are not hashable.
"""
```

SashankBhamidi (changes requested):

tox.ini
niccokunzmann: `pip install --upgrade 'pip>=25.1'` is how tox interprets it, with the single quotes. So, it should be fine, I think. If Windows devs get a problem, they can sanitize it but I would say: It seems like tox is in the middle taking care of cross-platform compatibility.

src/icalendar/prop/dt/list.py
niccokunzmann: Hm.  By meaning, it would be nice to put this next to `__eq__`.  Do you suggest we move this in this PR? Actually, it is a function and I think we are getting away from the PR purpose 2 steps: (0) refactor icalendar.prop (1 step) fix a bit of ruff complaints (2) where to put the fix that had nothing to do with (0) 
Could we just skip (2)? It is not perfect but better.

tox.ini
SashankBhamidi: Makes sense, tox handles the quoting. Thanks for clarifying.

src/icalendar/prop/dt/list.py
SashankBhamidi: Fair enough, keeping this PR focused on the refactor makes sense. Not worth the churn.

niccokunzmann:
@SashankBhamidi would you merge it?

SashankBhamidi:
I dont have permissions to bypass and merge

PR #1144, opened 2026-02-06
Add type to bug report template

niccokunzmann:
The type is `Bug` and is added to the template.
The title is kept so that we can still have the knowledge of it being a bug even if github.com/collective remove the type.


stevepiercy:
Bypassing rules to merge. LGTM.

PR #1146, opened 2026-02-07
Add type hints to remaining prop value classes

owl-tech-sui:
## Closes issue

- [ ] Partially addresses #938

## Description

Adds type hints to the remaining prop value classes: `vText`, `vCalAddress`, `vCategory`, `vGeo`, `vN`, `vOrg`, `vAdr`, `vBrokenProperty`, `vUid`, `Conference`, and `Image`. Also adds the missing `-> Self` return type on `vUri.__new__`.

Methods annotated:
- `__new__` / `__init__` parameter types and return types
- `to_ical()` return types
- `from_ical()` parameter and return types
- `__repr__()`, `__eq__()`, `__hash__()` where present
- `__iter__()` (vCategory)
- Factory methods (`new()`, `from_uri()`, `from_property_value()`, `from_parse_error()`)
- Property accessors (`ical_value`, etc.)

No behavioral changes — only type annotations were added.

## Checklist

- [x] ~~I've added or updated tests if applicable.~~ [Not applicable — type hints only]
- [x] I've run and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests). [`tox -e py312`]
- [x] ~~I've added or edited documentation.~~ [Not applicable]

----
📚 Documentation preview 📚: https://icalendar--1146.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,7 @@ Minor changes
 - Regroup dependencies in, and remove obsolete ones, from :file:`pyproject.toml`. :issue:`906`
 - Add type hints to internal helper functions. :issue:`938`
 - Add type hints to prop value classes (vBoolean, vFloat, vUri, vBinary, vInline). :issue:`938`
+- Add type hints to remaining prop value classes (vText, vCalAddress, vCategory, vGeo, vN, vOrg, vAdr, vBrokenProperty, vUid, Conference, Image). :issue:`938`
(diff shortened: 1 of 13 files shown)


niccokunzmann (approved):
Thanks! This is nice.

I was wondering sometimes if it is better to use the class as a return type or Self but since you use both, I think, this will work out well.

owl-tech-sui:
@niccokunzmann Thank you for the review and merge!

Regarding Self vs concrete class names — I used Self for methods that return an instance of the called class (via cls()), so subclasses get correct type inference. For methods that return a different type (like NamedTuple or tuple from from_ical), I used the concrete return type instead. Glad the approach works well!

PR #1168, opened 2026-02-10
ruff format src/icalendar/timezone/

niccokunzmann:
ruff format src/icalendar/timezone/

https://github.com/collective/icalendar/issues/672


SashankBhamidi (dismissed):
LGTM

PR #1169, opened 2026-02-10
Ruff: fix all lint warnings and format test files

SashankBhamidi:
## Summary
- Resolved all 180 ruff lint warnings in `src/icalendar/tests/`
- Ran `ruff format` on all 22 test files that needed it
- Updated `pyproject.toml` per-file-ignores glob from `tests/*` to `tests/**` to properly cover subdirectories
- Added `PLC0415`, `FBT003`, `INP001`, `B018` to test per-file-ignores for intentional test patterns

Related https://github.com/collective/icalendar/issues/672

### DIFF
--- a/src/icalendar/tests/test_issue_821_stdin_parse.py
+++ b/src/icalendar/tests/test_issue_821_stdin_parse.py
@@ -85,22 +85,22 @@ PROPER_OUTPUT = f"""    Organizer: organizer <organizer@test.test>
     End        : {_secondend}
     Duration   : 0:30:00
     Location   : New Amsterdam, 1010 Test Street
-    Comment    :
+    Comment    :{" "}
     Description:
      Test Description
      This one is multiline
 
-    Organizer:
+    Organizer:{" "}
     Attendees:
 
     Summary    : TEST
     Starts     : Wed May 11 00:00:00 2022
     End        : Mon May 16 00:00:00 2022
     Duration   : 5 days, 0:00:00
-    Location   :
-    Comment    :
+    Location   :{" "}
+    Comment    :{" "}
     Description:
-
+    {" "}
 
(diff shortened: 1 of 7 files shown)


src/icalendar/tests/test_issue_821_stdin_parse.py
```diff
@@ -86,22 +85,22 @@ def _local_datetime(dt: str) -> str:
     End        : {_secondend}
     Duration   : 0:30:00
     Location   : New Amsterdam, 1010 Test Street
-    Comment    : 
+    Comment    :
```
niccokunzmann: This string needs reverting, I think.

src/icalendar/tests/test_rfc_7953.py
```diff
@@ -131,15 +131,15 @@ def test_missing_info_for_end(rfc_7953_component):
     """We miss information and cannot compute the end."""
     rfc_7953_component.DURATION = timedelta(hours=1)
     with pytest.raises(IncompleteComponent) as e:
-        rfc_7953_component.end  # noqa: B018
```
niccokunzmann, line 134: Sometimes, one needs to add `RUF100` which then ignores this ignore so that it does not turn up again under different circumstances.

src/icalendar/tests/test_issue_821_stdin_parse.py
SashankBhamidi: Fixed — restored trailing spaces using f-string \ syntax so ruff won't strip them.

src/icalendar/tests/test_rfc_7953.py
SashankBhamidi: Good call — restored all B018 noqa comments with \ so they won't be stripped again.

PR #1192, opened 2026-02-11
Update maintenance notes on how to create a new release

niccokunzmann:
Changes 

- consider stable branch 7.x in the workflow
- reorder steps so that they do not clash
- use rebase for 7.x - should be checked - the commits should not be altered - must be a clean rebase or a branch delete+re-create.

Issue #1102, opened 2026-01-25
Switch to uv for development


stevepiercy:
I did not update the installation documentation as mentioned as a closing criteria in #906. That's because I found a way to install tox with a plugin uv-tox, which means I can run the tests across multiple Pythons even if I don't already have them installed on my system. And it's *fast*. It has only two prerequisites, uv and Make.

See https://github.com/sphinx-contrib/httpdomain/blob/main/CONTRIBUTING.md#install-httpdomain-for-development.

That's it. One command `make dev` and you're set up.

I'm also using uv to build and release a dist, and to run CI with a significant speed boost. I was skeptical about it, but it met my needs and was easy. See the full journey at https://github.com/sphinx-contrib/httpdomain/pull/91.

Anyway, I'd propose to overhaul dev set up to be similar to this. @niccokunzmann what do you think? 

 _Originally posted by @stevepiercy in [#906](https://github.com/collective/icalendar/issues/906#issuecomment-3796579739)_

---

Closing criteria include the following items.

- [ ] Edit `pyproject.toml`, GitHub workflows, tox configuration, and any other files for installation and development.
- [ ] Fully document the processes in https://icalendar.readthedocs.io/en/latest/contribute/development.html.

Any other items?

niccokunzmann:
This is included in v7.1.0.

PR #1193, opened 2026-02-11
Configure Git remote for collective/icalendar

niccokunzmann:
Add Git remote configuration for collective/icalendar

## Closes issue

Contributes to #1102 

## Description

The intention is that we can have documentation that allows people to easily update `main` and create PRs from that.
Another intention is to be able to update the maintenance section to a remote that is configured with the right settings.

## Checklist

## Additional information

See also:

- #1192 

## Follow up

- [ ] change maintenance section to use correct remotes

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- Add `collective` remotes for documentation in `make dev`. :pull:`1193`
+- Add `collective` remotes for documentation in `make dev`. :pr:`1193`
 
 Breaking changes
 ~~~~~~~~~~~~~~~~


Makefile
```diff
@@ -138,6 +138,10 @@ test: clean linkcheckbroken  ## Clean docs build, then run vale and linkcheckbro
 dev: .venv  ## Install required Python, create Python virtual environment, install package and development requirements
 	@uv sync --group dev
 	@pre-commit install
+	@git remote add collective https://github.com/collective/icalendar.git || git remote set-url collective https://github.com/collective/icalendar.git  # add collective/icalendar as a remote
+	@git remote add collective+ssh git@github.com:collective/icalendar.git || git remote set-url collective+ssh git@github.com:collective/icalendar.git  # add collective/icalendar as a remote
+	@git config branch.main.remote collective     # set remote tracking branch
```
stevepiercy: This breaks my git config.

stevepiercy (commented):
I'm -1 on this. It broke my git config. I had a backup, so no harm.

I use `origin`. Other people like `upstream`. I don't think we should force users into this pattern.

Makefile
niccokunzmann: Good. I removed that part. However setting the collective remote will allow the docs in the future to be copied&pasted. (without setting tracking)

Just 

> Update `main`:
> `git pull collective main
> Create a new branch
> ...

stevepiercy: Two things.

1. Why "collective"? That's the GitHub organization name, not the project name.
2. I think that most people just use the link to clone the repo from the home page, which results in the following git config. What more do users need?

```
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
        ignorecase = true
        precomposeunicode = true
[remote "origin"]
        url = https://github.com/collective/icalendar.git
        fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
        remote = origin
        merge = refs/heads/main
```

niccokunzmann: I will not justify this now. I put work in because it is of value to me. If it does not break your workflow, that is good.

stevepiercy: Yes, I understand it's good for you, but why is it good for anyone else?

It adds extra configuration that gets in my way. It's one more option to ignore.

PR #1199, opened 2026-02-14
Remove various Python standard library base classes from being documented

niccokunzmann:
str's, int's and other's methods do not need documenting in icalendar.prop and other places.

## Closes issue

Since introducing inherited members for components, we added too many methods to other classes.

Example:
grafik

## Description

The undocs members of Python std lib.

## Checklist

## Additional information

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-autoclass-inherited-members

----
📚 Documentation preview 📚: https://icalendar--1199.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,7 +47,7 @@ Bug fixes
 Documentation
 ~~~~~~~~~~~~~
 
-- Removed methods of ``str``, ``int``, and others from the documentation in :mod:`icalendar.prop`.
+- Removed methods of ``str``, ``int``, and other classes and methods in the Python standard library from the documentation.
 
 7.0.1 (2026-02-17)
 ------------------


stevepiercy:
> This is weird. The `count` is still present but not defined in the module.

There's no difference in rendering between before and after.

- https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.prop.adr.html#icalendar.prop.adr.AdrFields.count
- https://icalendar--1199.org.readthedocs.build/en/1199/reference/api/icalendar.prop.adr.html#icalendar.prop.adr.AdrFields.count

Perhaps it's inherited from elsewhere? Maybe `CaselessDict` in `recur.py`? `zoneinfo.py`? Python standard library? I couldn't a match to the string "number of occurrences of value" anywhere in icalendar or my local `.venv` directory. `/` is in the signature and represents positional parameters, so I expect it to show up, but as "count"? That _is_ weird.

Oh, wait a minute...

https://github.com/python/cpython/blob/645f5c4a737b3eab29d0b7bcd4ec5f8bd36f332d/Objects/clinic/tupleobject.c.h#L53-L57

> It works for float, but not for NamesTuple...

Maybe try lowercase `namedtuple`?

> -> `as_integer_ratio` is gone

As a data point, it's not part of icalendar, but in the Python standard library. I don't know if that helps.

https://github.com/python/cpython/blob/645f5c4a737b3eab29d0b7bcd4ec5f8bd36f332d/Objects/clinic/floatobject.c.h#L181-L195

I wonder whether we should document stuff from the standard library after all?

docs/conf.py
```diff
@@ -172,7 +172,7 @@
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
 autodoc_default_options = {
     "ignore-module-all": True,
-    "inherited-members": "CaselessDict",
+    "inherited-members": "object,CaselessDict,str,bool,NamedTuple,int,float,Enum",
```
stevepiercy: Try the negation of `inherited-members` to exclude their documentation. I think this is what you want, right?

```suggestion
    "no-inherited-members": "object,CaselessDict,str,bool,NamedTuple,int,float,Enum",
```

CHANGES.rst
```diff
@@ -47,7 +47,7 @@ Bug fixes
 Documentation
 ~~~~~~~~~~~~~
 
-- ...
+- Removed methods of ``str``, ``int``, and others from the documentation in :mod:`icalendar.prop`.
```
```suggestion
- Removed methods of ``str``, ``int``, and other classes and methods in the Python standard library from the documentation.
```

stevepiercy (changes requested):

stevepiercy (approved):

Issue #1093, opened 2026-01-20
Parametrize test suite to run with both Calendar and LazyCalendar


SashankBhamidi:
## Summary

Run all existing `Calendar` tests with both `Calendar` and `LazyCalendar` to ensure complete behavioral parity.

## Background

LazyCalendar (added in #1050) defers parsing of VEVENT, VTODO, VJOURNAL, VFREEBUSY, and VAVAILABILITY components until accessed. While it has comprehensive unit tests, running the entire Calendar test suite with LazyCalendar would provide stronger guarantees of behavioral compatibility.

## Proposed Implementation

1. **Update `conftest.py`** to parametrize the `calendars` fixture:
   ```python
   @pytest.fixture(params=[Calendar, LazyCalendar])
   def calendar_class(request):
       """Calendar class to use for tests - parametrized for Calendar and LazyCalendar."""
       return request.param
   
   @pytest.fixture(scope="module")
   def calendars(tzp, calendar_class):
       return DataSource(CALENDARS_FOLDER, calendar_class.from_ical)
   ```

2. **Mark LazyCalendar-specific tests** to skip parametrization:
   - Tests in `test_issue_1050_lazy_calendar.py` that check `_parsed_indices`, `_raw_components`, etc.
   - Use a marker like `@pytest.mark.lazy_only` to exclude from parametrization

3. **Handle expected failures**:
   - Some tests may fail with LazyCalendar initially
   - Document and fix any legitimate behavioral differences
   - Use `@pytest.mark.xfail` for known issues during migration

## Benefits

- Ensures LazyCalendar maintains full compatibility with Calendar
- Catches edge cases not covered by LazyCalendar-specific tests
- Provides confidence for users switching to LazyCalendar
- Tests both classes with real-world calendar files from the test suite

## Related

- https://github.com/collective/icalendar/pull/1090#discussion_r2709072318

## Context from Review

> Nicco suggested: "However what I recommend is adjusting the calendars fixture to be parametrized and then let all the tests run with the LazyCalendar AND the Calendar. This way, all tests run on both classes."

lcampanella98:
[...]

Issue #1050, opened 2025-12-23
Lazy subcomponent parsing for large calendars


SashankBhamidi:
`Calendar.from_ical()` parses all subcomponents (Events, Todos, etc.) upfront. For calendars with 1000+ events, this creates all Event objects immediately even if you only need metadata.

**Problem:**
- Checking a calendar's VERSION or counting events requires parsing everything
- vdirsyncer needs metadata access without full parsing overhead

**Approach:**
- Parse Calendar properties and VTIMEZONE eagerly (needed for datetime handling)
- Store raw bytes for Events/Todos/Journals
- Parse on access (`.walk()`, `.events`, etc.)

Split at `BEGIN:`/`END:` markers, parse chunks when accessed.

**Related:**
- #158 (performance aspect)
- PR #1044 (property lazy parsing)

niccokunzmann:
This is included in v7.1.0.
Can it be closed @SashankBhamidi ?

SashankBhamidi:
Yes, sorry, I may have forgotten to link this issue in the PR due to the rebase confusions.

PR #1201, opened 2026-02-16
Lazy parsing of calendars

niccokunzmann:
## Closes issue

- [ ] #158 
- [x] #1050 
- [x] #1093 

Inspired by #1090

## Description

- move `Component.from_ical` into a class.
- create subclass to re-parse components in case the calendar has timezone forward references

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1201.org.readthedocs.build/


src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+        self.initialize_parsing()
+        begin_line = "BEGIN:" + name.upper()
+        return any(
+            len(content_line) == len(begin_line) and content_line.upper() == begin_line
+            for content_line in self._content_lines
+        )
+
+    def contains_uid(self, uid: str) -> bool:
+        self.initialize_parsing()
+        return any(uid in line for line in self._content_lines)
```
SashankBhamidi, line 76: `uid in line` is a plain substring check across every raw content line. A UID like
`event-1` will match `DESCRIPTION:See event-1 for details`, causing unnecessary parsing.
`contains_component` just above it uses an exact length+equality check. Same care should
apply here, and there's no docstring either.

```suggestion
    def contains_uid(self, uid: str) -> bool:
        """Check if any line in this component matches the given UID property."""
        self.initialize_parsing()
        uid_line = f"UID:{uid}"
        return any(
            len(line) == len(uid_line) and line.upper() == uid_line.upper()
            for line in self._content_lines
        )
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+        if not self._stack:
+            # The stack is currently empty, the input must be invalid
+            raise ValueError("END encountered without an accompanying BEGIN!")
+
+        component = self._stack.pop()
+        if not self._stack:  # we are at the end
+            self._components.append(component)
+        else:
+            self._stack[-1].add_component(component)
+        if vals == "VTIMEZONE" and "TZID" in component:
```
SashankBhamidi: Every other component name check in this file goes through `.upper()` (lines 96, 147, etc.).
A hand-crafted `.ics` with `END:vtimezone` won't cache the timezone because of this.

```suggestion
        if vals.upper() == "VTIMEZONE" and "TZID" in component:
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        content_lines = [Contentline(f"BEGIN:{component_name}")]
+        for line in self._content_lines_iterator:
+            content_lines.append(line)
+            if (
+                line[:4].upper() == "END:"
+                and line[4:].strip().upper() == component_name
+            ):
+                break
+        assert self.component is not None, "We assume a component is already started."
```
SashankBhamidi: `handle_lazy_begin_component` is only called when `self.component` is already non-None
(the caller checks `not self.component` first), so the assert is unreachable today. But
it documents a fragile assumption and if someone calls `handle_lazy_begin_component`
directly or refactors `handle_begin_component`, it silently does nothing under `python -O`
instead of failing loudly.

```suggestion
        if self.component is None:
            raise ValueError(
                f"BEGIN:{component_name} encountered outside of a parent component."
            )
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Parameters:
+            name: The name of the component to walk to.
+        """
+        assert name is not None
+        if name == self.name or (
+            self._parser and self._parser.contains_component(name)
+        ):
+            return self.parse().walk(name)
+        return []
```
SashankBhamidi, line 153: `Component.walk` has `walk(self, name=None, select=lambda _: True)`. `LazySubcomponent.walk`
only accepts `name` with no default and no `select`. Anything calling
`subcomponent.walk(name, select)` on a `LazySubcomponent` raises `TypeError`. The `assert`
on line 141 also enforces `name is not None`, which is inconsistent with the base contract.

```suggestion
    def walk(self, name: str | None = None, select=lambda _: True) -> list[Component]:
        """Walk through this component, parsing only if necessary."""
        if name is None:
            return self.parse().walk(name, select)
        if name == self.name or (
            self._parser and self._parser.contains_component(name)
        ):
            return self.parse().walk(name, select)
        return []
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def walk(self, name: str) -> list[Component]:
+        """Walk through this component.
+
+        This only parses the component if necessary.
+
+        Parameters:
+            name: The name of the component to walk to.
+        """
+        assert name is not None
```
SashankBhamidi: Same issue as line 65. The suggestion above replaces both.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def set_components(
+        self, components: list[Component]
+    ) -> ParsedSubcomponentsStrategy:
+        """Set the subcomponents of the calendar."""
+        return ParsedSubcomponentsStrategy().set_components(components)
+
+    def add_component(
+        self, component: Component | LazySubcomponent
+    ) -> LazyCalendarIcalParser:
```
SashankBhamidi: The method returns `self`, which is a `LazySubcomponentsStrategy`. 
LazyCalendarIcalParser` is a parser class and completely unrelated.

```suggestion
    def add_component(
        self, component: Component | LazySubcomponent
    ) -> LazySubcomponentsStrategy:
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+        for component in self._components:
+            if component.name in self.initial_components_to_parse:
+                component.parse()
+
+    def walk(
+        self, name: str | None
+    ) -> tuple[LazySubcomponentsStrategy, list[Component]]:
+        """Get the subcomponents of the calendar with the given name.
+
+        Parse only the minumal number of subcomponents.
```
SashankBhamidi:
```suggestion
        Parse only the minimal number of subcomponents.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+        self.parse_initial_components()
+        result = []
+        for component in self._components:
+            result += component.walk(name)
+        return self, result
+
+    def with_uid(self, uid: str) -> tuple[LazySubcomponentsStrategy, list[Component]]:
+        """Get the subcomponents of the calendar with the given uid.
+
+        Parse only the minumal number of subcomponents.
```
SashankBhamidi:
```suggestion
        Parse only the minimal number of subcomponents.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+class InitialSubcomponentsStrategy:
+    """Initial strategy for the calendar.
+
+    No subcomponents.
+    """
+
+    def set_components(self, components: list[Component]) -> LazySubcomponentsStrategy:
+        """Set the subcomponents of the calendar."""
+        assert components == []
+        return LazySubcomponentsStrategy()
```
SashankBhamidi, line 158: Stripped with `-O`, and even when it does fire the `AssertionError` gives nothing useful.
If someone does `calendar.subcomponents = existing_list` on a freshly created `LazyCalendar`,
they get a confusing crash.

```suggestion
    def set_components(self, components: list[Component]) -> LazySubcomponentsStrategy:
        if components:
            raise ValueError(
                "Cannot set subcomponents on an uninitialised LazyCalendar. "
                "Parse it first or add components via add_component()."
            )
        return LazySubcomponentsStrategy()
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+            False
+
+    """
+
+    _subcomponents: (
+        LazySubcomponentsStrategy
+        | ParsedSubcomponentsStrategy
+        | InitialSubcomponentsStrategy
+    )
+    """The stategy pattern for subcomponents of the calendar."""
```
SashankBhamidi:
```suggestion
    """The strategy pattern for subcomponents of the calendar."""
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+    def is_lazy(self):
+        """Whether the subcomponents will be parsed lazily.
+
+        .. note:: If you believe that the calendar parses more than it should,
+            please open an issue.
+
+        Returns:
+            False if all subcomponents are parsed.
+            True if subcomponents are parsed before they get accessed.
+        """
```
SashankBhamidi: Missing `-> bool`. The Returns description is also backwards. `True` means they haven't
been parsed yet, not that they are "parsed before they get accessed."

```suggestion
    def is_lazy(self) -> bool:
        """Whether the subcomponents will be parsed lazily.

        .. note:: If you believe that the calendar parses more than it should,
            please open an issue.

        Returns:
            True if subcomponents are deferred and not yet parsed.
            False if all subcomponents have been parsed.
        """
        return self._subcomponents.is_lazy()
```

src/icalendar/parser/ical/calendar.py
```diff
@@ -0,0 +1,37 @@
...
+
+        This handles timezone forward references.
+        """
+        all_timezones_so_far = True
+        for comp in self._components:
+            for component in comp.subcomponents:
+                if component.name == "VTIMEZONE":
+                    if not all_timezones_so_far:
+                        # If a preceding component refers to a VTIMEZONE defined
+                        # later in the source st
```
SashankBhamidi: Comment got cut off mid-word. Should probably be "source string" or "source stream."

src/icalendar/tests/test_issue_1050_lazy_parsing.py
```diff
@@ -0,0 +1,221 @@
...
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+from icalendar import Alarm, Event, LazyCalendar
+from icalendar.parser.ical.lazy import LazySubcomponent
+
+if TYPE_CHECKING:
+    from calendar import Calendar
```
SashankBhamidi: This pulls in Python's stdlib `calendar.Calendar`, not `icalendar.Calendar`. The type
annotation on line 156 (`calendar: Calendar`) is pointing at the wrong class entirely.

```suggestion
    if TYPE_CHECKING:
        from icalendar import Calendar
```

src/icalendar/tests/test_issue_1050_lazy_parsing.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@pytest.fixture
+def lazy_subcomponent(mock_parser):
+    """A lazy subcomponent that uses the mock parser."""
+    return LazySubcomponent("VEVENT", mock_parser)
+
+
+def test_lazy_subcomponent_is_not_parsed_when_created(mock_parser, lazy_subcomponent):
+    """The component is inisially not parsed."""
```
SashankBhamidi:
```suggestion
    """The component is initially not parsed."""
```

src/icalendar/tests/test_issue_1050_lazy_parsing.py
```diff
@@ -0,0 +1,221 @@
...
+
+def test_when_parsed_parser_is_freed(lazy_subcomponent, mock_parser):
+    """Test that we can remove references to the content lines."""
+    component = lazy_subcomponent.parse()
+    assert component == "mock_component"
+    assert lazy_subcomponent._parser is None
+
+
+def test_big_calendar_returns_a_calendar(lazy_calendars):
+    """Test that the BigCalendar returns a Calendar."""
```
SashankBhamidi: The class was renamed to `LazyCalendar`. Both the function name and docstring are left over
from an earlier iteration.

```suggestion
def test_lazy_calendar_returns_a_calendar(lazy_calendars):
    """LazyCalendar.from_ical returns a LazyCalendar instance."""
```

src/icalendar/tests/test_issue_1050_lazy_parsing.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@pytest.mark.parametrize(
+    "component_name", ["VEVENT", "VTODO", "VALARM", "X-COMPONENT", "VTIMEZONE"]
+)
+def test_can_only_parse_calendar_components(component_name):
+    """Test that we still get a component even if the string contains other components."""
+    component = LazyCalendar.from_ical(f"BEGIN:{component_name}\nEND:{component_name}")
+    assert component.name == component_name
+    assert not component.is_lazy()
```
SashankBhamidi: `LazyCalendar.from_ical` is called with a non-VCALENDAR string. `LazyCalendarIcalParser`
handles this via the `not self.component` branch in `handle_begin_component` so there's no
parent on the stack, VEVENT/VTODO/etc. gets parsed instantly as a plain component and not
wrapped in a `LazyCalendar`. `is_lazy()` returns `False` because `Component.is_lazy()`
always returns `False`. The test name says "can only parse calendar components" but what it
actually tests is that non-VCALENDAR input degrades gracefully. If that's the intent, the
name and docstring should say so explicitly.

CHANGES.rst
SashankBhamidi, line 1: `main` currently has `7.0.3 (unreleased)`. This PR bumps to `7.1.0`. `LazyCalendar` is a
new public class so a minor bump is correct per semver. But `7.0.3` changes on `main` need
to either ship first or get folded into `7.1.0` here, otherwise they'd be skipped.

SashankBhamidi (changes requested):
Left some issues here. `contains_uid` does a substring match so it can trigger unnecessary parsing on description text that happens to contain the UID. There's also a case-sensitive `VTIMEZONE` check that misses lowercase input. A few `assert`s that get silently dropped with `python -O`, a signature mismatch on `LazySubcomponent.walk`, some typos, and the version on `main` needs to be reconciled before this merges.

I'm going to rebase #1090 on top of this once it merges and add tests there, so I'll hold off on adding them here.

docs/reference/api/icalendar.cal.rst
```diff
@@ -17,6 +17,7 @@ Submodules
    icalendar.cal.examples
    icalendar.cal.free_busy
    icalendar.cal.journal
+   icalendar.cal.lazy
```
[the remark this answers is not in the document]
stevepiercy: @niccokunzmann please add `icalendar.cal.lazy.rst` and push.

docs/reference/api/icalendar.parser.rst
```diff
@@ -1,6 +1,14 @@
...
 ========================
 
+Subpackages
+-----------
+
+.. toctree::
+   :maxdepth: 4
+
+   icalendar.parser.ical
+
```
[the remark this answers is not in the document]
stevepiercy: @niccokunzmann please add `icalendar.parser.ical.rst` and push.

stevepiercy (changes requested):
I did a quick cursory review. Without the two missing files, I don't feel comfortable doing a complete review.

docs/reference/api/icalendar.cal.rst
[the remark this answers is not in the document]
niccokunzmann: done

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+        return self._subcomponents.is_lazy()
+
+    def _walk(
+        self, name: str | None, select: callable[[Component], bool]
+    ) -> list[Component]:
+        self._subcomponents, result = self._subcomponents.walk(name)
```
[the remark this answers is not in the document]
niccokunzmann: 3757ab19

src/icalendar/parser/ical/calendar.py
niccokunzmann: b5a7b056

src/icalendar/parser/ical/component.py
niccokunzmann: That fails if the UID has parameters. I would rather parse more. Also, UID is supposed to be a UUID from now on. As such - if it appears it is alright. Your suggestion could skip cases. It is important that we find all. It is ok to parse more. What do you think?

I guess, at least tests should be added that if the UID is inside e.g. because it is referenced, then the component should not be returned.

Also, UID is case-sensitive and I would like to avoid transforming operations on the content lines to speed up execution.

Test: 81aee19e

src/icalendar/parser/ical/component.py
niccokunzmann: Nice catch!

src/icalendar/parser/ical/lazy.py
niccokunzmann: That is not inconsistent. The _subcomponents only have a subset of functionality. This reduces code duplication. The check is performed by the component. Also name is str.

src/icalendar/tests/test_equality.py
```diff
@@ -35,6 +35,26 @@ class UnknownTimeZoneError(Exception):
...
+            )
+    if isinstance(actual_value, Component) and isinstance(expected_value, Component):
+        assert len(actual_value.subcomponents) == len(expected_value.subcomponents), (
+            "Number of subcomponents is not equal."
+        )
+        for sub1, sub2 in zip(
+            actual_value.subcomponents, expected_value.subcomponents, strict=True
+        ):
+            print(f"Comparing subcomponents: \n\t{sub1} and \n\t{sub2}")
     assert actual_value == expected_value
```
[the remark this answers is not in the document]
niccokunzmann: f8d9761c

src/icalendar/tests/test_issue_1050_lazy_parsing.py
niccokunzmann: f877c8d5

src/icalendar/attr.py
```diff
@@ -900,6 +900,7 @@ def create_single_property(
     type_def: type,
     doc: str,
     vProp: type = vDDDTypes,  # noqa: N803
+    convert: callable[object, object] | None = None,
 ):
```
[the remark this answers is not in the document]
niccokunzmann: 55dec6be

CHANGES.rst
niccokunzmann: 87e57d80

src/icalendar/parser/ical/lazy.py
niccokunzmann: 721366e8

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,300 @@
...
+    def walk(
+        self, name: str | None
+    ) -> tuple[LazySubcomponentsStrategy, list[Component]]:
+        """Get the subcomponents of the calendar with the given name.
+
+        Parse only the minumal number of subcomponents.
+        """
+        if name is None:
+            return self.as_parsed.walk(name)
+        self.parse_initial_components()
```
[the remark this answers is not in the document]
niccokunzmann: 6e8bb7f2

src/icalendar/cal/lazy.py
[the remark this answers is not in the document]
niccokunzmann: 8c01774f

src/icalendar/parser/ical/component.py
SashankBhamidi: Makes sense. If a UID appears as a substring in a description it's a false positive but parsing one extra component is fine. I'll add tests for that case when I rebase #1090 on top of this.

src/icalendar/parser/ical/lazy.py
SashankBhamidi: Still think this is worth fixing, if `LazySubcomponent` ever leaks out of internal use or someone passes `None` expecting `Component.walk` behavior, it breaks silently. Up to you.

src/icalendar/parser/ical/lazy.py
SashankBhamidi: Thanks, cleaner.

src/icalendar/parser/ical/lazy.py
niccokunzmann: It is like this:

```
    def localize(
        self, dt: datetime.date, tz: datetime.tzinfo | str | None
    ) -> datetime.datetime:
        """Localize a datetime to a timezone."""
        if isinstance(tz, str):
            tz = self.timezone(tz)
        if tz is None:
            return dt.replace(tzinfo=None)
        return self.__provider.localize(to_datetime(dt), tz)
 ```
 
The idea is to have a compatibility interface for the developers (Component.walk) and reduce complexity inside to allow an easier adaptation because the different strategies have less complexity to deal with.
 
So, I would say, it is fine: The intention is to internally provide the minimal functionality, name is clearly documented as str and as component name.

Comment (niccokunzmann) [reply src/icalendar/parser/ical/lazy.py:153]: The strategy pattern is also used in the timezone subpackage and funny enough someone asked why I do this: https://github.com/collective/icalendar/pull/1276#discussion_r2959908354 At the same time! #synchronicity

**Review by SashankBhamidi — DISMISSED** [review]
There's a duplicate return `self._subcomponents.is_lazy()` at lines 263-264 in `cal/lazy.py`, looks like it crept in during the edits. One of them needs to go. 

Good work, thanks!

**Review by SashankBhamidi — DISMISSED** [review]
LGTM, can be merged unless @stevepiercy has something to add.

File: src/icalendar/cal/lazy.py [hunk 23]
```diff
@@ -0,0 +1,305 @@
...
+        self._components = components
+        return self
+
+    def add_component(self, component: Component) -> ParsedSubcomponentsStrategy:
+        """Add a component to the calendar."""
+        self._components.append(component.parse())
+        return self
+
+    def is_lazy(self) -> Literal[False]:
+        """Return whether the components are lazy."""
```
stevepiercy: I'm not clear. Does this mean that the subcomponents can be parsed lazily or were parsed lazily? I think the latter. Please verify.

```suggestion
        """Returns ``False`` because subcomponents are not lazily parsed."""
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+        for component in self._components:
+            result += component.with_uid(name)
+        return self, result
+
+
+class LazySubcomponentsStrategy:
+    """All the subcomponents are parsed lazily."""
+
+    initial_components_to_parse: tuple[str, ...] = ("VTIMEZONE",)
+    """The components that are parsed immediately, instead of lazily."""
```
stevepiercy: These two docstrings conflict with each other. It seems that the first should be revised for consistency.

```
    """Lazy parse all subcomponents except those explicitly set to immediately parse."""
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def __init__(self):
+        self._components: list[LazySubcomponent | Component] = []
+        self._initial_parsed: bool = False
+
+    @property
+    def as_parsed(self) -> ParsedSubcomponentsStrategy:
+        """The same components just fully parsed."""
```
stevepiercy:
```suggestion
        """Return the components as parsed."""
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def add_component(
+        self, component: Component | LazySubcomponent
+    ) -> LazySubcomponentsStrategy:
+        """Add a component to the calendar."""
+        self._components.append(component)
+        return self
+
+    def is_lazy(self) -> bool:
+        """Return whether the components are lazy."""
```
stevepiercy: With the foregoing exceptions according to line 67, correct?

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+        if name is None:
+            return self.as_parsed.walk(name)
+        self.parse_initial_components()
+        result = []
+        for component in self._components:
+            result += component.walk(name)
+        return self, result
+
+    def with_uid(self, uid: str) -> tuple[LazySubcomponentsStrategy, list[Component]]:
+        """Get the subcomponents of the calendar with the given uid.
```
stevepiercy:
```suggestion
        """Get the subcomponents of the calendar with the given ``uid``.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+class LazyCalendar(Calendar):
+    """A calendar that can handle big files.
+
+    Subcomponents of this calendar are evaluated lazily,
+    meaning that they are not parsed until they are accessed.
+    This allows the calendar to handle large files without
+    consuming too much memory or time.
+
+    All the properties of the calendar are parsed immediately,
+    just the subcomponents are not.
```
stevepiercy: Can you clarify? For example, if a subcomponent has a property, will its property be parsed immediatly?


```suggestion
    All properties of the calendar components are parsed immediately.
    Subcomponents and their properties are parsed lazily.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+
+    Examples:
+
+        By accessing the :attr:`~icalendar.cal.calendar.Calendar.events` of the calendar,
+        only :class:`~icalendar.cal.event.Event` and
+        :class:`~icalendar.cal.timezone.Timezone` are parsed.
```
stevepiercy:
```suggestion
        :class:`~icalendar.cal.timezone.Timezone` are immediately parsed.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+
+        .. code-block:: pycon
+
+            >>> from icalendar import LazyCalendar
+            >>> calendar = LazyCalendar.example("issue_1050_all_components")
+            >>> len(calendar.events) == 1
+            True
+
+        Other subcomponents are not parsed.
```
stevepiercy:
```suggestion
        The calendar's subcomponents were not parsed because they were not accessed.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+        The calendar is still lazy.
+
+            >>> calendar.is_lazy()
+            True
+
+        After you access all :attr:`subcomponents` of the calendar,
+        the whole calendar has been parsed.
```
stevepiercy:
```suggestion
        When you access all :attr:`subcomponents` of the calendar,
        for example by getting their count, the entire calendar is
        parsed and becomes not lazy.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+        factory = ComponentFactory()
+        factory.add_component_class(cls)
+        return factory
+
+    def add_component(self, component: Component) -> None:
+        """Add a component to the calendar.
+
+        Use this instead of appending to
+        :attr:`~icalendar.cal.lazy.LazyCalendar.subcomponents`'
+        as it does not parse the whole calendar.
```
stevepiercy:
```suggestion
        :attr:`~icalendar.cal.lazy.LazyCalendar.subcomponents`,
        as the latter does not parse the whole calendar.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        self._subcomponents = self._subcomponents.add_component(component)
+
+    def is_lazy(self) -> bool:
+        """Whether the subcomponents will be parsed lazily.
+
+        .. note:: If you believe that the calendar parses more than it should,
+            please open an issue.
```
stevepiercy:
```suggestion
            please `open an issue <https://github.com/collective/icalendar/issues/new?template=bug_report.md>`_.
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Returns:
+            True if subcomponents are deferred and not yet parsed.
+            False if all subcomponents have been parsed.
```
stevepiercy:
```suggestion
            ``True`` if subcomponents are deferred and not yet parsed.
            ``False`` if all subcomponents have been parsed.
```

src/icalendar/parser/ical/calendar.py
```diff
@@ -0,0 +1,37 @@
...
+                        # If a preceding component refers to a VTIMEZONE defined
+                        # later in the source string
+                        # (forward references are allowed by RFC 5545), then the
+                        # earlier component may have
+                        # the wrong timezone attached.
+                        # However, during computation of comps, all VTIMEZONEs
+                        # observed do end up in
+                        # the timezone cache. So simply re-running from_ical will
+                        # rely on the cache
+                        # for those forward references to produce the correct result.
```
stevepiercy: Exact same content, but without the jagged line endings, to make it more readable.

```suggestion
                        # If a preceding component refers to a VTIMEZONE
                        # defined later in the source string (forward
                        # references are allowed by RFC 5545), then the
                        # earlier component may have the wrong timezone
                        # attached.
                        # However, during computation of comps, all VTIMEZONEs
                        # observed do end up in the timezone cache. So simply
                        # re-running from_ical will rely on the cache for those
                        # forward references to produce the correct result.
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
+"""Parsing a component's ical data."""
```
stevepiercy:
```suggestion
"""Parsing a component's iCalendar data."""
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+from icalendar.timezone import tzp
+
+if TYPE_CHECKING:
+    from icalendar.cal import Component, ComponentFactory
+    from icalendar.parser.parameter import Parameters
+    from icalendar.prop import VPROPERTY, TypesFactory
+
+
+class ComponentIcalParser:
+    """A parser for a component's ical data.
```
stevepiercy:
```suggestion
    """A parser for a component's iCalendar data.
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+
+    This uses the template method pattern,
+    where the main parsing logic can be refined in the base class,
```
stevepiercy: Rewrap, punctuation.

```suggestion
    This uses the template method pattern, where the main parsing
    logic can be refined in subclasses.
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+        "DTSTART",
+        "DTEND",
+        "RECURRENCE-ID",
+        "DUE",
+        "RDATE",
+        "EXDATE",
+    )
+    """Names to check for TZID parameter when parsing datetimes.
+
+    Their from_ical methods take an optional tzid argument,
```
stevepiercy:
```suggestion
    Their ``from_ical`` methods take an optional ``tzid`` argument,
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+    def __init__(
+        self,
+        data: bytes | str | list[Contentline],
+        component_factory: ComponentFactory,
+        types_factory: TypesFactory,
+    ):
+        """Initialize the parser with the raw data.
+
+        Parameters:
+            data: The raw ical data to parse, as bytes or a list of content lines.
```
stevepiercy:
```suggestion
            data: The raw iCalendar data to parse, either as bytes or a list of content lines.
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+    def contains_component(self, name: str) -> bool:
+        """Check if the parser contains a component."""
+        self.initialize_parsing()
+        begin_line = "BEGIN:" + name.upper()
+        return any(
+            len(content_line) == len(begin_line) and content_line.upper() == begin_line
+            for content_line in self._content_lines
+        )
+
+    def contains_uid(self, uid: str) -> bool:
```
stevepiercy, line 69:
```suggestion
    def contains_uid(self, uid: str) -> bool:
    """Determines whether the component contains a ``uid``.
    
    Returns:
        ``True`` if the component contains a ``uid``, else ``False``.
    """
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+    def get_factory_for_property(self, name: str, params: Parameters) -> VPROPERTY:
+        """Get the factory for a property."""
+        return self._types_factory.for_property(name, params.value)
+
+    def handle_property(
+        self, name: str, params: Parameters, vals: str, line: Contentline
+    ) -> None:
+        """Handle a property line.
+
+        We are adding properties to the current top of the stack
```
stevepiercy:
```suggestion
        Add properties to the top of the current stack.
```

src/icalendar/parser/ical/component.py
```diff
@@ -0,0 +1,282 @@
...
+        self.component.errors.append((name, str(exception)))
+        self.component.add(name, broken_prop, encode=0)
+
+    def handle_categories(
+        self, params: Parameters, vals: str, line: Contentline
+    ) -> bool:
+        """Handle the special case of CATEGORIES property.
+
+        Returns:
+            True if handled, False if not.
```
stevepiercy:
```suggestion
            ``True`` if handled, else ``False``.
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+    """A parser for calendar components.
+
+    Instead of parsing the components, LazyComponents are created.
+    Parsing can happen on demand.
+
+    We assume that a calendar grows over time and that the subcomponents
+    of the calendar become more in number.
+    This optimization allows us to parse big calendar files without consuming
+    more memory than necessary and reducing the initial time it takes to access
+    meta data.
```
stevepiercy:
```suggestion
    A calendar may grow over time, with its subcomponents greatly increasing
    in number, and requiring more memory and time to fully parse. This
    optimization lazily parses calendar files without consuming more memory
    than necessary, reducing the initial time it takes to access meta data.
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    parse_instantly: ClassVar[tuple[str, ...]] = ("VCALENDAR",)
+    """Parse these components instantly, instead of lazily.
```
stevepiercy:
```suggestion
    """Parse these components immediately, instead of lazily.
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+
+    All other components are parsed lazily.
+    """
+
+    def handle_begin_component(self, vals):
+        """Begin a new component.
+
+        This could well be the first component.
```
stevepiercy:
```suggestion
        This may be the first component.
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        c_name = vals.upper()
+        if (
+            c_name in self.parse_instantly
+            or not self.component
+            or not self.component.is_lazy()
+        ):
+            # these components are parsed instantly
```
stevepiercy:
```suggestion
            # these components are parsed immediately
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+            super().handle_begin_component(vals)
+        else:
+            self.handle_lazy_begin_component(c_name)
+
+    def handle_lazy_begin_component(self, component_name: str) -> None:
+        """Begin a new component, but do not parse it yet.
+
+        Parameters:
+            component_name:
+                The upper case name of the component, for example, ``"VEVENT"``.
```
stevepiercy, line 54: This made me smile, no Latin abbreviations for "for example"!

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+        """Initialize the lazy subcomponent with the raw data."""
+        self._name = name
+        self._parser = parser
+        self._component: Component | None = None
+
+    @property
+    def name(self) -> str:
+        """The name of the subcomponent.
+
+        This is upper case, as per RFC 5545.
```
stevepiercy:
```suggestion
        The name is uppercased, per :rfc:`5545#section-2.1`.
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+                raise ValueError(
+                    f"Expected exactly one component in the subcomponent, "
+                    f"but got {len(components)}."
+                )
+            self._component = components[0]
+            self._parser = None  # free memory
+        return self._component
+
+    def is_lazy(self) -> bool:
+        """You need to call :meth:`parse` to get the fully parsed component."""
```
stevepiercy:
```suggestion
        """Return whether the subcomponents were accessed and parsed lazily.

        Call :meth:`parse` to get the fully parsed component.
        """
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+    def __repr__(self) -> str:
+        return f"LazySubcomponent(name={self._name}, parsed={self.is_parsed()})"
+
+    def walk(self, name: str) -> list[Component]:
+        """Walk through this component.
+
+        This only parses the component if necessary.
+
+        Parameters:
+            name: The name of the component to walk to.
```
stevepiercy:
```suggestion
            name: The name of the component to walk through.
```

src/icalendar/parser/ical/lazy.py
```diff
@@ -0,0 +1,165 @@
...
+        if not isinstance(name, str):
+            raise TypeError("name must be a string.")
+        if name == self.name or (
+            self._parser and self._parser.contains_component(name)
+        ):
+            return self.parse().walk(name)
+        return []
+
+    def with_uid(self, uid: str) -> list[Component]:
+        """Return the components with the uid.
```
stevepiercy:
```suggestion
        """Return the components containing the given ``uid``.
```

src/icalendar/tests/conftest.py
```diff
@@ -130,34 +150,46 @@ def multiple(self):
...
 
-@pytest.fixture(scope="module")
-def calendars(tzp):
-    return DataSource(CALENDARS_FOLDER, Calendar.from_ical)
+@pytest.fixture(scope="module", params=[Calendar, LazyCalendar])
+def calendars(tzp, request):
+    """Return the data source for calendar files.
+
+    We have two ways of parsing those:
+    - Instantly
```
stevepiercy:
```suggestion
    - Immediately
```

CHANGES.rst
```diff
@@ -24,6 +24,11 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
+- Added :attr:`Event.RECURRENCE_ID <icalendar.cal.event.Event.RECURRENCE_ID>`,
+  :attr:`Todo.RECURRENCE_ID <icalendar.cal.todo.Todo.RECURRENCE_ID>` and
```
```suggestion
  :attr:`Todo.RECURRENCE_ID <icalendar.cal.todo.Todo.RECURRENCE_ID>`, and
```

stevepiercy (changes requested):
To review:
 
- [ ] https://icalendar--1201.org.readthedocs.build/en/1201/reference/api/icalendar.cal.lazy.html
- [ ] https://icalendar--1201.org.readthedocs.build/en/1201/reference/api/icalendar.parser.ical.component.html
- [ ] https://icalendar--1201.org.readthedocs.build/en/1201/reference/api/icalendar.parser.ical.lazy.html

src/icalendar/cal/lazy.py
niccokunzmann:
```suggestion
        """Returns ``False`` as no deferred parsing of subcomponents takes place."""
```

How is that?

src/icalendar/cal/lazy.py
niccokunzmann: 93ac0a17 - how is that?

src/icalendar/cal/lazy.py
niccokunzmann:
```suggestion
        """Return the parsed components."""
```

How is that?

src/icalendar/cal/lazy.py
niccokunzmann:
```suggestion
        """Return whether deferred parsing of components can take place."""
```

How is this?

src/icalendar/cal/lazy.py
niccokunzmann:
```suggestion
    All properties of the calendar component are parsed immediately.
    Subcomponents and their properties are parsed lazily.
```

I would do it without the `s` because we expect to only have one calendar and I it is less confusing to read for me as a non-native speaker (apostrophe or without, with the s it sounds like calendars' components)

src/icalendar/cal/lazy.py
niccokunzmann: That is not quiet true because some subcomponents have not been parsed. Maybe it would be good to make an example here?
E.g. ``VTODO`` and ``VJOURNAL`` are still not parsed. Only the _other_ ones ....

src/icalendar/parser/ical/component.py
niccokunzmann: content correction: subclasses

src/icalendar/parser/ical/lazy.py
niccokunzmann: Only the components with this name are returned. All components might be walked though.


```suggestion
            name: All components in the result have this name.
```

How is this?

src/icalendar/cal/lazy.py
stevepiercy: For consistency, I'd prefer to use "lazily" or "lazy", instead of "deferred".

```suggestion
        """Returns ``False`` because subcomponents are not lazily parsed."""
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,306 @@
...
+    ) -> tuple[ParsedSubcomponentsStrategy, list[Component]]:
+        """Get the subcomponents of the calendar with the given uid."""
+        result = []
+        for component in self._components:
+            result += component.with_uid(name)
+        return self, result
+
+
+class LazySubcomponentsStrategy:
+    """Allow lazy parsing of subcomponents on access."""
```
stevepiercy: I think this aligns more closely as a strategy.

```suggestion
    """Parse subcomponents only when accessed."""
```

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,306 @@
...
+
+    initial_components_to_parse: tuple[str, ...] = ("VTIMEZONE",)
+    """Before parsing other subcomponents, these will be parsed first."""
```
stevepiercy: More compact, and puts the thing to do first.

```suggestion
    """Parse these subcomponents before any others."""
```

src/icalendar/cal/lazy.py
stevepiercy: I think this aligns with a clearer version of line 67.

```suggestion
        """Return whether the subcomponents may be lazily parsed."""
```

src/icalendar/cal/lazy.py
stevepiercy: Yes, I think an example to show this point would be good. The next sentence is a related, but separate point.

src/icalendar/cal/lazy.py
```diff
@@ -0,0 +1,306 @@
...
+            >>> from icalendar import LazyCalendar
+            >>> calendar = LazyCalendar.example("issue_1050_all_components")
+            >>> len(calendar.events) == 1
+            True
+
+        Other subcomponents are not parsed.
+        The calendar is still lazy.
+
+            >>> calendar.is_lazy()
+            True
```
stevepiercy, line 189: FTR, when I use PyCharm's Python console, I get `False`. When I use `.venv/bin/python3` to enter the console from the terminal, I get `True`. They both use the same Python in the virtual environment. I attempted to find the cause, but was not successful.


```pycon
# PyCharm console
>>> from icalendar import LazyCalendar
>>> calendar = LazyCalendar.example("issue_1050_all_components")
>>> calendar.is_lazy()
False
```

```
$ .venv/bin/python3
Python 3.14.2 (main, Jan 14 2026, 23:37:46) [Clang 21.1.4 ] on darwin
>>> from icalendar import LazyCalendar
>>> calendar = LazyCalendar.example("issue_1050_all_components")
>>> calendar.is_lazy()
True
```

Weird, right?

src/icalendar/cal/lazy.py
stevepiercy: I modified them in a couple of suggestions.

src/icalendar/cal/lazy.py
stevepiercy: That works. Thank you!

src/icalendar/cal/lazy.py
stevepiercy: Makes sense.

src/icalendar/parser/ical/lazy.py
stevepiercy: Ah, so there may be multiple components that match the name? Maybe this?

```suggestion
            name: The name to match for all components in the calendar, then walk through and parse the resulting matches.
```

stevepiercy (changes requested):
Thanks for the suggestions and feedback. I accepted a couple of suggestions, and made counterproposals for the others. Please let me know what you think. Thank you!

src/icalendar/cal/lazy.py
niccokunzmann: I get the second example. I don't know why this could be. I trust the tests.

If you like, you can open an issue after merge if it is still reproducible.

src/icalendar/parser/ical/lazy.py
niccokunzmann: Yes, that is good.

src/icalendar/cal/lazy.py
stevepiercy: It's most likely something with PyCharm's Python console. After this PR is merged, then released, I'll report it to JetBrains.

stevepiercy (dismissed):
Docstrings look good now. Thank you! Yay team!

Let's wait for @SashankBhamidi to respond.

SashankBhamidi (dismissed):
I've marked them resolved, again, this can be merged.

Thank you for working on this! I will rebase and pick this up after merge in the first week of April hopefully.

PR #1205, opened 2026-02-17
#876 vBoolean

ijsbol:
## Description

Intermediate pull request for issue https://github.com/collective/icalendar/issues/876

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- ...
+- Created an :meth:`~icalendar.prop.vBoolean.ical_value` method for the :class:`~icalendar.prop.vBoolean` component. See :issue:`876`.
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
@@ -57,7 +57,6 @@ Minor changes
 
 - Show colorful required code changes in the CI output to help contributors solve the formatting issues. :pr:`1216`
 - Use ruff 0.15.0 for code formatting in :file:`tox.ini`. :pr:`1215`
-- Created an :meth:`~icalendar.prop.vBoolean.ical_value` method for the :class:`~icalendar.prop.vBoolean` component. See :issue:`876`.
 
 New features
 ~~~~~~~~~~~~


src/icalendar/prop/boolean.py
```diff
@@ -60,8 +60,13 @@ def __new__(
         return self
 
     def to_ical(self) -> bytes:
+        """The iCalendar representation of the vBoolean."""
```
stevepiercy: This doesn't make sense to me. vBoolean is an icalendar (lower case "c") representation of iCalendar (upper case "C") boolean values. That's not what this method does.

I think this is more correct. See https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.2. @SashankBhamidi @niccokunzmann can you help clarify?

```suggestion
        """Converts a :class:`~icalendar.prop.boolean.vBoolean` to a BOOLEAN property type.

        This class method takes a ``vBoolean``—a Python boolean value—and converts it to an iCalendar BOOLEAN property type, in compliance with :rfc:`5545#section-3.3.2`.

        Returns:
            Either "TRUE" or "FALSE" as bytes, depending on the value of the ``vBoolean``.
        """
```

## Preview

Image

src/icalendar/prop/boolean.py
```diff
@@ -101,4 +106,6 @@ def from_jcal(cls, jcal_property: list) -> Self:
         )
 
 
+
+
```
stevepiercy: Why the extra line breaks? This should have been caught by ruff.

```suggestion
```

src/icalendar/prop/boolean.py
niccokunzmann: I would say

> BOOLEAN property type according to :rfc:`5545#section-3.3.2

#1244

niccokunzmann: The docstrings are all over the place. Let's unify the format in #1244 - that can become several first-timer issues for docs.

ijsbol: Would you like me to update the existing doc string in this PR, or leave it for a first-time issue?

niccokunzmann: If you have time, then, by all means - please make the documentation nicer! Thank you. Yes, I would like it of you could update the first line and the documentation.

ijsbol:
Awesome! I've updated the docstring as requested, let me know if there's anything else I can do ^-^

niccokunzmann (approved):
technically sound

src/icalendar/prop/boolean.py
stevepiercy: I updated my original suggestion. See revised example and the screenshot. @niccokunzmann @ijsbol what do you think of this example as a guide?

This example might not be perfect, but I think it improves a reader's understanding of what the method actually does.

stevepiercy: Adding commit for progress.

PR #1208, opened 2026-02-17
Auto format code on non-protected branches.

niccokunzmann:
## Closes issue

Contributors might push code that is badly formatted and then need to work more before the change can be accepted.
If you edit a file through the web UI, you do not have code formatting available. This automatically applies certain fixes.

## Description

This automatically pushes commits with formatted code to the pull requests.

branches like `.x` and `main` and tags are exempt.
 
## Checklist

## Additional information

The commit 98bca822 should be reverted automatically.

----
📚 Documentation preview 📚: https://icalendar--1208.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,7 @@ Minor changes
 ~~~~~~~~~~~~~
 
 - Branches pushed to the repository are formatted automatically with ruff. :pr:`1208`
+- Use ruff 0.15.0 for code formatting in :file:`tox.ini`. :pr:`1215`
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
@@ -37,7 +38,7 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
-- ...
+- Enabled :meth:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>` to read calendars from files. :issue:`756`
 
 Bug fixes
 ~~~~~~~~~
@@ -47,7 +48,7 @@ Bug fixes
 Documentation
 ~~~~~~~~~~~~~
 
-- ...
+- Removed methods of ``str``, ``int``, and other classes and methods in the Python standard library from the documentation.
 
 7.0.1 (2026-02-17)
 ------------------
@@ -55,7 +56,7 @@ Documentation
 Minor changes
 ~~~~~~~~~~~~~
 
-- Setting :attr:`~cal.calendar.Calendar.calendar_name` now also writes ``X-WR-CALNAME``, and setting :attr:`~cal.calendar.Calendar.description` now also writes ``X-WR-CALDESC``, for improved client compatibility. See `Issue #918 <https://github.com/collective/icalendar/issues/918>`_.
+- Setting :attr:`~cal.calendar.Calendar.calendar_name` now also writes ``X-WR-CALNAME``, and setting :attr:`~cal.calendar.Calendar.description` now also writes ``X-WR-CALDESC``, for improved client compatibility. :issue:`918`
 
 Bug fixes
 ~~~~~~~~~
@@ -66,6 +67,7 @@ Documentation
 ~~~~~~~~~~~~~
 
 - Fixed the version switcher on Read the Docs, and documented the process for a major release. :issue:`1194`
+- Added usage examples for reading calendars from files and URL. :issue:`756`
 
 
 7.0.0 (2026-02-11)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: debug-statements
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.0 # also update pyproject.toml
+    rev: v0.15.2 # also update pyproject.toml
     hooks:
       # Run the linter.
       - id: ruff-check
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -172,7 +172,7 @@ autoclass_content = "both"
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
 autodoc_default_options = {
     "ignore-module-all": True,
-    "inherited-members": "CaselessDict",
+    "no-inherited-members": "object,CaselessDict,str,bool,NamedTuple,int,float,Enum",
     "members": True,
     "show-inheritance": True,
     "undoc-members": True,
--- a/docs/contribute/development.rst
+++ b/docs/contribute/development.rst
@@ -42,6 +42,10 @@ This section describes the branches used in icalendar development.
     Active development takes place on this branch.
     It is compatible with Python versions 3.10 - 3.14, and PyPy3.10.
 
+``7.x``
+    icalendar version 7 is on the branch `7.x <https://github.com/collective/icalendar/tree/7.x/>`_.
+    It is compatible with Python versions 3.10 - 3.14, and PyPy3.10.
+
 ``6.x``
     icalendar version 6 is on the branch `6.x <https://github.com/collective/icalendar/tree/6.x/>`_.
     It is compatible with Python versions 3.8 - 3.13, and PyPy3.9.
--- a/docs/how-to/cli.rst
+++ b/docs/how-to/cli.rst
@@ -16,10 +16,11 @@ To get information about all the commands, use the ``-h`` option.
 
     icalendar -h
 
-View an event
-=============
 
-To view a readable summary of an event file that is in icalendar format, pass the name of the icalendar file to ``icalendar``.
+Read calendar from file
+=======================
+
+To view a readable summary of an event file that is in iCalendar format, pass the path of the iCalendar file to icalendar.
 
 .. code-block:: shell
 
@@ -40,6 +41,18 @@ The following is example output.
     Description:
 
     Your yearly evaluation is scheduled for next Tuesday. Please be on time.
+
+
+Read calendar from URL
+======================
+
+Download the :file:`.ics` file first, then pass it to icalendar.
+
+.. code-block:: shell
+
+    curl -fsSL "https://www.gov.uk/bank-holidays/england-and-wales.ics" -o cal.ics
+    icalendar cal.ics
+
 
 Shell-based mail client
 =======================
--- a/docs/how-to/usage.rst
+++ b/docs/how-to/usage.rst
@@ -233,22 +233,35 @@ Examples
 This section describes how to use icalendar's essential features.
 
 
-Inspect files
-'''''''''''''
+Read calendar from file
+'''''''''''''''''''''''
 
-Open an :file:`.ics` file and view its events.
+Read a calendar from a local :file:`.ics` file and view its events.
 
 .. code:: pycon
 
     >>> import icalendar
     >>> from pathlib import Path
     >>> ics_path = Path("src/icalendar/tests/calendars/example.ics")
-    >>> calendar = icalendar.Calendar.from_ical(ics_path.read_bytes())
+    >>> calendar = icalendar.Calendar.from_ical(ics_path)
     >>> for event in calendar.events:
     ...     print(event.get("SUMMARY"))
     New Year's Day
     Orthodox Christmas
     International Women's Day
+
+
+Read calendar from URL
+''''''''''''''''''''''
+
+To read a calendar from a URL, retrieve the data first, then parse it in icalendar.
+
+.. code-block:: python
+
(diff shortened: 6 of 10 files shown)


.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
+name: Automatically format the code with ruff
+on:
+  - push
+
+jobs:
+  ruff-format:
+    permissions:
+      contents: write
+      pull-requests: write
```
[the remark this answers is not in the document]
niccokunzmann: This needs to run without pull request, too. Whether a fork or not is ok.

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+          python-version: 3.14
+          cache: pip
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install . --group formatting
+      - name: Format the code with ruff
+        run: ruff format
+      - name: Format the code with ruff check
+        run: ruff check --fix
```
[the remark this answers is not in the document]
niccokunzmann: We want formatting as much as possible. That is it. It happens.

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+    env:
+      GIT_COMMIT_MESSAGE: |
+        Format with ruff
+
+        See https://icalendar.readthedocs.io/en/stable/contribute/development.html#code-format
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
```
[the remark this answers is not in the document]
niccokunzmann: I don't change this because it works.

CHANGES.rst
```diff
@@ -27,7 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- ...
+- Pull Requests are formatted automatically. :pr:`1208`
```
stevepiercy: If ruff reformats PRs in its own repo, then I would strongly oppose this PR. I already dislike the idea of rewriting PRs as I think it will confuse the contributor more when they can't figure out git, which will make more work for us. I much prefer that the contributor install ruff and pre-commit locally to handle this, instead of in CI.

```suggestion
- Pull requests from remote forks are formatted automatically. :pr:`1208`
```

niccokunzmann:
@stevepiercy On the matter of confusion, I would try it out and get feedback. If it is an issue, it can be solved.

For now, this would ease this PR: #1205

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+    name: Format the code with ruff and push the changes
+    runs-on: ubuntu-latest
+    if: ! startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/main' && ! (startsWith(github.ref, 'refs/heads/') && endsWith(github.ref, '.x')) && ! startsWith(github.ref, 'refs/heads/release')
+    env:
+      GIT_COMMIT_MESSAGE: |
+        Format with ruff
+
+        See https://icalendar.readthedocs.io/en/stable/contribute/development.html#code-format
+    steps:
+      - uses: actions/checkout@v4
```
SashankBhamidi: `actions/checkout@v4`, every other workflow in the repo uses `@v6`.
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
```
SashankBhamidi: `github.event.pull_request.head.repo.full_name` and `github.event.pull_request.head.ref` don't exist on `push` events, they evaluate to empty string. I checked the checkout source (`input-helper.ts`), it uses `core.getInput('repository') || context.repo` so empty string falls back to the current repo. Same for ref. These parameters do nothing here, they're dead code from when this was a `pull_request` trigger. Copilot flagged this, Nicco said "I don't change this because it works" but it works despite these lines, not because of them.
```suggestion
      - uses: actions/checkout@v6
        with:
          fetch-depth: 0
```

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+          python-version: 3.14
+          cache: pip
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install . --group formatting
+      - name: Format the code with ruff
+        run: ruff format
+      - name: Format the code with ruff check
+        run: ruff check --fix
```
SashankBhamidi, line 52: `ruff check --fix` auto-commits lint fixes beyond formatting. It can remove unused imports, rewrite expressions. That's different from whitespace fixes. Copilot and Steve both flagged this.

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Show the changes
+        run: |
+          git status
+          git diff
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git commit -a -m "${GIT_COMMIT_MESSAGE}" || echo "No changes to commit"
```
SashankBhamidi, line 61: Step is named "Show the changes" but it also configures git identity and creates a commit.

CHANGES.rst
```diff
@@ -27,7 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- ...
+- Pull requests from remote forks are formatted automatically. :pr:`1208`
```
SashankBhamidi: "Pull requests from remote forks are formatted automatically" is inaccurate. The `push` event only fires for pushes directly to `collective/icalendar` branches. When a fork contributor pushes to their fork, the push event fires in the fork, not in upstream. This workflow won't format fork PRs.
```suggestion
- Branches pushed to the repository are formatted automatically with ruff. :pr:`1208`
```

SashankBhamidi (changes requested):
A few things:

- Lines 21-22: `github.event.pull_request.*` is null on push events. Checkout falls back to defaults (verified in checkout's `input-helper.ts`: `core.getInput('repository') || context.repo`). These lines do nothing, should be removed.
- Copilot correctly flagged lines 21-22 and the `ruff check --fix` scope. Worth reconsidering those.

.github/workflows/ruff-format.yml
niccokunzmann: That is intended. The more fixes, the better. We check that in our CI. If these fixes are not applied, then the CI ruff.yml still fails.

.github/workflows/ruff-format.yml
niccokunzmann: I addressed this in 1ce254c0.

.github/workflows/ruff-format.yml
niccokunzmann: I addressed this in 1ce254c0.

.github/workflows/ruff-format.yml
niccokunzmann: It did not remove unused imports.

```
Run ruff check --fix
F401 `time` imported but unused
 --> src/icalendar/__init__.py:1:8
  |
1 | import time
  |        ^^^^
2 |
3 | from icalendar.alarms import (
  |
help: Remove unused import: `time`

Found 2 errors (1 fixed, 1 remaining).

```

But it reordered imports.

SashankBhamidi: Fair point on __init__.py specifically. F401 is unsafe there so it won't auto-remove. But in regular .py files unused imports are a safe fix and will be auto-committed. The concern still stands for those files.

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+  - push
+
+jobs:
+  ruff-format:
+    permissions:
+      contents: write
+      pull-requests: write
+    name: Format the code with ruff and push the changes
+    runs-on: ubuntu-latest
+    if: ! startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/main' && ! (startsWith(github.ref, 'refs/heads/') && endsWith(github.ref, '.x')) && ! startsWith(github.ref, 'refs/heads/release')
```
stevepiercy: What is the intent here? It would be good to add a comment and possibly a link to a reference to explain the intent of this statement, as its arguments aren't really clear. That will help make sure what you want is what you get, and help other people understand it.

.github/workflows/ruff-format.yml
```diff
@@ -0,0 +1,48 @@
...
+  - push
+
+jobs:
+  ruff-format:
+    permissions:
+      contents: write
+      pull-requests: write
+    name: Format the code with ruff and push the changes
+    runs-on: ubuntu-latest
+    if: github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/') && !(startsWith(github.ref, 'refs/heads/') && endsWith(github.ref, '.x')) && !startsWith(github.ref, 'refs/heads/release')
```
niccokunzmann, line 30: The ! at the start was ignored by GitHub. I re-ordered the condition to not  start with that.

In the downloaded log zip, there is a system.txt showing:

```
(
  true && (
  ('refs/heads/auto-format' != 'refs/heads/main')
  && !startsWith('refs/heads/auto-format', 'refs/tags/') 
  && !(
    startsWith('refs/heads/auto-format', 'refs/heads/')
    && endsWith('refs/heads/auto-format', '.x')
  )
  && !startsWith('refs/heads/auto-format', 'refs/heads/release')))
```

That should exclude `main`, all `.x` branches and tags as well as the release-... branches created by the maintenance tasks of releasing.

.github/workflows/ruff-format.yml
stevepiercy: heh, and you just updated it, but my suggestion stands.

.github/workflows/ruff-format.yml
niccokunzmann: I think, that is ok. If tests run, the import was likely useless.

.github/workflows/ruff-format.yml
niccokunzmann: Yes, comments and purpose are in [73a97fe](https://github.com/collective/icalendar/pull/1208/commits/73a97fefced5d42233b425d8ac2275b32204c2ce). Please have a look.

.github/workflows/ruff-format.yml
```diff
@@ -9,6 +17,16 @@ jobs:
       pull-requests: write
     name: Format the code with ruff and push the changes
     runs-on: ubuntu-latest
+    #
+    # Run conditions as safeguards
+    #
+    # This workflow EXCLUDES:
+    #
+    # - branch "main"
```
stevepiercy, line 25: This sounds like I can fork to `stevepiercy/icalendar`, push to my `main` and it won't run.

Then when I open a PR against `collective/icalendar` on its `main` branch, it still won't run.

Is that what you want?

Same for the remaining exclusions below.

niccokunzmann: I think, that is a good start.

It can be extended to format main on other repos as that is where first -timers often start like in #1205. I would do that another time, and it requires more consideration.

I think, I prefer it if `main` stays `main` and can be fast-forwarded to `collective/calendar`'s main.

stevepiercy: Honestly, I think automatically rewriting pull requests is a bad idea all around. Pull requests should be intentional and deliberate, which is something only humans can do.

niccokunzmann: That is ok. I will not talk about the value. I see value. I would like to see if this is technically correct. Then, I will merge it or someone else does it and we see if it makes sense. We can remove it if it gets in the way of contribution.

SashankBhamidi: > Honestly, I think automatically rewriting pull requests is a bad idea all around. Pull requests should be intentional and deliberate, which is something only humans can do.

+1

niccokunzmann:
Something like this could help:

https://github.com/orgs/community/discussions/25100

It would be nice to just comment "Please format" and that PR gets formatted.

I mean... One part of it is ease, the other one being perfectionist and the other one is having fun tinkering and exploring.

niccokunzmann:
This is stale. I will close this now. This can be taken up and changed when needed.

PR #1209, opened 2026-02-17
feat: also set X-WR-CALNAME in Calendar.calendar_name setter

### LINKED ISSUE #918: Getting out an ical with NAME and X-WR-CALNAME
It would be good if there was an option for the generated ical file to have both X-WR-CALNAME and NAME

This is because Google Calendar only supports X-WR-CALNAME

I don't know if it should be on by default but given how popular Google Calendar is it would be good to have an easy option to enable support.

Workaround:
    
    calendar.calendar_name = "Demo"
    calendar.add("X-WR-CALNAME", "Demo")

From https://github.com/collective/icalendar/discussions/913

kibanana:
## Closes issue

- [x] Closes #918

## Description

Setting calendar_name now also writes X-WR-CALNAME for improved compatibility with clients that rely on the non-standard property.

**Changes:**

**1. Core fix** (`src/icalendar/attr.py`)
- `multi_language_text_property` setter now also writes to `compatibility_prop` if one is specified

**2. Docstring update** (`src/icalendar/cal/calendar.py`)
- Updated examples to reflect that both `NAME` and `X-WR-CALNAME` (and `DESCRIPTION` + `X-WR-CALDESC`) are written

**3. Test update** (`src/icalendar/tests/test_rfc_7986.py`)
- Updated `test_set_calendar_name` to assert `X-WR-CALNAME` is also set
- Renamed and rewrote `test_setting_the_name_deletes_the_non_standard_attribute` → `test_setting_the_name_sets_both_properties` to reflect the new behavior

**4. Changelog** (`CHANGES.rst`)
- Added entry in "Minor changes" section

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1209.org.readthedocs.build/

### DIFF
--- a/src/icalendar/attr.py
+++ b/src/icalendar/attr.py
@@ -357,6 +357,8 @@ def multi_language_text_property(
         fdel(self)
         if value is not None:
             self.add(main_prop, value)
+            if compatibility_prop is not None:
+                self.add(compatibility_prop, value)
 
     def fdel(self: Component):
         """Delete the property."""
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,7 @@ We still recommend checking out the new features and giving feedback in the repo
 Minor changes
 ~~~~~~~~~~~~~
 
-- ...
+- Setting :attr:`~cal.calendar.Calendar.calendar_name` now also writes ``X-WR-CALNAME``, and setting :attr:`~cal.calendar.Calendar.description` now also writes ``X-WR-CALDESC``, for improved client compatibility. See `Issue #918 <https://github.com/collective/icalendar/issues/918>`_.
 
 Breaking changes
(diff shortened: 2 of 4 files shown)


src/icalendar/attr.py
```diff
@@ -357,6 +357,8 @@ def fset(self: Component, value: str | None):
         fdel(self)
         if value is not None:
             self.add(main_prop, value)
+            if compatibility_prop is not None:
+                self.add(compatibility_prop, value)
```
niccokunzmann, line 361: Hm. While this is ok. It is also not deleted before.

Can you check these cases in the file before:

- set `NAME:`, then set .calendar_name
- set `X-WR-CALNAME`, then set .calendar_name

making sure that these names are not duplicated?
This might be tested already. Maybe, this is not an issue. I would still like to know it that is covered.

Issue #1062, opened 2025-12-28
Test windows & macos


niccokunzmann:
It appears that icalendar can sometimes misbehave [under Windows](https://github.com/collective/icalendar/issues/1056) and maybe under macOS. It would be good to run the CI tests under these platforms.

- [ ] do not run additional environments (Python 3.X)
- [ ] choose to run some of the environments, especially those not shipped by LTS linux distribution Debian under macOS and Windows
- [ ] for that, add an additional matrix parameter "runner"\
- [ ] Edit https://github.com/collective/icalendar/settings/branch_protection_rules, see #1247

Related:

- https://github.com/collective/icalendar/discussions/813

niccokunzmann:
We are also running three tests always. I would suggest:

- update the test matrix so that the minimal tests are spread over all 3 OS
- the full tests = `python versions x OS`. They run rarely anyway.

PR #1239, opened 2026-02-25
Run tests conditionally on pull request approval

niccokunzmann:
## Closes issue

Contributes to #1062

## Description

This is a start: the matrix will only run specific tests when the pull request is approved.

## Checklist

## Additional information

- [ ] Add the OS conditionally
- [x] Always run on main, `*.x` and tags

### DIFF
--- a/.github/workflows/generate-matrix.py
+++ b/.github/workflows/generate-matrix.py
@@ -22,6 +22,7 @@ import sys
 #
 
 PYTHON_MINOR_VERSION_MIN = 10
+# Update this when a new Python minor version is released
 PYTHON_MINOR_VERSION_MAX = 14
 
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,7 +55,7 @@ jobs:
     - name: Coveralls Parallel
       uses: coverallsapp/github-action@v2
       with:
-        flag-name: run-${{ matrix.tox_env }}
+        flag-name: run-${{ matrix.python_version }}-${{ matrix.tox_env }}
         parallel: true
         file: coverage.xml
         allow-empty: true


.github/workflows/tests.yml
```diff
@@ -46,11 +51,11 @@ jobs:
         python -m pip install --upgrade pip
         pip install tox coveralls coverage-python-version
     - name: Test
-      run: tox -e ${{ matrix.config[1] }}
+      run: tox -e ${{ matrix.tox_env }}
     - name: Coveralls Parallel
       uses: coverallsapp/github-action@v2
       with:
-        flag-name: run-${{ matrix.config[1] }}
+        flag-name: run-${{ matrix.tox_env }}
```
SashankBhamidi: `flag-name` is now `run-py` for every Python version. The old matrix used `${{ matrix.config[1] }}` which gave each job a unique name (`run-py310`, `run-py311`, etc.). Now all `py` jobs emit `run-py` — coveralls parallel mode uses flag names to merge uploads, so duplicate flags mean later uploads overwrite earlier ones and the final report reflects only one Python version's coverage.
```suggestion
        flag-name: run-${{ matrix.python_version }}-${{ matrix.tox_env }}
```

.github/workflows/generate-matrix.py
```diff
@@ -0,0 +1,95 @@
...
+import json
+import re
+import sys
+
+#
+# Configuration
+#
+
+PYTHON_MINOR_VERSION_MIN = 10
+PYTHON_MINOR_VERSION_MAX = 14
```
SashankBhamidi, line 26: `PYTHON_MINOR_VERSION_MAX = 14` with no comment. When Python 3.15 ships, it is silently never tested.
```suggestion
# Update this when a new Python minor version is released
PYTHON_MINOR_VERSION_MAX = 14
```

.github/workflows/generate-matrix.py
```diff
@@ -0,0 +1,96 @@
...
+#
+
+if RUN_ALL_JOBS:
+    for minor_python_version in range(
+        PYTHON_MINOR_VERSION_MIN + 1, PYTHON_MINOR_VERSION_MAX
+    ):
+        matrix.append({"tox_env": "py", "python_version": f"3.{minor_python_version}"})
+
+    # pypy is slow but good to test
+    matrix.append({"tox_env": "pypy3", "python_version": "pypy3.10"})
```
stevepiercy, line 84: Because PyPy3 is slow, I'd make this the first job to make it more likely to be triggered first.

Issue #1056, opened 2025-12-27
[BUG] Windows: dateutil.tz.gettz("UTC") produces TZID parameter violating RFC 5545


SashankBhamidi:
## Bug
On Windows, `dateutil.tz.gettz("UTC")` returns timezone with name "Coordinated Universal Time" instead of "UTC", causing icalendar to add TZID parameter to UTC datetimes, violating RFC 5545.

## Reproduce
**Platform:** Windows
**Python:** 3.11
**icalendar:** 7.0.0a3

```python
from icalendar import Event
from dateutil import tz
import datetime

utc = tz.gettz("UTC")
dt = datetime.datetime(2012, 7, 16, 0, 0, 0, tzinfo=utc)
event = Event()
event.add("dtstart", dt)
print(event.to_ical())
```

**Expected (RFC 5545 compliant):**
```
DTSTART:20120716T000000Z
```

**Actual on Windows:**
```
DTSTART;TZID=Coordinated Universal Time:20120716T000000
```

## Root Cause
`dateutil.tz.gettz("UTC")` returns `tzwin('UTC')` on Windows. Its `tzname()` returns "Coordinated Universal Time", which icalendar doesn't recognize as UTC.

## Test Failures
4 failures in `test_parsing.py::test_no_tzid_when_utc` with utc4 fixture.

## RFC 5545
Section 3.3.5: "The TZID property parameter MUST NOT be applied to DATE-TIME or TIME properties whose time values are specified in UTC."

## Fix
UTC detection should recognize "Coordinated Universal Time" as equivalent to "UTC".

PR #1245, opened 2026-02-26
fix: do not set TZID=UTC on UTC datetimes, fix KeyError in get_missing_tzids

### LINKED ISSUE #1124: [BUG] icalendar generates UTC timezone
## Describe the bug

icalendar should not generate UTC timezone information.

## To reproduce

```python
    >>> from icalendar import Calendar, Event
    >>> from datetime import datetime
    >>> event = Event.new(
    ...     summary="Meeting in Zurich",
    ...     start=datetime(2022, 1, 1, 12, 0, 0, tzinfo=ZoneInfo("Europe/Zurich")),
    ...     end=datetime(2022, 1, 1, 13, 0, 0, tzinfo=ZoneInfo("Europe/Zurich")),
    ...     location="Zurich, Switzerland",
    ... )
    >>> calendar = Calendar.new(
    ...     subcomponents=[event],    
    ... )
    >>> calendar.add_missing_timezones()
    >>> calendar.timezones[0].tz_name
    "UTC"
    >>> calendar.timezones[1].tz_name
    'Europe/Zurich'
```

## Expected behavior

UTC should not be generated.

## Environment

 - [x] Operating system:      Xubuntu 24   
 - [x] Python version:      3.14     
 - [x] `icalendar` version: `main`@52b5142a160b94b411ad42183f2396daae2f0bc0  7.0.0a4.dev441  

## Additional context

SashankBhamidi:
Fixes #1124.

Two bugs caused `add_missing_timezones()` to generate a spurious VTIMEZONE for UTC:

1. `Parameters.update_tzid_from()` was setting `TZID=UTC` on UTC datetimes. RFC 5545 section 3.2.19 prohibits TZID on UTC datetimes — they use the Z suffix. Fixed by checking `is_utc()` and skipping TZID in that case.

2. `Calendar.get_missing_tzids()` used `set.remove()`, which raises `KeyError` when a VTIMEZONE exists for a timezone not referenced by any event TZID. Fixed by using `discard()` instead.

### DIFF
--- a/src/icalendar/cal/calendar.py
+++ b/src/icalendar/cal/calendar.py
@@ -249,8 +249,11 @@ class Calendar(Component):
 
         To create a :rfc:`5545` compatible calendar,
         all of these timezones should be added.
+
+        UTC is excluded: per :rfc:`5545` section 3.2.19, UTC datetimes use
+        the ``Z`` suffix and never require a VTIMEZONE component.
         """
-        tzids = self.get_used_tzids()
+        tzids = self.get_used_tzids() - {"UTC"}
         for timezone in self.timezones:
             # discard (not remove) — a VTIMEZONE may exist for a timezone not
             # referenced by any event TZID (e.g. added by x-wr-timezone conversion)
--- a/src/icalendar/tests/test_issue_1124_no_utc_vtimezone.py
+++ b/src/icalendar/tests/test_issue_1124_no_utc_vtimezone.py
@@ -30,15 +30,20 @@ def test_add_missing_timezones_does_not_generate_utc():
     assert "Europe/Zurich" in tz_names
 
 
-def test_utc_not_in_used_tzids():
-    """UTC must not appear in get_used_tzids() — it uses the Z suffix, not a TZID."""
-    event = Event.new(
-        summary="test",
-        start=datetime(2022, 1, 1, 12, 0, 0, tzinfo=timezone.utc),
-        end=datetime(2022, 1, 1, 13, 0, 0, tzinfo=timezone.utc),
-    )
-    calendar = Calendar.new(subcomponents=[event])
-    assert "UTC" not in calendar.get_used_tzids()
+def test_utc_not_in_missing_tzids():
+    """UTC must not appear in get_missing_tzids() — no VTIMEZONE should be generated.
+
+    Even if UTC were present in get_used_tzids() (e.g. from a legacy calendar
+    with TZID=UTC), get_missing_tzids() must filter it out so that
+    add_missing_timezones() never creates a VTIMEZONE for UTC.
+    Per :rfc:`5545` section 3.2.19, UTC datetimes use the Z suffix instead.
+    """
+    calendar = Calendar()
+    event = Event()
+    event.add("dtstart", datetime(2022, 1, 1, 12, 0, 0, tzinfo=timezone.utc))
+    event.add("dtend", datetime(2022, 1, 1, 13, 0, 0, tzinfo=timezone.utc))
+    calendar.add_component(event)
+    assert "UTC" not in calendar.get_missing_tzids()
 
 
 def test_get_missing_tzids_does_not_crash_with_extra_vtimezone():
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,10 @@ Bug fixes
 - Fixed :meth:`Calendar.get_missing_tzids <icalendar.cal.calendar.Calendar.get_missing_tzids>`
   raising ``KeyError`` when a VTIMEZONE exists for a timezone not referenced by any event TZID
   (e.g. added by x-wr-timezone conversion). :issue:`1124`
+- Fixed :meth:`Calendar.get_missing_tzids <icalendar.cal.calendar.Calendar.get_missing_tzids>`
+  and :meth:`Calendar.add_missing_timezones <icalendar.cal.calendar.Calendar.add_missing_timezones>`
+  generating a spurious ``VTIMEZONE`` for UTC. :rfc:`5545` section 3.2.19 requires UTC datetimes
+  to use the ``Z`` suffix; no ``VTIMEZONE`` component is needed or permitted. :issue:`1124`
 - Fixed :meth:`Parameters.update_tzid_from <icalendar.parser.parameter.Parameters.update_tzid_from>`
   incorrectly setting ``TZID=UTC`` on UTC datetimes. :rfc:`5545` section 3.2.19 requires UTC datetimes to
   use the ``Z`` suffix without a ``TZID`` parameter. :issue:`1124`


src/icalendar/parser/parameter.py
```diff
@@ -463,9 +463,14 @@ def update_tzid_from(self, dt: datetime | time | Any) -> None:
         """Update the TZID parameter from a datetime object.
 
         This sets the TZID parameter or deletes it according to the datetime.
+        RFC 5545 §3.2.19 prohibits TZID on UTC datetimes — those use Z suffix.
```
niccokunzmann: Please apply formatting.

src/icalendar/parser/parameter.py
```diff
@@ -463,9 +463,15 @@ def update_tzid_from(self, dt: datetime | time | Any) -> None:
...
+        :rfc:`5545` section 3.2.19 prohibits TZID on UTC datetimes,
+        which use the Z suffix instead.
         """
         if isinstance(dt, (datetime, time)):
-            self.tzid = tzid_from_dt(dt)
+            if is_utc(dt):
+                # UTC uses Z suffix; TZID must not be set per :rfc:`5545` section 3.2.19
+                self.tzid = None
+            else:
+                self.tzid = tzid_from_dt(dt)
```
niccokunzmann:
```suggestion
            tzid = tzid_from_dt(dt)
            if tzid != "UTC":
                # UTC uses Z suffix and does not appear as TZID parameter
                self.tzid = tzid
```

This is a bit more efficient.

SashankBhamidi:
> Also, please prove that the tests do what they should be committing them first, have them fail and then apply the changes.

WIll do next time!

src/icalendar/tests/prop/test_date_and_time.py
```diff
@@ -32,7 +32,7 @@ def test_setting_a_utc_datetime_value_does_not_include_a_TZID(
     """Check that the VALUE parameter is correctly determined."""
     if convert:
         utc_prop = convert(utc_prop.dt)
-    assert utc_prop.params.is_utc()
+    assert not utc_prop.params.tzid, "UTC datetimes must not carry a TZID parameter"
```
niccokunzmann, line 35: We must remove params.is_utc() in this PR. It will always return False and the parameters do not know then if it actually is UTC.

We have a discrepancy here. Maybe, that responsibility of whether something is UTC should be held in the vProp.
It is good to remove the TZID=UTC in the parameters to be more according to standard.

src/icalendar/tests/test_issue_1124_no_utc_vtimezone.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+from datetime import datetime, timezone
+from zoneinfo import ZoneInfo
+
+from icalendar import Calendar, Event, Timezone
+
+
+def test_add_missing_timezones_does_not_generate_utc():
+    """add_missing_timezones() must not add a VTIMEZONE for UTC."""
```
niccokunzmann:
```suggestion
    """add_missing_timezones() must not add a VTIMEZONE for UTC.
    
    Using new(), we set DTSTAMP which is in UTC.
    This tests that the UTC timezone is not added.
    """
```

src/icalendar/tests/test_issue_1124_no_utc_vtimezone.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_utc_not_in_used_tzids():
+    """UTC should never appear in get_used_tzids() even when DTSTAMP is UTC."""
+    event = Event.new(
+        summary="test",
+        start=datetime(2022, 1, 1, 12, 0, 0, tzinfo=timezone.utc),
+        end=datetime(2022, 1, 1, 13, 0, 0, tzinfo=timezone.utc),
+    )
+    calendar = Calendar.new(subcomponents=[event])
+    assert "UTC" not in calendar.get_used_tzids()
```
niccokunzmann: Well. I would say that "UTC" should be in the used TZIDs because it is a timezone and it has an id.
However, it should not be generated. So, it must be removed from the tzids before they are generated.

niccokunzmann: That should be in Calendar.

src/icalendar/tests/prop/test_date_and_time.py
SashankBhamidi: `params.is_utc()` still serves jCal parsing where `TZID=UTC` can arrive from external data. `vDatetime.is_utc()` / `vTime.is_utc()` stay correct via the `is_utc(self.dt)` fallback. 

Side note: on Windows, `tz.gettz("UTC")` returns `tzwin('UTC')` with `tzname()` = `'Coordinated Universal Time'`, so `params.is_utc()` was already broken there before this PR, separate issue in `tzids_from_tzinfo`.

src/icalendar/tests/test_issue_1124_no_utc_vtimezone.py
SashankBhamidi: Done. UTC is now filtered in `get_missing_tzids()` with `- {"UTC"}`, so `add_missing_timezones()` never generates a VTIMEZONE for it regardless of what `get_used_tzids()` returns.

src/icalendar/cal/calendar.py
```diff
@@ -249,10 +249,15 @@ def get_missing_tzids(self) -> set[str]:
 
         To create a :rfc:`5545` compatible calendar,
         all of these timezones should be added.
+
+        UTC is excluded: per :rfc:`5545` section 3.2.19, UTC datetimes use
+        the ``Z`` suffix and never require a VTIMEZONE component.
         """
-        tzids = self.get_used_tzids()
+        tzids = self.get_used_tzids() - {"UTC"}
```
niccokunzmann, line 256: Yes! I thought, this would be an easier approach...

src/icalendar/tests/prop/test_date_and_time.py
niccokunzmann: Wow! This is why we need #1062 ! Fancy working on it? I will do the CI but could you report that error as an issue, please?

niccokunzmann: Generally, what I remember: It is ok to set the TZID in params to UTC. However, when we serialize the calendar, then it should be removed.

That should be tested, can be a different PR.

src/icalendar/parser/parameter.py
```diff
@@ -463,9 +463,14 @@ def update_tzid_from(self, dt: datetime | time | Any) -> None:
...
         This sets the TZID parameter or deletes it according to the datetime.
+        :rfc:`5545` section 3.2.19 prohibits TZID on UTC datetimes,
+        which use the Z suffix instead.
         """
         if isinstance(dt, (datetime, time)):
-            self.tzid = tzid_from_dt(dt)
+            tzid = tzid_from_dt(dt)
+            if tzid != "UTC":
+                # UTC uses Z suffix and does not appear as TZID parameter
+                self.tzid = tzid
```
niccokunzmann, line 473: Yes, I think that is alright.

niccokunzmann (approved):
Yes, this change works as I would expect it.

TODO open issues

- [ ] @SashankBhamidi report UTC recognition problems on Windows
- [ ] @niccokunzmann create an issue to deprecate is_utc() in params

niccokunzmann:
Tests running!!!..... Yay :) If you have time, can you have a look at #1246

src/icalendar/tests/prop/test_date_and_time.py
SashankBhamidi: Already reported in #1056.

PR #1213, opened 2026-02-17
Work through release process with docs

niccokunzmann:
- Add release glossary
- update release process notes for docs
- merge into stable branch to keep the edit of the version switcher in it.
- order release process for docs to show if the step applies at the start of the list item

## Description

I just did a release and worked through the documentation notes.

## Checklist

- [ ] I've added a change log entry to `CHANGES.rst`.
- [x] I've added or updated tests if applicable.
- [x] I've run and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [x] I've added or edited documentation, both as docstrings to be rendered in the API documentation and narrative documentation, as necessary.

----
📚 Documentation preview 📚: https://icalendar--1213.org.readthedocs.build/

PR #1246, opened 2026-02-26
Do not run tests if review requests changes

niccokunzmann:
This skips running tests when not required (review negative)

## Closes issue

Refers to tests running here: 
https://github.com/collective/icalendar/pull/1245#pullrequestreview-3862796807

## Description

- add tests for all the cases
- change ruff formatting
- change run conditions
- put logic into function

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,23 +4,6 @@ Change log
 
 .. py:currentmodule:: icalendar
 
-We use `Semantic Versioning <https://semver.org>`_.
-
-- Breaking changes increase the **major** version number.
-- New features increase the **minor** version number.
-- Minor changes and bug fixes increase the **patch** version number.
-
-7.x series
-----------
-
-Developers may be concerned about upgrading to a **new major** release.
-Upgrading from 6.x to 7.x should have **no complications for most developers**, because:
-
-- the core API stays compatible with 4.x
-- the breaking changes likely affect you only if you are an icalendar expert, not a normal user
-
-We still recommend checking out the new features and giving feedback in the repository.
-
 7.0.4 (unreleased)
 ------------------
 
--- a/docs/contribute/development.rst
+++ b/docs/contribute/development.rst
@@ -32,38 +32,9 @@ You'll pull code from the upstream icalendar repository, push your work from you
 
 #.  `Configure Git to sync your fork with the upstream repository <https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo#configuring-git-to-sync-your-fork-with-the-upstream-repository>`_.
 
-Branches
---------
-
-This section describes the branches used in icalendar development.
-
-``main``
-    The `main <https://github.com/collective/icalendar/tree/main/>`_ branch receives the latest updates and features.
-    Active development takes place on this branch.
-    It is compatible with Python versions 3.10 - 3.14, and PyPy3.10.
-
-``7.x``
-    icalendar version 7 is on the branch `7.x <https://github.com/collective/icalendar/tree/7.x/>`_.
-    It is compatible with Python versions 3.10 - 3.14, and PyPy3.10.
-
-``6.x``
-    icalendar version 6 is on the branch `6.x <https://github.com/collective/icalendar/tree/6.x/>`_.
-    It is compatible with Python versions 3.8 - 3.13, and PyPy3.9.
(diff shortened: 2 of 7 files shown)


niccokunzmann:
These will run when a PR is approved.
grafik

And I renamed Pypy3, so I will

- [ ] adjust the branch protection rules after merge.

.github/workflows/test_generate_matrix.py
```diff
@@ -0,0 +1,66 @@
...
+@pytest.mark.parametrize(
+    ("arg_pr", "expected", "arg_ref"),
+    [
+        # push
+        ("", COUNT_ALL, "refs/heads/main"),
+        ("", COUNT_ALL, "refs/heads/7.x"),
+        ("", COUNT_ALL, "refs/heads/6.x"),
+        ("", COUNT_ALL, "refs/heads/5.x"),
+        ("", COUNT_ALL, "refs/heads/release-1"),
+        ("", COUNT_ALL, "refs/tags/v7.0.2"),
```
stevepiercy: Now that I see this test, I think that running on tags is redundant to the other jobs, based on the assumption that you would only tag a commit that has already passed all tests. What do you think?

As a bonus, it would reduce maintenance of this test file and reduce the number of workflow runs.

.github/workflows/test_generate_matrix.py
```diff
@@ -0,0 +1,66 @@
...
+        ("", COUNT_MIN, "refs/heads/pr-branch"),
+        # review
+        ("changes_requested", 0, "refs/heads/main"),
+        ("changes_requested", 0, "refs/heads/7.x"),
+        ("changes_requested", 0, "refs/heads/6.x"),
+        ("changes_requested", 0, "refs/heads/5.x"),
+        ("changes_requested", 0, "refs/heads/release-2"),
```
stevepiercy: Just a note, we're also discussing removal of `release-*` in https://github.com/collective/icalendar/pull/1213#issuecomment-3979702450. This might need to be updated accordingly.

.github/workflows/test_generate_matrix.py
```diff
@@ -0,0 +1,66 @@
...
+
+@pytest.mark.parametrize(
+    ("arg_pr", "arg_ref"),
+    [
+        # push
+        ("", "refs/heads/main"),
+        ("approved", "refs/heads/6.x"),
+    ],
+)
+def test_pypy_is_first(arg_ref, arg_pr):
```
stevepiercy, line 65: Out of curiosity, did it actually run first? I guess we won't know until approval.

.github/workflows/test_generate_matrix.py
```diff
@@ -0,0 +1,66 @@
...
+    """PyPy should be first so that it gets triggered first.
```
stevepiercy: This explains its purpose: why should it be first?

```suggestion
    """PyPy takes longest to run, so it should be triggered first.
```

.github/workflows/test_generate_matrix.py
niccokunzmann: Perfect, yes. I added it to the TODO there in case this gets merged before that.

.github/workflows/test_generate_matrix.py
niccokunzmann: I don't know. I cannot verify it as mostly they all start within seconds. But I just go with what you suggest.

.github/workflows/test_generate_matrix.py
niccokunzmann: I would always run on tags, too because this is the most important place to run them - just before shipping. For convenience, these tests are a dependency to the release job and I would not count on main/7x always to work before a release nor do tests passing in main imply that a tag cannot have an influence on the tests. It probably does not but e.g. hatch-vcs actually changes code based on the tag and so will maybe our new release process. I would like all tests to pass on a tag.

stevepiercy: Thanks for the explanation. Makes sense. I didn't know that hatch-vcs changes code.

stevepiercy (dismissed):

.github/workflows/test_generate_matrix.py
stevepiercy: The first few checks passed in seconds and the PyPy check queued before I could see if it was first. Oh, well, maybe next time.

PR #1255, opened 2026-02-28
Fix change log to reflect what actually happened in the proper version

stevepiercy:
This fixes an issue with the change log from PR #1241.

@japinderofficial-hub please don't rewrite change log entries after a release, especially where it changes its meaning and what was actually done. It's OK to fix typos and minor mistakes, but changing history is not OK.

@niccokunzmann I'd like to look at a better change log management system. The current one is prone to human error and annoying merge conflicts. I like [towncrier](https://towncrier.readthedocs.io/en/stable/). It's used throughout Plone and Collective.

No change log entry needed for this PR because it fixes a previous entry.

----
📚 Documentation preview 📚: https://icalendar--1255.org.readthedocs.build/


niccokunzmann:
Yes, a better changelog management is good to have. Can you open an issue?
Additional issues with this one:

- needs editing in release
- old PRs go into the wrong place after a release

PR #1276, opened 2026-03-13
1142/timezone support

### LINKED ISSUE #1142: Add timezone support to vTime
vTime does not support timezones at the moment as per [comment].

To close this issue:

- [ ] create an example calendar with vTime, preferably from RFC 5545
- [ ] remove the comment
- [ ] create a test with a calendar that checks that the correct timezone is assigned after parsing
- test
  - [ ] a named timezone
  - [ ] UTC
  - [ ] no timezone

[comment]: https://github.com/collective/icalendar/pull/1136/files#diff-0d19f5e03fd9aa2e6eb340a79f5ac5e76528893b9e1f96531e52c3bab6689700R154

clara-rodrigues:
## Closes issue

- [x] Closes #1142 

## Description
In src/icalendar/prop/dt/time.py file, more specifically function from_ical, added a new argument in order to send timezone related information and updated the parsing logic in order to take into account the three different kinds of timezones (UTC, known timezone, floating timezone)
Created a new instance of a calendar as well as a new test in order to assure the correct behavior of the added functionality.
Adapted the from_jcal function logic in src/icalendar/prop/dt/types.py to also be able to parse timezone information.

## Checklist

- [x] I've run and ensured all tests pass locally by following [Run tests] (https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests). (

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,7 @@ Minor changes
 - Created an :meth:`~icalendar.prop.integer.vInt.ical_value` property for the :class:`~icalendar.prop.integer.vInt` component. :issue:`876`
 - Created an :meth:`~icalendar.prop.binary.vBinary.ical_value` property for the :class:`~icalendar.prop.binary.vBinary` component. :issue:`876`
 - Put the link check as the last documentation CI task, allowing the documentation build and Vale to run first and fail faster. :pr:`1295`
+- Extended :func:`~icalendar.timezone.tzp.TZP.localize` to support localizing both :class:`datetime.datetime` and :class:`datetime.time` objects, returning timezone-aware :class:`datetime.time` objects for the latter. :issue:`1142`
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
@@ -26,11 +27,8 @@ Breaking changes
 New features
 ~~~~~~~~~~~~
 
-- Added ``subcomponents`` parameter to
-  :meth:`Component.new <icalendar.cal.component.Component.new>`,
-  :meth:`Event.new <icalendar.cal.event.Event.new>`,
-  :meth:`Todo.new <icalendar.cal.todo.Todo.new>`, and
-  :meth:`Availability.new <icalendar.cal.availability.Availability.new>`. :issue:`1065`
+- Updated :func:`icalendar.prop.dt.time.vTime.from_ical` to support parsing time values with TZID parameters, returning timezone-aware :class:`datetime.time` objects. :issue:`1142`
+- Added ``subcomponents`` parameter to :meth:`Component.new <icalendar.cal.component.Component.new>`, :meth:`Event.new <icalendar.cal.event.Event.new>`, :meth:`Todo.new <icalendar.cal.todo.Todo.new>`, and :meth:`Availability.new <icalendar.cal.availability.Availability.new>`. :issue:`1065`
 
 Bug fixes
 ~~~~~~~~~
@@ -62,15 +60,10 @@ Minor changes
 
 - Show colorful required code changes in the CI output to help contributors solve the formatting issues. :pr:`1216`
 - Use ruff 0.15.0 for code formatting in :file:`tox.ini`. :pr:`1215`
-- Extended :func:`~icalendar.timezone.tzp.TZP.localize` to support localizing both 
-  :class:`datetime.datetime` and :class:`datetime.time` objects, returning timezone-aware 
-  :class:`datetime.time` objects for the latter. :issue:`1142`
 
 New features
 ~~~~~~~~~~~~
 
-- Updated :func:`icalendar.prop.dt.time.vTime.from_ical` to support parsing time values 
-  with TZID parameters, returning timezone-aware :class:`datetime.time` objects. :issue:`1142`
 - Added :attr:`Event.RECURRENCE_ID <icalendar.cal.event.Event.RECURRENCE_ID>`, :attr:`Todo.RECURRENCE_ID <icalendar.cal.todo.Todo.RECURRENCE_ID>`, and :attr:`Journal.RECURRENCE_ID <icalendar.cal.journal.Journal.RECURRENCE_ID>` properties, including support in their ``new()`` constructors. :issue:`1231`
 
 Bug fixes


src/icalendar/tests/test_time.py
```diff
@@ -28,3 +28,25 @@ def test_create_to_ical(x_sometime):
     cal = Calendar()
     cal.add("X-SOMETIME", datetime.time(17, 20, 10))
     assert b"X-SOMETIME;VALUE=TIME:172010" in cal.to_ical().splitlines()
+
+def test_vtime_multiple_timezones():
+    directory = os.path.dirname(__file__)
```
niccokunzmann: Please use the calendars fixture for this as in the other tests.

src/icalendar/tests/test_time.py
```diff
@@ -28,3 +28,25 @@ def test_create_to_ical(x_sometime):
...
+        cal = Calendar.from_ical(f.read())
+
+    events = list(cal.walk("VEVENT"))
+
+    named = events[0]["RDATE"].dts[0].dt
+    utc = events[1]["RDATE"].dts[0].dt
+    floating = events[2]["RDATE"].dts[0].dt
+
+    # Named timezone
+    assert named.tzinfo is not None
```
niccokunzmann, line 44: currently, this does not test that the timezone is actually the one that is expected. Could you add a test that checks that it is the correct timezone?

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,24 @@ def is_utc(self) -> bool:
...
 
     @staticmethod
-    def from_ical(ical):
-        # TODO: timezone support
+    def from_ical(ical, timezone=None):
+        tzinfo = None
+        if isinstance(timezone, str):
+            tzinfo = tzp.timezone(timezone)
+        elif timezone is not None:
+            tzinfo = timezone
```
niccokunzmann, line 174: I wonder if this code path is checked. Can you add a test for this?

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,24 @@ def is_utc(self) -> bool:
         return self.params.is_utc() or is_utc(self.dt)
 
     @staticmethod
-    def from_ical(ical):
-        # TODO: timezone support
+    def from_ical(ical, timezone=None):
```
niccokunzmann:
```suggestion
    def from_ical(ical:str, timezone:str|None|datetime.tzinfo=None) -> time:
        """Convert an ical string into a time."""
```

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,25 @@ def is_utc(self) -> bool:
         return self.params.is_utc() or is_utc(self.dt)
 
     @staticmethod
-    def from_ical(ical):
-        # TODO: timezone support
+    def from_ical(ical:str, timezone:str|None|datetime.tzinfo=None) -> time:
+        """Convert an ical string into a time."""
```
stevepiercy: Would you please enhance the docstring with a `Returns` and `Raises`, and include a more detailed description, per https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure?

This sentence is an excellent summary and should stay the same.

`Return type:` gets automatically generated, now that you've added the type hint. Yay!

You can inspect the result by building the docs locally, or when checking the pull request preview after Read the Docs builds it at https://icalendar--1276.org.readthedocs.build/en/1276/reference/api/icalendar.prop.dt.time.html#icalendar.prop.dt.time.vTime.from_ical

Thank you!

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,42 @@ def is_utc(self) -> bool:
...
+                ical = ical.decode()
+            utc = ical.endswith("Z")
+            if utc:
+                ical = ical[:-1]
             timetuple = (int(ical[:2]), int(ical[2:4]), int(ical[4:6]))
+            if tzinfo:
+                dt = datetime.combine(
+                    date(2020, 1, 1), time(*timetuple)
+                )  # reference date is arbitrary, we only care about the time and tzinfo
+                localized = tzp.localize(dt, tzinfo)
```
niccokunzmann: This code duplication (3x the same code) in front of tzp.localize hints at that:

tzp.localize should be changed to allow datetime.time to be passed to it and localized and take on this responsibility.

Could you do that refactoring? It might need a [type hint overload](https://docs.python.org/3/library/typing.html#typing.overload) to correctly state that if you pass a datetime.time, you get a datetime.time back.

clara-rodrigues: Sounds good, thank you. I will work on it today

src/icalendar/tests/test_time.py
```diff
@@ -28,3 +29,28 @@ def test_create_to_ical(x_sometime):
...
+
+    assert floating.tzinfo is None
+
+
+def test_vtime_with_tzinfo_object():
+    tz = datetime.timezone.utc
+
+    parsed = vTime.from_ical("083000", timezone=tz)
+
+    assert parsed.tzinfo == tz
```
niccokunzmann, line 57: Please add a new test for the localization of the time object:

```
def test_localize_time(tzp):

def test_localize_time_utc(tzp):

```

src/icalendar/timezone/pytz.py
```diff
@@ -29,8 +29,18 @@ def localize_utc(self, dt: datetime) -> datetime:
         # assume UTC for naive datetime instances
         return pytz.utc.localize(dt)
 
-    def localize(self, dt: datetime, tz: tzinfo) -> datetime:
-        """Localize a datetime to a timezone."""
+    @overload
```
niccokunzmann: The overloads work!

Image

This is not the right place, though.
The right place is in icalendar.timezone.tzp - wich is the public interface everyone uses.

- [ ] in localize_utc()
- [ ] in localize()

If you put it here, then it only works for pytz, but not zoneinfo. Hence, could you also add the two tests?

The fix has gotten way bigger than I thought! Sorry for that! I am really happy with your work :) Thank you!

clara-rodrigues: No problem! I can add it to icalendar.timezone.tzp. Nevertheless, this logic should also be present in pytz, right?

niccokunzmann: No, because pytz is used internally and has a restricted scope to reduce complexity. Pytz and zoneinfo only need to operate internally on datetime.

The strictness of that [strategy pattern](https://refactoring.guru/design-patterns/strategy) allows to have a nice interface to reduce complexity in the icalendar code, where the common timezone logic is handeled in the tzp class and only the logic that differs goes to pytz and to zoneinfo.

src/icalendar/timezone/tzp.py
```diff
@@ -71,14 +70,26 @@ def localize_utc(self, dt: datetime.date) -> datetime.datetime:
         """
         return self.__provider.localize_utc(to_datetime(dt))
 
+    @overload
+    def localize(self, dt: datetime, tz: datetime.tzinfo | str | None) -> datetime: ...
```
niccokunzmann:
```suggestion
    def localize(self, dt: datetime.datetime, tz: datetime.tzinfo | str | None) -> datetime.datetime: ...
```

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,38 @@ def is_utc(self) -> bool:
...
-    def from_ical(ical):
-        # TODO: timezone support
+    def from_ical(ical: str, timezone: str | None | tzinfo = None) -> time:
+        """Convert an ical string into a time.
+
+        Description:
+            This method supports parsing the three forms of time values defined in :rfc:`5545`:
+                - Local time (floating)
+                - UTC time
+                - Local time with time zone reference
```
stevepiercy: Description doesn't need a section heading. I also found the specific section that describes the three forms.

```suggestion
        This method supports parsing the three forms of time values defined in :rfc:`5545#section-3.3.12`:
            - Local time (floating)
            - UTC time
            - Local time with time zone reference
```

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,38 @@ def is_utc(self) -> bool:
...
+
+        Returns:
+            A datetime.time object representing the parsed time, with timezone information if applicable.
```
stevepiercy: Let's link to the Python standard library.

```suggestion
            A :class:`datetime.time` object representing the parsed time, with timezone information if applicable.
```

src/icalendar/prop/dt/time.py
```diff
@@ -152,10 +153,38 @@ def is_utc(self) -> bool:
...
+
+        Raises:
+            ~error.ValueError: if the provided string cannot be parsed as a time.
```
stevepiercy: `ValueError` is part of the Python standard library. It's imported in the icalendar package. Sphinx will automatically link to the Python standard library when the Returns class has correct syntax.

```suggestion
            ValueError: if the provided string cannot be parsed as a time.
```

src/icalendar/timezone/tzp.py
```diff
@@ -71,14 +70,28 @@ def localize_utc(self, dt: datetime.date) -> datetime.datetime:
...
+        self, dt: datetime.time, tz: datetime.tzinfo | str | None
+    ) -> datetime.time: ...
+
     def localize(
-        self, dt: datetime.date, tz: datetime.tzinfo | str | None
-    ) -> datetime.datetime:
-        """Localize a datetime to a timezone."""
+        self, dt: datetime.date | datetime.time, tz: datetime.tzinfo | str | None
+    ) -> datetime.datetime | datetime.time:
+        """Localize a datetime or time to a timezone."""
```
stevepiercy: Nice with the overloads. https://icalendar--1276.org.readthedocs.build/en/1276/reference/api/icalendar.timezone.tzp.html#icalendar.timezone.tzp.TZP.localize

Now let's add a Returns.

```suggestion
        """Localize a datetime or time to a timezone.
        
        Returns:
            -   A localized :class:`datetime.datetime` when a :class:`datetime.datetime` is given.
            -   A localized :class:`datetime.time` when a :class:`datetime.time` is given.
        """
```

CHANGES.rst
```diff
@@ -53,6 +53,9 @@ Minor changes
 
 - Show colorful required code changes in the CI output to help contributors solve the formatting issues. :pr:`1216`
 - Use ruff 0.15.0 for code formatting in :file:`tox.ini`. :pr:`1215`
+- Extended :func:`~icalendar.timezone.tzp.localize` to support localizing both
```
stevepiercy: Almost!

```suggestion
- Extended :func:`~icalendar.timezone.tzp.TZP.localize` to support localizing both 
```

stevepiercy (changes requested):
Thank you for your contribution and patience. I updated the branch to `main`, and now the change log entries are under an already released version. Would you please take care to move them under the unreleased version?

Other than that, I've made a few suggestions.

PR #1279, opened 2026-03-15
Add tests for Event.new() implicit end/duration per RFC 5545 §3.6.1

### LINKED ISSUE #1275: [Bug] Event.new doesn't set a DTEND when neither it nor DURATION are provided
## Describe the bug

According to the [iCalendar specification for VEVENT](https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.6.1):

> For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE value type but no "DTEND" nor "DURATION" property, the event's duration is taken to be one day. For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE-TIME value type but no "DTEND" property, the event ends on the same calendar date and time of day specified by the "DTSTART" property.

icalendar doesn't set this property.

## To reproduce

```python
import icalendar
import datetime as dt
import zoneinfo
from icalendar import Event

calendar = Calendar.new()
event = Event.new(
    start=dt.datetime(2026, 3, 21, 6, 30,0, tzinfo=zoneinfo.ZoneInfo("Europe/Berlin")),
)
calendar.add_component(event)
print(calendar.to_ical().decode())
```

```calendar
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//collective//icalendar//7.0.3.dev78//EN
UID:898b4925-3576-4893-a82f-0f01eaedb6a9
BEGIN:VEVENT
DTSTART;TZID=Europe/Berlin:20260321T063000
DTSTAMP:20260312T203408Z
UID:e07a90f5-bee8-4fab-8622-9202eb2d9090
CATEGORIES:
END:VEVENT
END:VCALENDAR
```

## Expected behavior

A calendar event should set its DTEND property to the DTSTART property when neither DTEND nor DURATION are set.

## Environment

 - [ ] Operating system: macOS 13.7.8 (22H730)
 - [ ] Python version: Python 3.14
 - [ ] `icalendar` version: 7.0.3.dev78

## Additional context

See https://github.com/plone/volto/issues/7993#issuecomment-4042453088

@SashankBhamidi @niccokunzmann @tobixen @abe-101 I'd like your interpretation of the RFC to verify what I think is the correct behavior before anyone works on this issue. Thank you!

tobixen:
## Closes issue

- [x] Closes #1275

## Description

This pull request was AI-generated.  

Asking Claude about opinions on #1275 and also asking if it's sufficient test code, it claimed that tests for `Event.new()` was missing.  It added tests with and without time zone as stevepiercy requested.  I've looked through the changes, LGTM.  Are those tests useful, or do they just add clutter and increased test run time?  Honestly, I don't know the answer to that.

This change introduces new tests, and no code changes.  I didn't spent much effort on this, and I promise not to break down and cry if the pull request is rejected.

Claude also wanted to add a line to CHANGES.rst, but I deem this to be too insignificant to clutter that file.

## Checklist

- [ ] I did not add a change log entry to `CHANGES.rst`.
- [ ] I did not add any documentation

### DIFF
--- a/src/icalendar/tests/test_issue_662_component_properties.py
+++ b/src/icalendar/tests/test_issue_662_component_properties.py
@@ -377,7 +377,7 @@ def test_event_new_datetime_no_dtend_implicit_zero_duration():
     assert "DTEND" not in event
 
 
-def test_event_new_datetime_tzaware_no_dtend_implicit_zero_duration():
+def test_event_new_datetime_tzaware_no_dtend_implicit_zero_duration(tzp):
     """Event.new() with a timezone-aware datetime and no end/duration: implicit zero duration.
 
     See https://github.com/collective/icalendar/issues/1275
@@ -385,8 +385,7 @@ def test_event_new_datetime_tzaware_no_dtend_implicit_zero_duration():
     This is the exact case reported in the issue: timezone-aware DTSTART with
     neither DTEND nor DURATION should yield zero duration, not raise an error.
     """
-    tz = ZoneInfo("Europe/Berlin")
-    start = datetime(2026, 3, 21, 6, 30, 0, tzinfo=tz)
+    start = tzp.localize(datetime(2026, 3, 21, 6, 30, 0), "Europe/Berlin")
     event = Event.new(start=start)
     assert event.end == start
     assert event.duration == timedelta(0)


src/icalendar/tests/test_issue_662_component_properties.py
```diff
@@ -361,6 +361,54 @@ def test_event_duration_one_day():
...
+
+    RFC 5545 §3.6.1: "For cases where a VEVENT specifies a DTSTART with a
+    DATE-TIME value type but no DTEND property, the event ends on the same
+    calendar date and time of day specified by DTSTART."
+    """
+    start = datetime(2026, 3, 21, 6, 30, 0)
+    event = Event.new(start=start)
+    assert event.end == start
+    assert event.duration == timedelta(0)
+    assert "DTEND" not in event
```
stevepiercy, line 377: Because we set `event.end`, should we also set `"DTEND"` when not given?

tobixen: Not in my opinion.  The user did not set any explicit end, so we do not add any explicit end to the event.

It's important to understand the difference between semantics and representation here.

An icalendar string representation of an event may either have a DURATION, a DTEND or none of them.  If the user didn't specify a explicit end or duration, I think it should not be included in the event.

The `.end`-property is not part of the representation, it's calculated - either DTEND, DTSTART+DURATION or DTSTART+1 DAY or DTSTART+0 seconds depending on the circumstances.  It's giving the semantic value, "what end does should this event actually have?".

niccokunzmann: Yes, I agree with @tobixen

src/icalendar/tests/test_issue_662_component_properties.py
```diff
@@ -361,6 +361,54 @@ def test_event_duration_one_day():
...
+def test_event_new_datetime_tzaware_no_dtend_implicit_zero_duration():
+    """Event.new() with a timezone-aware datetime and no end/duration: implicit zero duration.
+
+    See https://github.com/collective/icalendar/issues/1275
+
+    This is the exact case reported in the issue: timezone-aware DTSTART with
+    neither DTEND nor DURATION should yield zero duration, not raise an error.
+    """
+    tz = ZoneInfo("Europe/Berlin")
+    start = datetime(2026, 3, 21, 6, 30, 0, tzinfo=tz)
```
niccokunzmann:
```suggestion
def test_event_new_datetime_tzaware_no_dtend_implicit_zero_duration(tzp):
    """Event.new() with a timezone-aware datetime and no end/duration: implicit zero duration.

    See https://github.com/collective/icalendar/issues/1275

    This is the exact case reported in the issue: timezone-aware DTSTART with
    neither DTEND nor DURATION should yield zero duration, not raise an error.
    """
    start = tzp.localize(datetime(2026, 3, 21, 6, 30, 0), "Europe/Berlin")
```

This makes use the the tzp fixture and as such tests across pytz + zoneinfo.

PR #1281, opened 2026-03-16
Explained import shortcuts in API design documentation

### LINKED ISSUE #1161: Explain black magic of `Calendar.from_jcal` in API design reference docs
This looks like black magic to me because the link to the method does not align with the code sample. There's no `Calendar.from_jcal` method in the `icalendar.cal.calendar` module, yet the code example works! Is the black magic in `icalendar.cal.component`?

I think we need to explain this black magic somewhere in the [API design document](https://icalendar.readthedocs.io/en/latest/reference/design.html), and reference it here in the description of the code example.

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/1122#discussion_r2778670614_

stevepiercy:
Closes #1161

## Description

Explained import shortcuts in API design documentation.

## Checklist

- ~~[ ] I've added or updated tests if applicable.~~
- [ ] After merging, create redirect on Read the Docs of reference/component-api.html to explanation/api-design.html

----
📚 Documentation preview 📚: https://icalendar--1281.org.readthedocs.build/

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -47,11 +47,11 @@ jobs:
       - name: Create Python virtual environment, install requirements
         run: make .venv PYTHONVERSION=${{ matrix.python-version }}
 
-      - name: Check for broken links
-        run: make linkcheckbroken PYTHONVERSION=${{ matrix.python-version }}
-
       - name: Build HTML documentation
         run: make html PYTHONVERSION=${{ matrix.python-version }}
 
       - name: Run vale
         run: make vale VALEOPTS="--minAlertLevel='warning'" PYTHONVERSION=${{ matrix.python-version }}
+
+      - name: Check for broken links
+        run: make linkcheckbroken PYTHONVERSION=${{ matrix.python-version }}
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,10 +10,13 @@ Change log
 Minor changes
 ~~~~~~~~~~~~~
 
+- Make icalendar an explicit editable install for clarity. :pr:`1268`
 - Do not run some tests until a pull request is approved. :pr:`1246`
-- Created an :meth:`~icalendar.prop.vBoolean.ical_value` method for the :class:`~icalendar.prop.vBoolean` component. See :issue:`876`.
-- Created an :meth:`~icalendar.prop.vFloat.ical_value` method for the :class:`~icalendar.prop.vFloat` component. See :issue:`876`.
-- Created an :meth:`~icalendar.prop.vInt.ical_value` method for the :class:`~icalendar.prop.vInt` component. See :issue:`876`.
+- Created an :meth:`~icalendar.prop.boolean.vBoolean.ical_value` property for the :class:`~icalendar.prop.boolean.vBoolean` component. :issue:`876`
+- Created an :meth:`~icalendar.prop.float.vFloat.ical_value` property for the :class:`~icalendar.prop.float.vFloat` component. :issue:`876`
+- Created an :meth:`~icalendar.prop.integer.vInt.ical_value` property for the :class:`~icalendar.prop.integer.vInt` component. :issue:`876`
+- Created an :meth:`~icalendar.prop.binary.vBinary.ical_value` property for the :class:`~icalendar.prop.binary.vBinary` component. :issue:`876`
+- Put the link check as the last documentation CI task, allowing the documentation build and Vale to run first and fail faster. :pr:`1295`
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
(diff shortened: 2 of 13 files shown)


docs/reference/component-api.rst
```diff
@@ -4,15 +4,24 @@ API design
 
 This chapter describes the API design of icalendar.
 
+icalendar's API aligns with iCalendar components and properties as defined in :rfc:`5545`.
+
 At their core, iCalendar components are dictionaries with case insensitive keys.
 The preferred way to access their values is through lower case properties as described in :ref:`property-access`.
 Advantages and disadvantages of the other approaches are discussed below.
 
+As a convenience, icalendar offers shortcuts to import its classes instead of using their fully qualified Python path.
```
niccokunzmann: The interface that should be used is from icalendar import X. That is the stable, public interface. From icalendar.Y import X is discouraged.

docs/reference/component-api.rst
```diff
@@ -4,15 +4,24 @@ API design
...
+
+
 .. _property-access:
 
 Property access
 ===============
 
+icalendar builds components and subcomponents.
+It then adds, modifies, or removes properties to components and subcomponents, all within the RFC requirements.
+This is done in the :doc:`prop <../reference/api/icalendar.prop>` module.
```
niccokunzmann: In a development reference, I would add that for resposibility clarity. This is not quite right though and I would remove this or clarify the intention of why this us needed.

docs/explanation/api-design.rst
```diff
@@ -215,3 +224,26 @@ The following example sets the ``DESCRIPTION`` from  :rfc:`7986#section-5.1`, an
     PRODID:-//My calendar product//mxm.dk//
     DESCRIPTION;LANGUAGE=en:This is my personal calendar.
     END:VCALENDAR
+
+
+.. _import-shortcuts:
+
+Import shortcuts
```
niccokunzmann, line 242: While this section is right and still needs a focus on the !do not import from icalender subpackages!  Then, it would be more suitable.in the design decisions document as it has nothing to do with the decisions around how components relate to the rfc.

how components relate and map to the rfcs is the focus of this document.

Import shortcuts are a Python API structuring for maintenance and development and stability.

docs/reference/component-api.rst
stevepiercy: Addressed in b3127aae574f0110535279ef0774f799b3d9ae3a.

docs/reference/component-api.rst
stevepiercy: It's required because you can't understand properties without an understanding of components and subcomponents. Modified in b3127aae574f0110535279ef0774f799b3d9ae3a.

docs/explanation/api-design.rst
stevepiercy: This document describes the API design. It's an explanation of the reasoning behind the choices made. This section is an explanation of both how and why you implemented shorter imports. Where else would it go, beside API design?

stevepiercy: Actually, this file doesn't belong in `reference`, but should be moved into a new section, `explanation`, to align with Diátaxis. It's not at all like a dictionary, glossary, or API reference, but more of an explanation of the design decisions. We have nothing that explains choices made at this time. What do you think? See https://diataxis.fr/reference-explanation/.

niccokunzmann: Right! That makes total sense... It is an explanation

stevepiercy: This also brings into question [Design](https://icalendar.readthedocs.io/en/stable/reference/design.html), which is both a reference to all the RFCs and an explanation. I created a new issue https://github.com/collective/icalendar/issues/1292 proposing how to handle that.

stevepiercy: > Right! That makes total sense... It is an explanation

Done in 83b5852dd16ea2373b2ecb415101e7ef9b3dad80.

docs/explanation/api-design.rst
```diff
@@ -4,15 +4,35 @@ API design
...
+
+
 .. _property-access:
 
 Property access
 ===============
 
+As described in the foregoing section, icalendar builds components and subcomponents.
+It then adds, modifies, or removes properties to components and subcomponents, all within the RFC requirements.
+This is done in the :doc:`prop <../reference/api/icalendar.prop>` module.
```
niccokunzmann:
```suggestion
All property value data types specified by :rfc:`5545#section-3.3` and subsequent RFCs can be found in :mod:`icalendar.prop`.
```

niccokunzmann (changes requested):
Thanks, here are my suggestions.

docs/explanation/api-design.rst
stevepiercy: I made a slight wording change and linked to the specific section of the RFC.

PR #1284, opened 2026-03-17
fix(ci): mark conditional tests as skipped instead of pending

### LINKED ISSUE #1264: [GitHub Actions] "pending" tests should be "skipped"
Currently, our tests say that they are "pending" but actually, we would like them to not show up at all.

Is there a way that we can adjust the tests so that they get skipped if a certain condition is met? This message is a bit confusing.

grafik

## Documentation

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions

## See also

Originally posted by @stevepiercy in https://github.com/collective/icalendar/issues/1246#issuecomment-3981211987_

jigangz:
## Summary

Fixes #1264

Previously, tests that only run on main/approved PRs (Python 3.11-3.13, PyPy3) were simply **omitted from the matrix** on unapproved feature branches. This caused them to appear as **"pending"** ("Expected — Waiting for status to be reported") in GitHub required status checks, which was confusing.

### `.github/workflows/generate-matrix.py`
- Always include all jobs in the matrix (previously conditional jobs were skipped entirely)
- Added a `skip` boolean field to each matrix entry: `False` for always-run jobs, and `not RUN_ALL_JOBS` for conditional jobs

### `.github/workflows/tests.yml`
- Added `if: ${{ !matrix.skip }}` at the job level so GitHub correctly reports conditional jobs as **"skipped"** instead of **"pending"** when conditions aren't met

## Behavior

| Trigger | Before | After |
|---------|--------|-------|
| Feature branch (unapproved PR) | Extra jobs: **pending** | Extra jobs: **skipped** ✅ |
| Main / approved PR / stable branch | All jobs run | All jobs run ✅ |

## Checklist
- [x] No test changes needed (this is a CI configuration fix)

----
📚 Documentation preview 📚: https://icalendar--1284.org.readthedocs.build/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,6 @@ jobs:
 
   run-tests:
     needs: generate-matrix
-    if: ${{ !matrix.skip }}
     strategy:
       matrix:
         include: ${{ fromJSON(needs.generate-matrix.outputs.include) }}
@@ -36,7 +35,9 @@ jobs:
     name: ${{ matrix.test_name }}
     steps:
     - uses: actions/checkout@v6
+      if: ${{ !matrix.skip }}
     - name: Set up Python
+      if: ${{ !matrix.skip }}
       uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python_version }}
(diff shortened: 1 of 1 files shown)


jigangz:
Thanks for the review! I've addressed both issues:

- Fixed the ruff line-length errors in `generate-matrix.py`
- Moved the `if: ${{ !matrix.skip }}` condition from the step level to the job level — that's why the skipped jobs were still showing as "pending" rather than "skipped". With the condition at the job level, GitHub treats them as properly skipped.

Let me know if anything still looks off!

PR #1287, opened 2026-03-17
[Fix issue 1224] Components not necessarily equal

### LINKED ISSUE #1224: [Bug] Components not necessarily equal
## Describe the bug

It can be that equality depends on the order of equality.

## To reproduce

```python
>>> from icalendar import Calendar, Event
>>> calendar1 = Calendar()
>>> calendar2 = Calendar()
>>> calendar1.subcomponents = [event1, event2]
>>> calendar2.subcomponents = [event1, event1]
>>> calendar1 == calendar2
False
>>> calendar2 == calendar1
True

```

## Expected behavior

False always

## Environment

 - [x] Operating system:   Linux     
 - [x] Python version:      3.13     
 - [x] `icalendar` version:   7.0.1   

## Additional context

- [ ] The tests for this should be in test_equality.
- [ ] Make sure we have a linear run time where possible.
- [ ] Please test this with the calendars and then also the deep copy `.copy(recursive=True)`

551E5:
## Closes issue
- [ ] Closes #1224

## Description
Component's `__eq__` method (which Calendar inherits), after checking for same properties and same subcomponent amount, only checks one way whether subcomponents are the same. 

Therefore it misses inequality of subcomponents if the left side of the equality has a strict subset of the right side's subcomponents.

```python
>>> calendar1.subcomponents = [event1, event2]
>>> calendar2.subcomponents = [event1, event1]
>>> calendar1 == calendar2 # (event1 in [event1, event1]) and (event2 in [event1, event1])
False
>>> calendar2 == calendar1 # event1 in [event1, event2]
True
```
Fixed by also checking whether the other component's subcomponents are also in self's subcomponents.

## Checklist

## Additional information
No extra tests fail, and added tests pass only with the added changes.

----
📚 Documentation preview 📚: https://icalendar--1287.org.readthedocs.build/

### DIFF
--- a/src/icalendar/tests/test_equality.py
+++ b/src/icalendar/tests/test_equality.py
@@ -110,7 +110,7 @@ def test_calendar_equality_order():
     event2 = Event()
     calendar1.subcomponents = [event1, event2]
     calendar2.subcomponents = [event1, event1]
-    assert_equal(calendar1, calendar2)
+    assert_not_equal(calendar1, calendar2)
 
 
 def test_component_equality_order():
@@ -121,7 +121,7 @@ def test_component_equality_order():
     event2 = Event()
     component1.subcomponents = [event1, event2]
     component2.subcomponents = [event1, event1]
-    assert_equal(component1, component2)
+    assert_not_equal(component1, component2)
 
 
 def test_copy_does_not_copy_subcomponents(calendars, tzp):


src/icalendar/tests/test_equality.py
```diff
@@ -100,6 +102,38 @@ def test_copies_are_equal(source_file, tzp):
...
+    event1 = Event.new(start=datetime(2021, 1, 1, 12, 30, 0))
+    event2 = Event()
+    calendar1.subcomponents = [event1, event2]
+    calendar2.subcomponents = [event1, event1]
+    copy1 = calendar1.copy(recursive=True)
+    copy2 = calendar2.copy(recursive=True)
+    assert_equal(calendar1 == calendar2, False)
+    assert_equal(calendar2 == calendar1, False)
+    assert_equal(copy1 == copy2, False)
+    assert_equal(copy2 == copy1, False)
```
niccokunzmann: Have a look at how assert_equal works here. It does the check in both directions. I think, you will only need one line.

```suggestion
    assert_equal(calendar1, calendar2)
```

Same below.

src/icalendar/cal/component.py
```diff
@@ -692,6 +692,12 @@ def __eq__(self, other):
             if subcomponent not in other.subcomponents:
                 return False
 
+        # We now know the other component's subcomponents are not a strict subset
```
niccokunzmann, line 695: This is definitely a fix.

I am wondering whether this can be made a bit faster where the second loop only checks those components that have not been equal before. Let me know what you think, if you find a way that makes comparing a bit faster.

src/icalendar/tests/test_equality.py
551E5: Yeah, you are right, it checks in both directions, and other tests already cover the copies.

src/icalendar/cal/component.py
551E5: I have tried marking the positions matched in the second array by the first array, and simply checking whether all have been marked in the second loop. This is the same O(n^2), but should be a bit better (O(2n^2) -> O(n^2 + 2n)). 

In a quick test it performed worse (for lists of 10k elements, worst case scenario, 0.7s vs 0.5s), probably due to python overhead. For smaller lists they were similar. I don't think we can go much faster in python without hashing.

Issue #116, opened 2013-11-19
How to add 'X-APPLE-STRUCTURED-LOCATION'


joshhunt:
I'm looking to embed the new X-APPLE-STRUCTURED-LOCATION used in Mavericks to specify a rich location into .ics files.

Calendar in Mavericks builds it like this:

```
[...]
DTSTART:20131104T070000
CREATED:20131119T222157Z
X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS=367 George     Street\\nSydn
 ey CBD NSW 2000;X-APPLE-RADIUS=72;X-TITLE=367 George Street:geo:-33.8689
 97,151.206775
END:VEVENT
```

How can I add this into an Event object? I'm able to form the string myself, but I'm unsure how to add this into the Event object, of if it's even the right way to do it.

grosch:
Note that if you also have a LOCATION element that it must match exactly the location you specify in the X-APPLE-STRUCTURED-LOCATION.  If it differs, iCal won't display the location in a map.

timgentry:
The following worked for me, it's basically a 'geo' URI with additional properties
```
event.x_apple_structured_location = Icalendar::Values::Uri.new(
  "geo:-33.868900,151.207000",
  "X-ADDRESS": "367 George Street Sydney CBD NSW 2000",
  "X-APPLE-RADIUS": "72",
  "X-TITLE": "367 George Street"
)
```

PR #1288, opened 2026-03-17
bug: X-properties with VALUE param now parsed as correct type (fixes …

### LINKED ISSUE #1238: [Bug] X-Properties lose their type when parsed from ical
## Describe the bug

X- properties can have their type set. However, we lose that information on runtime.

## To reproduce

```python
import icalendar
cal = """
BEGIN:VCALENDAR
X-FILTER-DATE-RANGE;VALUE=PERIOD:20250202T000000/20250203T000000
END:VCALENDAR
"""
calendar  = icalendar.Calendar.from_ical(cal)
print(calendar["x-filter-date-range"])
```

Output:

```pycon
vText(b'20250202T000000/20250203T000000')

```

## Expected behavior

This should be `vPeriod`

## Environment

 - [x] Operating system:  linux       
 - [x] Python version:  3.13         
 - [x] `icalendar` version:    7.0.1  

## Additional context

See also:
- https://github.com/waldbaer/icalendar-events-cli/pull/42#issuecomment-3954399003

FilipAtanasovski-dev:
…#1238)

## Closes issue

Replace `ISSUE_NUMBER` with the issue number that your pull request fixes. Then GitHub will link and automatically close the related issue.

- [ 1238] Closes #1238

## Description

Write a description of the fixes or improvements.

## Checklist

## Additional information

Upload screenshots, videos, links to documentation, or any other relevant information.

----
📚 Documentation preview 📚: https://icalendar--1288.org.readthedocs.build/


src/icalendar/prop/factory.py
```diff
@@ -234,6 +234,10 @@ def for_property(self, name, value_param: str | None = None) -> type:
         # For unknown/custom properties, always use the default type from types_map
         if value_param and name in self.types_map and value_param in self:
             return self[value_param]
+        
+        if(value_param and value_param.lower() in self):
+            return self[value_param.lower()]
```
niccokunzmann:
```suggestion
        if(value_param and value_param in self):
            return self[value_param]
```

This is a caseless dict

src/icalendar/tests/test_issue_1238.py
```diff
@@ -0,0 +1,25 @@
...
+not fall back to vUnknown.
+"""
+import pytest
+from icalendar import Calendar
+from icalendar.prop.dt import vPeriod
+
+
+@pytest.mark.parametrize(("cal_str", "prop_name", "expected_type"), [
+    (
+        "BEGIN:VCALENDAR\r\nX-FILTER-DATE-RANGE;VALUE=PERIOD:20250202T000000/20250203T000000\r\nEND:VCALENDAR\r\n",
```
niccokunzmann: Please put this calendar into a file. Additional tests will run on it, then.

src/icalendar/tests/test_issue_1238.py
```diff
@@ -0,0 +1,25 @@
...
+        "X-FILTER-DATE-RANGE",
+        vPeriod,
+    ),
+])
+
+def test_x_property_respects_value_param(cal_str, prop_name, expected_type):
+    """X-properties with a VALUE parameter should be parsed as that type (issue #1238)."""
+    calendar = Calendar.from_ical(cal_str)
+    result = calendar[prop_name]
```
niccokunzmann: Please also check that the values are correct.

src/icalendar/tests/test_issue_1238.py
```diff
@@ -0,0 +1,25 @@
...
+    assert isinstance(result, expected_type), (
+        f"Expected {expected_type.__name__}, got {type(result).__name__}"
+    )
```
niccokunzmann: We need another test to check that if the value is broken - e.g. it says VALUE=PERIOD but the content is just text, that we get a vBroken from this.

That is good for future development and stability.

src/icalendar/prop/factory.py
FilipAtanasovski-dev: Done!

src/icalendar/tests/test_issue_1238.py
FilipAtanasovski-dev: Done!

src/icalendar/tests/test_issue_1238.py
FilipAtanasovski-dev: Done!

src/icalendar/tests/test_issue_1238.py
FilipAtanasovski-dev: Done! Though for this one im not sure if this is what you had in mind, as I had to change Calendar.ignore_exeptions as otherwise it just raised an exception and failed.
```python


def test_x_property_broken_value():
    """X-properties with VALUE=PERIOD but invalid content should yield vBroken (issue #1238)."""

    cal_str = "BEGIN:VCALENDAR\r\nX-FILTER-DATE-RANGE;VALUE=PERIOD:this-is-not-a-period\r\nEND:VCALENDAR\r\n"

    Calendar.ignore_exceptions = True
    try:
        calendar = Calendar.from_ical(cal_str)
    finally:
        Calendar.ignore_exceptions = False  # always restore default

    result = calendar["X-FILTER-DATE-RANGE"]
    assert isinstance(result, vBroken)
    assert result.expected_type == "vPeriod"

```

CHANGES.rst
```diff
@@ -32,6 +32,7 @@ New features
 Bug fixes
 ~~~~~~~~~
 
+- X-properties with a ``VALUE`` parameter are now parsed using the correct type instead of falling back to ``vUnknown`` (fixes `#1238 <https://github.com/collective/icalendar/issues/1238>`_).
```
```suggestion
- X-properties with a ``VALUE`` parameter are now parsed using the correct type instead of falling back to :class:`~icalendar.prop.unkown.vUnknown`. :issue:`1238`
```

FilipAtanasovski-dev:
I think I fixed it! Sorry for the inconvenience

niccokunzmann:
Now, there are test failures... O.O

How did that happen.

FilipAtanasovski-dev:
Yup, just found the issue in test_image.py in the rfc_7986_image.ics file that the test uses this is how it looks like
BEGIN:VCALENDAR
BEGIN:VEVENT
UID:uri-event
IMAGE;VALUE=URI;DISPLAY=BADGE;FMTTYPE=image/png:http://example.com/images/party.png
END:VEVENT
BEGIN:VTODO
UID:uri-todo
IMAGE;VALUE=URI;DISPLAY=BADGE:http://example.com/images/party.jpg
END:VTODO
BEGIN:VJOURNAL
UID:uri-journal
IMAGE;VALUE=URI;DISPLAY=ICON;FMTTYPE=image/jpg:http://example.com/images/party.jpg
END:VJOURNAL
BEGIN:VEVENT
UID:data-event
IMAGE;ENCODING=BASE64;VALUE=BINARY;FMTTYPE=image/png:iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAACXBIWXMAAAAnAAAAJwEqCZFPAAAAGXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAAA1JREFUCJlj+P//PwMACPwC/oXNqzQAAAAASUVORK5CYII=
END:VEVENT
BEGIN:VTODO
UID:data-todo
IMAGE;ENCODING=BASE64;VALUE=BINARY;FMTTYPE=image/png;ALTREP="http://example.com/images/party.jpg":iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAACXBIWXMAAAAnAAAAJwEqCZFPAAAAGXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAAA1JREFUCJlj+P//PwMACPwC/oXNqzQAAAAASUVORK5CYII=
END:VTODO
BEGIN:VJOURNAL
UID:data-journal
IMAGE;ENCODING=BASE64;VALUE=BINARY;FMTTYPE=image/png;DISPLAY=BADGE:iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAACXBIWXMAAAAnAAAAJwEqCZFPAAAAGXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAAA1JREFUCJlj+P//PwMACPwC/oXNqzQAAAAASUVORK5CYII=
END:VJOURNAL
END:VCALENDAR

the parameter name is IMAGE which is not one of the allowed types in factory.py

Thus when it was parsed in the test the type was automatically set to vUnknown in the original test, which isnt what was supposed to happen, thus the test was made around the vUknown type then when the new correct type gets introduced the test fails.

FilipAtanasovski-dev:
Tommorrow I can update those tests? If you agree with my assesment?

niccokunzmann:
In test issue 116, the important part is that all the information is preserved.
```
b"BEGIN:VEVENT\r\nX-APPLE-STRUCTURED-LOCATION;VALUE=URI;"
        b'X-ADDRESS="367 George Street Sydney \r\n CBD NSW 2000";'
        b'X-APPLE-RADIUS=72;X-TITLE="367 George Street":'
        b"geo:-33.868900\r\n \\,151.207000\r\nEND:VEVENT\r\n"
```

It appears that the difference is here:

`geo:-33.868900\r\n \\,151.207000` -> `geo:-33.868900\r\n ,151.207000` without the comma. The change is better that what was there before because #116 clearly does not want a \\ in the geo URI. So, the test could be changed to better reflect the issue. Good find!


---

> the parameter name is IMAGE which is not one of the allowed types in factory.py

Yes, the problem that we have is that IMAGE does not have a default type because it was introduced later that RFC 5545 and has two types possible.

the tests in test_image.py are testing the Image class and the interface that allows programmers to access images in a unified way. As such, the interface should stay the same. If now the implementation has to be adjusted, then that is alright. 

> Thus when it was parsed in the test the type was automatically set to vUnknown in the original test, which isnt what was supposed to happen, thus the test was made around the vUknown type then when the new correct type gets introduced the test fails.

Yes. I would say do the changes as you need them and I will review to make sure it will work with the whole icalendar library.

src/icalendar/tests/rfc_7265_jcal/test_section_5_3_examples.py
```diff
@@ -16,29 +16,6 @@ def test_convert_coffee(calendars):
...
-    [
-        ("SUMMARY", "TEXT", "TEXT"),
-        ("SUMMARY", "URI", "URI"),
-        ("DESCRIPTION", "URI", "URI"),
-        ("X-ALT-DESC", "UNKNOWN", "UNKNOWN"),
-        ("X-ALT-DESC", "TEXT", "UNKNOWN"),
-        ("DESCRIPTION", "TEXT", "TEXT"),
-    ],
-)
-def test_know_attributes_do_not_receive_a_VALUE(prop_name, value, expected_value):
```
niccokunzmann, line 30: Hi, this test should not be removed.

src/icalendar/prop/factory.py
```diff
@@ -234,10 +234,13 @@ def for_property(self, name, value_param: str | None = None) -> type:
         # For unknown/custom properties, always use the default type from types_map
         if value_param and name in self.types_map and value_param in self:
             return self[value_param]
-
-        if value_param and value_param in self:
+                 
+        if name.upper() == "IMAGE":
```
niccokunzmann, line 238: Hm. Instead of doing this, it would be better to fix the vBinary class. If the image cannot use it, it means that nobody can use it. So, avoiding it is a fix for the image class but actually does not solve the underlying problem that vBinary is broken. Did you try removing to_unicode in the assignment to self.obj in vBinary? That is just a remnant from Python 2 days.

src/icalendar/tests/rfc_7265_jcal/test_section_5_3_examples.py
```diff
@@ -16,30 +16,6 @@ def test_convert_coffee(calendars):
...
 
-@pytest.mark.parametrize(
-    ("prop_name", "value", "expected_value"),
-    [
-        ("SUMMARY", "TEXT", "TEXT"),
-        ("SUMMARY", "URI", "URI"),
-        ("DESCRIPTION", "URI", "URI"),
-        ("X-ALT-DESC", "UNKNOWN", "UNKNOWN"),
-        ("X-ALT-DESC", "TEXT", "UNKNOWN"),
-        ("DESCRIPTION", "TEXT", "TEXT"),
```
niccokunzmann, line 27: Hi, this test should not be removed.
Can you switch it over so that these values are correct? This is a test for the the jcal RFC.

FilipAtanasovski-dev: Hi! Yes I will do so!

FilipAtanasovski-dev: Done!

src/icalendar/prop/factory.py
```diff
@@ -234,6 +234,15 @@ def for_property(self, name, value_param: str | None = None) -> type:
...
             return self[value_param]
+
+        if name.upper() == "IMAGE":
+            return self[
+                "unknown"
+            ]  # IMAGE is always unknown, even if VALUE is URI or BINARY
+
+        if value_param and (value_param in self) and value_param != "IMAGE":
+            return self[value_param]
+
```
niccokunzmann, line 245: I mark this here. For solving the issue #1289 , these changes need reverting.

src/icalendar/tests/test_issue_1238.py
```diff
@@ -0,0 +1,36 @@
...
+def test_x_property_broken_value():
+    """X-properties with VALUE=PERIOD but invalid content should yield vBroken (issue #1238)."""
+
+    cal_str = "BEGIN:VCALENDAR\r\nX-FILTER-DATE-RANGE;VALUE=PERIOD:this-is-not-a-period\r\nEND:VCALENDAR\r\n"
+
+    Calendar.ignore_exceptions = True
+    try:
+        calendar = Calendar.from_ical(cal_str)
+    finally:
+        Calendar.ignore_exceptions = False  # always restore default
```
niccokunzmann:
```suggestion
    calendar = Calendar.from_ical(cal_str)
```

To be compliant with the standard, this would need to be changed.
Also, the content of vBroken needs checking.

Would you like to take this on or would you like me to open another issue for this?

FilipAtanasovski-dev: How would you like it to be changed? the only way to generate vBroken is to enable ignore_exeptions? otherwise it just throws an error? What would the expected behaviour be here? I could check if it throws an error instead?

FilipAtanasovski-dev: something like this?
```python
def test_x_property_broken_value():
    """X-properties with VALUE=PERIOD but invalid content should yield an error (issue #1238)."""

    cal_str = "BEGIN:VCALENDAR\r\nX-FILTER-DATE-RANGE;VALUE=PERIOD:this-is-not-a-period\r\nEND:VCALENDAR\r\n"

    with pytest.raises(ValueError):
        calendar = Calendar.from_ical(cal_str)
```

FilipAtanasovski-dev: also i found this test in test_unit_cal.py
```python
def test_cal_strict_parsing(calendars):
    """If components are damaged, we raise an exception."""
    with pytest.raises(ValueError):
        calendars.parsing_error_in_UTC_offset
```
does this not need to return vBroken?

niccokunzmann: There is a difference for me with X- attributes and RFC specified attributes. An invalid RFC specified attribute is an error in the calendar. A custom attribute should be ignored if it is not understood.

I am feeling into how this affects backwards compatibility.

It appears that these attributes are recorded:

```
$ grep -rE '^X-.*VALUE' src/icalendar/tests/calendars/
src/icalendar/tests/calendars/period_with_timezone.ics:X-WR-CALNAME;VALUE=TEXT:Test RDATE
src/icalendar/tests/calendars/x_location.ics:X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS="Röadstar 16\n12764 Happyvi
src/icalendar/tests/calendars/time.ics:X-SOMETIME;VALUE=TIME:172010
```

Since the tests run with your change, it should be alright to leave it like this and open another issue. Maybe, we don't get any complaints. Also, if someone bothers to specify a type on an X- attribute, that might mean that its form complies to the type - however, this is not according to standard.

src/icalendar/tests/test_issue_1238.py
```diff
@@ -0,0 +1,36 @@
...
+    """X-properties with a VALUE parameter are parsed as the correct type (issue #1238)."""
+    calendar = calendars.issue_1238
+    result = calendar["X-FILTER-DATE-RANGE"]
+    assert isinstance(result, vPeriod), f"Expected vPeriod, got {type(result).__name__}"
+    assert result.start == datetime(2025, 2, 2, 0, 0)
+    assert result.end == datetime(2025, 2, 3, 0, 0)
+
+
+def test_x_property_broken_value():
+    """X-properties with VALUE=PERIOD but invalid content should yield vBroken (issue #1238)."""
```
niccokunzmann:
```suggestion
    """X-properties with VALUE=PERIOD but invalid content should yield vBroken (issue #1238).

    View the discussion:
    https://github.com/collective/icalendar/pull/1288#discussion_r2962648845
    """
```

src/icalendar/tests/test_issue_1238.py
FilipAtanasovski-dev: Hmmm. Intresting, would you maybe like it to be something like this?
```python
                        if not component.ignore_exceptions and not name.startswith("X-"):
                            raise
                        # Error-tolerant mode: create vBroken
                        expected_type = getattr(factory, "__name__", "unknown")
                        broken_prop = vBroken.from_parse_error(
                            raw_value=val,
                            params=params,
                            property_name=name,
                            expected_type=expected_type,
                            error=e,
                        )
```
with this change, everything works like you expected it to and in my test the flag is no longer necessary as it correctly returns vBroken but the remaining tests also work because the test i indicated does not use the X- variables

FilipAtanasovski-dev: I tested it and it should pass all tests let me know if i should commit!

niccokunzmann (dismissed):
Thanks for the discussion and the patience. This brings us closer to the goal and as such, I will merge this. We have a minor change in the way that icalendar works but that will be recorded as an issue.

src/icalendar/tests/test_issue_1238.py
niccokunzmann: Oh. Yes, if you could do that, I do not need to open another issue. I think, it should be `name[:2].upper() == "X-"`
Could you do that?

niccokunzmann: I was just about to merge and open an issue. Thanks for responding. I added a link to the discussion in the test docsstring.

FilipAtanasovski-dev: Yup done, I added it!

PR #1294, opened 2026-03-22
fix(ci): exclude skipped tasks from test matrix and report them as skipped

### LINKED ISSUE #1286: [GitHub Actions] Mark skipped tasks as skipped
Currently, we run all tasks even if they should be skipped.

There is a possibility to skip tasks:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions

grafik
grafik

To fix this issue:

- [ ] mark tasks as skipped if they are skipped 
- [ ] do not run the tests in them

See also:

- #1284 
- #1264

mvanhorn:
## Closes issue

- [x] Closes #1286

## Description

Follow-up to #1284. Previously, matrix entries with `skip=True` still created
runner jobs where every step was individually skipped via `if: !matrix.skip`.
This made them show as "success" (green check) instead of "skipped", and they
still consumed runner capacity.

This PR splits the matrix output into two lists:
- `include`: entries that should actually run (no `skip` key)
- `skipped`: test names for entries that should be skipped

The `run-tests` job now only receives non-skipped entries, so no runners are
allocated for skipped tests. A new `report-skipped` job uses the GitHub Checks
API (`checks.create` with `conclusion: 'skipped'`) to report skipped entries,
keeping branch protection rules satisfied.

### Changes

**`.github/workflows/generate_matrix.py`**
- `generate_matrix()` returns `{"include": [...], "skipped": [...]}` instead of a flat list
- Non-skipped entries no longer carry the `skip` key

**`.github/workflows/tests.yml`**
- `generate-matrix` job outputs both `include` and `skipped`
- `run-tests` uses only non-skipped entries; step-level `if: !matrix.skip` removed
- New `report-skipped` job creates check runs with `conclusion: 'skipped'`

**`.github/workflows/test_generate_matrix.py`**
- Updated for new dict return format
- Added assertions: running and skipped sets are disjoint and together equal CASES_ALL

## Checklist

## Additional information

All 23 existing tests pass with the updated format. The `report-skipped` job
requires `checks: write` permission, which is available by default for workflows
triggered by the same repository.

This contribution was developed with AI assistance (Claude Code).


.github/workflows/generate_matrix.py
```diff
@@ -151,7 +151,15 @@ def generate_matrix(git_ref, review):
             if run["tox_env"] != "py":
                 run["test_name"] += f" ({run['tox_env']})"
 
-    return matrix
+    include = []
+    skipped = []
+    for run in matrix:
+        if run["skip"]:
+            skipped.append(run["test_name"])
```
niccokunzmann: I think, for easy, it would be cool to append the whole `run` variable. Also, we do not need to remove the `skip` content.

.github/workflows/tests.yml
```diff
@@ -18,12 +18,15 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       include: ${{ steps.set-matrix.outputs.include }}
+      skipped: ${{ steps.set-matrix.outputs.skipped }}
```
niccokunzmann: Is there documentation for this? If you can find it, please add a comment link above this line to the docs.

.github/workflows/tests.yml
```diff
@@ -66,6 +64,45 @@ jobs:
         allow-empty: true
         fail-on-error: false
 
+  report-skipped:
+    needs: generate-matrix
+    if: ${{ needs.generate-matrix.outputs.skipped != '[]' }}
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
```
niccokunzmann: That is the problem... We really need it for PRs. Is there a way that a test case can report itself as skipped?

Image

.github/workflows/tests.yml
```diff
@@ -66,6 +64,45 @@ jobs:
...
+                conclusion: 'skipped',
+                output: {
+                  title: name,
+                  summary: 'Skipped by the workflow test matrix.',
+                },
+              });
+            } catch (err) {
+              // checks.create requires write permission, which is not
+              // available for workflows triggered by fork pull requests.
+              core.info(`Could not report ${name} as skipped: ${err.message}`);
```
niccokunzmann: This errors, have a look in the output.
I would prefer to remove this job totally if possible because that reduces maintenance overhead.

.github/workflows/generate_matrix.py
mvanhorn: Fixed - keeping the full `run` dict now. Simplified to a list comprehension.

.github/workflows/tests.yml
mvanhorn: Added a link to the GitHub Actions job outputs docs.

.github/workflows/tests.yml
mvanhorn: You're right - `checks: write` doesn't work for fork PRs. Removed the job entirely. The tradeoff is skipped tests are invisible in CI status, but no runners are wasted on them. If branch protection requires those specific test names to appear, that would need a config adjustment.

.github/workflows/tests.yml
mvanhorn: Removed.

mvanhorn:
Addressed in d90a739:

1. `generate_matrix()` keeps the full `run` dict (simplified to list comprehension)
2. Removed `report-skipped` job entirely - no runners wasted, no permissions issues on fork PRs
3. Added doc link for workflow job outputs

One question: does branch protection require specific test names to be present? If so, skipped tests won't appear at all now. If protection checks the `run-tests` job overall, this should work fine.

.github/workflows/tests.yml
niccokunzmann: Let's have a look at the CI run. Branch protection might "expect" them to run.

niccokunzmann: I wonder if GitHub even allows branch protection to require those jobs and the jobs to be skipped if they are protected.

niccokunzmann:
I was wondering if one can do something like this:

https://matthewrich.com/2022/10/13/github-actions-job-output/

And set job.result to skipped or something alike.

> jobs.<job_id>.result	string	The result of a job in the reusable workflow. Possible values are success, failure, cancelled, or skipped.

```
echo "result='skipped'" >> $GITHUB_OUTPUT
```

PR #1314, opened 2026-04-11
deprecate q_join, q_split and foldline for icalendar 8 (#1011)

### LINKED ISSUE #1011: Make private functions private by adding a _ at the start
`icalendar` allows importing everything that is designated for the public API with `from icalendar import *`. Everything that is not mentioned in `icalendar.__all__`, is private. As such, the convention is to prefix those with `_`. 

- [ ] rename all private functions and methods to start with `_`
- [ ] Rather create a small Pull Request several times than a big one that takes ages to merge.
- [ ] check back about unclear functions/classes

See also:
- https://github.com/collective/icalendar/issues/1004
- https://github.com/collective/icalendar/pull/979#discussion_r2550928113

siddhirajkatkar:
Closes #1011

Deprecate the following public functions by renaming them to private 
and adding deprecation wrappers:

- `q_join` → `_q_join`
- `q_split` → `_q_split`  
- `foldline` → `_foldline`

Changes:
- Added `deprecate_for_version_8` helper in `compatibility.py`
- Updated all internal call sites to use private names
- Added deprecation wrappers that warn users about removal in v8
- Added changelog entries in `CHANGES.rst`
- Updated tests to use private names

----
📚 Documentation preview 📚: https://icalendar--1314.org.readthedocs.build/


CHANGES.rst
```diff
@@ -11,6 +11,9 @@ Minor changes
 ~~~~~~~~~~~~~
 
 - Deprecate ``icalendar.parser.escape_string`` and ``icalendar.parser.unescape_string`` for icalendar version 8. Use ``_escape_string`` and ``_unescape_string`` internally. :issue:`1011`
+- Deprecate ``icalendar.parser.q_split`` for icalendar version 8. Use the private ``_q_split`` internally.
+- Deprecate ``icalendar.parser.q_join`` for icalendar version 8. Use the private ``_q_join`` internally.
+- Deprecate ``icalendar.parser.foldline`` for icalendar version 8. Use the private ``_foldline`` internally.
```
niccokunzmann: Please put them in one comment and mention the issue.
It seams that you are missing deprecated methods.

niccokunzmann (changes requested):
Could youplease explain the difference between the issue, the code and why you don't use the helper you created?
If you used an AI model, please disclose it and read our contribution policy.

CHANGES.rst
siddhirajkatkar: Hi @niccokunzmann,

I want to be transparent — I used AI to help 
navigate the codebase and implement the changes. I have read your 
contribution policy.

Here is my understanding of the changes:

**The issue (#1011):** Public functions that are not part of the public API 
should be renamed to start with _ and have deprecation wrappers so existing 
users get a warning before they are removed in v8.

**What we changed:**
- `q_join`, `q_split` (in parser/parameter.py) and `foldline` (in parser/string.py) 
  were still fully public with no deprecation wrappers
- Renamed each to `_q_join`, `_q_split`, `_foldline`
- Added public wrappers that call `deprecate_for_version_8` from compatibility.py
- Updated all internal call sites to use the private names directly

**Regarding the helper:** The `deprecate_for_version_8` helper in compatibility.py 
IS being used in all 3 wrappers. Happy to clarify if something looks wrong.

I am learning open source contribution and take full responsibility for 
reviewing and understanding every change in this PR.

src/icalendar/parser/parameter.py
```diff
@@ -50,10 +52,29 @@ def param_value(
...
+    value: Sequence[str] | str | HasToIcal, always_quote: bool = False
+) -> str:
+    """Convert a parameter value to its iCalendar representation.
+
+    .. deprecated:: 7.0.0
+        Use the private :func:`_param_value` internally. For external use,
+        this function is deprecated. Please contact the maintainers if you
+        rely on this function.
+    """
+    warnings.warn(
```
niccokunzmann, line 71: Hi, this would be where you can use the wrapper.

src/icalendar/compatibility.py
```diff
@@ -7,8 +7,28 @@
 Members will be added and removed without deprecation warnings.
 """
 
+import warnings
 from typing import TYPE_CHECKING
 
+
+def deprecate_for_version_8(name: str) -> None:
```
niccokunzmann: The best use of this with the least duplication would be to create a wrapper using functools.wraps.

This allows
- copying the docstring
- copying the type hints
- the docstring can automatically replace the function name with _ at the start, so the string is adjusted.
Then, all the code changes would look like this:

```
def _to_unicode(...): ...
    """... docstring for
    _to_unicode ...
    """
    ...
to_unicode = deprecate_for_version_8(_to_unicode)
```

And we can replace `_to_unicode` in the docstring with `to_unicode` and also add a generic comment to the end of the docstring.

That should remove a lot of code changes.

CHANGES.rst
niccokunzmann: Hi, thanks for disclosing this. Which tools are you using?
I left a comment about which direction I would like this to go in :)

src/icalendar/compatibility.py
siddhirajkatkar: Hi @niccokunzmann, thank you for the explanation and your patience!

I understand now — I should use the deprecate_for_version_8 wrapper 
instead of writing the warning manually inside each function. So the 
pattern should be:

def _q_join(...):
    ...

q_join = deprecate_for_version_8(_q_join)

This is cleaner and removes a lot of duplicate code. I will update 
the PR with this approach. Please let me know if I am on the right track!

CHANGES.rst
siddhirajkatkar: I am using Claude .

I have read your comments and understand the direction — using 
deprecate_for_version_8 as a functools.wraps wrapper instead of 
manually writing warnings inside each function. I am working on 
updating the PR now!

siddhirajkatkar:
Hi, I have updated the PR:

1. deprecate_for_version_8 now uses functools.wraps
2. q_join, q_split and foldline now use the pattern: public_name = deprecate_for_version_8(_private_name)
3. CHANGES.rst combined into one entry with :issue:`1011`
4. No manual warnings.warn calls left inside the private functions

All tests passing and ruff clean. Please review!

src/icalendar/compatibility.py
niccokunzmann: Yes! I think, you got it :)

niccokunzmann:
Hi, we created a new release, please move your edits in CHANGES.rst to the new section.

stevepiercy:
@siddhirajkatkar thanks for your contribution. However, you haven't responded in over a month, and other contributors have stepped up to complete the work you started, creating many merge conflicts. Closing as abandoned.

You're welcome to reopen and resume work on this PR at any time, but it might be easier to start from a clean branch.

PR #1268, opened 2026-03-05
Make icalendar an explicit editable install for clarity, add docs for Sphinx extension sphinx-icalendar usage

stevepiercy:
- See https://github.com/collective/icalendar/pull/1254#issuecomment-3977169488

## Description

- Make icalendar an explicit editable install for clarity.
- Add documentation for usage of the Sphinx extension [sphinx-icalendar](https://sphinx-icalendar.readthedocs.io/en/latest/)

## Checklist

- [X] I've added a change log entry to `CHANGES.rst`.
- [ ] I've added or updated tests if applicable.
- [X] I've run and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [X] I've added or edited documentation, both as docstrings to be rendered in the API documentation and narrative documentation, as necessary.

See documentation preview at:

https://icalendar--1268.org.readthedocs.build/en/1268/contribute/documentation/style-guide.html#calendar-file-code-blocks

PR #1254, opened 2026-02-28
Add syntax highlighting for icalendar code block

Parth-Vasave:
## Closes issue

- [x] Closes #1253

## Description

Add the [sphinx-icalendar](https://sphinx-icalendar.readthedocs.io) extension for syntax highlighting of iCalendar code blocks.

- Add `sphinx-icalendar>=0.1.0` to docs dependencies in [pyproject.toml](cci:7://file:///Users/parth/HomeFolder/temp/OpenContributions/icalendar/pyproject.toml:0:0-0:0)
- Add `sphinx_icalendar` to extensions in [docs/conf.py](cci:7://file:///Users/parth/HomeFolder/temp/OpenContributions/icalendar/docs/conf.py:0:0-0:0)
- Replace `.. code-block:: text` → `.. code-block:: ics` in [docs/reference/design.rst](cci:7://file:///Users/parth/HomeFolder/temp/OpenContributions/icalendar/docs/reference/design.rst:0:0-0:0) and across 20 source code files (54 blocks total)
- Blocks containing ABNF grammars, ASCII art, or non-iCalendar content are left as [text](cci:1://file:///Users/parth/HomeFolder/temp/OpenContributions/icalendar/src/icalendar/attr.py:1024:0-1054:43)

## Checklist

- [x] I've added a change log entry to `CHANGES.rst`.
- [x] I've added or updated tests if applicable.
- [x] I've run and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [x] I've added or edited documentation, both as docstrings to be rendered in the API documentation and narrative documentation, as necessary.

## Additional information

No functional code changes.. only the Sphinx code-block language hint is updated from [text](cci:1://file:///Users/parth/HomeFolder/temp/OpenContributions/icalendar/src/icalendar/attr.py:1024:0-1054:43) to `ics` for proper syntax highlighting.

----
📚 Documentation preview 📚: https://icalendar--1254.org.readthedocs.build/

Issue #1101, opened 2026-01-25
Completely remove pytz module and its tests and docs


stevepiercy:
We [deprecated pytz](https://icalendar.readthedocs.io/en/latest/how-to/upgrade.html#pytz-support). The closing criteria for this issue are the following.

- [ ] Announce in the next release of 7.x.x that `pytz` support will be removed in 8.0.0.
- [ ] Remove the module `icalendar.timezone.pytz`.
- [ ] Remove or replace with `zoneinfo` any tests, classes, methods, or functions that rely on `pytz`.
- [ ] ~~Run `make apidoc` to remove the page of docs automatically.~~ Obsolete, as Sphinx now regenerates the Reference API stub files automatically on each build.
- [ ] Check the PR preview to make sure its page is gone from https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.timezone.pytz.html (this is the latest version)
- [ ] Create a redirect on RTD from that file to the root of the Reference API https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.html.
- [ ] Remove or replace with `zoneinfo` any usage examples that use `pytz` from the documentation
- [ ] Remove any dependencies from `pyproject.toml`, GitHub workflows, and tox that use `pytz`.

Any others?

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/1100#discussion_r2725515655_

niccokunzmann:
Awr.. I feel this thing that developers have when they are a bit precious over the code they wrote :) I will feel into it - it is not a no, it just means that I will look into how to remove it and if I want to do it. Also, we can use the current structure to support other timezone implementations. Like for different Python implementations that are not file based. However, nobody really requests different timezone implementations... #1073 Hm.

PR #1315, opened 2026-04-11
Switch to uv for development

### LINKED ISSUE #1102: Switch to uv for development

stevepiercy:
## Closes issue

- Closes #1102

## Description

Switch to uv for development.

Update documentation with new commands.

## Checklist

- [ ] ~~I've added or updated tests if applicable.~~

----
📚 Documentation preview 📚: https://icalendar--1315.org.readthedocs.build/


stevepiercy:
@niccokunzmann I can't remember what we did to fix this error https://github.com/collective/icalendar/actions/runs/24280322355/job/70901092097?pr=1315#step:6:74. Do you remember?

niccokunzmann:
> warning: `icalendar` defines settings for `uv_build` in `tool.uv.build-backend`, but uses `hatchling.build` as build backend instead
 
This is one indicator.

```
Because only sphinx-icalendar<=0.1.0 is available and
      sphinx-icalendar==0.1.0 depends on your project, we can conclude that
      sphinx-icalendar>=0.1.0 depends on your project.
      And because icalendar:dev depends on sphinx-icalendar>=0.1.0 and your
      project requires icalendar:dev, we can conclude that your project's
      requirements are unsatisfiable.

      hint: The package `sphinx-icalendar` depends on the package `icalendar`
      but the name is shadowed by your project. Consider changing the name of
      the project.

```

These are our edits for the problem:
- https://github.com/collective/icalendar/pull/1268/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R301
- https://github.com/collective/icalendar/pull/1254/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

I think, it somehow instructs UV to use icalendar from that source.
Fetching tags was one idea but that should not be it, really.

Also the clash between uv and hatchling is something I don't know much about. We could also change from hatchling to uv as a build backend. I am not attached. The only thing that I like having is the ability to release a new version based on tags. But that could be possible anyway with this: https://stackoverflow.com/a/32965521/1320237, so we can keep the same interface we have having at the moment, with `icalendar.version`.

.github/workflows/tests.yml
```diff
@@ -57,7 +59,8 @@ jobs:
       if: ${{ !matrix.skip }}
       run: |
         python -m pip install --upgrade pip
-        pip install tox coverage-python-version
+        pip install tox coverage-python-version tox-uv
```
niccokunzmann: Would a `uv install` do the whole bunch or too much?

docs/_include/prerequisites.inc
```diff
@@ -0,0 +1,56 @@
...
+You'll need to first install prerequisites on, and clone the icalendar repository to, your local computer.
+
+Tests, code formatting, and documentation builds and checks require that you install :program:`GNU Make`, :program:`uv`, and :program:`Git`.
+
+.. note::
+
+    :program:`uv` handles the installation of Python and all dependencies for icalendar.
+
+
+:program:`Make`
```
niccokunzmann, line 13: If `uv` is lowercase - make could also be?

docs/_include/prerequisites.inc
```diff
@@ -0,0 +1,56 @@
...
+
+    .. tab-item:: Windows
+
+        .. code-block:: shell
+
+            powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
+
+.. seealso::
+
+    `Other uv installation methods <https://docs.astral.sh/uv/getting-started/installation/>`_
```
niccokunzmann: I think, this could be another tab. Maybe now.

docs/contribute/development.rst
```diff
@@ -5,55 +5,36 @@ Development
...
-.. code-block:: shell
+:program:`Make` commands
+------------------------
 
-    pip install tox
+With all :ref:`prerequisites <development-prerequisites>` installed, you can run tests, reformat code, make various documentation builds, perform quality checks, and clean your environment.
+All :program:`Make` commands use the file :file:`Makefile` at the root of the repository.
 
-From now on, tox will manage Python versions and test commands for you.
+The :program:`Make` commands may invoke :program:`uv`, :program:`tox`, :program:`Sphinx`, :program:`Vale`, shell commands, and other programs.
```
niccokunzmann, line 37: Personally, I think it is ok to use lowercase for programs but maybe there is a reason.

docs/contribute/development.rst
```diff
@@ -105,35 +86,34 @@ The screenshot below shows how to configure PyCharm to disable pre-commit by sea
...
 
-You might not have all Python versions installed or you may want to run a specific one.
-The following command show how to run tox with Python 3.12:
+The following command shows how to run tests in the Python 3.14.
 
-.. code-block:: shell
+..  code-block:: shell
 
-    tox -e py312
+    make test TOXENV=py314
```
niccokunzmann: for tox, I use `TOX_ENV`

pyproject.toml
```diff
@@ -300,10 +302,17 @@ filterwarnings = [
     "ignore::DeprecationWarning",
 ]
 
+[tool.uv]
+python-preference = "managed"
+
+[tool.uv.build-backend]
+namespace = true
+module-name = "icalendar"
```
niccokunzmann: The only difference for the build error that I can see are these lines here. So, maybe they can be experimented with.

tox.ini
```diff
@@ -33,23 +31,17 @@ commands =
     coverage xml
 
 [testenv:docs]
-deps =
-    -r {toxinidir}/requirements_docs.txt
-    setuptools
-changedir = docs
+dependency_groups = docs
```
niccokunzmann, line 30: We could remove this command. If you need the matrix to be adapted then, we can change it to run based on `make` or on `uv` or provide the full command.

niccokunzmann:
Thanks for doing the big load. Whenever you feel like it, I can start tinkering with this branch :)

docs/_include/prerequisites.inc
stevepiercy: Its name is title case, whereas the command is a lowercase inline literal. https://www.gnu.org/software/make/

docs/contribute/development.rst
stevepiercy: We use proper name casing of the program in narrative text. These are all correct.

docs/contribute/development.rst
stevepiercy: Done in 993c8def78a204e31e631afa01c5a832a19547d4

docs/_include/prerequisites.inc
stevepiercy: Good idea. Done in c806654a493253cc5f8ff95b71191743d229ca38.

niccokunzmann:
The test environment should not need the sphinx-icalendar package.

stevepiercy:
Can we move icalendar as a hard dependency to a dev dependency of sphinx-icalendar?

https://github.com/pycalendar/sphinx-icalendar/blob/main/pyproject.toml#L30

stevepiercy:
@niccokunzmann I give up for now. Maybe we remove tox entirely. I was reluctant to go there, but we could do the same thing with uv.

niccokunzmann:
I guess it might be enough to run uv with all the python versions in the CI. Most contributors do not need more than 1 version. I have pyenv and will find a way to make the test cases work on my machine. E.g. I can just develop based on python 3.10.
So if you want to remove tox, that is fine with me. Let's try out something else that works for you, too.

niccokunzmann:
In sphinx-icalendar, we can remove the >= of icalendar. That might help. The dependency is still important. If I use that in another project, icalendar is required.

You can experiment with a branch in sphinx-icalendar if you like, see what works.

stevepiercy:
The problem is the circular dependency where `icalendar:dev` and `sphinx-icalendar` depend on each another.

```console
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version >= '3.12'):
  ╰─▶ Because only sphinx-icalendar<=0.1.0 is available and
      sphinx-icalendar==0.1.0 depends on your project, we can conclude that
      sphinx-icalendar>=0.1.0 depends on your project.
      And because icalendar:dev depends on sphinx-icalendar>=0.1.0 and your
      project requires icalendar:dev, we can conclude that your project's
      requirements are unsatisfiable.

      hint: The package `sphinx-icalendar` depends on the package `icalendar`
      but the name is shadowed by your project. Consider changing the name of
      the project.

      hint: While the active Python version is 3.10, the resolution failed for
      other Python versions supported by your project. Consider limiting your
      project's supported Python versions using `requires-python`.
```

I tried to find where `icalendar:dev` comes from by toggling bits off, but that was not successful.

SashankBhamidi:
I'd like to review this once it's ready, please wait for my approval before mering, thanks!

stevepiercy:
This is frustrating. According to https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups:

> By default, uv includes the `dev` dependency group in the environment (e.g., during `uv run` or `uv sync`).
> ...
> To disable this behaviour during `uv run` or `uv sync`, use `--no-default-groups`. To exclude a specific default group, use `--no-group <name>`.

Neither of these had an affect on `icalendar:dev`. Where the heck is it coming from, if not uv? It's definitely in the `Install dependencies` step because that's where it fails.

stevepiercy:
@niccokunzmann @SashankBhamidi the circular dependency between icalendar and sphinx-icalendar is the blocker.

We can either:

1. Abandon the work to move to uv.
2. Omit sphinx-icalendar as a dependency from the docs dependency group in icalendar.
3. Move icalendar from a hard to a soft dependency in sphinx-icalendar, such as in its test or docs dependency group.
4. Something else unknown.

I prefer option 3 because it removes the circular dependency. I dislike option 2 because it makes it harder to run tests and build docs successfully as we'll need to specifically add it everywhere we build docs. There are far more users of icalendar than sphinx-icalendar.

What do you all think?

niccokunzmann:
I will give it a go and try out these options:

- remove version pinning in sphinx-icalendar
- extra install step in docs for icalendar
- extra install step in docs for sphinx-icalendar
- remove dependency icalendar, document install as an extra and give proper error message when importing icalendar in sphinx-icalendar
- I had another idea. Forgot...

pyproject.toml
```diff
@@ -125,7 +127,7 @@ docs = [
     "sphinx-autodoc-typehints",
     "sphinx-copybutton>=0.5.2",
     "sphinx-design>=0.5.0",
-    "sphinx-icalendar>=0.1.0",
+#    "sphinx-icalendar",
```
niccokunzmann:
```suggestion
    "sphinx-icalendar",
```

I fixed this now by removing the version restriction of icalendar in sphinx-icalendar>=0.1.1.

docs/contribute/development.rst
```diff
@@ -105,35 +86,34 @@ The screenshot below shows how to configure PyCharm to disable pre-commit by sea
...
 
-You might not have all Python versions installed or you may want to run a specific one.
-The following command show how to run tox with Python 3.12:
+The following command shows how to run tests in the Python 3.14.
 
-.. code-block:: shell
+..  code-block:: shell
 
-    tox -e py312
+    make test TOX_ENV=py314
```
niccokunzmann, line 101:
```
$ make test TOX_ENV=py314
  File "/home/nicco/.pyenv/versions/3.11.6/bin/tox", line 6, in <module>
    sys.exit(run())
             ^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/run.py", line 23, in run
    result = main(sys.argv[1:] if args is None else args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/run.py", line 49, in main
    return handler(state)
           ^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/session/cmd/legacy.py", line 115, in legacy
    return run_sequential(state)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/session/cmd/run/sequential.py", line 25, in run_sequential
    return execute(state, max_workers=1, has_spinner=False, live=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/session/cmd/run/common.py", line 169, in execute
    state.envs.ensure_only_run_env_is_active()
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/session/env_select.py", line 437, in ensure_only_run_env_is_active
    envs, active = self._defined_envs, self._env_name_to_active()
                   ^^^^^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/session/env_select.py", line 259, in _defined_envs
    run_env = self._build_run_env(name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/session/env_select.py", line 320, in _build_run_env
    runner = REGISTER.runner(cast("str", env_conf["runner"]))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nicco/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tox/tox_env/register.py", line 74, in runner
    return self._run_envs[name]
           ~~~~~~~~~~~~~~^^^^^^
KeyError: 'uv-venv-lock-runner'
make: *** [Makefile:138: test] Fehler 1

```

This is the same for `make test`.

niccokunzmann (changes requested):
Good news - sphinx-icalendar can now be installed.
I cannot run the tests however...

docs/contribute/development.rst
niccokunzmann: It appears that I run tox not from the uv source...
```
/home/nicco/.pyenv/versions/3.11.6/bin/tox
```

Makefile
```diff
@@ -128,9 +130,12 @@ doctest: .venv  ## Test snippets and docstrings in the documentation
...
 	@pytest src/icalendar/tests/test_with_doctest.py
 
+.PHONY: docs-all
+docs-all: .venv clean vale doctest html linkcheckbroken  ## Clean docs build, then run vale, doctest, html, and linkcheckbroken
+
 .PHONY: test
-#test: clean vale linkcheckbroken doctest  ## Clean docs build, then run vale and linkcheckbroken
-test: clean linkcheckbroken  ## Clean docs build, then run vale and linkcheckbroken
+test:  ## Run code tests and coverage
+	@tox
```
niccokunzmann:
```suggestion
test: .venv ## Run code tests and coverage
	@uv run tox
```

This works for me.
Making me think about pre-commit above.

```
$ uv run pre-commit install
pre-commit installed at .git/hooks/pre-commit
```

That works for me, too. You might use your global pre-commit.

pyproject.toml
stevepiercy: Done in 30c117e7a271c398a511ae0b70de627e33a8db4b. Let's see if docs build now.

stevepiercy: Well, docs may build, and that took care of one circular dependency, but now we have a new circular dependency issue for `recurring-ical-events<=3.8.1`. See https://github.com/collective/icalendar/actions/runs/24370219211/job/71171939218?pr=1315. Can you handle that one as well?

docs/contribute/development.rst
stevepiercy: We need to decide which tox we want to use. If we go all in with uv, it makes sense to use tox from the project virtual environment, not the user's Python or any other Python not managed by uv.

Makefile
stevepiercy: See what I did in 68082cffd1050c8b5936955d21f9d8b53844c9ce to address both comments.

docs/contribute/development.rst
stevepiercy: Actually, after reading your next comment and checking the uv docs, I think I resolved this in 68082cffd1050c8b5936955d21f9d8b53844c9ce.

niccokunzmann: I think running all commands from the project venv is good. That reduces miunderstandings for new setups. Uv all the way!
BTW if I can easily run a certain Python version or command from the CI tests with uv, I don't need tox. It is also ok to keep toxbutif it is simpler without, we can remove it. I should check that.

stevepiercy: Yes!

```console
$ make clean-python  # Otherwise it uses the existing Python.
$ make dev PYTHONVERSION=3.12
Installed Python 3.12.9 in 75ms
 + cpython-3.12.9-macos-aarch64-none (python3.12)
Using CPython 3.12.9
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
...
pre-commit installed at .git/hooks/pre-commit
```

I did not want to remove tox, as I know that's a bigger ask of you, but it could be replaced by uv for this purpose. That's a little sad, as I used to love tox and pyenv, but nowawadays uv makes them almost obsolete. There may be cases where tox and pyenv are still useful, but I haven't found any yet.

.github/workflows/tests.yml
stevepiercy: I went with:

```yaml
    - name: Install dependencies
      if: ${{ !matrix.skip }}
      run: |
        uv sync --no-dev --group test
        uv tool install tox --with tox-uv
```

stevepiercy: Actually, this was part of a workaround to avoid the circular import with sphinx-icalendar. Updated in 87ba4cccdd48ecdd8c783bcd4cbe18a6176fa54c. Let's see what that does for CI.

stevepiercy:
@niccokunzmann remaining issue is the circular import for `recurring-ical-events`. See https://github.com/collective/icalendar/pull/1315#discussion_r3076184953.

niccokunzmann:
I am working on the tests and the circular import.

- [ ] adjust test
- [ ] adjust test matrix
- [ ] remove tox

niccokunzmann:
Problem: I cannot reproduce this error on my machine.

stevepiercy:
@niccokunzmann I created a PR that might handle the issue in https://github.com/pycalendar/sphinx-icalendar/pull/6 that might address the circular import.

niccokunzmann:
The tests are running. We need to open an issue that the pytz tests are not working.
Or, we can just remove the nopytz tests fully and remove the tox dependency.

.github/workflows/tests.yml
```diff
@@ -41,31 +41,28 @@ jobs:
...
-          setup.*
-          tox.ini
-          requirements*.txt
     - name: Install dependencies
       if: ${{ !matrix.skip }}
       run: |
-        python -m pip install --upgrade pip
-        pip install tox coverage-python-version
+        make dev PYTHONVERSION=${{ matrix.python_version }}
+        uv run pip list
```
stevepiercy: Is this a debug statement that should be removed?

pyproject.toml
```diff
@@ -300,10 +301,17 @@ filterwarnings = [
     "ignore::DeprecationWarning",
 ]
 
+[tool.uv]
+python-preference = "managed"
+
+#[tool.uv.build-backend]
+#namespace = true
+#module-name = "icalendar"
+#
```
stevepiercy: This should be removed, since we're not using it as a build backend.
```suggestion
```

.github/workflows/test_generate_matrix.py
```diff
@@ -42,26 +41,87 @@
...
-    all_in_matrix = {case["test_name"] for case in result["include"]}
+
+
+def test_running_is_not_skipped(running_names, skipped_names):
+    """Running means not skipped."""
+    assert running_names.isdisjoint(skipped_names)
+
+
+def test_all_cases_are_always_included(running_names, skipped_names, matrix):
+    """The matirx always needs to be complete or we might wait for required checks."""
```
stevepiercy:
```suggestion
    """The matrix always needs to be complete or we might wait for required checks."""
```

.github/workflows/generate_matrix.py
```diff
@@ -30,6 +30,9 @@
 # Update this when a new Python minor version is released
 PYTHON_MINOR_VERSION_MAX = 14
 
+COMMAND_TEST = "uv run tox -e py"
+COMMAND_INSTALL_REMOVE_PYTZ = "uv run pip uninstall -y pytz"
```
stevepiercy: Why not this?

https://docs.astral.sh/uv/reference/cli/#uv-remove

```suggestion
COMMAND_INSTALL_REMOVE_PYTZ = "uv remove pytz"
```

stevepiercy (commented):
Please see my suggestions and questions.

stevepiercy:
> The tests are running. We need to open an issue that the pytz tests are not working. Or, we can just remove the nopytz tests fully and remove the tox dependency.

Sorry, I don't understand which would be better.

Do we need nopytz tests? Removing tests would not remove the feature, but would remove testing the absence of the feature. Additionally, the feature has been deprecated since 6.0.0, although we said it "may still be used" per https://icalendar.readthedocs.io/en/stable/how-to/upgrade.html#pytz-support.

I think we should announce that pytz support will be removed in 8.0.0, since that would be a change in the promise. I think that we should make an announcement in the 7.0.4 release that pytz support, deprecated since 6.0.0, shall be removed in 8.0.0. That would give fair warning.

Does that help?

.github/workflows/tests.yml
niccokunzmann: It is a debug statement. It is nice to know which versions we are actually running because we do not pin them. So, if tests fail with no significant change, we can check versions. So, it is handy.

docs/contribute/development.rst
niccokunzmann: Let's not remove tox in this PR. We can open an issue to evaluate this. It would blur the focus.

niccokunzmann (dismissed):
This is good to go.
@stevepiercy have a final look if you like and merge it if you are satisfied!

niccokunzmann:
> Does that help?

Yes, that helps. There is the pytz issue and it is good if we drop pytz. It is outdated now. #1101 - can you update that on how you would like this to go or put in a PR?

.github/workflows/tests.yml
stevepiercy: I like [`uv tree`](https://docs.astral.sh/uv/reference/cli/#uv-tree) for debugging to show where each dependency comes from. That especially helps with circular imports.

`uv run pip list` will run `pip list`, which is a table of the dependencies at only the root level, and no deeper in the dependency tree.

[`uv pip list`](https://docs.astral.sh/uv/reference/cli/#uv-pip-list) will provide a complete list of dependencies in a tabular format.

There's also [`uv pip tree`](https://docs.astral.sh/uv/reference/cli/#uv-pip-tree) which gives a slightly different tree format from `uv tree` and appears not to resolve as well as `uv tree`.

```suggestion
        uv tree
```

Isn't it great how uv gives us so many options? Or maybe they already exist, but their documentation is so good, that it's easy to discover new things.

stevepiercy:
> > Does that help?
> 
> Yes, that helps. There is the pytz issue and it is good if we drop pytz. It is outdated now. #1101 - can you update that on how you would like this to go or put in a PR?

Updated!

https://github.com/collective/icalendar/issues/1101

Please feel free to edit the description.

docs/contribute/development.rst
stevepiercy: Agreed. Removing tox is opening a can of worms, unless you like the taste of worms.

tox.ini
stevepiercy: Let's defer to the removal of tox.

stevepiercy:
@SashankBhamidi would you please do a review, per your request? I think this is ready to merge.

We decided to defer the replacement of tox with uv for another PR. This PR moves us forward, and makes it a lot easier to get started for contributors.

.github/workflows/tests.yml
```diff
@@ -41,31 +41,28 @@ jobs:
...
-          tox.ini
-          requirements*.txt
     - name: Install dependencies
       if: ${{ !matrix.skip }}
       run: |
-        python -m pip install --upgrade pip
-        pip install tox coverage-python-version
+        make dev PYTHONVERSION=${{ matrix.python_version }}
+        uv tree
+        ${{ matrix.install_command }}
```
SashankBhamidi: `${{ matrix.install_command }}` is never set in `generate_matrix.py`, is this scaffolding for a follow-up? If not, why not remove it?

```suggestion
```

Makefile
```diff
@@ -128,21 +131,23 @@ doctest: .venv  ## Test snippets and docstrings in the documentation
...
 	@pytest src/icalendar/tests/test_with_doctest.py
 
+.PHONY: docs-all
+docs-all: .venv clean vale doctest html linkcheckbroken  ## Clean docs build, then run vale, doctest, html, and linkcheckbroken
+
 .PHONY: test
-#test: clean vale linkcheckbroken doctest  ## Clean docs build, then run vale and linkcheckbroken
-test: clean linkcheckbroken  ## Clean docs build, then run vale and linkcheckbroken
+test: .venv  ## Run code tests and coverage
+	@uv run tox
```
SashankBhamidi, line 139: `TOX_ENV` is declared at line 8 but the `test` target doesn't use it. `make test TOX_ENV=py314` runs all envs. Why not pass it through?

```suggestion
test: .venv  ## Run code tests and coverage
	@uv run tox $(if $(TOX_ENV),-e $(TOX_ENV))
```

docs/contribute/development.rst
```diff
@@ -105,35 +86,34 @@ The screenshot below shows how to configure PyCharm to disable pre-commit by sea
...
 
-.. code-block:: shell
+..  code-block:: shell
 
-    tox
+    make test
 
-You might not have all Python versions installed or you may want to run a specific one.
-The following command show how to run tox with Python 3.12:
+The following command shows how to run tests in the Python 3.14.
```
SashankBhamidi:
```suggestion
The following command shows how to run tests in Python 3.14.
```

CHANGES.rst
SashankBhamidi, line 1: Why no entry for #1102?

tox.ini
```diff
@@ -33,23 +31,17 @@ commands =
     coverage xml
```
SashankBhamidi, line 27: Umm... the main `[testenv]` uses `runner = uv-venv-lock-runner` but `nopytz` still uses old-style `deps`. 

I this Intentional?

SashankBhamidi (changes requested):
Thanks! Few comments:

.github/workflows/tests.yml
stevepiercy: Good catch!

CHANGES.rst
stevepiercy: Whoopsie. Added in 6407c9eda51b2018d089118401c6e44bc2cba572. Thank you!

Makefile
stevepiercy: Did you try this locally?

Compare to my results.

```console
$ make test TOX_ENV=py314
      Built icalendar @ file:///Users/stevepiercy/projects/icalendar
Uninstalled 1 package in 1ms
Installed 1 package in 1ms
py314: uv-sync> .venv/bin/uv sync --locked --python-preference system --no-default-groups --group test -p cpython3.14
...
  py314: OK (32.73=setup[0.59]+cmd[27.80,0.82,2.51,1.00] seconds)
  congratulations :) (32.77 seconds)
```

Similarly...

```console
$ make test TOX_ENV=py310
py310: uv-sync> .venv/bin/uv sync --locked --python-preference system --no-default-groups --group test -p cpython3.10
...
  py310: OK (80.77=setup[0.03]+cmd[75.04,1.07,3.42,1.21] seconds)
  congratulations :) (80.79 seconds)
```

No `TOX_ENV` runs all the tox environments.

```console
$ make test
...
  py310: OK (77.97=setup[0.03]+cmd[74.19,1.09,1.43,1.24] seconds)
  py311: OK (75.60=setup[0.03]+cmd[71.01,0.96,2.47,1.13] seconds)
  py312: OK (87.67=setup[1.82]+cmd[81.99,0.88,1.90,1.09] seconds)
  py313: OK (73.17=setup[1.71]+cmd[68.03,0.84,1.57,1.02] seconds)
  py314: OK (32.23=setup[0.03]+cmd[27.13,0.83,2.51,1.74] seconds)
  pypy3: OK (1018.01=setup[2.58]+cmd[1006.29,2.17,4.67,2.30] seconds)
  docs: OK (107.27=setup[3.25]+cmd[104.01] seconds)
  nopytz: OK (82.96=setup[0.55]+cmd[76.74,1.05,3.41,1.21] seconds)
  congratulations :) (1554.90 seconds)
```

tox.ini
stevepiercy: It was ignored, due to future removal of tox, but I went ahead and removed the `deps` here in e85f7e2be217b76be40de6418b521f092f07019a. Good catch, and thank you!

stevepiercy:
@SashankBhamidi thanks for the review! I double-checked `make test TOX_ENV=py310` as well as for py314 and no `TOX_ENV`, and it works as I expect, but if it doesn't work for you, then please open an issue.

I'm going ahead and merge this monster. Thanks everyone!

SashankBhamidi:
@stevepiercy You're right, apologies.

 I didn't actually test the command locally. Thanks for the patience.

SashankBhamidi (commented):
Perfect.

PR #1316, opened 2026-04-11
Improved docstring for string_parameter (Google Style)

Aayan0207:
## Description
Improved the docstring for `string_parameter` to follow Google style.

## Changes
- Rewrote docstring using Args and Returns format

Contributes to #1072

----
📚 Documentation preview 📚: https://icalendar--1316.org.readthedocs.build/


src/icalendar/param.py
```diff
@@ -45,7 +45,22 @@ def string_parameter(
     convert: Callable[[str], T] | None = None,
     convert_to: Callable[[T], str] | None = None,
 ) -> property:
-    """Return a parameter with a quoted value (case sensitive)."""
+    """Create a property for a string parameter with optional conversion.
+
+    Args:
```
niccokunzmann: Thanks for contributing!
We use this style guide here to comply:
https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure

Args-> Parameters

Please also don't put the types in the docstring. You can check if they appear in the documentation build of readthedocs, in the API section.

Aayan0207: Updated the docstring to follow the style guide (Parameters + removed types). Let me know if it's correct now.

niccokunzmann:
Thanks for the update. I hope you read through our documentation on the doc strings and I will pass it over to @stevepiercy, our documentation lead.

src/icalendar/param.py
```diff
@@ -45,7 +45,21 @@ def string_parameter(
...
-    """Return a parameter with a quoted value (case sensitive)."""
+    """Create a property for a string parameter with optional conversion.
+
+    Parameters:
+        name: Name of the parameter in the params dictionary.
+        doc: Documentation for the property.
+        default:
+            Function that returns a default value if the parameter is not found.
+        convert:
+            Function that converts stored string value to desired type for usage.
```
stevepiercy:
```suggestion
            Function that converts the stored string value to the desired type for usage.
```

src/icalendar/param.py
```diff
@@ -45,7 +45,21 @@ def string_parameter(
...
+        convert_to:
+            Function to convert value back to string for storage.
```
stevepiercy:
```suggestion
            Function to convert a value back to a string for storage.
```

CHANGES.rst
```diff
@@ -48,6 +48,7 @@ Bug fixes
 Documentation
 ~~~~~~~~~~~~~
 
+- Convert docstring of ``string_parameter`` to Google style. :issue:`1072`, :pr:`1316`
```
stevepiercy: Let's make it link to the function.

```suggestion
- Convert docstring of :func:`~icalendar.param.string_parameter` to Google style. :issue:`1072`, :pr:`1316`
```

src/icalendar/param.py
```diff
@@ -45,7 +45,21 @@ def string_parameter(
     convert: Callable[[str], T] | None = None,
     convert_to: Callable[[T], str] | None = None,
 ) -> property:
-    """Return a parameter with a quoted value (case sensitive)."""
+    """Create a property for a string parameter with optional conversion.
+
```
stevepiercy, line 49: Would you like to insert a description here with more details, such as why would someone use this function?

src/icalendar/param.py
```diff
@@ -45,7 +45,21 @@ def string_parameter(
...
+        default:
+            Function that returns a default value if the parameter is not found.
+        convert:
+            Function that converts stored string value to desired type for usage.
+        convert_to:
+            Function to convert value back to string for storage.
+
+    Returns:
+        A property object with a getter, setter, and deleter for the parameter.
+    """
```
stevepiercy, line 98: Would you like to add an Example section?

stevepiercy (commented):
Thank you! This is good.

I made suggestions to fix grammar and linkify the function in the change log.

Also it would be nice to have a description and usage examples, if you would like to take that on.

src/icalendar/param.py
Aayan0207: Yes, I've added it

src/icalendar/param.py
Aayan0207: I tried to add the following section:

Example:
  Define a parameter that is stored as a string but used as an integer:
  
      priority = string_parameter(
          "PRIORITY",
          "Priority of the component",
          default=lambda: 0,
          convert=int,
          convert_to=str,
      )
  
      Accessing the property converts the stored value:
  
      obj.params["PRIORITY"] = "5"
      obj.priority  # 5
  
      Setting the property stores it as a string:
  
      obj.priority = 10
      obj.params["PRIORITY"]  # "10"

However, the automated "Build documentation, check links, spelling, grammar, and style / Documentation (3.11)" and "docs/readthedocs.org:icalendar" checks continued to fail despite multiple attempts to adjust the formatting and repeated commits.

I've removed it in the latest commit.

stevepiercy: I found the error message to be helpful to correct the issue.

https://github.com/collective/icalendar/actions/runs/24403647836/job/71280800920#step:6:135

First, please use `pycon` as the lexer, instead of `python`. This allows code examples to be tested, both locally and in CI, and users can copy paste the console code into a Python interpreter to run it.

Your example looked good to me, but I couldn't find the exact issue causing the failure. Would you please restore and adapt your example to `pycon` including the `>>> ` prompts? Then I can give further guidance.

As an aside, I think writing examples in the API reference via docstrings is a fun way to learn how to use icalendar. It's bonus fun to know that you're sharing that knowledge with future readers of documentation. Pretty soon, you'll be an expert and can charge your clients accordingly! 🤑

Aayan0207: I have restored and updated the example to use pycon style with a self-contained runnable example. Let me know if any further adjustments are needed.

Aayan0207: All checks seem to be passing now.

src/icalendar/param.py
```diff
@@ -45,7 +45,56 @@ def string_parameter(
...
+        >>> Dummy.priority = string_parameter(
+        ...     "PRIORITY",
+        ...     "Priority of the component",
+        ...     default=lambda: 0,
+        ...     convert=int,
+        ...     convert_to=str,
+        ... )
+        >>> obj = Dummy()
+
+        Accessing the property converts the stored value:
```
stevepiercy:
```suggestion
        Accessing the property converts the stored string value to the type specified by the ``convert`` parameter, in this case, an ``int``.
```

src/icalendar/param.py
```diff
@@ -45,7 +45,56 @@ def string_parameter(
...
+
+        >>> obj.params["PRIORITY"] = "5"
```
stevepiercy: Let's use the icalendar API for property access, instead of the native Python dict, per https://icalendar.readthedocs.io/en/latest/explanation/api-design.html#property-access. This encourages best practice and is more convenient for developers.

```suggestion
        >>> obj.priority = 5
```

src/icalendar/param.py
```diff
@@ -45,7 +45,56 @@ def string_parameter(
...
+        >>> obj.priority
+        5
+
+        Setting the property stores it and keeps access consistent:
```
stevepiercy:
```suggestion
        Setting the property stores the new value.
```

src/icalendar/param.py
stevepiercy: Thanks for sticking with it!

stevepiercy (changes requested):
Excellent work! I have a few suggestions. What do you think?

stevepiercy:
@niccokunzmann would you please take a look at the Examples section, and my suggestions?

src/icalendar/param.py
niccokunzmann: The function is about converting  between the params (str) and the desired property value and type (int) in the example. So, it is good how it is done with the self-contained example. I think, this line should stay to illustrate the use-case.

niccokunzmann (dismissed):
Thanks for the documentation! This depicts well what the function does. Could you add Steve's changes?

Aayan0207:
Hi, I've merged the changes suggested by @stevepiercy and ensured all the automated checks are successful. Let me know if any other changes are required. Thanks.

src/icalendar/param.py
stevepiercy: That makes sense. I'll make a revision.

src/icalendar/param.py
```diff
@@ -45,7 +45,57 @@ def string_parameter(
...
+        ...     default=lambda: 0,
+        ...     convert=int,
+        ...     convert_to=str,
+        ... )
+        >>> obj = Dummy()
+
+        Accessing the property converts the stored string value to the type
+        specified by the ``convert`` parameter, in this case, an ``int``.
+
+        >>> obj.priority = 5
```
stevepiercy: Per @niccokunzmann's [comment](https://github.com/collective/icalendar/pull/1316#discussion_r3098263520), this illustrates the conversion of a `str` to an `int`, while using the icalendar API instead of a Python dict.

```suggestion
        >>> obj.priority = "5"
```

stevepiercy (changes requested):

src/icalendar/param.py
Aayan0207: I've merged the commit. Let me know if anything else is required. Thanks.

stevepiercy (approved):
Thank you! This is awesome. Thank you for your patience to work through this together.

Aayan0207:
Thank you for the opportunity! Looking forward to more in the future!

niccokunzmann:
This is included in v7.1.0.

Issue #1226, opened 2026-02-24
use to_unicode in escape_char


niccokunzmann:
escape_char in icalendar.parser should also accept bytes.

- Please follow the development setup: https://icalendar.readthedocs.io/en/stable/contribute/development.html
- [ ] Please use to_unicode before doing the replacements
- [ ] add a test that checks that bytes work

> `_escape_char` is annotated to accept `bytes`, but the current replacement chain uses `str` patterns/replacements, which will raise `TypeError` if `text` is actually `bytes`. Either add a `bytes` branch (similar to `_unescape_char`) or restrict the signature/validation to `str` only so the type hints match runtime behavior. 

 _Originally posted by @Copilot in [#1222](https://github.com/collective/icalendar/pull/1222/changes#r2844480507)_

kokowa194:
Hi @niccokunzmann 

Hi, I see @mackarovak was planning to work on this. If you're still working on it, no rush at all! But if you don't mind, I'd love to give this a try.

niccokunzmann:
Hm. Maybe it got solved but the issue is still open. Please have a look. If it still needs doing, please give it a go. There was enough time to work on it. Let me know if we can close it, please.

adamressom:
Hi! I looked into this issue and the linked PR (#1227). It seems like a solution was started but the issue is still open. Is there still work needed here, or should this be closed? I’d be happy to help finish or test the fix if needed.

tolunigo:
Hi, Will I be able to solve this issue?

stevepiercy:
@tolunigo welcome to icalendar! You don't need to ask to start work on an open issue. However, there's already an open PR awaiting review, and its link is in this issue. It's best not to duplicate effort.

niccokunzmann:
This is included in v7.1.0.

PR #1325, opened 2026-04-15
fix(parser): use to_unicode in _escape_char to support bytes input

pureqin:
## Summary\n\nCloses #1226\n\nThe `_escape_char` function was annotated to accept `bytes` but only used `str` replacement patterns, which would raise `TypeError` when actually passed bytes.\n\n## Changes\n\n- **Modified**: `src/icalendar/parser/string.py` — `_escape_char()`\n  - Added `to_unicode()` import\n  - Convert input to str before applying escape replacements\n  - Updated return type from `str | bytes` to `str`\n\n- **Modified**: `src/icalendar/tests/test_parsing.py`\n  - Added `test_escape_char()` with both str and bytes test cases\n  - Added `_escape_char` to imports\n\n## Test Plan\n\n- All new tests pass\n- Lint checks pass (ruff)\n- No breaking changes — existing behavior for str input is preserved\n\n---\n*This PR was generated by an autonomous AI agent ('Han - Champion Hou') focused on fixing open-source bugs.*

### DIFF
--- a/src/icalendar/parser/string.py
+++ b/src/icalendar/parser/string.py
@@ -3,10 +3,10 @@
 import re
 import warnings
 
-from icalendar.parser_tools import DEFAULT_ENCODING
+from icalendar.parser_tools import DEFAULT_ENCODING, to_unicode
 
 
-def _escape_char(text: str | bytes) -> str | bytes:
+def _escape_char(text: str | bytes) -> str:
     r"""Format value according to iCalendar TEXT escaping rules.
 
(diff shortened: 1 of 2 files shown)


niccokunzmann:
We have no interest in AI bots trying to help. Thank you. Please do not come back.

stevepiercy:
I reported the user to GitHub and permanently blocked them from the collective GitHub organization.

PR #1326, opened 2026-04-15
fix: use to_unicode in _escape_char to handle bytes input

### LINKED ISSUE #1226: use to_unicode in escape_char

Ai-chan-0411:
## Summary

Fixes #1226

`_escape_char` was annotated to accept `str | bytes` but used str-only replacement patterns (`r"\N"`, `"\\"`, etc.), which raises `TypeError` when bytes are passed.

**Fix:** Call `to_unicode(text)` at the start of `_escape_char` to decode bytes to str before applying the replacement chain. This is the approach requested in the issue.

**Changes:**
- Import `to_unicode` from `icalendar.parser_tools` in `parser/string.py`
- Update `_escape_char` to call `to_unicode(text)` before replacements
- Update return type annotation from `str | bytes` to `str` (bytes are always decoded to str)
- Update `escape_char` public wrapper return type annotation accordingly
- Add `test_escape_char_bytes` test to verify bytes input works correctly

## Test

```python
assert _escape_char(b"hello") == "hello"
assert _escape_char(b"hello,world") == r"hello\,world"
assert _escape_char(b"line\nbreak") == r"line\nbreak"
```

### DIFF
--- a/src/icalendar/parser/string.py
+++ b/src/icalendar/parser/string.py
@@ -3,10 +3,10 @@
 import re
 import warnings
 
-from icalendar.parser_tools import DEFAULT_ENCODING
+from icalendar.parser_tools import DEFAULT_ENCODING, to_unicode
 
 
-def _escape_char(text: str | bytes) -> str | bytes:
+def _escape_char(text: str | bytes) -> str:
     r"""Format value according to iCalendar TEXT escaping rules.
 
     Escapes special characters in text values according to :rfc:`5545#section-3.3.11`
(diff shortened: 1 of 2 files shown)


stevepiercy:
Closing as AI bot and reporting to GitHub.

PR #1327, opened 2026-04-16
Remove misleading `TOX_ENV` variable from `Makefile`

stevepiercy:
This is a follow up to the just merged #1315, per a discussion with @SashankBhamidi in Signal.

tox actually handles this black magic, not Make.

https://github.com/tox-dev/tox/blob/4dcde44c0746138421bef38f1c0dac42ee505223/src/tox/config/cli/env_var.py#L24

No change log is needed, since it's removing dead code.


SashankBhamidi (approved):
LGTM, thanks!

PR #1330, opened 2026-04-19
fix(parser): handle bytes input in escape_char using to_unicode

### LINKED ISSUE #1226: use to_unicode in escape_char

Alm0stSurely:
## Problem

`escape_char` in `icalendar.parser.string` was annotated to accept `str | bytes` but used `str.replace()` operations directly, causing `TypeError` when `bytes` input contained characters that needed escaping (comma, semicolon, backslash, newline).

This was flagged by Copilot in the review of #1222.

## Analysis

The function signature allowed `bytes` input, but the implementation used string literals for all replacements. When `bytes` was passed, Python would raise `TypeError: a bytes-like object is required, not str` (or similar, depending on the replacement).

The `unescape_char` function already handles this correctly by branching on `isinstance(text, str)` vs `isinstance(text, bytes)`. A more consistent approach is to use `to_unicode()`, which is already available in the codebase and converts `bytes` to `str` using the default encoding.

## Solution

- Use `to_unicode(text)` at the start of `escape_char` to normalize input to `str`
- Update the return type from `str | bytes` to `str`
- Add `test_escape_char_bytes` verifying all escape patterns work with bytes input

## Benchmarks

No performance impact expected — `to_unicode` is a fast encoding check/conversion that is already used extensively in the parser.

## Notes

- This aligns `escape_char` with the rest of the parser, which uses `to_unicode` for input normalization.
- The return type change from `str | bytes` to `str` is technically a narrowing, but since the function previously crashed on `bytes` input for most cases, this is a bugfix rather than a breaking change.

Closes #1226

----
📚 Documentation preview 📚: https://icalendar--1330.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,7 @@ New features
 Bug fixes
 ~~~~~~~~~
 
+- Fixed :func:`~icalendar.parser.string.escape_char` to accept :class:`bytes` input by converting it to :class:`str` with :func:`~icalendar.parser_tools.to_unicode` before applying replacements. :issue:`1226`
 - ...
 
 Documentation
--- a/src/icalendar/parser/string.py
+++ b/src/icalendar/parser/string.py
@@ -2,10 +2,10 @@
 
 import re
 
-from icalendar.parser_tools import DEFAULT_ENCODING
+from icalendar.parser_tools import DEFAULT_ENCODING, to_unicode
 
 
-def escape_char(text: str | bytes) -> str | bytes:
+def escape_char(text: str | bytes) -> str:
     r"""Format value according to iCalendar TEXT escaping rules.
 
     Escapes special characters in text values according to :rfc:`5545#section-3.3.11`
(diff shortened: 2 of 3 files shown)


stevepiercy:
@Alm0stSurely would you please resolve the merge conflicts, so we can review? Thank you!

stevepiercy:
Closing due to no response and superseded by #1332

PR #1335, opened 2026-04-23
fix(vBinary): preserve raw bytes in binary round-trips

### LINKED ISSUE #1289: [Bug] vBinary class causes image bytes to become corrupt
## Describe the bug

The vBinary class corrupts image files if loaded in.
## To reproduce

```python
from icalendar import vBinary
import base64

# 1. Create a small "image" payload (pretend PNG header + fake data)
original_bytes = (
    b"\x89PNG\r\n\x1a\n"  # PNG signature
    b"\x00\x00\x00\rIHDR"  # fake chunk
    b"\x00\x00\x00\x01\x00\x00\x00\x01"
)

# 2. Wrap it into vBinary
vb = vBinary(original_bytes)

# 3. Serialize to iCal base64 form
ical_encoded = vb.to_ical()

# 4. Decode back using the library method
decoded_bytes = vBinary.from_ical(ical_encoded)

# 5. Save both to files for comparison
print("Original bytes:", original_bytes)
print("Decoded bytes:", decoded_bytes)

print("Match:", original_bytes == decoded_bytes)
### ...
```

Output:

```pycon
Original bytes: b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01'
Decoded bytes: b'\xef\xbf\xbdPNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01'
Match: False
```

## Expected behavior

Original bytes: b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01'
Decoded bytes: b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01'
Match: True
## Environment

 - [ ] Operating system:     Windows 11    
 - [ ] Python version:    3.13.1       
 - [ ] `icalendar` version:   7.0.4.dev127   

## Additional context

- [ ] I tested it with the latest version. For example, I used the command `pip3 install --upgrade https://github.com/collective/icalendar.git`.
- [ ] I attached the `.ics` sample file, or there is no `.ics` sample file.

baodq06:
## Closes issue

Replace `ISSUE_NUMBER` with the issue number that your pull request fixes. Then GitHub will link and automatically close the related issue.

- [x] Closes #1289

## Description

After looking more into the issue at hand, I saw that vBinary was converting the input through Unicode before doing a base64 encoding; I think that that would work for text-like inputs such as plain ASCII strings or byte sequences that could be cleanly decoded as text, but it corrupted the arbitrary binary data, which included the PNG-style bytes from the issue report. The PR I am working on right now fixes that by just keeping the internal vBinary value as raw bytes and do the base64 encoding directly in to_ical(). 

It also updated the related tests to make sure that the expected behavior matches the corrected raw-bytes representation. Please look over and let me know if there are any changes I should make. 

Changes that I made in this PR includes:
- storing vBinary internal values as raw bytes instead of storing them as Unicode text
- base64 encoding directly in to_ical()
- updating ical_value/jcal behavior to match the corrected internal representation
- updated/added regression tests for non-UTF-8 byte round-trips
- made sure the jCal test expectations were updated to use raw bytes internally

## Checklist

## Additional information

Local QA performed included:
- I made sure to reproduce the bug locally before implementing the solution/fix using minimal PNG-like byte sequence
- also verified that the repro passes after I fixed it
- ran the targeted tests for vBinary and jCal conversion
- ran the broader local test suite (14475 passes, 28 skipped)

----
📚 Documentation preview 📚: https://icalendar--1335.org.readthedocs.build/


stevepiercy:
@baodq06 please resolve the merge conflicts, so we can review this PR.

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -48,9 +46,20 @@ def test_from_ical():
...
+
+def test_round_trip_preserves_non_utf8_bytes():
+    original_bytes = (
+        b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01"
+    )
+
+    vb = vBinary(original_bytes)
+    decoded_bytes = vBinary.from_ical(vb.to_ical())
+
+    assert decoded_bytes == original_bytes
```
niccokunzmann, line 62: Thanks for adding more tests.
I think, this is a difficult one to approach.

We should add example calendar files with binary content, too.
See https://github.com/collective/icalendar/issues/1299

The problem is that we need to somehow make sure that people can use it properly and also not break too much backward compatibility. I would value your view on that, too.

Why did you choose this way, based on your perspective of the problem.

Thanks for taking this on!

src/icalendar/tests/rfc_7265_jcal/test_section_3_6_values.py
```diff
@@ -30,7 +30,7 @@
 )
 
 JCAL_PAIRS = [
-    (["attach", {}, "binary", "SGVsbG8gV29ybGQh"], vBinary("SGVsbG8gV29ybGQh")),
+    (["attach", {}, "binary", "SGVsbG8gV29ybGQh"], vBinary(b"Hello World!")),
```
niccokunzmann, line 33: Yeah, this also requires that we get a jcal file for the tests so that we can be sure that this is actually working when parsed.

As well as to_jcal().

niccokunzmann:
Hi, we created a new release, please move your edits in CHANGES.rst to the new section.

stevepiercy:
@baodq06 would you please respond to @niccokunzmann's comments? Do you need any assistance to proceed? Please let us know. Thank you!

stevepiercy:
@baodq06 thanks for starting work on this PR. However there are now merge conflicts and no response to our requests, so I'm closing this PR as abandoned. There's also PR #1356 that addresses our requests and which supersedes this one.

You're welcome to pick up any open issue and contribute to improve icalendar. Thank you!

PR #1338, opened 2026-04-26
Improve docstring for categories_property per #1244

stevepiercy:
This provides a rich example of how to improve docstrings as mentioned in #1244.

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1338.org.readthedocs.build/

### DIFF
--- a/src/icalendar/attr.py
+++ b/src/icalendar/attr.py
@@ -616,27 +616,36 @@ categories_property = property(
     _del_categories,
     """This property defines the categories for a component.
 
-Property Parameters:
-    IANA, non-standard, and language property parameters can be specified on this
-    property.
+The categories property is used to specify categories or subtypes of the
+calendar component. The categories are useful to search for a calendar
+component of a particular type and category.
 
-Conformance:
-    The property can be specified within "VEVENT", "VTODO", or "VJOURNAL" calendar
-    components.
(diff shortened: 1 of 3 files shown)


src/icalendar/attr.py
```diff
@@ -616,27 +616,36 @@ def _del_categories(component: Component) -> None:
...
+
+The categories property for ``Event``, ``Journal``, and ``Todo`` complies
+with :rfc:`5545#section-3.8.1.2`, and for ``Calendar`` with :rfc:`7986#section-5.6`.
+
+Note:
+    At present, icalendar doesn't take the LANGUAGE parameter as defined
+    in :rfc:`5545#section-3.2.10` into account.
+
+Parameters:
+    categories(list[str]): A list of categories as strings.
```
niccokunzmann, line 641:
```suggestion
    categories: A list of categories as strings.
```

Is the typing inserted automtically?

niccokunzmann (approved):
Thanks, I have a minor suggestion. Apply or not - as you like. And merge it.
Nice!

src/icalendar/attr.py
stevepiercy: No. All four classes that have this property lack type hints. And I don't know whether it's possible to add type hints to a class (I'm pretty sure it's not possible), and if so whether a property can inherit type hints from its parent class.

PR #1341, opened 2026-04-28
Issue 884

### LINKED ISSUE #884: Probable bug is DURATION getter
https://github.com/collective/icalendar/blob/2f6ce64975d4c7500b30c58b15f222b0cf51cc74/src/icalendar/attr.py#L956-L960

While reading the source code I think I found a little error.

If the duration property exist (so it’s not `default` object) and it’s already a `timedelta` instance, then currently `None` is returned.

Maybe it should be something like that:
```python
    if isinstance(duration, timedelta):  # it cannot be default as default is not a timedelta instance
        return duration
    elif duration is not default:
        raise InvalidCalendar(
            f"DURATION must be a timedelta, not {type(duration).__name__}."
        )
    return None
```

But maybe I didn’t understand the code correctly…

niccokunzmann:
## Closes issue

- [ ] Closes #884

## Description

timedelta without a wrapper will render into wrong ical values. As such, it is not used for DURATION.

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1341.org.readthedocs.build/


src/icalendar/tests/test_issue_884_get_duration.py
```diff
@@ -0,0 +1,25 @@
+"""If the duration property exist (so it's not default object) and it's already a timedelta instance,
+then currently None is returned.
```
stevepiercy:
```suggestion
"""If the duration property exists, and it's not a ``default`` object but a
``timedelta`` instance, then it should return a ``duration``, not ``None``.
```

CHANGES.rst
```diff
@@ -43,7 +43,8 @@ Bug fixes
 
 - Allow lenient parsing of content lines with optional whitespace around property and parameter delimiters (for example, ``REFRESH - INTERVAL; VALUE = DURATION:PT48H``) when parsing calendars with ``strict=False``. :issue:`351`
 - X-properties with a ``VALUE`` parameter are now parsed using the correct type instead of falling back to :class:`~icalendar.prop.unkown.vUnknown`. :issue:`1238`
-- Fixed :func:`~icalendar.attr.get_end_property` to avoid allowing the creation of VEVENT components with negative durations. Only VTODO components are allowed to have negative durations. :issue:`999`
+- Test the the ``DURATION`` property catches :class:`datetime.timedelta` objects without vProperty wrappers. :issue:`884`
```
```suggestion
- Test that the ``DURATION`` property catches :class:`datetime.timedelta` objects without vProperty wrappers. :issue:`884`
```

stevepiercy (changes requested):

stevepiercy (approved):

niccokunzmann:
Thanks!

PR #1346, opened 2026-04-29
Avoid recursion-limit crash when walking deeply nested components

uwezkhan:
This PR prevents denial of service from deeply nested iCalendar component trees by making Component.walk() use iterative traversal instead of recursive traversal.

### DIFF
--- a/src/icalendar/tests/test_unit_cal.py
+++ b/src/icalendar/tests/test_unit_cal.py
@@ -1,5 +1,6 @@
 import itertools
 import re
+import sys
 from datetime import datetime, timedelta
 
 import pytest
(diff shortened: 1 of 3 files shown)


niccokunzmann (dismissed):
@uwezkhan Thanks for your PR! Could you do the code formatting? Then, it is ready for merge.

src/icalendar/tests/test_unit_cal.py
```diff
@@ -127,6 +128,23 @@ def test_filter_walk(calendar_component, filled_event_component):
         "20000101T000000"
     ]
```
niccokunzmann, line 130:
```suggestion


```

niccokunzmann:
I just pushed a formatting fix, so you can concentrate on the other PR.

niccokunzmann:
Thanks for adding this! You might need to rebase your other PR.

PR #1347, opened 2026-04-29
Fix RecursionError in Component.walk() for deeply nested components

uwezkhan:
`Component.walk()` currently relies on recursive traversal through `_walk()`, which can raise `RecursionError` when processing deeply nested component trees.

This change replaces the recursive traversal with an iterative stack-based implementation while preserving the existing traversal order and filtering behavior.

A regression test has been added to cover deeply nested components and ensure `walk()` no longer fails due to Python recursion limits.


niccokunzmann:
Thanks for this. I will close it in favour of #1346 - as that has tests.

PR #1348, opened 2026-04-29
Fix RecursionError in property_items() for deeply nested components

uwezkhan:
`property_items()` currently relies on recursive traversal when `recursive=True`, which can raise `RecursionError` for deeply nested component trees.

This change replaces the recursive traversal with an iterative stack-based implementation while preserving the existing output order and behavior.

The iterative implementation has been verified against the original recursive logic across flat, nested, and multi-sibling component structures with all combinations of `recursive` and `sorted` flags to ensure identical output.

A regression test has been added to cover deeply nested components and prevent future recursion-related regressions.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,7 @@ Bug fixes
 - Fixed :meth:`Component.__eq__ <icalendar.cal.component.Component.__eq__>` method not being commutative when comparing subcomponents. :issue:`1224`
 - Verified that the ``VALUE`` parameter of jCal components is used for the type of the component property. :issue:`1237`
 - Fix :func:`~icalendar.parser.string.escape_char` handling of ``bytes`` input by converting with :func:`icalendar.parser_tools.to_unicode` before escaping. :issue:`1226`
+- Fixed ``RecursionError`` in ``walk()``, ``property_items()``, and ``to_ical()`` by using iterative implementations for component traversal and property extraction. :pr:`1348`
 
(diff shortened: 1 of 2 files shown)


niccokunzmann:
Thanks for your PR!
When #1346 is merged - do you think, you can use the new walk() functionality inside of property items to list all the property items? It should make this much shorter. My hunch is that you do not need to duplicate the stack in this method if you can use walk.

What do you think?

niccokunzmann (dismissed):
Ok. I had a deeper look. Actually, this is how it needs to be done.

uwezkhan:
Thanks @niccokunzmann  — that makes sense.

Reusing the new iterative walk() traversal here should help keep the traversal logic centralized and avoid maintaining separate stack implementations.

I’ll update this PR once #1346 is merged and rework property_items() to build on top of that.

niccokunzmann:
Thanks for your comment. I will wait for your evaluation. Both ways are acceptable.

uwezkhan:
I re-evaluated this after #1346 merged.

While reusing `walk()` is possible, I think the current dedicated stack is clearer here because `property_items()` needs additional traversal state to emit `BEGIN / properties / child components / END` in the correct order.

Since both approaches are acceptable, I’ve kept the current implementation and updated the branch to latest main.

niccokunzmann (approved):

niccokunzmann:
This is included in v7.1.0.

PR #1351, opened 2026-04-30
donate -> fund

niccokunzmann:
## Closes issue

Issue: Wording is important. This is not about giving donations but making a dependency sustainable, taking responsibility. I prefer "fund" over "donate".

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1351.org.readthedocs.build/


docs/_templates/donate.html
```diff
@@ -1,2 +1,2 @@
 <p>You can support icalendar with a donation through Open Collective.</p>
```
stevepiercy:
```suggestion
<p>You can support icalendar by funding our work through Open Collective.</p>
```

stevepiercy (changes requested):
I'll merge this change, since you're away. Agree that donation is not the right term, which implies charity and a tax deduction.

PR #1353, opened 2026-05-02
Fix binary data corruption in vBinary for non-UTF-8 bytes

uwezkhan:
vBinary currently converts byte input through UTF-8 text coercion, which corrupts arbitrary non-UTF-8 binary payloads.

This change preserves raw byte input in vBinary and encodes it directly when serializing to iCal, allowing binary values such as images and attachments to round-trip without data loss.

Tests were added to cover non-UTF-8 binary payloads and verify compatibility with existing image handling.

----
📚 Documentation preview 📚: https://icalendar--1353.org.readthedocs.build/

### DIFF
--- a/src/icalendar/prop/binary.py
+++ b/src/icalendar/prop/binary.py
@@ -18,10 +18,7 @@ class vBinary:
 
     def __init__(self, obj: str | bytes, params: dict[str, str] | None = None) -> None:
         if isinstance(obj, str):
-            try:
-                self.obj = self.from_ical(obj)
-            except ValueError:
-                self.obj = obj.encode("utf-8")
+            self.obj = obj.encode("utf-8")
         else:
             self.obj = obj
         self.params = Parameters(encoding="BASE64", value="BINARY")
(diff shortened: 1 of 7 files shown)


stevepiercy:
@uwezkhan thank you for the contribution!

Let's get @niccokunzmann, @SashankBhamidi, or @angatha for a technical review.

This looks good to me, but I'd like a second pair of eyes.

src/icalendar/prop/binary.py
```diff
@@ -7,18 +7,20 @@
...
-from icalendar.parser_tools import to_unicode
 
 
 class vBinary:
     """Binary property values are base 64 encoded."""
 
     default_value: ClassVar[str] = "BINARY"
     params: Parameters
-    obj: str
+    obj: str | bytes
```
angatha, line 17: This is always `bytes`?

src/icalendar/prop/image.py
```diff
@@ -56,7 +56,11 @@ def from_property_value(cls, value: vUri | vBinary | vText) -> Image:
         if value_type == "URI" or isinstance(value, vUri):
             params["uri"] = str(value)
         elif isinstance(value, vBinary):
-            params["b64data"] = value.obj
+            if isinstance(value.obj, bytes):
```
angatha, line 59: should always be true?

angatha:
Im a little bit confused how this relates to #1356. @uwezkhan should we review the other one instead of this?

uwezkhan:
Closing this in favor of the refined follow-up PR.

PR #1354, opened 2026-05-02
DOC: Remove print() calls from tutorial documentation

### LINKED ISSUE #1257: Replace print() in documentation
The print statement in the documentation has been modified. However, this causes confusion.

## Current solution

-  to convert bytes to string - Python 2 compatible
- provide an easy way to display the resulting component as ics

## Current Problems

- when you copy & paste the code, you get a different output
- that can be confusing

## Future ideas

I would like people to be able to view and see, grasp and have documented well the resulting ICS output and calendar.

- it would be nice to use ics syntax highlighting on the `<output>`
- it would be nice to do something like this in the docs:

```python
>>> calendar = Calendar.new()
>>> calendar.view()
<output>
```

- When using the above statement, what I would like to be able to switch between ics and jcal output in the documentation, like [sphinx-icalendar](https://sphinx-icalendar.readthedocs.io/en/latest/usage.html#calendar-rendering).

## Implementation

- test/conftest.py needs changing
- if we go with `view()`, then, this goes into Component
- A new solution should solve the problems and discuss the best way forward for the ideas

## Inspiration

[vObject](https://py-vobject.github.io/#usage)

```
>>> cal.prettyPrint()
 VCALENDAR
    VEVENT
       SUMMARY: This is a note
       params for  SUMMARY:
          X-RANDOM ['Random parameter']
```

Sevda2661:
## Closes issue

Closes #1257

## Description

Replace print() statements with direct expression calls 
in create-event-with-attendees tutorial documentation.
This makes the code examples cleaner and more consistent
with Python's interactive shell output style.

## Checklist
- [x] I've added or edited documentation, both as docstrings 
to be rendered in the API documentation and narrative 
documentation, as necessary.

----
📚 Documentation preview 📚: https://icalendar--1354.org.readthedocs.build/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,7 +33,9 @@ Bug fixes
 Documentation
 ~~~~~~~~~~~~~
 
-- ...
+- Remove ``print()`` calls from the create-event-with-attendees tutorial,
+  replacing them with direct expression calls for cleaner output.
+  :issue:`1257`
 
 7.1.0 (2026-04-30)
 ------------------
--- a/docs/tutorials/create-event-with-attendees.rst
+++ b/docs/tutorials/create-event-with-attendees.rst
@@ -20,7 +20,7 @@ After creating a new calendar, you can view its properties by using :meth:`~ical
 
 ..  code-block:: pycon
 
-    >>> print(calendar.to_ical().decode())
+    >>> calendar.to_ical().decode()
     BEGIN:VCALENDAR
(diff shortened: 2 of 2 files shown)


stevepiercy:
@Sevda2661 issue #1257 is not as simple as removing the `print()` statement, and in fact causes the tests to fail. Please read its Implementation section for guidance. Closing.

PR #1356, opened 2026-05-03
Fix binary data corruption in vBinary for non-UTF-8 byte input

uwezkhan:
vBinary currently coerces byte input through Unicode conversion before Base64 serialization, which corrupts arbitrary non-UTF-8 binary payloads.

This change preserves raw byte input in vBinary and serializes it directly to Base64, ensuring binary values round-trip losslessly.

Related image handling and jCal serialization were updated accordingly to maintain compatibility with the new raw-bytes storage.

Regression tests were added to cover non-UTF-8 binary payloads and verify that image handling continues to work as expected.


uwezkhan:
Thanks for pointing that out — this PR is the refined follow-up to my earlier attempt.

I’ll close #1353 to avoid duplicate review and keep this PR as the active version of the fix.

stevepiercy:
@angatha @SashankBhamidi @niccokunzmann would y'all please do a technical review? Thank you!

SashankBhamidi:
I'd like to learn if you're using assistance of any AI tool since this is interesting, could you let us know? Thanks!

uwezkhan:
Thankyou for the positive response . 

I manually research, validate and test the program before raising any PR . also i do use AI for the purpose of wordings to enhance my description of PR.

uwezkhan:
any update on this PR?

stevepiercy:
@uwezkhan thanks for your patience and for adding the change log entry per the 7.1.1 release.

This PR will require technical review. @SashankBhamidi @angatha or @niccokunzmann would one of you please take a look and let @uwezkhan know? Thank you!

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -23,8 +21,8 @@ def test_binary():
 
 
 def test_param():
-    assert isinstance(vBinary("txt").params, Parameters)
```
angatha, line 26: Please keep the old test cases in place. They should still be valid or do you want to do **a breaking change**?

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -61,14 +59,9 @@ def test_from_ical_rejects_non_base64_characters(value):
 
 
 def test_ical_value():
-    """ical_value property returns the string value."""
-    magic_string = base64.b64encode(b"magic string")
```
angatha, line 65: Same: Please keep the old test cases in place. They should still be valid or do you want to do **a breaking change**?

angatha (changes requested):

SashankBhamidi (dismissed):
Thanks. The bug is real, confirmed it locally. `vBinary(bytes(range(256)))` round-trips to 512 bytes on main and 256 on this branch. The jCal output is fixed too, the old `'Hello World!'` becomes `'SGVsbG8gV29ybGQh'`, which matches the RFC 7265 example.

@angatha's concern stands. The PR isn't just refactoring internal storage; it also changes the contract of `ical_value`, and that isn't called out in the news entry.

- `ical_value` used to decode the stored value as base64. `vBinary("SGVsbG8=").ical_value` returned `b"Hello"` on main, and returns `b"SGVsbG8="` here.
- `vBinary("!!!invalid!!!").ical_value` used to raise `ValueError`. Now it silently returns the bytes.

`ical_value` was only added in 7.1.0, so the external blast radius is small, but the new behavior is the opposite of what the previous version did. This needs to be deliberate.

The new "always raw bytes" semantics are cleaner than what main had, the old design was internally inconsistent (`to_ical` treated `self.obj` as raw data while `ical_value` treated it as base64). I'd lean toward keeping the new semantics, but:

- The news entry should explicitly mention the `ical_value` change.
- The category should be `breaking`, not `bugfix`. Pinging @niccokunzmann for a sanity check on that.

On the test removals @angatha flagged:
- `test_param` should stay as it was. `vBinary("txt")` still works on this branch, the original assertions still pass.
- `test_ical_value_rejects_non_base64_characters` can go, but the news entry should explain why.

SashankBhamidi:
The underlying bug is tracked in #1289, where @niccokunzmann proposed a specific migration path (introduce `vBinary.bytes`, keep `vBinary.obj` as a property returning string data, optionally deprecate). This PR takes a different approach by changing `.obj` to bytes directly. Worth aligning with that thread before continuing.

uwezkhan:
Thanks both. I’ve reworked the PR to follow the migration path discussed in #1289: added `vBinary.bytes`, kept `.obj` as a deprecated compatibility view, restored the original tests, and updated the news entries accordingly.

@niccokunzmann does this align with what you had in mind for #1289?

src/icalendar/prop/binary.py
```diff
@@ -14,11 +14,13 @@ class vBinary:
     """Binary property values are base 64 encoded."""
 
     default_value: ClassVar[str] = "BINARY"
-    params: Parameters
-    obj: str
+    bytes: bytes
```
niccokunzmann, line 18: Nice. It would probably be good to add a docstring for this somehow, so it turns up in the docs. I have not had a look at the docs.

niccokunzmann (changes requested):
This change is technically sound and well made. It is an improvement.

I cannot say wether it solves the issue. I would say: we should get a review from the one who actually uses vBinary.

For backward compatibility, we also need a setter for obj.

src/icalendar/prop/binary.py
```diff
@@ -36,13 +38,25 @@ def from_ical(ical: str | bytes) -> bytes:
         except (binascii.Error, ValueError) as e:
             raise ValueError("Not valid base 64 encoding.") from e
 
+    @property
+    @deprecate_for_version_8
+    def obj(self) -> str:
+        """Deprecated string view of the value.
+
+        .. deprecated:: 7.1.1
```
stevepiercy:
```suggestion
        .. deprecated:: 7.1.3
```

stevepiercy: Verify in https://icalendar--1356.org.readthedocs.build/en/1356/reference/api/icalendar.prop.binary.html#icalendar.prop.binary.vBinary.obj

src/icalendar/prop/binary.py
stevepiercy: https://icalendar--1356.org.readthedocs.build/en/1356/reference/api/icalendar.prop.binary.html#icalendar.prop.binary.vBinary.bytes

The most preferred method of adding a docstring for class attributes is in the class's docstring.

```python
class vBinary:
    """Binary property values are base 64 encoded.

    Attributes:
        bytes: description of bytes.
    """
```

See https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure.

src/icalendar/prop/binary.py
```diff
@@ -57,12 +71,23 @@ def to_jcal(self, name: str) -> list:
...
+            base64.b64encode(self.bytes).decode("ascii"),
+        ]
 
     @property
     def ical_value(self) -> bytes:
-        """The bytes value of the BINARY property."""
-        return self.from_ical(self.obj)
+        """The raw ``bytes`` value of the BINARY property.
+
+        .. versionchanged:: 7.1.1
```
stevepiercy:
```suggestion
        .. versionchanged:: 7.1.3
```

stevepiercy (changes requested):
I did a brief review on docstrings only.

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -61,14 +59,44 @@ def test_from_ical_rejects_non_base64_characters(value):
...
 
-def test_ical_value_rejects_non_base64_characters():
-    with pytest.raises(ValueError, match=r"Not valid base 64 encoding\."):
-        vBinary("!!!!dGV4dA==@@@@").ical_value
+def test_ical_value_returns_raw_bytes_not_decoded():
+    """ical_value returns the raw stored bytes; it does not base64-decode them.
+
+    This is the breaking change from #1356. Previously ical_value decoded the
+    stored value as base64 (so ``vBinary("SGVsbG8=").ical_value`` was ``b"Hello"``)
+    and raised ValueError for non-base64 input. See news/1356.breaking.
```
stevepiercy: We can't refer to files in `news/` because upon release, their content gets written to `CHANGES.rst` then the file is deleted. Instead refer to the PR. I also added when the bug was introduced, and wordsmithed it a bit for clarity. Also its proper spelling is "Base64".

```suggestion
    """Test that ``ical_value`` returns the raw stored bytes.
    
    With the release of icalendar 7.1.0, and previous to PR #1356,
    ``ical_value`` Base64-decoded the raw stored bytes. For example,
    ``vBinary("SGVsbG8=").ical_value`` was decoded to ``b"Hello"``
    and raised ``ValueError`` for non-Base64 input.
```

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -61,14 +59,44 @@ def test_from_ical_rejects_non_base64_characters(value):
...
+    """
+    assert vBinary(b"SGVsbG8=").ical_value == b"SGVsbG8="
+    # Non-base64 input no longer raises; it is just stored and returned as-is.
+    assert vBinary(b"!!!!dGV4dA==@@@@").ical_value == b"!!!!dGV4dA==@@@@"
+
+
+def test_bytes_holds_raw_lossless_data():
+    """The .bytes attribute exposes the raw value, including non-UTF-8 data."""
```
stevepiercy: I think that it's worth mentioning the PR here to explain why this test exists.

```suggestion
    """The .bytes attribute exposes the raw value, including non-UTF-8 data.
    
    See PR #1356.
    """
```

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -61,14 +59,44 @@ def test_from_ical_rejects_non_base64_characters(value):
...
+    raw = bytes(range(256))
+    binary = vBinary(raw)
+    assert binary.bytes == raw
+    # round-trips losslessly through the wire format
+    assert vBinary.from_ical(binary.to_ical()) == raw
+
+
+def test_obj_is_deprecated_string_view():
+    """.obj is kept for backward compatibility but deprecated in favour of .bytes."""
```
stevepiercy:
```suggestion
    """.obj is kept for backward compatibility but deprecated in favour of .bytes.
    
    See PR #1356.
    """
```

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -61,14 +59,44 @@ def test_from_ical_rejects_non_base64_characters(value):
...
+    with pytest.warns(DeprecationWarning, match="obj is deprecated"):
+        assert vBinary(b"txt").obj == "txt"
+
+
+def test_obj_setter_updates_bytes():
+    """Setting .obj still works for backward compatibility and writes .bytes."""
```
stevepiercy:
```suggestion
    """Setting .obj still works for backward compatibility and writes .bytes.
    
    See PR #1356.
    """
```

news/1356.breaking
```diff
@@ -0,0 +1 @@
+:class:`~icalendar.prop.binary.vBinary` now stores the raw ``bytes`` of the value internally, so non-UTF-8 binary data round-trips losslessly (previously it was decoded as text and corrupted). As a consequence, :attr:`vBinary.ical_value` now returns the raw stored bytes instead of base64-decoding them, and no longer raises :class:`ValueError` for non-base64 input. ``ical_value`` was introduced in 7.1.0, so the external impact is limited. @uwezkhan
```
stevepiercy: This will link properly, after towncrier runs upon release. I also wordsmithed a bit.

```suggestion
:class:`~icalendar.prop.binary.vBinary` now stores the raw ``bytes`` of the value internally, so non-UTF-8 binary data round-trips losslessly. With the release of icalendar 7.1.0, and previous to this bug fix, :attr:`vBinary.ical_value <icalendar.prop.binary.vBinary.ical_value>` Base64-decoded the raw stored bytes and raised :class:`ValueError` for non-Base64 input. :attr:`vBinary.ical_value <icalendar.prop.binary.vBinary.ical_value>` now returns the raw stored bytes instead of Base64-decoding them, and no longer raises :class:`ValueError` for non-Base64 input. @uwezkhan
```

news/1356.feature
```diff
@@ -0,0 +1 @@
+Added :attr:`icalendar.prop.binary.vBinary.bytes` exposing the raw binary value of a BINARY property. @uwezkhan
```
stevepiercy:
```suggestion
Added :attr:`icalendar.prop.binary.vBinary.bytes`, exposing the raw binary value of a BINARY property. @uwezkhan
```

news/1356.removal
```diff
@@ -0,0 +1 @@
+:attr:`icalendar.prop.binary.vBinary.obj` is deprecated in favour of :attr:`~icalendar.prop.binary.vBinary.bytes`. It is retained as a (lossy) string view of the value for backward compatibility and will be removed in icalendar 8. The ``test_ical_value_rejects_non_base64_characters`` test was removed because ``ical_value`` no longer validates or decodes base64. @uwezkhan
```
stevepiercy: American English spelling and punctuation, some markup tweaks.

```suggestion
:attr:`~icalendar.prop.binary.vBinary.obj` is deprecated in favor of :attr:`~icalendar.prop.binary.vBinary.bytes`. It is retained as a lossy string view of the value for backward compatibility, and will be removed in icalendar 8. The ``test_ical_value_rejects_non_base64_characters`` test was removed because :attr:`vBinary.ical_value <icalendar.prop.binary.vBinary.ical_value>` no longer validates or decodes Base64. @uwezkhan
```

src/icalendar/prop/binary.py
stevepiercy:
```suggestion
    """Binary property values are Base64 encoded."""
```

src/icalendar/prop/binary.py
```diff
@@ -57,12 +82,23 @@ def to_jcal(self, name: str) -> list:
...
 
     @property
     def ical_value(self) -> bytes:
-        """The bytes value of the BINARY property."""
-        return self.from_ical(self.obj)
+        """The raw ``bytes`` value of the BINARY property.
+
+        .. versionchanged:: 7.1.3
+            Returns the raw stored bytes. Previously the stored value was
+            base64-decoded, which raised :class:`ValueError` for non-base64
```
stevepiercy:
```suggestion
            Base64-decoded, which raised :class:`ValueError` for non-Base64
```

src/icalendar/prop/binary.py
```diff
@@ -57,12 +82,23 @@ def to_jcal(self, name: str) -> list:
...
+            input. See ``news/1356.breaking``.
```
stevepiercy:
```suggestion
            input. See :pr:`1356`.
```

src/icalendar/prop/binary.py
```diff
@@ -57,12 +82,23 @@ def to_jcal(self, name: str) -> list:
...
+            params,
+            self.VALUE.lower(),
+            base64.b64encode(self.bytes).decode("ascii"),
+        ]
 
     @property
     def ical_value(self) -> bytes:
-        """The bytes value of the BINARY property."""
-        return self.from_ical(self.obj)
+        """The raw ``bytes`` value of the BINARY property.
```
stevepiercy, line 117: Better late than never. I think this helps clarify what happened.

```suggestion
        """The raw ``bytes`` value of the BINARY property.

        .. versionadded:: 7.1.0
```

stevepiercy (changes requested):
I did a thorough review on this pass. The code looks great, and I made some suggestions to docstrings and news fragments. Would you please take care? Thank you!

stevepiercy:
@niccokunzmann @SashankBhamidi would you please do another technical review?

@FilipAtanasovski-dev and @kmurphy4 does this pull request solve the issue you each reported?

- #1289
- #1299

uwezkhan:
Done — applied all of these. Switched to "Base64" everywhere, pointed the news fragments at the PR instead of the news files, and added the 7.1.0 versionadded. Thanks for the detailed review!

news/1356.removal
uwezkhan: Thanks! Fixed. Base64 and American English win this round..

stevepiercy (dismissed):
I approve the docs. Thank you for the changes!

@niccokunzmann @SashankBhamidi @angatha would you please do another technical review to ensure that @uwezkhan has addressed all of your concerns?

@FilipAtanasovski-dev and @kmurphy4 does this pull request solve the issue you each reported?

-   https://github.com/collective/icalendar/issues/1289
-   https://github.com/collective/icalendar/issues/1299

Please let us know. Thank you!

src/icalendar/prop/image.py
```diff
@@ -56,9 +56,13 @@ def from_property_value(cls, value: vUri | vBinary | vText) -> Image:
         if value_type == "URI" or isinstance(value, vUri):
             params["uri"] = str(value)
         elif isinstance(value, vBinary):
-            params["b64data"] = value.obj
+            params["b64data"] = base64.b64encode(value.bytes).decode("ascii")
```
angatha: Just if you want, you can add a property for the Base64 string view since it is used here and in `binary.py`

angatha (dismissed):

src/icalendar/prop/image.py
uwezkhan: Good suggestion. I think I'd prefer to keep this PR focused on the raw-bytes fix and avoid introducing another API surface here, but I can see the value of a dedicated Base64 string property if we find more call sites that need it.

uwezkhan:
any update?

kmurphy4:
> @kmurphy4 does this pull request solve the issue you each reported?

Yes, I think so.  I verified with

```py
import base64
import icalendar
import importlib.metadata


ICALENDAR_VERSION = importlib.metadata.version("icalendar")
EXPECTED_HEADER = b"\x89PNG\r\n"


with open("example.ical", "rb") as fin:
    ical = icalendar.Calendar.from_ical(fin.read())

attach = ical.subcomponents[0]["ATTACH"]
if ICALENDAR_VERSION == "6.3.2":  # most recent 6.x version upstream
    actual_header = base64.b64decode(attach)[:6]
elif ICALENDAR_VERSION == "7.2.0":  # most recent 7.x version upstream
    actual_header = attach.obj[:6]
elif ICALENDAR_VERSION == "7.0.4.dev726":  # https://github.com/collective/icalendar/pull/1356
    actual_header = attach.bytes[:6]
else:
    raise NotImplementedError(ICALENDAR_VERSION)

assert actual_header == b"\x89PNG\r\n", repr(actual_header)
print("ok")
```
and confirmed with

```
$ uv pip install icalendar==6.3.2
$ python example.py
ok

$ uv pip install icalendar==7.2.0
$ python example.py
  File "example.py", line 23, in <module>
    assert actual_header == b"\x89PNG\r\n", repr(actual_header)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: '�PNG\r\n'

$ uv pip install .
$ python example.py
ok
```


where `example.ical` is
```ical
BEGIN:VCALENDAR
VERSION:2.0
BEGIN:VEVENT
DTSTAMP:19960704T120000Z
UID:uid1@example.com
ORGANIZER:mailto:jsmith@example.com
DTSTART:19960918T143000Z
DTEND:19960920T220000Z
STATUS:CONFIRMED
CATEGORIES:CONFERENCE
SUMMARY:Summary
DESCRIPTION:Description
ATTACH;VALUE=BINARY;ENCODING=BASE64;X-FILENAME=image.png:
 iVBORw0KGgoAAAANSUhEUgAAAKwAAACyCAMAAADoM9QBAAAAXVBMVEUAAAD////2/ezlxRL7ztH2
 +Pq06PXJ6e7n8PDH29JInHJfbGV1gnu3xr6Jn5KhsqfZ59w0SS7P6cLH4qattm3y8Lnt4oDp00WK
 dUn+8O774N+oenrClJTgtrb00Q1TIg0UAAAIfUlEQVR4XuzOMQ0AAADCsCmZf5mo4CChCopDsOfZ
 sGNvPY7iUBCAByYOds7NF5LBhOz//5lbJ8lotrXvdEainLhNP31dOkIN41+Rvxd7YA/sgT2wB/bA
 HtgDe2AP7IE9sAf2wB7YA3tgmZmI/g6s5bAmeYJ9+2gsrMMw9CGEnkX5dz4SK2voMULr5CH0DrIx
 E3vPTB+FtRBiSjmnHAO6BRfoDnlPpq5FXP0BWE2BLU0C1ytqSUQmwGMYsEK27U1+/vhGrMbIpGma
 0oRy44SCk5qJMRXSlJOkGDqyitE7Duf9sbBOrClNIIJZU821Sc7TO6Ie4BhV+5DELL8nY3esRVFU
 l6fk2Iq95mutfpimCO1r9+LNheY9hxwS79+sBGVwWmvVV23vONe9aPicztP5HZyMEBajvjc2iWpG
 3sS39fbCVgwxAi7Ab6u7veN7gHVfrGWy4JPqyaBWxNGvvCsWSP8EdiUOu98NTCiOkrOk+s61/i/t
 jHL/qxUmgnVfLCf2WrOX6qh5WealYJuXWaSm6vFq5QvWVCSOO2M5qGYxk1qTNFn41m7PtHmeuWgd
 K6AV6mlKf4bA2CyGnbE6GampXkqhi7V2vf26/UKwAXxt1zyOMUfE7wjT2RfUbGcGdV+sBSl6gfRi
 tWkp8+16gxP5+fPXz4VmXlgZMWJWNRGBWPgsAdZ9sbCiz8srelEleNvNxxVj0OarzVU0J7UslCdh
 BZqUJwV1Z6zkrJqg/BPT8mNBufMzrYEGqVGqRaMVFSIyt+6NlTBIyurYL9xLAa00VaZqZa5kMy04
 NfthVkiy7v4vIuchJCpMdPkatUvRStpK4wKj4dB+zNdCuC48Jh33xmoMyLr2sNrGVIgvr+cuRBla
 loWha0yVqCrlWljJhrz/k4L0AYFU7rzl2EMSUXPrBatpGeusRRqxFZOiWnAlxXrqu2MlDiGG2E/3
 5b5tj/sJFSM5xiTGxFTYqpnXCW2Vwkxc6REF1p2xyR9S/NmqP5y6PLb79kD89z4eGVDVhYuako+I
 KGdjGWz/B0b02uMwhPR4nE6nFV+Y7/dluT96H/o4YI15zDURlKUwqazMY9Rxd2yG0ytMvNwd+RSv
 jl5RNj5DyIODwzjgk5OossYYuO+N9VsWAnCXE6j37f42P9a+9t8JvTs8DT7dOWAL3/SSo0t28NN2
 Wn0GgN0WTIIf1yfZ/5oQYI04AZu+741M968mVPaEgb0+YPWifW0YjRVWBNKAdu3bXx+BjDzW4OL1
 Obm4MbyzndYXdRiGaOGj3nVJ6mF1/oo8ZGNNHc7gSdrHj8L24KDtsUI8jD7QrwkAOPInv/k2yVAi
 AxJl7J+MfenUsosTLj4X+9VM/V92zWjHdRQIoruU/AlVJeXh/v9nrtZASoSxcu1oNNe7+CHYdHVz
 wD0oaeb/dwCyYBfsgl2wC3bBLtgFu2AX7IJdsAt2wS7YBbtgF+yCla0/H5ZgKQIA3AFWxZBktQ5r
 El22fA5LSpJIllIEGvy3e6tWA5PHRcvnsDJyASbhwQ4N+suWz2EJWCQJiHVt/TrSVnJdtnwOmyUg
 cgeW774o8CwsrSc1cvf9sIZPr6w8eOfuuy9cgAXjrQGWrtHU2kLZXf3GQpm1VWsYYVLtJCx3l3gn
 eYXaGEBPbmeEI0vrqI9doRchbQO2eRLWGSDYpGWAmyBXiaBhmzi2EDZQCk1VhaFRKAGWtJ1MA/A1
 UQVQtaGeyeVdGKJDS4G23aJnUECDMF3nYA3NKWvlL4AAI8z1xhKkSRjfs7CF03YFOAMm36bwh5Zt
 7wmS4FehrsF+mbIJl/QAHPk7S5+Gt/HtefC9Cmt4C3bCEVDvzwDHlg61lVBOwnRdgmW8/ZoF7Ibc
 HlvmlM1Li5AAL8HOIytpl5GG5XhreURBwIMwvrwG66Qsshxte6KE+cvYscVwpkPvARlhV9C6BEvA
 W5aloch1KaA6AYMD7LHFz1YWwKIIu6+sK2lAIe9RcBvIrrMwi+GX5Tu2JKX6N3lIjjCWKzkr7L62
 lWUxwK0isb5LDETHliRzSwAA2iKM5SxsvtOT5PMhyc9ny24pbyyGhxjULOSfUuQAdJuKDAHeBlZw
 uQ0swNvAEi53gJVZil3uAQvAvkt9VhLvU0ze/pOVb8rmPWBpAPAtYGnoNitraCvbPXJWwH3+wAzf
 BpaAbgMrgJ/D8vSwuuJtoHwES9l2v60NZ03rawoD+tqb1i6r5nbTNyvAQ8QTsNmjoX5bG48iW7VP
 VSrAmr3hXkM2ADa9y2NXJWUT8RysrJ5JMgEXeoJtmn1ZvUs4edMCTMtAsdxCyhAhAWCXJuI5WKYs
 yVrVUG0GVnSNtkKggLO3HwWQWQCbfW69sxBgUjYRz8BCqSLuEbNUmdAuAliiOfAG22cx9JSzdFhU
 aSKegTWG3896rfzFajDb+qE3e1oSYDorbLUNEc/AEnqtonxdMKTFTFBH3q4x5s6eRcAQ8RSskMpm
 KDnCumVdnn3sHWT3TmaKhoeI509rEpkA5ywwzL/HU/RD74ac+XIQA07Ec7B5K4k8nxJk3PbYICdv
 AR15EzD+NlemwEQ8D0sDLKobS6ejBtic85o9aWdvt952ryR4edo2YYhI/T6sWsFXfkYWQFpbYF2K
 bGYz9+ydlO33UmD1zC+ZiZileA9LADCFHH41CHBLYu+XyqMDEDj27p92kzq7lAE7EXOE8xa2e++f
 ymnbNNmdhhVODZ+Td3ZZjgczSJ1bkWI6H38LW/joheFcx8+9FTR7b18fXXoj0zuPQH7zl28Cv6k0
 9LNFjjCYJyqgPwIbWkMnXsHPwRbDhk/UwX8StlDibxU07lKYI+ByD1gKAGDdArZd6z+TP78W7IJd
 sAt2wS7YBbtgF+w/7d3BAAAAAIPAkeSPuUcaRZDBpbJaswdUrHtR+3in6gAAAABJRU5ErkJggg==
END:VEVENT
END:VCALENDAR
```

stevepiercy:
@niccokunzmann @SashankBhamidi would you please review this PR to ensure the changes you requested have been satisfied? Thank you!

niccokunzmann (dismissed):
This is ready for merge. Thanks for all the effort, time and feedback!

niccokunzmann:
@kmurphy4 Thanks for the example. I value having examples from real use cases for the tests. I will open an issue to add that code as a test.

niccokunzmann:
@SashankBhamidi have a look again the old behaviour was badly tested and inconsistent. I do not know if we are breaking anything but a buggy behaviour. The one example provided now works.

I would like to see it merged and get feedback rather than block it and keep confusion.

In case you approve, please merge it.

niccokunzmann:
There is a new commit in `main` which can be improved when merging.

uwezkhan:
Merged main into this branch and adapted the new base64data property from #1550 to the raw-bytes storage, as its commit message anticipated.

Before: the setter wrote through .obj, which coerces the decoded bytes to text. On this branch that would be lossy for non-UTF-8 payloads and would also emit the .obj deprecation warning.
After: the setter writes the decoded bytes to .bytes directly, so non-UTF-8 data round-trips losslessly and no warning fires. The getter is unchanged.

to_jcal and Image.from_property_value now use base64data instead of manual b64encode calls, which is also what @angatha suggested earlier. All base64data tests from main pass unchanged, and I added one for the lossless setter path.

niccokunzmann (approved):
Thanks!

niccokunzmann:

PR #1357, opened 2026-05-04
docs(alarm): rewrite REPEAT and DURATION docstrings (refs #1244)

tmchow:
## Closes issue

- [x] Refs #1244

This PR addresses two of the unchecked items in the #1244 checklist (`Alarm.REPEAT` and `Alarm.DURATION`). It does not close the umbrella issue.

## Description

The existing docstrings for `Alarm.REPEAT` and `Alarm.DURATION` are paragraph-style copy-paste from RFC 5545. The umbrella issue asks for these to be rewritten as Pythonic docstrings.

The pattern matches PR #1338 for `Calendar.categories`:

- Lead with a one-line summary of what the property does in Python terms (default value, accepted types, return type).
- Note that the two are paired: a repeating alarm sets both, and `REPEAT` counts additional triggers, not total triggers (worth calling out because the iCal wording is easy to misread).
- Reference the RFC sections via `:rfc:` directives so Sphinx renders linked anchors instead of bare quotation.
- End with a working `pycon` example that pairs `TRIGGER`, `DURATION`, and `REPEAT` to produce the same behavior described in the prose.

The RFC text itself is no longer copy-pasted into the docstring; the `:rfc:` link is the canonical reference.

The CHANGES entry slots in next to the other `:issue:` 1244 entries under the 7.0.4 Documentation section.

## Checklist

The doctest suite (`pytest src/icalendar/tests/test_with_doctest.py`) passes locally with the new examples (343/343).

## Additional information

The other unchecked items under `Alarm` in #1244 (`ACKNOWLEDGED`, `TRIGGER`, `TRIGGER_RELATED`, `Triggers`, `attendees`, `description`, `triggers`, `uid`) are intentionally left for separate PRs to keep this one reviewable and to mirror the per-property cadence of #1338.

----
📚 Documentation preview 📚: https://icalendar--1357.org.readthedocs.build/

### DIFF
--- a/src/icalendar/cal/alarm.py
+++ b/src/icalendar/cal/alarm.py
@@ -97,10 +97,9 @@ class Alarm(Component):
         triggers, so a ``REPEAT`` of ``2`` produces three alarms in total
         (the initial trigger plus two repeats).
 
-        Conformance:
-            Defined by :rfc:`5545#section-3.8.6.2`. The property can appear
-            once in a "VALARM" component and must be paired with
-            :attr:`DURATION`.
+        Conforming with :rfc:`5545#section-3.8.6.2`, this property can appear
+        once in an :class:`~icalendar.cal.alarm.Alarm` component and must be
+        paired with :attr:`DURATION`.
 
         Example:
             Build an alarm that fires once and then repeats twice at
@@ -126,7 +125,7 @@ class Alarm(Component):
         """The delay between repeated triggers of a repeating alarm.
 
         Returns a :class:`datetime.timedelta` or ``None`` when the alarm
-        has no DURATION set. Setting this attribute accepts a
+        has no :attr:`DURATION` set. Setting this attribute accepts a
         :class:`~datetime.timedelta`; deleting it removes the property
         from the component.
 
@@ -135,9 +134,8 @@ class Alarm(Component):
         ``REPEAT`` additional triggers, each spaced by ``DURATION`` after
         the initial trigger.
 
-        Conformance:
-            Defined by :rfc:`5545#section-3.8.2.5`. The property can
-            appear once in a "VALARM" component.
+        Conforming with :rfc:`5545#section-3.8.2.5`, the ``DURATION`` property
+        can appear once in an :class:`~icalendar.cal.alarm.Alarm` component.
 
         Example:
             Pair :attr:`DURATION` with :attr:`REPEAT` to produce three
--- a/src/icalendar/cal/alarm.py
+++ b/src/icalendar/cal/alarm.py
@@ -89,15 +89,33 @@ class Alarm(Component):
     REPEAT = single_int_property(
         "REPEAT",
         0,
-        """The REPEAT property of an alarm component.
+        """The number of additional times the alarm is triggered after the initial trigger.
 
-        The alarm can be defined such that it triggers repeatedly.  A
-        definition of an alarm with a repeating trigger MUST include both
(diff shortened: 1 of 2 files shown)


src/icalendar/cal/alarm.py
```diff
@@ -89,29 +89,71 @@ class Alarm(Component):
...
+        Defaults to ``0``, meaning the alarm fires once. To repeat the alarm,
+        set both :attr:`REPEAT` and :attr:`DURATION`. The :attr:`DURATION`
+        sets the gap between repetitions. ``REPEAT`` is the count of *additional*
+        triggers, so a ``REPEAT`` of ``2`` produces three alarms in total
+        (the initial trigger plus two repeats).
+
+        Conformance:
+            Defined by :rfc:`5545#section-3.8.6.2`. The property can appear
+            once in a "VALARM" component and must be paired with
+            :attr:`DURATION`.
```
stevepiercy: Avoid using RFC text for a Python implementation.

Don't use unsupported docstring section headers. This is part of the description.

```suggestion
        Conforming with :rfc:`5545#section-3.8.6.2`, this property can appear
        once in an :class:`~icalendar.cal.alarm.Alarm` component and must be
        paired with :attr:`DURATION`.
```

src/icalendar/cal/alarm.py
```diff
@@ -89,29 +89,71 @@ class Alarm(Component):
...
-
-    The alarm can be defined such that it triggers repeatedly.  A
-    definition of an alarm with a repeating trigger MUST include both
-    the "DURATION" and "REPEAT" properties.  The "DURATION" property
-    specifies the delay period, after which the alarm will repeat.
-    """,
+        """The delay between repeated triggers of a repeating alarm.
+
+        Returns a :class:`datetime.timedelta` or ``None`` when the alarm
+        has no DURATION set. Setting this attribute accepts a
```
stevepiercy: I think this should link, since it's the first mention and it discusses the Python object. Subsequent mentions don't need to be linked, but it's OK if they are.

```suggestion
        has no :attr:`DURATION` set. Setting this attribute accepts a
```

src/icalendar/cal/alarm.py
```diff
@@ -89,29 +89,71 @@ class Alarm(Component):
...
+        from the component.
+
+        :attr:`DURATION` is meaningful only for repeating alarms and must
+        be paired with :attr:`REPEAT`. The two together produce
+        ``REPEAT`` additional triggers, each spaced by ``DURATION`` after
+        the initial trigger.
+
+        Conformance:
+            Defined by :rfc:`5545#section-3.8.2.5`. The property can
+            appear once in a "VALARM" component.
```
```suggestion
        Conforming with :rfc:`5545#section-3.8.2.5`, the ``DURATION`` property
        can appear once in an :class:`~icalendar.cal.alarm.Alarm` component.
```

stevepiercy (changes requested):
This is excellent work, and I expect the doctests will pass in CI.

Would you please take a look at my suggestions?

Also would you take care of the merge conflicts?

Thank you!

tmchow:
Applied all three suggestions in 4a78922 and rebased on main to clear the CHANGES.rst conflict (kept upstream's recent docs entries; the `:issue:`1244`` line slots in after them).

Doctests still green: `pytest src/icalendar/tests/test_with_doctest.py` 346/346.

stevepiercy (approved):
Thank you! Will merge when CI passes.

PR #1358, opened 2026-05-04
Add type hints to vFrequency

arjunrawal1:
## Summary
- Adds explicit type hints to `vFrequency.__new__`, `to_ical`, and `from_ical`.
- Keeps runtime behavior unchanged while continuing the rolling type-hinting work from #938.

Refs #938

## Testing
- `python3 -m pytest src/icalendar/tests/prop/test_constructors.py src/icalendar/tests/prop/test_unit.py src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py -q` — 751 passed

Note: I also tried `python3 -m mypy src/icalendar/prop/recur/frequency.py`, but mypy is not installed in this local environment.

### DIFF
--- a/src/icalendar/prop/recur/frequency.py
+++ b/src/icalendar/prop/recur/frequency.py
@@ -29,11 +29,11 @@ class vFrequency(str):
 
     def __new__(
         cls,
-        value,
-        encoding=DEFAULT_ENCODING,
+        value: Any,
+        encoding: str = DEFAULT_ENCODING,
         /,
         params: dict[str, Any] | None = None,
-    ):
+    ) -> Self:
         value = to_unicode(value, encoding=encoding)
         self = super().__new__(cls, value)
         if self not in vFrequency.frequencies:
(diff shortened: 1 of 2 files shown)


src/icalendar/prop/recur/frequency.py
```diff
@@ -29,23 +29,23 @@ class vFrequency(str):
...
         self.params = Parameters(params)
         return self
 
-    def to_ical(self):
+    def to_ical(self) -> bytes:
         return self.encode(DEFAULT_ENCODING).upper()
 
     @classmethod
-    def from_ical(cls, ical):
+    def from_ical(cls, ical: Any) -> Self:
```
stevepiercy, line 48: Same as above.

```suggestion
    def from_ical(cls, ical: ICAL_TYPE) -> Self:
```

src/icalendar/prop/recur/frequency.py
```diff
@@ -29,23 +29,23 @@ class vFrequency(str):
 
     def __new__(
         cls,
-        value,
-        encoding=DEFAULT_ENCODING,
+        value: Any,
```
stevepiercy, line 32: I think the use of `Any` is too loose. We have a special variable that you can import. Try this out, and let me know.

Modify the import.

```python
from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE, to_unicode
```

Then:

```suggestion
        value: ICAL_TYPE,
```

stevepiercy (changes requested):
Thanks for the PR! Let me know what you think.

stevepiercy:
@arjunrawal1 since there are unresolved merge conflicts, and you haven't responded to my suggestions, I'm closing this PR as abandoned. Feel free to reopen and update your PR, and we'll review it. Thank you!

PR #1361, opened 2026-05-06
Add strict length validation to vDatetime.from_ical()

uwezkhan:
This PR fixes a low-severity parsing vulnerability in vDatetime.from_ical(). It now validates the minimum expected datetime string format before slicing components, preventing short malformed inputs from being parsed incorrectly


src/icalendar/prop/dt/datetime.py
```diff
@@ -6,11 +6,14 @@
...
 from icalendar.error import JCalParsingError
 from icalendar.parser import Parameters
+from icalendar.parser_tools import to_unicode
 from icalendar.timezone import tzp
 from icalendar.timezone.tzid import is_utc
 
 from .base import TimeBase
 
+__all__ = ["vDatetime"]
+
```
stevepiercy: Please revert this change, and restore it at the end of the file. It's unrelated to the purpose of this PR and is not necessary. Thank you!

src/icalendar/prop/dt/datetime.py
```diff
@@ -192,6 +199,3 @@ def from_jcal(cls, jcal_property: list) -> Self:
             dt,
             params=params,
         )
-
-
-__all__ = ["vDatetime"]
```
stevepiercy, line 197: Revert.

src/icalendar/tests/prop/test_vDatetime.py
```diff
@@ -29,6 +29,11 @@ def test_bad_ical():
         vDatetime.from_ical("20010101T000000A")
 
 
+def test_bad_ical_short_string():
```
stevepiercy: Would you please add a second test for the position of `T`?

src/icalendar/prop/dt/datetime.py
```diff
@@ -110,12 +113,16 @@ def to_ical(self):
...
     def from_ical(ical, timezone=None):
         """Create a datetime from the RFC string."""
+        ical = to_unicode(ical)
         tzinfo = None
         if isinstance(timezone, str):
             tzinfo = tzp.timezone(timezone)
         elif timezone is not None:
             tzinfo = timezone
 
+        if len(ical) < 15 or ical[8] != "T":
```
stevepiercy: This is not correct. There are three valid forms of the date-time value type. See https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.5. This check will fail on form #3.

- `19980118T230000`
- `19980119T070000Z`
- `TZID=America/New_York:19980119T020000`

CHANGES.rst
```diff
@@ -10,7 +10,7 @@ Change log
 Minor changes
 ~~~~~~~~~~~~~
 
-- ...
+- Cleaned up duplicate code and fixed import ordering in :file:`datetime.py`. :pr:`1361`
```
stevepiercy: There's no duplicate code that I can see. I see that you moved a couple of lines around. And this entry doesn't describe the purpose of this PR. I'd suggest this instead, depending on the outcome of other changes.

```suggestion
- Add strict length validation to :method:`vDatetime.from_ical() <icalendar.prop.dt.datetime.vDatetime.from_ical>`. :pr:`1361`
```

stevepiercy (changes requested):
Would you please address my comments? Thank you!

stevepiercy:
Would you also please resolve the merge conflicts? Thank you!

uwezkhan:
Thanks for the review and the detailed feedback.

I’ll revert the unrelated __all__/import-order changes, update the validation logic to correctly handle all valid RFC5545 DATE-TIME forms, add an additional test covering the T position, update the changelog entry to better describe the actual fix and resolve the merge conflicts.

src/icalendar/prop/dt/datetime.py
```diff
@@ -110,26 +111,34 @@ def to_ical(self):
...
     def from_ical(ical, timezone=None):
         """Create a datetime from the RFC string."""
+        ical = to_unicode(ical)
         tzinfo = None
         if isinstance(timezone, str):
             tzinfo = tzp.timezone(timezone)
         elif timezone is not None:
             tzinfo = timezone
 
+        # Extract the value part if parameters are present (Form #3)
```
stevepiercy:
```suggestion
        # Extract the value part if parameters are present per
        # https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.5
```

src/icalendar/prop/dt/datetime.py
```diff
@@ -110,26 +111,34 @@ def to_ical(self):
...
+        # Form #3: TZID=America/New_York:19980119T020000
+        ical_value = ical.rpartition(":")[2]
```
stevepiercy, line 124: Nice! I was today years old when I learned about rpartition.

src/icalendar/prop/dt/datetime.py
```diff
@@ -110,26 +111,34 @@ def to_ical(self):
...
+                int(ical_value[11:13]),  # minute
+                int(ical_value[13:15]),  # second
             )
             if tzinfo:
                 return tzp.localize(datetime(*timetuple), tzinfo)
-            if not ical[15:]:
+            if not ical_value[15:]:
                 return datetime(*timetuple)
-            if ical[15:16] == "Z":
+            if ical_value[15:16] == "Z":
```
stevepiercy: Since there's no 16th index value. It should always return either '' or 'Z', and if it doesn't, then the final `raise` should handle it.

```suggestion
            if ical_value[15:] == "Z":
```

news/1361.bugfix
```diff
@@ -0,0 +1 @@
+Strictly validate :class:`~icalendar.prop.dt.datetime.vDatetime` values in :meth:`~icalendar.prop.dt.datetime.vDatetime.from_ical` and reject malformed input. This also improves handling of values with a ``TZID`` prefix (Form #3). @uwezkhan
```
stevepiercy: Let's link to the RFC.

```suggestion
Strictly validate :class:`~icalendar.prop.dt.datetime.vDatetime` values in :meth:`~icalendar.prop.dt.datetime.vDatetime.from_ical` and reject malformed input. This also improves handling of values with a ``TZID`` prefix, per :rfc:`5545#section-3.3.5`, Form #3. @uwezkhan
```

stevepiercy (changes requested):
This is much better! I added a couple of suggestions to link to the RFC section, and a minor tweak for clarity. Also your branch needs to be updated against `main`. Would you please take care? Then I think it's ready to merge. Thank you!

stevepiercy (approved):
Thank you! Will merge when CI passes.

PR #1362, opened 2026-05-06
Avoid implicit file reads from string calendar input

uwezkhan:
This PR hardens `Component.from_ical()` by removing ambiguous string-path handling.

Previously, a newline-free `str` passed to `from_ical()` was treated as a possible filesystem path. If that path existed, the parser read the file contents automatically. This made it easy for applications that pass untrusted strings into `Calendar.from_ical()` / `Component.from_ical()` to accidentally allow local file reads or file-content disclosure through parser errors.

After this change:

- `Path` input still reads an iCalendar file explicitly.
- `bytes` and `str` input are treated only as iCalendar payload data.
- Regression tests verify string paths are not read and file contents are not leaked.
-


src/icalendar/cal/component.py
```diff
@@ -523,14 +525,6 @@ def from_ical(
...
         if isinstance(st, Path):
             st = st.read_bytes()
-        elif isinstance(st, str) and "\n" not in st and "\r" not in st:
-            path = Path(st)
-            try:
-                is_file = path.is_file()
-            except OSError:
-                is_file = False
-            if is_file:
-                st = path.read_bytes()
```
stevepiercy: I don't understand why you removed this. Can you explain?

I would expect that if the data is a `str` and doesn't have either a `\n` or `\r`, then it's intended to be a `Path`. The subsequent code ensures it's a file on the system. This feature allows passing in a `str` object, but this change removes that feature.

@SashankBhamidi or @angatha can you verify my interpretation?

If that's true, then do we want to remove this feature?

stevepiercy: @uwezkhan would you please respond to my question?

@SashankBhamidi @angatha or @niccokunzmann do you have any thoughts on this removal?

src/icalendar/tests/test_issue_756_read_calendar_from_file.py
```diff
@@ -28,12 +28,25 @@ def test_reading_cal_from_path(dummy_cal):
     assert actual_cal.to_ical() == expected_cal.to_ical()
 
 
-def test_reading_cal_from_string_path(dummy_cal):
-    """Test reading a calendar from a valid string path."""
-    expected_cal, path = dummy_cal
-    actual_cal = Calendar.from_ical(str(path))
```
stevepiercy, line 34: Why did you remove this test? This is a breaking change.

stevepiercy (changes requested):

src/icalendar/cal/component.py
uwezkhan: Fair question. To be clear up front: I'm not claiming this is an exploitable vulnerability, and I don't want to overstate it.

The reason I removed it is the ambiguity it introduces into from_ical's contract. Today a str argument means either "parse this iCal text" or "open this file and read it" — and which one happens depends on whether the string, with no \n/\r, happens to match an existing file on disk. So the same call can do completely different things depending on the filesystem state at runtime, including an unintended file read in code that only meant to parse data.

My aim was to make the two intents explicit: Path → read a file, str/bytes → parse as data. That keeps the #756 file-reading use case fully working, just non-implicitly.

That said — I recognize the implicit behavior was added deliberately for #756, so removing it is a real tradeoff, not an obvious win. If you all feel the convenience outweighs the ambiguity, I'm happy to drop the removal and instead just keep Path support documented as the recommended way. Your call.

src/icalendar/tests/test_issue_756_read_calendar_from_file.py
uwezkhan: You're right that it's a breaking change, which is why I removed the test rather than keeping it — it asserted the exact behavior the PR changes. I replaced it with tests covering the new contract (string is parsed as data).

If we keep the string-path behavior, the original test should obviously stay and I'll revert that. If we do go ahead with the change, I think it warrants a deprecation cycle: keep str-path reading with a DeprecationWarning for one release, then remove. I'd also move the news fragment out of .bugfix into a .breaking/deprecation entry to match. Whichever direction you prefer.

uwezkhan:
sorry for late reply

src/icalendar/cal/component.py
stevepiercy: @uwezkhan thanks for the explanation. I'd propose restoring the code and creating a new issue with your response to explain your concern. Then we can proceed with this PR.

We'll need @niccokunzmann to comment on that new issue to determine a path (ha! get it?) forward.

src/icalendar/tests/test_issue_756_read_calendar_from_file.py
stevepiercy: See previous comment.

niccokunzmann:
Valid ical always has newlines in it. I never saw a path that had them. What you can test instead is that the calendar paths on the file system that exist do not cause a file read.

Also, you can add a test that invalid paths on major operating systrms do not try to open a file but raise filenotfound. Would that also settle the matter?

stevepiercy:
@uwezkhan in my personal opinion, closing your own PR without comment harms your reputation and erodes my trust in your willingness to work with me. Why should I bother to volunteer my time, when you'll just throw it away? Please consider that when you contribute to open source software in the future.

uwezkhan:
@stevepiercy You're right, and I apologize. I shouldn't have closed the PR without leaving an explanation, especially after you and others had taken the time to review it.

The reason I closed it was that the discussion evolved and the narrower approach in #1444 addressed the issue, so I thought this broader PR was no longer the right direction. I realize now I should have explained that instead of closing it silently.

Thank you for taking the time to review my work and for pointing this out. I'll make sure to communicate clearly before closing or abandoning a PR in the future.

niccokunzmann:
Hi, this is a breaking change, bumping to icalendar version 8. 
I do not see the necessity to do that now. How do we handle that?
Also, it is not needed in my eyes. As I explained, either we have a newline or not in a string. Could you cite a few other sources that say that this approach is a cleaner design decision?

angatha:
Windows seems to forbid control characters in path names. However linux..

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-a
BEGIN:VEVENT
UID:a
DTSTAMP:19970701T100000Z
DTSTART:19970714T170000Z
DTEND:19970715T040000Z
SUMMARY:a
END:VEVENT
END:VCALENDAR
```
Or more specific
```python
Component.from_ical("BEGIN:VCALENDAR\r\nVERSION:2.0\r\nPRODID:-a\r\nBEGIN:VEVENT\r\nUID:a\r\nDTSTAMP:19970701T100000Z\r\nDTSTART:19970714T170000Z\r\nDTEND:19970715T040000Z\r\nSUMMARY:a\r\nEND:VEVENT\r\nEND:VCALENDAR")
```
parses but that string is a valid file name in linux.

However, who has such a file?

On the other hand, when the input is controled by some external user, he may pass a path as calendar. That may allow for reading local files so I see a point in removing the automatic string as path handling. Not because of an unintentionall read of a file but to harden parsing of untrusted data.

And for thouse who are interested, Windows can handle thouse paths even if you can not create one

niccokunzmann:
Thanks, that is a fair point.
Then, passing a path only to read data from the file system is perfect.

stevepiercy:
> @stevepiercy You're right, and I apologize. I shouldn't have closed the PR without leaving an explanation, especially after you and others had taken the time to review it.

Thank you, I appreciate it.

It looks like there was more discussion. Would you like to respond to any subsequent comments? Your opinion matters.

> Hi, this is a breaking change, bumping to icalendar version 8.
> I do not see the necessity to do that now. How do we handle that?

I think at this time, if a `string` is passed in instead of a `path`, then use a deprecation warning stating it will be removed in v8.

Also we use milestones, and add this PR to that milestone. I've created 7.3 and 8.0 milestones, and assigned this PR to 8.0. When we feel ready, we'll quickly find what should go into a v8 release.

uwezkhan:
Cleaned up the branch first: the merge with main pulled the implicit string-path read back in, so the code no longer matched the docstring or the new tests (they were failing). I re-dropped that branch and moved the news fragment to `removal`, since this is a breaking change rather than a bugfix.

On handling the break: keeping this on the 8.0 milestone as the hard removal makes sense to me, and Nicco's point about not breaking the 7.x line is fair. The clean migration is a DeprecationWarning on str-path reads for one 7.x release, then this removal in v8. I already had that deprecation in #1436; happy to reopen/redo it targeted at 7.3 so users get a release of warning before the behavior changes.

On why this is a cleaner contract, the precedent I'd point to is the stdlib splitting data-vs-file by API instead of sniffing the input: `json.loads` vs `json.load`, `tomllib.loads` vs `load`, `pickle.loads` vs `load`. None of them inspect the content to decide whether to touch the filesystem. Here `Path` means read a file and `str`/`bytes` means parse data, same split. As angatha said, the practical payoff is hardening: a `str` from untrusted input can't silently turn into a local file read.

src/icalendar/tests/test_issue_756_read_calendar_from_file.py
```diff
@@ -28,12 +28,25 @@ def test_reading_cal_from_path(dummy_cal):
...
+def test_string_path_is_parsed_as_calendar_data(dummy_cal):
+    """Test that string input is not interpreted as a filesystem path."""
+    _, path = dummy_cal
 
-    assert actual_cal.to_ical() == expected_cal.to_ical()
+    with pytest.raises(ValueError) as exc_info:
+        Calendar.from_ical(str(path))
+
+    assert "BEGIN:VCALENDAR" not in str(exc_info.value)
```
angatha: This is testing the same thing as the other test except is it not that obvious that the file contains the string `"BEGIN_VCALENDAR"`?

Another test would be to have a file that is named like a valid calendar but then the test is not cross platform.

uwezkhan: Good point, the two were checking the same thing and the BEGIN:VCALENDAR assertion leaned on the fixture's contents, which isn't obvious. Reworked both to test behavior instead of error strings, matching what Nicco asked for earlier:

- existing calendar file, path passed as str: asserts it raises rather than parsing. If the file were still read this would succeed and return that calendar, so the raise is what proves it isn't opened.
- non-existent path string: asserts ValueError, not FileNotFoundError, so a path that doesn't exist is parsed as data instead of triggering a file open.

Both are cross-platform since they never rely on a filename that looks like calendar data. Pushed in f7c35d0.

SashankBhamidi:
I haven't reviewed this PR at all, but while auditing the codebase I filed #1531 which tracks the same underlying issue. @niccokunzmann can you take a look at both and decide the direction here?

src/icalendar/cal/component.py
```diff
@@ -556,22 +558,6 @@ def from_ical(
         """
         if isinstance(st, Path):
             st = st.read_bytes()
-        elif isinstance(st, str) and "\n" not in st and "\r" not in st:
```
niccokunzmann, line 563: For now, this PR only removes reading from files. We can change this check to check for pathlib.Path.

niccokunzmann (changes requested):
The suggested type check is valid. We can also assume that icalendar files are read from external sources. As such blocking the file read based on a string check is important, a fix for more security.

Please do not remove all code to read from files but make it conditional on a pathlib.Path being the argument. Tests can cover that str cannot handle a path that exists.

src/icalendar/cal/component.py
uwezkhan: The branch already matches this. The `isinstance(st, Path)` check on the line above is kept, so reading from files stays in, it's just conditional on the type now:

- `Path`: reads the file, unchanged (`test_reading_cal_from_path` still passes).
- `str` / `bytes`: always parsed as calendar data, never checked against the filesystem.

The only removal is the elif that probed newline-free strings as paths. `test_existing_calendar_path_as_string_is_not_read` covers the case from your review: a str naming a real calendar file raises ValueError instead of returning that calendar, which shows the file is never opened.

The tradeoff is still the compat break for anyone passing str paths today, which is why this sits on the 8.0 milestone, with the 7.x DeprecationWarning planned separately.

niccokunzmann (dismissed):
Thanks, I think, this works!

niccokunzmann:
@stevepiercy @SashankBhamidi , I would treat this as a security fix. The change has probably almost no effect on anyone.
The code is valid. Is there anything, we need to do for this to get merged?

stevepiercy:
> @stevepiercy @SashankBhamidi , I would treat this as a security fix.

FTR, it wasn't disclosed properly, per our [icalendar's security policy](https://github.com/collective/icalendar/blob/main/SECURITY.md). It should have been from the start.

For @uwezkhan, in the future, please follow our security policy. If you're not certain that something may be a security issue, then err on the side of caution and submit a security report. The icalendar team will investigate and let you know the next steps.

Nonetheless, I think it's appropriate to follow the [security advisory workflow](https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/fix-reported-vulnerabilities/create-repository-advisory). GitHub doesn't support converting an open PR to a security advisory, so I'm not sure of the exact process. Anyone not on vacation can do that. 1-2-3, not it! 😉

> The change has probably almost no effect on anyone.

There's no way to know for certain whether someone uses this feature. The only data point we have is that no one has reported it as an issue until this PR, and that's insufficient data. We don't even know who uses icalendar! Regardless of usage, icalendar follows semantic versioning. When a feature is removed, as this PR does, then that's a breaking change and requires a full version bump to 8.0.0.

news/1362.removal
```diff
@@ -0,0 +1 @@
+Passing a file path as a ``str`` to :meth:`~icalendar.cal.component.Component.from_ical` no longer reads that file implicitly. ``str`` and ``bytes`` are always parsed as iCalendar data; use :class:`pathlib.Path` to read from a file. This hardens parsing of untrusted input against unintended local file reads. @uwezkhan
```
stevepiercy:
```suggestion
Passing a file path as a ``str`` to :meth:`~icalendar.cal.component.Component.from_ical` no longer reads that file implicitly. ``str`` and ``bytes`` values are now parsed as iCalendar data only. Instead, use :class:`pathlib.Path` to read data from a file. This hardens parsing of untrusted input against unintended local file reads. @uwezkhan
```

stevepiercy (changes requested):
Tweak the news.

news/1362.removal
SashankBhamidi: Steve's suggestion from the previous review round is still open and makes the wording cleaner.

SashankBhamidi (dismissed):
Thanks for sticking with this one through the review cycle @uwezkhan 

The rationale you laid out (hardening against untrusted input, matching the `json.loads`/`json.load` split from stdlib) is solid and the implementation is clean. 

One thing to note inline on the news fragment before this goes in.

news/1362.removal
uwezkhan: Applied your wording in 34f66e0.

uwezkhan:
Applied Steve's suggested wording to the news fragment, pushed in 34f66e0.

On the security policy: understood. I wasn't sure this rose to a reportable issue since it needs an attacker-controlled string that matches an existing path, but that's exactly the borderline case where SECURITY.md says to err toward a private report. I'll do that going forward.

On semver, no disagreement. It's a removal, so 8.0.0, which is where the milestone already has it.

On the AI note: I do use an LLM a bit for research and sanity checking, nothing heavy. Noted on the policy.

stevepiercy (dismissed):
Although I approve, I think we should put this PR through the security report process to give it appropriate visibility.

We also should wait to merge it until we're ready to bump a whole version to 8.0.0.

@SashankBhamidi @niccokunzmann how do you want to proceed?

niccokunzmann:
It is added to the milestone.
The code change is small. We should not ecpect merge conflicts.
So, yes, we can hold off merging until the release 8.0 is ready.

Thank you for the work on the PR! We will merge it to release 8.0.

uwezkhan:
Holding for 8.0 works for me. The branch is small and self-contained, so it should rebase cleanly whenever the release window opens, and if you decide to route it through the advisory process first I can help with the writeup.

stevepiercy:
@uwezkhan would you please initiate a security report? I recently updated our security policy. It would be good practice to follow it. See https://github.com/collective/icalendar/blob/main/SECURITY.md. Thank you!

uwezkhan:
Filed it: https://github.com/collective/icalendar/security/advisories/GHSA-j72g-xg6j-fq68

Scoped the report to the two ways content comes back: the full read when the target parses as a calendar, and the error message echoing file content when it does not, which is the half Sashank tracked in #1531. Rated it low, since it needs the calling application to pass an attacker-influenced `str`. Affected range is 7.0.2 through 7.2.2, since the str-path read landed with #756 in 7.0.2.

No change to the branch, it is still the commit you approved. One note from the report worth repeating here: if you want something in the 7.x line ahead of 8.0, the smaller option is to keep the read but put the path in the error message instead of the file content. That closes #1531 without changing the accepted types, and leaves this PR as the full removal for 8.0. Happy to do that as a separate PR if it is useful.

angatha:
Does anyone keep track of this? I do not have access to the advisory and want to make sure it is not lost.

stevepiercy:
> Does anyone keep track of this? I do not have access to the advisory and want to make sure it is not lost.

@angatha Maintainers and Owners can view it. I just now granted you access to it, since you're a trusted user already with Write access. Can you verify?

niccokunzmann:
@SashankBhamidi There is a requested change from you that would block auto-merge. Can you have a look?
@uwezkhan we will need an entry in the upgrade guide for readers who upgrade to v8.
This can best the introduced without merge conflict after #1700. Would you like to write this? It does not need to be included in this PR.

uwezkhan:
Done. Merged main now that #1700 is in and added the entry under the 8.0.0 breaking changes section in this PR, df103a8. It follows the 7.0.0 entries: what changed, why, then a before (str path) and after (pathlib.Path) code block. Diff against main is still the same three files plus upgrade.rst.

For Sashank's requested change: his only inline point was Steve's news wording, which is applied in 34f66e0, so it just needs a re-review to clear the block.

stevepiercy (dismissed):
Nice upgrade guide entry! I approve.

stevepiercy:
https://github.com/collective/icalendar/actions/runs/32461821256/job/96710104947?pr=1362#step:8:98

>  docs/how-to/upgrade.rst
> 30:15  error    Did you really mean 'untrusted'?                   Vale.Spelling 

Yes, vale, we did. 🙄

@uwezkhan would you please add the term per https://icalendar.readthedocs.io/en/stable/contribute/documentation/build-check.html#vale-configuration? Thank you!

niccokunzmann:
@stevepiercy Please merge it after the docs changes. We reviewed the code already. Branch `main` will continue to work towards an 8.0.0 release.

uwezkhan:
Done in 6d5087a. Put `untrusted` in `Base/accept.txt` rather than the icalendar vocab since it's a generic spelling, which is where the build-check docs say those go. `make vale` on `docs/how-to/upgrade.rst` is clean now.

stevepiercy (dismissed):
Thanks for checking the correct location for the term.

stevepiercy:
@niccokunzmann I'm hesitant to merge this because I don't know your release plans. Do you plan 7.3.1, 7.4.0, or 8.0.0?

For anything that won't make it to 8.0.0, but gets merged to `main`, should we backport to the `7.x` branch?

Looking for clarity and direction here.

stevepiercy:
## DON'T MERGE!

We haven't dealt with the proposed but not yet accepted GHSA. Will post link in Matrix chat.

niccokunzmann:
`main` from now on is for 8.x. We can just merge things.

I do not do the work to backport it. It breaks the API and bumps to the major version. This is a nice fix.  There should not be a problem with this.

A backport could just affect the docs and add a warning when str is passed and that code path is being used. I don't know what the right way is to approach this. I would not want to wait with shipping a fix to 8.0. 

The communication path is not clear for me or what we are waiting for.
What is the harm of merging this now?

uwezkhan:
Updated the branch against main since it had fallen behind. The diff is unchanged: str/bytes parse as data, Path reads the file, and the regression tests still pass.

On what's blocking: everything in GHSA-j72g-xg6j-fq68 has been public in this PR since May, so merging doesn't disclose anything new. The advisory can be finalized and published whenever the fixed release ships, which is the usual order anyway. From my side nothing depends on the advisory landing first, but I'll defer to Steve on the process hold.

If a 7.x backport is still wanted, the smaller option from #1436 stands: keep the str read there, add a DeprecationWarning, and let 8.0 carry the removal. I can put that up as a separate PR against the 7.x branch if useful.

stevepiercy:
> The communication path is not clear for me or what we are waiting for.
> What is the harm of merging this now?

First a maintainer must request a CVE and communicate the fact. GitHub is many weeks behind in issuing CVEs, but we can proceed without one being issued. You might also include in what version this will be released, which is why a maintainer needs to do this step.

Otherwise there is nothing else blocking the merge of this PR.

uwezkhan:
Added the AI use disclosure to the news fragment in 21aa90a, with the same sentence the other entries in CHANGES.rst use, since the responsible AI policy Sashank linked asks for it in the change log entry. Only the news file changed, so the code is what was already reviewed, but the push dismissed the approval and it needs one more click. Nothing is pending on my side for the advisory. The CVE request is the maintainer step Steve described.

niccokunzmann (approved):

niccokunzmann:
@stevepiercy Thanks for the clarification. The CVE is requested. I will take a last look.
@uwezkhan Thanks for the update. This is ready for merge. I hope it happens automatically.

PR #1366, opened 2026-05-07
Add ical_value property to vFrequency, vMonth, and vInline

xxiaoxiong:
This PR adds the `ical_value` property to three more value type classes as part of Issue #876.

### Added `ical_value` property to:
1. **vFrequency** (src/icalendar/prop/recur/frequency.py)
   - Returns `str(self)` (frequency string)
   - Type hint: `str`
   - RFC reference: :rfc:`5545#section-3.3.10`
   - Example: `vFrequency("DAILY").ical_value == "DAILY"`

2. **vMonth** (src/icalendar/prop/recur/month.py)
   - Returns `int(self)` (month number)
   - Type hint: `int`
   - RFC references: :rfc:`5545#section-3.3.10` and :rfc:`7529`
   - Example: `vMonth(1).ical_value == 1`, `vMonth("5L").ical_value == 5`

3. **vInline** (src/icalendar/prop/inline.py)
   - Returns `str(self)` (unparsed text)
   - Type hint: `str`
   - Example: `vInline("text").ical_value == "text"`

### Tests
- Created `test_vFrequency.py` with ical_value test
- Created `test_vMonth.py` with ical_value test (including leap month case)
- Created `test_vInline.py` with ical_value test
- All tests pass: 9447 passed

## Implementation Details
- All properties use `@property` decorator
- Include type hints for return values
- Include docstrings with RFC references where applicable
- Follow the same pattern as existing implementations (vBoolean, vInt, vFloat)
- These classes inherit from basic Python types (str, int), so ical_value returns the converted value

Contributes to #876

### DIFF
--- a/src/icalendar/prop/inline.py
+++ b/src/icalendar/prop/inline.py
@@ -29,6 +29,11 @@ class vInline(str):
     def to_ical(self) -> bytes:
         return self.encode(DEFAULT_ENCODING)
 
+    @property
+    def ical_value(self) -> str:
+        """Inline value type (unparsed text)"""
+        return str(self)
+
     @classmethod
     def from_ical(cls, ical: ICAL_TYPE) -> Self:
         return cls(ical)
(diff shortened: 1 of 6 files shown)


xxiaoxiong:
Closing this PR due to quality issues discovered during review
Sorry for the noise. Will do better next time.

Issue #1158, opened 2026-02-08
[Beginner] Correct broken links to classes & functions in documentation


niccokunzmann:
Choose a documentation file and correct all the links to Python objects.

We have a lot of documentation files. Sometimes, we link to components, functions, and other objects and the links don't work.

- Working link: <img width="56" height="27" alt="Image" src="https://github.com/user-attachments/assets/9ce10aa2-45c4-4c70-9b8d-d7211b3026cd" />
- Broken link: <img width="56" height="27" alt="Image" src="https://github.com/user-attachments/assets/70fd1921-8710-4387-bab2-343dc79d4fe3" />

Please read through as many files as you like and do the following:

- If the file has no broken links, comment below that it is perfect! Thanks!
- If the file contains one or more broken links, please fix them (see Guidance section below).
- If the file contains a prefix of `:py` on to the Python object, remove it.
- Verify that the links that you fix are actually fixed by viewing the pull request preview that Read the Docs will generate after about one or two minutes. The link should appear in the pull request description automatically. If not, then visit https://app.readthedocs.org/projects/icalendar/builds/ to find your pull request's preview build, click it, and click the link to preview the docs. Also edit your PR description to include the link if it doesn't appear. This makes your PR easier to review.

## Guidance

(GitHub only) For some files, you can [Make small edits via GitHub](https://icalendar.readthedocs.io/en/latest/contribute/documentation/small-edits.html).

(Local setup) For the [API Reference in `/reference/`](https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.html), you need to find the file in [src/icalendar](https://github.com/collective/icalendar/blob/main/src/icalendar/) and [build the documentation on your computer](https://icalendar.readthedocs.io/en/latest/contribute/documentation/build-check.html). See also #1072 for important details of how to hyperlink Python objects in docstrings.

Have a look at the [different styles of links in the documentation](https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#cross-reference-python-objects).

Additionally, we value changes to the documentation to make it easier to understand for first-time users!

## Checklist

All of these files need checking before closing the issue. If you create your first pull request, please only modify one file!

- [ ] [404.html](https://icalendar.readthedocs.io/en/latest/404.html)
- [ ] [about.html](https://icalendar.readthedocs.io/en/latest/about.html)
- [ ] [api.html](https://icalendar.readthedocs.io/en/latest/api.html)
[...]

PR #1384, opened 2026-05-13
Fix Python object links in usage docs

kyriakikrimitza:
Fixes Python object links in docs/how-to/usage.rst as part of #1158.

----
📚 Documentation preview 📚: https://icalendar--1384.org.readthedocs.build/

### DIFF
--- a/docs/how-to/usage.rst
+++ b/docs/how-to/usage.rst
@@ -32,10 +32,10 @@ The following example shows how to set two properties for it, then display them.
 
 .. note::
 
-    To add components to the calendar, create the :ref:`subcomponent <subcomponents>`, then add it via :meth:`icalendar.cal.component.Component.add`.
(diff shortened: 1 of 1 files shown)


stevepiercy:
@kyriakikrimitza thank you for your effort, but the links you changed already work as is. You don't need to append `()`, as Sphinx will do that as needed automatically. `~` is used deliberately to show only the method name instead of the full dotted path to the Python object. Finally, all pull requests also require a change log entry. I'd suggest that you first check whether your changes actually fix something or are an improvement.

If you find things that are broken or that you can improve, we'd welcome your contribution.

kyriakikrimitza:
@stevepiercy  Thank you for the clarification and feedback!
I misunderstood the intended use of `~` in Sphinx references and assumed the fully qualified paths were preferred in all cases. I also did not realize that Sphinx automatically renders method parentheses.

This was very helpful for understanding the project's documentation conventions and contribution expectations. I’ll look into smaller functional improvements or tests that address a real issue before submitting another PR.

PR #1385, opened 2026-05-14
docs: replace RFC copy-paste docstrings with Pythonic docstrings

ykd007:
See #1244

Rewrites RFC copy-paste docstrings across `Alarm`, `Component`, and `Calendar` into clean Pythonic format with concise summaries, examples, and see-also references.

**Updated (12 properties across 3 files):**
- `Alarm`: `ACKNOWLEDGED`, `TRIGGER`
- `Component`: `DTSTAMP`/`stamp`, `LAST_MODIFIED`, `CREATED`
- `Calendar`: `calendar_name`, `description`, `color`, `prodid`, `version`, `calscale`, `method`, `refresh_interval`

Validation:
- `make format`
- `uv run pytest src/icalendar/tests/test_with_doctest.py`
- `make html`


stevepiercy:
@ykd007 would you please resolve the merge conflicts by moving your change log entry into the new format that we adopted in 7.1.2, as documented at https://icalendar.readthedocs.io/en/stable/contribute/index.html#change-log-entry-format?

src/icalendar/cal/alarm.py
```diff
@@ -156,32 +156,29 @@ class Alarm(Component):
...
-    When an alarm is triggered on a client, clients can check to see if an "ACKNOWLEDGED"
-    property is present. If it is, and the value of that property is greater than or
-    equal to the computed trigger time for the alarm, then the client SHOULD NOT trigger
-    the alarm. Similarly, if an alarm has been triggered and
-    an "alert" has been presented to a calendar user, clients can monitor
-    the iCalendar data to determine whether an "ACKNOWLEDGED" property is added or
-    changed in the alarm component. If the value of any "ACKNOWLEDGED" property
-    in the alarm changes and is greater than or equal to the trigger time of the alarm,
-    then clients SHOULD dismiss or cancel any "alert" presented to the calendar user.
+        """The UTC datetime at which this alarm was last sent or acknowledged.
```
stevepiercy: When I looked at the [rendered docs](), I found that there was a snippet prepended to this docstring. Would you please revise the following:

https://github.com/stevepiercy/icalendar/blob/8c26d2b22ce33c37bd2bfce7997ca240641236ab/src/icalendar/attr.py#L401-L415

Specifically, and possibly wrapped after using `make format`:

```diff
-        f"""The {name} property. datetime in UTC
+        f"""The {name} property with all values converted to a :class:`~datetime.datetime` in UTC.

-    All values will be converted to a datetime in UTC.
+ 
```

The two blank lines will ensure proper concatenation with a new paragraph to begin the description of the docstring.


```suggestion
        """This property is the UTC datetime at which this alarm was last sent or acknowledged as defined in :rfc:`9074`.
```

src/icalendar/cal/alarm.py
```diff
@@ -156,32 +156,29 @@ class Alarm(Component):
...
+
+    Defined in :rfc:`9074`. Setting this property allows calendar clients to
```
stevepiercy: Moved into previous paragraph.

```suggestion
    Setting this property allows calendar clients to
```

src/icalendar/cal/alarm.py
```diff
@@ -156,32 +156,29 @@ class Alarm(Component):
...
+    dismiss or suppress an alarm across multiple devices. Once set to a value
+    greater than or equal to the alarm's computed trigger time, conforming clients
+    will not re-fire the alarm.
```
```suggestion
    will not refire the alarm.
```

src/icalendar/cal/alarm.py
```diff
@@ -156,32 +156,29 @@ class Alarm(Component):
...
+    Example:
+        Mark an alarm as acknowledged at the current time.
+
+        .. code-block:: pycon
+
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Alarm
+            >>> alarm = Alarm()
+            >>> alarm.ACKNOWLEDGED = datetime(2024, 1, 15, 10, 0, 0, tzinfo=timezone.utc)
+            >>> alarm.ACKNOWLEDGED  # doctest: +ELLIPSIS
```
stevepiercy: Inline comments don't render. Is the comment necessary?

```suggestion
            >>> alarm.ACKNOWLEDGED
```

src/icalendar/cal/alarm.py
```diff
@@ -156,32 +156,29 @@ class Alarm(Component):
...
+            datetime.datetime(2024, 1, 15, 10, 0, tzinfo=...)
+
+    See also:
+        :attr:`TRIGGER` — the time at which the alarm fires.
```
```suggestion
        :attr:`TRIGGER`, the time at which the alarm fires.
```

src/icalendar/cal/alarm.py
```diff
@@ -190,17 +187,33 @@ class Alarm(Component):
...
+            >>> from datetime import timedelta
+            >>> from icalendar import Alarm
+            >>> alarm = Alarm()
+            >>> alarm.TRIGGER = timedelta(minutes=-15)
+            >>> alarm.TRIGGER  # doctest: +ELLIPSIS
+            datetime.timedelta(...)
+
+    See also:
+        :attr:`TRIGGER_RELATED`, :attr:`DURATION`, :attr:`REPEAT`
+    """,
```
stevepiercy, line 235: This docstring is almost perfect.

However, the property has documentation that gets prepended and appended, as defined in https://github.com/stevepiercy/icalendar/blob/8c26d2b22ce33c37bd2bfce7997ca240641236ab/src/icalendar/attr.py#L958-L966.

Let's clean that up, too.

```python
    p_doc = f"""The {prop} property.

    {doc}

    You can also delete the value with ``del`` or by setting it to ``None``.

    Parameters:
        {", ".join(t.__name__ for t in value_type)}

    Raises:
        InvalidCalendar: if the attribute has invalid values.

    Returns:
        A :class:`~datetime.datetime` or :class:`~datetime.timedelta` if the value is valid. ``None`` if the value is absent.
    """
```

src/icalendar/cal/calendar.py
```diff
@@ -256,30 +256,16 @@ def add_missing_timezones(
     calendar_name = multi_language_text_property(
         "NAME",
         "X-WR-CALNAME",
-        """This property specifies the name of the calendar.
+        """The display name of this calendar, per :rfc:`7986#section-5.3`.
```
stevepiercy: Please double-check the link.

```suggestion
        """The display name of this calendar, per :rfc:`7986#section-5.1`.
```

src/icalendar/cal/calendar.py
```diff
@@ -256,30 +256,16 @@ def add_missing_timezones(
...
 
-    This implements :rfc:`7986` ``NAME`` and ``X-WR-CALNAME``.
+    Implements both the ``NAME`` property (from :rfc:`7986`) and the widely-used
+    ``X-WR-CALNAME`` extension for broader client compatibility.
```
```suggestion
    Implements both the ``NAME`` property from :rfc:`7986#section-5.3` and the widely used
    ``X-WR-CALNAME`` extension for broader client compatibility.
```

src/icalendar/cal/calendar.py
stevepiercy: This was broken. I'm surprised it passed tests before.

```suggestion
            >>> print(calendar.to_ical().decode())
```

src/icalendar/cal/calendar.py
```diff
@@ -291,31 +277,25 @@ def add_missing_timezones(
...
-        """This property specifies the description of the calendar.
-
-    This implements :rfc:`7986` ``DESCRIPTION`` and ``X-WR-CALDESC``.
+        """A human-readable description of this calendar's content, per :rfc:`7986#section-5.2`.
 
-    Conformance:
-        This property can be specified multiple times in an
-        iCalendar object.  However, each property MUST represent the
-        description of the calendar in a different language.
+    Implements both ``DESCRIPTION`` (from :rfc:`7986`) and ``X-WR-CALDESC``
```
```suggestion
    Implements both ``DESCRIPTION`` from :rfc:`7986#section-5.2` and ``X-WR-CALDESC``
```

src/icalendar/cal/calendar.py
```diff
@@ -327,34 +307,25 @@ def add_missing_timezones(
...
-        """This property specifies a color used for displaying the calendar.
-
-    This implements :rfc:`7986` ``COLOR`` and ``X-APPLE-CALENDAR-COLOR``.
-    Please note that since :rfc:`7986`, subcomponents can have their own color.
+        """A CSS3 color name or value used to visually distinguish this calendar, per :rfc:`7986#section-5.9`.
 
-    Property Parameters:
-        IANA and non-standard property parameters can
-        be specified on this property.
+    Implements both ``COLOR`` (from :rfc:`7986`) and ``X-APPLE-CALENDAR-COLOR``.
```
```suggestion
    Implements both ``COLOR`` from :rfc:`7986#section-5.9` and ``X-APPLE-CALENDAR-COLOR``.
```

src/icalendar/cal/calendar.py
```diff
@@ -327,34 +307,25 @@ def add_missing_timezones(
...
+    The value is a case-insensitive CSS3 color name (e.g. ``"turquoise"``) or
+    a hex code (e.g. ``"#ffffff"``), drawn from the
+    `CSS3 color specification <https://www.w3.org/TR/css-color-3/>`_.
```
stevepiercy: Don't use Latin abbreviations per https://learn.microsoft.com/en-us/style-guide/word-choice/use-us-spelling-avoid-non-english-words.

```suggestion
    The value is a case-insensitive CSS3 color name, for example, ``"turquoise"``, or
    a hex code, for example, ``"#ffffff"``, drawn from the
    `CSS3 color specification <https://www.w3.org/TR/css-color-3/>`_.
```

src/icalendar/cal/calendar.py
```suggestion
            >>> print(calendar.to_ical().decode())
```

src/icalendar/cal/calendar.py
```diff
@@ -365,118 +336,137 @@ def add_missing_timezones(
...
     categories = categories_property
     uid = uid_property
     prodid = single_string_property(
         "PRODID",
-        """PRODID specifies the identifier for the product that created the iCalendar object.
+        """The product identifier for the software that created this iCalendar object.
 
-Conformance:
-    The property MUST be specified once in an iCalendar object.
+Defined in :rfc:`5545#section-3.7.3` and required exactly once per calendar object.
```
```suggestion
This property is defined in :rfc:`5545#section-3.7.3`.
It's required exactly once per calendar object.
```

src/icalendar/cal/calendar.py
```diff
@@ -365,118 +336,137 @@ def add_missing_timezones(
...
-    The vendor of the implementation SHOULD assure that
-    this is a globally unique identifier; using some technique such as
-    an FPI value, as defined in [ISO.9070.1991].
-
-    This property SHOULD NOT be used to alter the interpretation of an
-    iCalendar object beyond the semantics specified in this memo.  For
-    example, it is not to be used to further the understanding of non-
-    standard properties.
+The value should be a globally unique string. The conventional format is a
+Formal Public Identifier (FPI), e.g. ``-//My Company//My Product//EN``, but any
```
```suggestion
Formal Public Identifier (FPI), for example, ``-//My Company//My Product//EN``, but any
```

src/icalendar/cal/calendar.py
```diff
@@ -365,118 +336,137 @@ def add_missing_timezones(
...
+    :attr:`version`
 """,
     )
     version = single_string_property(
         "VERSION",
-        """VERSION of the calendar specification.
+        """The iCalendar specification version required to interpret this object.
 
-The default is ``"2.0"`` for :rfc:`5545`.
+Defined in :rfc:`5545#section-3.7.4` and required exactly once per calendar object.
```
```suggestion
This property is defined in :rfc:`5545#section-3.7.4`.
It's required exactly once per calendar object.
```

src/icalendar/cal/calendar.py
```diff
@@ -365,118 +336,137 @@ def add_missing_timezones(
...
-
-Description:
-    This memo is based on the Gregorian calendar scale.
-    The Gregorian calendar scale is assumed if this property is not
-    specified in the iCalendar object.  It is expected that other
-    calendar scales will be defined in other specifications or by
-    future versions of this memo.
+        """The calendar scale for date and time values in this iCalendar object.
+
+Defined in :rfc:`5545#section-3.7.1`. The only value currently defined is
```
```suggestion
This property is defined in :rfc:`5545#section-3.7.1`. The only value currently defined is
```

src/icalendar/cal/calendar.py
```diff
@@ -365,118 +336,137 @@ def add_missing_timezones(
...
-
-    If this property is not present in the iCalendar object, then a
-    scheduling transaction MUST NOT be assumed.  In such cases, the
-    iCalendar object is merely being used to transport a snapshot of
-    some calendar information; without the intention of conveying a
-    scheduling semantic.
+        """The iTIP scheduling method associated with this calendar object, per :rfc:`5545#section-3.7.2`.
+
+When present, ``METHOD`` indicates that this object is part of a scheduling
+transaction (e.g., a meeting invitation or cancellation). Scheduling methods
```
```suggestion
transaction, such as a meeting invitation or cancellation. Scheduling methods
```

src/icalendar/cal/calendar.py
```diff
@@ -365,118 +336,137 @@ def add_missing_timezones(
...
-
-        Description:
-            This property specifies a positive duration that gives
-            a suggested minimum polling interval for checking for updates to
-            the calendar data.  The value of this property SHOULD be used by
-            calendar user agents to limit the polling interval for calendar
-            data updates to the minimum interval specified.
+        Returns a :class:`~datetime.timedelta` or ``None`` when not set.
+        Calendar clients should not poll more frequently than this interval.
+        The value must be a positive duration.
```
```suggestion
        Calendar clients should not poll more frequently than this interval.
        The value must be a positive duration.

        Returns:
            A :class:`~datetime.timedelta`, or ``None`` when not set.
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
-        property, this property is equivalent to the "LAST-MODIFIED"
-        property.
+    Example:
+        .. code-block:: pycon
+
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Event
+            >>> event = Event()
+            >>> event.DTSTAMP = datetime(2024, 6, 1, 12, 0, 0, tzinfo=timezone.utc)
+            >>> event.DTSTAMP  # doctest: +ELLIPSIS
```
stevepiercy: Same.

```suggestion
            >>> event.DTSTAMP
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
+    Example:
+        .. code-block:: pycon
 
-        Note: This is analogous to the modification date and time for a
-        file in the file system.
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Event
+            >>> event = Event()
+            >>> event.LAST_MODIFIED = datetime(2024, 6, 1, 9, 0, 0, tzinfo=timezone.utc)
+            >>> event.LAST_MODIFIED  # doctest: +ELLIPSIS
```
stevepiercy: Same.

```suggestion
            >>> event.LAST_MODIFIED
```

src/icalendar/cal/component.py
```diff
@@ -704,17 +718,28 @@ def _utc_now():
...
 
-        """,
+    Example:
+        .. code-block:: pycon
+
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Event
+            >>> event = Event()
+            >>> event.CREATED = datetime(2024, 1, 1, 8, 0, 0, tzinfo=timezone.utc)
+            >>> event.CREATED  # doctest: +ELLIPSIS
```
```suggestion
            >>> event.CREATED
```

src/icalendar/cal/alarm.py
```diff
@@ -190,17 +187,33 @@ class Alarm(Component):
...
+    Example:
+        Set an alarm to fire 15 minutes before the start of an event.
+
+        .. code-block:: pycon
+
+            >>> from datetime import timedelta
+            >>> from icalendar import Alarm
+            >>> alarm = Alarm()
+            >>> alarm.TRIGGER = timedelta(minutes=-15)
+            >>> alarm.TRIGGER  # doctest: +ELLIPSIS
```
```suggestion
            >>> alarm.TRIGGER
```

stevepiercy:
@ykd007 would you also edit your PR description from `Closes #1244` to `See #1244`? We don't want to close it until all the docstrings are checked off the list.

CHANGES.rst
```diff
@@ -11,6 +11,7 @@ Minor changes
 ~~~~~~~~~~~~~
 
 - Bump PyPy from 3.10 to 3.11 for testing. :pr:`1383`
+- Replaced RFC copy-paste text in docstrings with Pythonic docstrings for calendar components. :pr:`1385`
```
stevepiercy: Please move this to `/news/1244.documentation`?

```suggestion
- Replaced RFC copy-paste text in docstrings with Pythonic docstrings for calendar components. :issue:`1244`
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
 
     DTSTAMP = stamp = single_utc_property(
         "DTSTAMP",
-        """RFC 5545:
+        """The UTC datetime stamp recording when this component instance was created or last revised.
```
stevepiercy, line 732: See other note about the prepending of text to your docstring. Let's confirm it looks good after editing the parent object.

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
 
-        Conformance:  This property MUST be included in the "VEVENT",
-        "VTODO", "VJOURNAL", or "VFREEBUSY" calendar components.
+    Defined in :rfc:`5545#section-3.8.7.2` and required in
+    ``VEVENT``, ``VTODO``, ``VJOURNAL``, and ``VFREEBUSY`` components.
```
```suggestion
    This property is defined in :rfc:`5545#section-3.8.7.2`. It's required
    in ``VEVENT``, ``VTODO``, ``VJOURNAL``, and ``VFREEBUSY`` components.
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
-        Description: In the case of an iCalendar object that specifies a
-        "METHOD" property, this property specifies the date and time that
-        the instance of the iCalendar object was created.  In the case of
-        an iCalendar object that doesn't specify a "METHOD" property, this
-        property specifies the date and time that the information
-        associated with the calendar component was last revised in the
-        calendar store.
+    When the calendar object carries a ``METHOD`` property (e.g., for
+    scheduling), this value is the creation time of *this particular revision*.
+    Without a ``METHOD`` property it is equivalent to :attr:`LAST_MODIFIED`.
```
```suggestion
    When the calendar object carries a ``METHOD`` property, such as for
    scheduling, this value is the creation time of *this particular revision*.
    Without a ``METHOD`` property, it's equivalent to :attr:`LAST_MODIFIED`.
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
 
-        The value MUST be specified in the UTC time format.
+    The value is always in UTC. Also accessible as :attr:`stamp`.
```
```suggestion
    The value is always in UTC. It's also accessible as :attr:`stamp`.
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
     LAST_MODIFIED = single_utc_property(
         "LAST-MODIFIED",
-        """RFC 5545:
+        """The UTC datetime when this component's information was last revised, per :rfc:`5545#section-3.8.7.3`.
 
-        Purpose:  This property specifies the date and time that the
-        information associated with the calendar component was last
-        revised in the calendar store.
+    Analogous to a file's modification timestamp. This property is optional;
+    when absent, :attr:`last_modified` falls back to :attr:`DTSTAMP`.
```
```suggestion
    It's analogous to a file's modification timestamp. This property is optional.
    When it's absent, :attr:`last_modified` falls back to :attr:`DTSTAMP`.
```

src/icalendar/cal/component.py
```diff
@@ -616,39 +616,53 @@ def __eq__(self, other):
...
+
+    Applicable to ``VEVENT``, ``VTODO``, ``VJOURNAL``, and ``VTIMEZONE``
```
```suggestion
    This property is applicable to ``VEVENT``, ``VTODO``, ``VJOURNAL``, and ``VTIMEZONE``
```

src/icalendar/cal/component.py
```diff
@@ -704,17 +718,28 @@ def _utc_now():
 
     CREATED = single_utc_property(
         "CREATED",
-        """
-        CREATED specifies the date and time that the calendar
-        information was created by the calendar user agent in the calendar
-        store.
+        """The UTC datetime when this calendar component was first created, per :rfc:`5545#section-3.8.7.1`.
+
+    Records when the calendar user agent originally stored the component.
```
```suggestion
    This property records when the calendar user agent originally stored the component.
```

src/icalendar/cal/component.py
```diff
@@ -704,17 +718,28 @@ def _utc_now():
...
+    This property is optional; when absent, :attr:`created` falls back to
```
```suggestion
    This property is optional. When it's absent, :attr:`created` falls back to
```

src/icalendar/cal/component.py
```diff
@@ -704,17 +718,28 @@ def _utc_now():
...
+    :attr:`DTSTAMP`.
 
-        Conformance:
-            The property can be specified once in "VEVENT",
-            "VTODO", or "VJOURNAL" calendar components.  The value MUST be
-            specified as a date with UTC time.
+    Applicable to ``VEVENT``, ``VTODO``, and ``VJOURNAL`` components.
```
```suggestion
    This property is applicable to ``VEVENT``, ``VTODO``, and ``VJOURNAL`` components.
```

stevepiercy (changes requested):
This is phenomenal work! Thank you!

There are some weird prepended and appended docstrings from a parent method that I'd like to see added to this PR, which will improve other docstrings automatically.

The change log needs to be moved, per the release of 7.1.1, and described in https://icalendar.readthedocs.io/en/stable/contribute/index.html#change-log-entry-format.

Other than that, I pretty much had only grammar and style fixes.

stevepiercy:
@ykd007 in addition to addressing my comments, would you please:

- Follow our new change log format, as described in https://icalendar.readthedocs.io/en/stable/contribute/index.html#change-log-entry-format?
- Revert your changes to `CHANGES.rst`? This will also resolve the merge conflicts.

Please let us know if you need assistance with moving this forward. Thank you!

ykd007:
Thank you for the detailed feedback, @stevepiercy!

I've pushed two commits addressing your requests:

1. Reverted `CHANGES.rst` to match upstream and added a towncrier news fragment at `news/1385.documentation` following the new format.
2. Updated the `single_utc_property` prepended docstring to `The {name} property with all values converted to a ``:class:`~datetime.datetime`` in UTC.`` with proper blank-line separation.
3. Updated the ACKNOWLEDGED `docs` first line to match your suggested text.
4. Cleaned up the `p_doc` boilerplate in `create_single_property` to use ``del``/``None`` literals and restructured the parameter note.

Please let me know if anything else needs attention!

stevepiercy:
@ykd007 GitHub shows that there are merge conflicts that need to be resolved. Would you please take care of that? Then I can complete a review and proceed. Thank you!

ykd007:
Hi @stevepiercy — I've rebased on current main and resolved all merge conflicts. The branch is clean now. Also tightened a few things during the rebase:

- Removed the duplicate `:rfc:`9074`` reference in the ACKNOWLEDGED docstring
- Fixed the invalid `:rfc:`5545#section-3.8.6.3`` role (section anchors aren't supported) to just `:rfc:`5545``
- Cleaned up the `p_doc` boilerplate in `create_single_property` to plain prose instead of a misapplied Parameters section

Should be conflict-free now — please let me know if anything else needs attention.

src/icalendar/attr.py
```diff
@@ -960,10 +960,9 @@ def p_del(self: Component):
     {doc}
 
     Accepted values: {", ".join(t.__name__ for t in value_type)}.
```
```suggestion
    Parameters:
        {", ".join(t.__name__ for t in value_type)}.
```

src/icalendar/attr.py
```diff
@@ -960,10 +960,9 @@ def p_del(self: Component):
...
-    If the attribute has invalid values, we raise
-    :exc:`~icalendar.error.InvalidCalendar`.
-    If the value is absent, we return None.
-    You can also delete the value with del or by setting it to None.
+    Raises :exc:`~icalendar.error.InvalidCalendar` if the value is invalid.
+    Returns ``None`` if absent.
+    You can also delete the value with ``del`` or by setting it to ``None``.
```
stevepiercy: Please follow the docstring format.

```suggestion
    Raises:
        InvalidCalendar: if the attribute has invalid values.

    Returns:
        A :class:`~datetime.datetime` or :class:`~datetime.timedelta` if the
        value is valid. ``None`` if the value is absent.
    """
```

src/icalendar/attr.py
```diff
@@ -960,10 +960,9 @@ def p_del(self: Component):
     {doc}
```
stevepiercy, line 1008:
```suggestion
    {doc}

    You can also delete the value with ``del`` or by setting it to ``None``.

```

stevepiercy (changes requested):
I started to review your changes, but it appears you missed about two dozen of my suggestions and comments. I stopped. Would you please take care of them all? Thank you!

niccokunzmann:
@ykd007 Thanks for your contribution so far. Let us know if you need anything. I know it can sometimes take a bit longer but we would like to see this added :) Please have a look again!

stevepiercy:
@ykd007 there's also some merge conflicts to resolve. I'd like to see your work make it into a release of icalendar. Please let us know if you need any help to get this across the finish line. Thank you!

ykd007:
Thanks for the nudge, @niccokunzmann and @stevepiercy. I rebased the branch on current main and pushed the remaining review cleanup.

I also moved the news fragment to `news/1244.documentation`, kept `CHANGES.rst` untouched, applied the pending wording/link/example suggestions, and removed the inline doctest comments from the rendered examples.

Checked locally with:
- `make format`
- `uv run pytest src/icalendar/tests/test_with_doctest.py`
- `make html`

src/icalendar/attr.py
```diff
@@ -1002,14 +1002,17 @@ def p_del(self: Component):
 
     {doc}
 
-    To delete the value, either use ``del`` or set it to ``None``.
+    You can also delete the value with ``del`` or by setting it to ``None``.
 
-    Returns:
-        If the value is absent, return ``None``.
+    Parameters:
+        {", ".join(t.__name__ for t in value_type)}.
```
stevepiercy: A trailing `.` appeared in the list. Let's remove it.

```suggestion
    Parameters:
        {", ".join(t.__name__ for t in value_type)}
```

I also think that this is too clever. It only gives us the name of the `value_type`s passed in, returned as an unordered list. If we want to support type hints or links to the types, then this is not helpful. I think we ought to move this into each of the properties' docstring instead.

@niccokunzmann can I get your thoughts on this?

src/icalendar/attr.py
```diff
@@ -1002,14 +1002,17 @@ def p_del(self: Component):
...
 
     Raises:
-        :exc:`~icalendar.error.InvalidCalendar`
-            If the attribute has invalid values.
+        InvalidCalendar: if the attribute has invalid values.
+
+    Returns:
+        A :class:`~datetime.datetime` or :class:`~datetime.timedelta` if the
+        value is valid. ``None`` if the value is absent.
```
stevepiercy: Oops. The Returns section is not correct for all its usages. See:

- https://icalendar--1385.org.readthedocs.build/en/1385/reference/api/icalendar.cal.free_busy.html#icalendar.cal.free_busy.FreeBusy.start
- https://icalendar--1385.org.readthedocs.build/en/1385/reference/api/icalendar.cal.free_busy.html#icalendar.cal.free_busy.FreeBusy.end
- https://icalendar--1385.org.readthedocs.build/en/1385/reference/api/icalendar.cal.free_busy.html#icalendar.cal.free_busy.FreeBusy.DTSTART
- https://icalendar--1385.org.readthedocs.build/en/1385/reference/api/icalendar.cal.free_busy.html#icalendar.cal.free_busy.FreeBusy.DTEND

And that's just one module. There is a total of 14 places where `create_single_property` is called.

And now that I think about it, does each property actually return anything when it's called? Can it even be called? What do you think?

@niccokunzmann can you add your thoughts on this?

src/icalendar/cal/alarm.py
stevepiercy: After seeing the result, I now have doubts this is a good idea. See my review.

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
     calendar_name = multi_language_text_property(
         "NAME",
         "X-WR-CALNAME",
-        """This property specifies the name of the calendar.
+        """The display name of this calendar, per :rfc:`7986#section-5.1`.
 
-    This implements :rfc:`7986` ``NAME`` and ``X-WR-CALNAME``.
+    Implements both the ``NAME`` property from :rfc:`7986#section-5.3` and the widely used
```
stevepiercy: This was a bad copy-paste by me.

```suggestion
    Implements both the ``NAME`` property from :rfc:`7986#section-5.1` and the widely used
```

src/icalendar/cal/component.py
stevepiercy: It doesn't. 😢 See other comments in attr.py.

stevepiercy (changes requested):
Thanks for your patience. I found a couple of incorrect RFC section references. I also realized that including section headings in `{doc}` isn't as helpful as I would like. I'd like your thoughts on this. Please let me know. Thank you!

src/icalendar/attr.py
niccokunzmann: There is also `__qualname__`. That might allow referencing them. That could be worth trying out.

```
f":class:`{x.__qualname__}`"
```

niccokunzmann: The initial idea for me was to create a doc template to reduce time spent on writing, copy&paste errors and ensure consistency. Since you are the docs lead, it is your call if you would rather have proper functions with docstrings and maybe wrappers or calls to ensure good docs.
I am not attached to keeping it like this.
One positive argument for change is that we now start testing the docs for consistency, based on the AST, not generated values.

stevepiercy: Yes, also, I think trying to test generated docs will be a layer on top of a layer, which would be an unsupported use case.

src/icalendar/attr.py
stevepiercy: > And now that I think about it, does each property actually return anything when it's called? Can it even be called?

@niccokunzmann can you respond to these questions, too? Thank you!

src/icalendar/attr.py
stevepiercy: Also, I didn't see any type hints on the properties that call `create_single_property`, only on `create_single_property` itself. There's also no return type hint on `create_single_property`, which makes me wonder whether properties should have a `Returns` docstring heading. What do you think?

stevepiercy: > There is also `__qualname__`. That might allow referencing them. That could be worth trying out.
> 
> ```
> f":class:`{x.__qualname__}`"
> ```

```python
    Parameters:
        {", ".join(t.__qualname__ for t in value_type)}
```

Same result as `__name__`.

```python
    Parameters:
        {", ".join(":class:`" + t.__qualname__ + "`" for t in value_type)}
```

Screenshot 2026-07-16 at 2 03 56 PM

Obliterates the `Parameters` section heading, too. I experimented a little bit to no avail. Ultimately, I think generated docstrings are more trouble than using static docstrings.

src/icalendar/attr.py
niccokunzmann: Sorry, I missed this.

So. I created the generator to reduce copy & paste work. It is totally cool to remove generating docstrings all together. It is cool to replace them with well written docstrings. The returns section is indeed wrong.

stevepiercy:
@ykd007 do you need anything further from us to finish this PR? Please let us know. Thank you!

src/icalendar/attr.py
SashankBhamidi: The trailing period makes this invalid RST. More importantly, as new properties get added via `create_single_property`, this will silently generate wrong parameter docs for anything that doesn't return `datetime | timedelta`. 

Drop the `Parameters` block from the template; callers can document their own types in the `doc` arg.

src/icalendar/attr.py
SashankBhamidi: Same issue: the hardcoded `Returns` is wrong for `FreeBusy.start`, `FreeBusy.end`, and the date/datetime variants of `DTSTART`/`DTEND`. 

Either remove it from the template, or derive it from the `type_def` parameter that's already passed in.

SashankBhamidi (changes requested):
Two unresolved threads from Steve's last round still need attention before this can merge.

src/icalendar/attr.py
stevepiercy: To get this PR merged, I think it's OK to let this issue detail go through as is, but create a new issue to replace the generated docstrings with static docstrings. The other unresolved comments need to be addressed, though. Either @ykd007 can do that or I will if they abandon this PR and someone assigns it to me to complete.

stevepiercy:
I created a new issue to handle the generated docstrings in https://github.com/collective/icalendar/issues/1650 so that this PR can move forward.

src/icalendar/cal/calendar.py
ykd007: Corrected this to section 5.1. Thanks for catching it.

src/icalendar/attr.py
ykd007: Leaving the generated Parameters text unchanged in this PR per the latest direction. The static-docstring replacement is now tracked in #1650.

src/icalendar/attr.py
ykd007: Leaving the generated Returns text unchanged here per the latest direction. Replacing these generated docstrings with static ones is tracked in #1650.

src/icalendar/cal/component.py
ykd007: This rendering concern comes from the helper-generated prepended text. Per the latest direction, that broader cleanup is now tracked in #1650 so this PR can move forward.

ykd007:
Thanks for clearing the generated-docstring concern into #1650, @stevepiercy.

I’ve brought the branch up to date with `main`, corrected the remaining `NAME` reference from RFC 7986 section 5.3 to section 5.1, and replied to the open threads. The generated `Parameters` and `Returns` text is unchanged here as agreed and can be handled through #1650.

Local checks are clean:

- `uv run ruff check src/icalendar/attr.py src/icalendar/cal/alarm.py src/icalendar/cal/calendar.py src/icalendar/cal/component.py`
- `uv run pytest -q src/icalendar/tests/test_with_doctest.py` — 374 passed
- `uv run pytest -q src/icalendar/tests/test_docstrings.py` — 1694 passed, 541 expected xfails

src/icalendar/cal/alarm.py
```diff
@@ -159,32 +159,29 @@ class Alarm(Component):
...
+
+    Setting this property allows calendar clients to
+    dismiss or suppress an alarm across multiple devices. Once set to a value
+    greater than or equal to the alarm's computed trigger time, conforming clients
+    will not refire the alarm.
+
+    Returns ``None`` when no acknowledgment has been recorded.
+
+    Example:
+        Mark an alarm as acknowledged at the current time.
```
stevepiercy: This should include a note to align with the example. Also add a Python snippet for usage. Rewrap as needed.

```suggestion
        Mark an alarm as acknowledged. Note that the example uses an arbitrary time
        for the purpose of passing doctests. In actual practice, clients should
        use the current time in UTC, such as ``datetime.now(UTC)``.
        
```

src/icalendar/cal/alarm.py
```diff
@@ -193,17 +190,33 @@ class Alarm(Component):
         "dt",
         (datetime, timedelta),
         timedelta | datetime | None,
-        """Purpose:  This property specifies when an alarm will trigger.
+        """The time at which this alarm fires, per :rfc:`5545`.
```
stevepiercy: Let's add the section reference.

```suggestion
        """The time at which this alarm fires, per :rfc:`5545#section-3.8.6.3`.
```

src/icalendar/cal/alarm.py
```diff
@@ -193,17 +190,33 @@ class Alarm(Component):
...
-    associated start or end of the event or to-do.""",
+    A negative :class:`~datetime.timedelta` fires *before* the related
+    component boundary (start or end); a positive one fires *after* it.
+    Use :attr:`TRIGGER_RELATED` to choose whether the offset is measured from
+    the start or the end of the parent event or to-do.
+    An absolute trigger fires at an exact UTC point in time regardless of the
+    parent component's dates.
+
+    Example:
+        Set an alarm to fire 15 minutes before the start of an event.
```
stevepiercy, line 209: Would you please add an event to the example, and show that the trigger actually is set to 15 minutes before the event?

src/icalendar/cal/alarm.py
```diff
@@ -193,17 +190,33 @@ class Alarm(Component):
...
+
+        .. code-block:: pycon
+
+            >>> from datetime import timedelta
+            >>> from icalendar import Alarm
+            >>> alarm = Alarm()
+            >>> alarm.TRIGGER = timedelta(minutes=-15)
+            >>> alarm.TRIGGER
+            datetime.timedelta(days=-1, seconds=85500)
+
```
stevepiercy, line 232: Would you please add a second example showing how to set an absolute trigger? And would you change the Example section heading to its plural accordingly?

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
     """,
     )
 
     description = multi_language_text_property(
         "DESCRIPTION",
         "X-WR-CALDESC",
-        """This property specifies the description of the calendar.
-
-    This implements :rfc:`7986` ``DESCRIPTION`` and ``X-WR-CALDESC``.
+        """A human-readable description of this calendar's content, per :rfc:`7986#section-5.2`.
```
stevepiercy: "human-readable" is obvious. Move the RFC mention into the description with more details.

```suggestion
        """A description of the calendar's content.
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
 
-    Conformance:
-        This property can be specified multiple times in an
-        iCalendar object.  However, each property MUST represent the
-        description of the calendar in a different language.
+    Implements both ``DESCRIPTION`` from :rfc:`7986#section-5.2` and ``X-WR-CALDESC``
+    for broader calendar client compatibility.
```
stevepiercy: There are actually two relevant RFCs.

```suggestion
    Implements both ``DESCRIPTION`` from :rfc:`5545#section-3.8.1.5` and
    :rfc:`7986#section-5.2` and ``X-WR-CALDESC`` for broader calendar client compatibility.
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
     uid = uid_property
     prodid = single_string_property(
         "PRODID",
-        """PRODID specifies the identifier for the product that created the iCalendar object.
+        """The product identifier for the software that created this iCalendar object.
 
-Conformance:
-    The property MUST be specified once in an iCalendar object.
+This property is defined in :rfc:`5545#section-3.7.3`.
+It's required exactly once per calendar object.
```
stevepiercy: iCalendar is a more precise term in this context.

```suggestion
It's required exactly once per iCalendar object.
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
-    specified in the iCalendar object.  It is expected that other
-    calendar scales will be defined in other specifications or by
-    future versions of this memo.
+        """The calendar scale for date and time values in this iCalendar object.
+
+This property is defined in :rfc:`5545#section-3.7.1`. The only value currently defined is
+``"GREGORIAN"`` (the default). When this property is absent, Gregorian is assumed.
+
+Per :rfc:`7529`, non-Gregorian calendar systems are expressed via ``RRULE``
+transformations rather than a different ``CALSCALE`` value.
```
stevepiercy: I couldn't find an implementation of a transformation in icalendar, only the ability to parse. See `src/icalendar/tests/test_rfc_7529.py`. @ykd007 @niccokunzmann can you verify?

If in fact there is no transformation, then let's use the following text, and create a new issue to actually add such a transformation for CHINESE, ETHIOPIC, and HEBREW.

```suggestion
Per :rfc:`7529`, non-Gregorian calendar systems are expressed via ``RRULE``
transformations rather than a different ``CALSCALE`` value. However, icalendar
currently implements only the parsing of this value, not a transformation."
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
             >>> calendar.calendar_name = "My Calendar"
-            >>> print(calendar.to_ical())
+            >>> print(calendar.to_ical().decode())
             BEGIN:VCALENDAR
             NAME:My Calendar
             X-WR-CALNAME:My Calendar
             END:VCALENDAR
+
+    See also:
+        :attr:`description`
```
stevepiercy: `description` isn't really a good `seealso` for `name`.

```suggestion
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
             >>> calendar.description = "This is a calendar"
-            >>> print(calendar.to_ical())
+            >>> print(calendar.to_ical().decode())
             BEGIN:VCALENDAR
             DESCRIPTION:This is a calendar
             X-WR-CALDESC:This is a calendar
             END:VCALENDAR
+
+    See also:
+        :attr:`calendar_name`
```
stevepiercy: The calendar's name and description aren't closely related enough to merit a seealso.

```suggestion
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
             >>> calendar = Calendar()
             >>> calendar.color = "black"
-            >>> print(calendar.to_ical())
+            >>> print(calendar.to_ical().decode())
             BEGIN:VCALENDAR
             COLOR:black
             END:VCALENDAR
 
+    See also:
+        :attr:`calendar_name`
```
stevepiercy: Not related at all.

```suggestion
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
 
+        >>> from icalendar import Calendar
+        >>> cal = Calendar()
+        >>> cal.version = "2.0"
+        >>> str(cal.version)
+        '2.0'
 
-      """,
+See also:
+    :attr:`prodid`, :attr:`calscale`
```
stevepiercy: Version and prodid are closely related, but calscale is an outlier.

```suggestion
    :attr:`prodid`
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
+Example:
+    .. code-block:: pycon
+
+        >>> from icalendar import Calendar
+        >>> cal = Calendar()
+        >>> cal.calscale
+        'GREGORIAN'
+
+See also:
+    :attr:`version`
```
stevepiercy: Not related.

```suggestion
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
-    If this property is not present in the iCalendar object, then a
-    scheduling transaction MUST NOT be assumed.  In such cases, the
-    iCalendar object is merely being used to transport a snapshot of
-    some calendar information; without the intention of conveying a
-    scheduling semantic.
+        """The iTIP scheduling method associated with this calendar object, per :rfc:`5545#section-3.7.2`.
+
+When present, ``METHOD`` indicates that this object is part of a scheduling
+transaction, such as a meeting invitation or cancellation. Scheduling methods
+are defined by :rfc:`5546` (iTIP), with values such as ``"REQUEST"``,
```
```suggestion
are defined by :rfc:`5546#section-1.4` (iTIP), with values such as ``"REQUEST"``,
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
+    .. code-block:: pycon
+
+        >>> from icalendar import Calendar
+        >>> cal = Calendar()
+        >>> cal.method = "REQUEST"
+        >>> str(cal.method)
+        'REQUEST'
+
+See also:
+    :attr:`version`, :rfc:`5546`
```
stevepiercy: Version is not related, and the RFC is already mentioned.

```suggestion
```

src/icalendar/cal/calendar.py
```diff
@@ -287,227 +287,221 @@ def add_missing_timezones(
...
+
+                >>> from datetime import timedelta
+                >>> from icalendar import Calendar
+                >>> cal = Calendar()
+                >>> cal.refresh_interval = timedelta(hours=1)
+                >>> cal.refresh_interval
+                datetime.timedelta(seconds=3600)
+
+        See also:
+            :attr:`source`
```
stevepiercy: Not related.

```suggestion
```

src/icalendar/cal/component.py
```diff
@@ -729,46 +729,55 @@ def __eq__(self, other: Component) -> bool:
...
+
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Event
+            >>> event = Event()
+            >>> event.DTSTAMP = datetime(2024, 6, 1, 12, 0, 0, tzinfo=timezone.utc)
+            >>> event.DTSTAMP
+            datetime.datetime(2024, 6, 1, 12, 0, tzinfo=ZoneInfo(key='UTC'))
+
+    See also:
+        :attr:`LAST_MODIFIED`, :attr:`CREATED`, :attr:`stamp`
```
stevepiercy: Because this currently gets included to both upper and lower cased properties, let's include all the stamp related properties.

```suggestion
        :attr:`CREATED`, :attr:`DTSTAMP`, :attr:`LAST_MODIFIED`, :attr:`created`, :attr:`stamp`, :attr:`last_modified`
```

src/icalendar/cal/component.py
```diff
@@ -729,46 +729,55 @@ def __eq__(self, other: Component) -> bool:
...
-        """,
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Event
+            >>> event = Event()
+            >>> event.LAST_MODIFIED = datetime(2024, 6, 1, 9, 0, 0, tzinfo=timezone.utc)
+            >>> event.LAST_MODIFIED
+            datetime.datetime(2024, 6, 1, 9, 0, tzinfo=ZoneInfo(key='UTC'))
+
+    See also:
+        :attr:`last_modified`, :attr:`DTSTAMP`, :attr:`CREATED`
```
```suggestion
        :attr:`CREATED`, :attr:`DTSTAMP`, :attr:`LAST_MODIFIED`, :attr:`created`, :attr:`stamp`, :attr:`last_modified`
```

src/icalendar/cal/component.py
```diff
@@ -823,17 +832,28 @@ def _utc_now() -> datetime:
...
+            >>> from datetime import timezone, datetime
+            >>> from icalendar import Event
+            >>> event = Event()
+            >>> event.CREATED = datetime(2024, 1, 1, 8, 0, 0, tzinfo=timezone.utc)
+            >>> event.CREATED
+            datetime.datetime(2024, 1, 1, 8, 0, tzinfo=ZoneInfo(key='UTC'))
 
-        """,
+    See also:
+        :attr:`created`, :attr:`DTSTAMP`, :attr:`LAST_MODIFIED`
```
```suggestion
        :attr:`CREATED`, :attr:`DTSTAMP`, :attr:`LAST_MODIFIED`, :attr:`created`, :attr:`stamp`, :attr:`last_modified`
```

src/icalendar/attr.py
```diff
@@ -1006,14 +1006,17 @@ def p_del(self: Component):
 
     {doc}
 
-    To delete the value, either use ``del`` or set it to ``None``.
+    You can also delete the value with ``del`` or by setting it to ``None``.
```
stevepiercy: Revert this one.


```suggestion
    To delete the value, either use ``del`` or set it to ``None``.
```

src/icalendar/attr.py
```diff
@@ -1006,14 +1006,17 @@ def p_del(self: Component):
...
 
-    Returns:
-        If the value is absent, return ``None``.
+    Parameters:
+        {", ".join(t.__name__ for t in value_type)}.
```
stevepiercy: Sadly, it's not to be in this PR. We'll get there eventually.

```suggestion
```

src/icalendar/attr.py
```diff
@@ -1006,14 +1006,17 @@ def p_del(self: Component):
...
 
     Raises:
-        :exc:`~icalendar.error.InvalidCalendar`
-            If the attribute has invalid values.
+        InvalidCalendar: if the attribute has invalid values.
+
+    Returns:
+        A :class:`~datetime.datetime` or :class:`~datetime.timedelta` if the
+        value is valid. ``None`` if the value is absent.
```
stevepiercy: Let's remove this for now, and we can address it in #1650.

```suggestion
```

stevepiercy (changes requested):
@ykd007 thank you for persevering. Now that we have some clarity, I took more time to comb through this carefully, and made more suggestions. Would you please take care? Thank you!

src/icalendar/cal/alarm.py
ykd007: Updated the note and example to distinguish the fixed doctest time from using `datetime.now(UTC)` in practice.

src/icalendar/cal/alarm.py
ykd007: Added the section reference.

src/icalendar/cal/alarm.py
ykd007: The example now creates an event and shows the alarm firing 15 minutes before its start.

src/icalendar/cal/alarm.py
ykd007: Added the absolute UTC trigger example and changed the heading to “Examples.”

src/icalendar/cal/calendar.py
ykd007: Simplified the opening sentence and moved the RFC details into the explanation below.

src/icalendar/cal/calendar.py
ykd007: Added both relevant RFC references.

src/icalendar/cal/calendar.py
ykd007: Changed this to “iCalendar object.”

src/icalendar/cal/calendar.py
ykd007: Confirmed that the current support parses and serializes these values but does not perform the transformation. I updated the wording and opened #1658 for implementing the CHINESE, ETHIOPIC, and HEBREW transformations.

src/icalendar/cal/calendar.py
ykd007: Removed this `See also` section.

src/icalendar/cal/calendar.py
ykd007: Removed this `See also` section.

src/icalendar/cal/calendar.py
ykd007: Removed this `See also` section.

src/icalendar/cal/calendar.py
ykd007: Removed `calscale` and kept only `prodid` here.

src/icalendar/cal/calendar.py
ykd007: Removed this `See also` section.

src/icalendar/cal/calendar.py
ykd007: Added the iTIP section reference.

src/icalendar/cal/calendar.py
ykd007: Removed this `See also` section.

src/icalendar/cal/calendar.py
ykd007: Removed this `See also` section.

src/icalendar/cal/component.py
ykd007: Included all six upper- and lowercase stamp properties.

src/icalendar/cal/component.py
ykd007: Included all six upper- and lowercase stamp properties.

src/icalendar/cal/component.py
ykd007: Included all six upper- and lowercase stamp properties.

src/icalendar/attr.py
ykd007: Reverted to the original deletion wording.

src/icalendar/attr.py
ykd007: Removed the generated `Parameters` section and left that work out of this PR.

src/icalendar/attr.py
ykd007: Removed the `Returns` section as requested; #1650 can handle it separately.

ykd007:
Thanks for the careful pass, Steve. I’ve addressed all of the comments, verified the RFC 7529 behavior, and opened #1658 for the recurrence transformations. The branch is also up to date with `main`, and the focused docstring and RFC tests pass locally.

stevepiercy:
@ykd007 in the future, would please take advantage of GitHub's [Add suggestion to batch](https://docs.github.com/en/pull-requests/how-tos/review-pull-requests/incorporating-feedback-in-your-pull-request) feature? Doing that would have saved spamming subscribers with 22 messages within a minute. Also it's not necessary to add comments that state you accepted the suggestion as is to every single suggestion, and especially not use AI to generate those messages. The Add suggestion to batch feature will send one notification, and resolve all the unresolved suggestions. It's a kinder, gentler, human process.

stevepiercy (dismissed):
Thanks for persevering, @ykd007. This looks good to me. Will merge when CI passes.

stevepiercy:
Whoops, @SashankBhamidi, I forgot to request another review from you, and ask whether your concerns have been addressed. Done. I also enabled auto-merge.

SashankBhamidi (approved):
Thanks for persevering through all those rounds, @ykd007. 

I resolved the one merge conflict.

One thing I noticed: the `See also` on `DTSTAMP` includes a link back to `DTSTAMP` itself. It makes sense for `stamp` since both share the same docstring, but it self-links when rendered on `DTSTAMP`. I'll open a follow-up issue rather than hold this PR.

PR #1391, opened 2026-05-18
deprecate foldline for icalendar 8 (#1011)

IoannaGiag:
Deprecates `icalendar.parser.string.foldline` as per #1011.

- Added `deprecate_for_version_8` decorator to `icalendar.compatibility`
- Renamed `foldline` to `_foldline` in `parser/string.py`
- Kept `foldline` as a deprecated wrapper that will be removed in icalendar 8
- Updated all internal usages to use `_foldline`
- Added changelog entry in `news/1011.removal`

----
📚 Documentation preview 📚: https://icalendar--1391.org.readthedocs.build/


src/icalendar/compatibility.py
```diff
@@ -33,8 +35,23 @@
     # we cannot use a TypeGuard = "TypeGuard" hack since it's used with a parameter
     TypeGuard = TypeIs = Self = None
 
+
+def deprecate_for_version_8(func):
+    @functools.wraps(func)
```
stevepiercy, line 43: A minimal docstring is helpful.

```suggestion
    """Issue a warning for deprecated functions to be removed in version 8.
    """
    public_name = func.__name__.removeprefix("_")
    @functools.wraps(func)
```

src/icalendar/parser/string.py
```diff
@@ -197,6 +198,9 @@ def foldline(line: str, limit: int = 75, fold_sep: str = "\r\n ") -> str:
     return "".join(ret_chars)
 
 
+foldline = deprecate_for_version_8(_foldline)
```
stevepiercy, line 201: See my [comment in the original issue](https://github.com/collective/icalendar/issues/1011#issuecomment-4525009633). I don't think this works as @niccokunzmann expects.

To verify, would you please call `foldline` to see what it returns?

src/icalendar/compatibility.py
```diff
@@ -33,8 +35,23 @@
...
+    def wrapper(*args, **kwargs):
+        warnings.warn(
+            f"{func.__name__} is deprecated and will be removed in icalendar 8",
```
stevepiercy: This might need to be changed. See my comment above and at https://github.com/collective/icalendar/issues/1011#issuecomment-4525009633.

stevepiercy (changes requested):

src/icalendar/compatibility.py
stevepiercy: @IoannaGiag my comment was confirmed. I've updated my suggestion here. Please see the discussion at https://github.com/collective/icalendar/issues/1011#issuecomment-4525009633. Thank you!

stevepiercy (changes requested):

IoannaGiag:
Hi @stevepiercy, I've addressed all the requested changes. Please let me know if there's anything else to fix. Thank you for your guidance.

src/icalendar/compatibility.py
```diff
@@ -33,8 +35,27 @@
     # we cannot use a TypeGuard = "TypeGuard" hack since it's used with a parameter
     TypeGuard = TypeIs = Self = None
 
+
+def deprecate_for_version_8(func):
```
stevepiercy, line 39: Docs look good:

https://icalendar--1391.org.readthedocs.build/en/1391/reference/api/icalendar.compatibility.html#icalendar.compatibility.deprecate_for_version_8

Thanks for responding to the ongoing discussion with the team by updating this method.

stevepiercy (approved):
Thank you!

Issue #1352, opened 2026-04-30
Choose version - version not displayed


niccokunzmann:
When I go to https://icalendar.readthedocs.io/
I see the text "Choose Version"

Instead, it should show the version that is stable.
<img width="225" height="51" alt="Image" src="https://github.com/user-attachments/assets/981c4285-0b7e-4e30-bf6b-15c1c4f3339c" />

stevepiercy:
- `7.1.0` tag version was not built. I built it manually.
- `stable` build was not triggered after the `7.1.0` tag, and therefore picking up its commit hash automatically. I triggered it to rebuild, and it now has the correct hash of 1300b76fb3fe40d987cc3715735c026b3c1a2191.
- I created a new automation rule, [Create new stable version](https://app.readthedocs.org/dashboard/icalendar/rules/3188/), on Read the Docs. We'll see if that handles the above two steps automatically going forward.
- I verified that https://icalendar.readthedocs.io/en/stable/ pulls in https://icalendar.readthedocs.io/en/latest/_static/version-switcher.json in web developer tools --> Network --> XHR, and that the content is correct.
- Visiting https://icalendar.readthedocs.io/en/v7.1.0/ selects `7.x (stable)` in the version switcher.
- Visiting https://icalendar.readthedocs.io/en/stable/ still doesn't select `7.x (stable)` but shows `Choose version`. I even tried this in a private browser session.
- I checked out `v7.1.0` and built the docs. I think I found something. How does `src/icalendar/_version.py` actually get updated on RTD? Right now I see `__version__ = version = '7.1.1.dev0'`, which is a mismatch.

@niccokunzmann any idea?

I've tried everything I can think of, and I don't know why it doesn't work as expected.

niccokunzmann:
Ah.Hatch vcs might have an option for that appending or not the dev0 thing. It is because we are not running on the tag but there is a commit after that. Might just resolve with the nextreleasse.

stevepiercy:
Well, it still doesn't work for 7.1.1, just released. I still don't know what the problem is, but I think it comes down to the release process doesn't set a value of `release` that is a simple `7.1.1`, but appends `.dev###`, so it never matches. We'll need to debug this some more with tags and lots of releases. That sucks. Unfortunately, I'm not a maintainer, so it's out of my control.

@SashankBhamidi do you want to try?

Is there a better way to manage versions? This `_version.py` file thing with hatch is not working with Sphinx or RTD. In fact, mine shows:

```python
[...]

PR #1393, opened 2026-05-20
Update maintenance docs and fix version switcher for "stable"

stevepiercy:
## Closes issue

If this is successful, then close the related issue.

- See #1352

## Description

Update maintenance docs and fix version switcher for "stable".

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1393.org.readthedocs.build/

### DIFF
--- a/docs/contribute/maintenance.rst
+++ b/docs/contribute/maintenance.rst
@@ -337,14 +337,7 @@ However, only people with ``Environments/Configure PyPI`` access can approve an 
         git commit -m "Show version warning banner for previous major version 6.x on Read the Docs"
         git push  # to collective/icalendar
 
-#.  Configure `Read the Docs <https://app.readthedocs.org/projects/icalendar/>`_.
-
-    a.  Verify that the "stable" version was activated automatically on Read the Docs.
-
-    #.  Deactivate previous releases in that current stable release line.
-        Click on the ellipsis icon, and select :guilabel:`Configure version`.
-        Toggle the :guilabel:`Active` switch to deactivate the version.
-
+#.  Verify that the "stable" version was activated automatically on `Read the Docs <https://app.readthedocs.org/projects/icalendar/>`_.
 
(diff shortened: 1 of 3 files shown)


docs/contribute/maintenance.rst
```diff
@@ -337,14 +337,7 @@ However, only people with ``Environments/Configure PyPI`` access can approve an
...
 
-#.  Configure `Read the Docs <https://app.readthedocs.org/projects/icalendar/>`_.
-
-    a.  Verify that the "stable" version was activated automatically on Read the Docs.
-
-    #.  Deactivate previous releases in that current stable release line.
-        Click on the ellipsis icon, and select :guilabel:`Configure version`.
-        Toggle the :guilabel:`Active` switch to deactivate the version.
-
+#.  Verify that the "stable" version was activated automatically on `Read the Docs <https://app.readthedocs.org/projects/icalendar/>`_.
```
SashankBhamidi, line 340: You dropped the "Configure Read the Docs" step that included deactivating previous releases. Intentional? I'm assuming active-vs-inactive on RTD only affects whether it wastes builds on old versions, not the switcher UX. If that's right, fine to drop.

SashankBhamidi (approved):
LGTM. Nails the root cause!

Approving. I'll cut the new release once this merges, address my comment before merge.

docs/contribute/maintenance.rst
stevepiercy: Yup. I reconfigured RTD so that it doesn't activate the `v#.#.#` build, but instead will merely trigger a "stable" build, which now does the `git fetch --tags` magic. Since it's not automatically activated, there's no longer a need to manually deactivate it.

stevepiercy:
All CI tasks passed, except RTD hasn't responded. This happens from time to time. It passed previously without any change since, so I'll merge this.

stevepiercy:
@SashankBhamidi release away, at your convenience, please. Thank you!

PR #1369, opened 2026-05-07
docs: fix broken cross-references in documentation

wahajahmed010:
## Summary

Fix broken cross-references in documentation by removing `:py:` prefixed roles and qualifying ambiguous references.

## Changes

- Remove `:py:class:`, `:py:meth:`, `:py:mod:` prefixed roles that break Sphinx cross-references
- Qualify ambiguous `:attr:` references with full module path in `cal/lazy.py`
- Fix `:meth:` reference in `parser/property.py` to use full path
- Qualify `:class:` and `:mod:` references in `prop/dt/*.py` files

See #1158

## Review feedback addressed

- Reverted unnecessary changes per @stevepiercy review
- Moved change log entry to `news/1158.documentation` per project convention

PR #1394, opened 2026-05-21
Revise pull request template for issue linking

stevepiercy:
With many first-time contributors, we don't tell them that they should use `See`, instead of `Closes`, when linking an issue that should remain open because their PR fixes only one or a few items in the original issue. This change supports that option. Example https://github.com/collective/icalendar/pull/1369#issuecomment-4447341005.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,4 @@
-## Closes issue
+## Linked issue
 
 Replace `ISSUE_NUMBER` with the issue number that your pull request addresses. This links the pull request to the related issue. Choose the appropriate form.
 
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -1,9 +1,17 @@
 ## Closes issue
```
SashankBhamidi: Now misleading since "See" doesn't close anything. Maybe `## Linked issue` or `## Related issue`?

```suggestion
## Linked issue
```

SashankBhamidi (changes requested):
Useful change. One thought:

.github/pull_request_template.md
stevepiercy: Good catch.

SashankBhamidi (approved):
LGTM, thanks!

PR #1396, opened 2026-05-22
Align workflow with .readthedocs.yaml

stevepiercy:
- This workflow creates a link as a comment in the pull request. Sometimes it fired when it shouldn't have, or didn't fire when it should.

## Checklist

### DIFF
--- a/.github/workflows/rtd-pr-preview.yml
+++ b/.github/workflows/rtd-pr-preview.yml
@@ -9,7 +9,7 @@ on:
     # documentation files.
     paths:
       - "docs/**"
-      - "src/icalendar/*.py"
+      - "src/icalendar/**/*.py"
       - .github/workflows/documentation.yml
       - .github/workflows/rtd-pr-preview.yml
       - .readthedocs.yaml
--- a/.github/workflows/rtd-pr-preview.yml
+++ b/.github/workflows/rtd-pr-preview.yml
@@ -12,7 +12,7 @@ on:
       - "src/icalendar/**/*.py"
       - .github/workflows/documentation.yml
       - .github/workflows/rtd-pr-preview.yml
-      - .readthedocs.yaml
+      - .readthedocs.yml
       - .vale.ini
       - CHANGES.rst
       - Makefile
--- a/.github/workflows/rtd-pr-preview.yml
+++ b/.github/workflows/rtd-pr-preview.yml
@@ -1,4 +1,5 @@
 # .github/workflows/rtd-pr-preview.yml
+# Usage: https://github.com/readthedocs/actions/blob/v1/preview/README.md
 name: readthedocs/actions
 on:
   pull_request_target:
@@ -8,10 +9,17 @@ on:
     # documentation files.
     paths:
       - "docs/**"
-      - "*.rst"
       - "src/icalendar/*.py"
+      - .github/workflows/documentation.yml
+      - .github/workflows/rtd-pr-preview.yml
       - .readthedocs.yaml
-      - requirements_docs.txt
+      - .vale.ini
+      - CHANGES.rst
+      - Makefile
+      - pyproject.toml
+      - "!src/icalendar/fuzzing"
+      - "!src/icalendar/tests"
+      - "!src/icalendar/timezone"
 
 permissions:
   pull-requests: write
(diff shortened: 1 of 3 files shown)


stevepiercy:
The expected RTD build is not responding, but that's separate from this PR's purpose. I sent a support request to RTD. This PR won't fix that issue.

.github/workflows/rtd-pr-preview.yml
```diff
@@ -8,10 +9,17 @@ on:
     # documentation files.
     paths:
       - "docs/**"
-      - "*.rst"
       - "src/icalendar/*.py"
```
SashankBhamidi: `.readthedocs.yml` (line 24) uses `src/icalendar/**/*.py` to detect Python changes anywhere under the package. This workflow uses `src/icalendar/*.py` which only matches top-level files. So an edit to e.g. `src/icalendar/cal/calendar.py` would cancel the RTD build correctly but skip the preview comment.

```suggestion
      - "src/icalendar/**/*.py"
```

.github/workflows/rtd-pr-preview.yml
```diff
@@ -8,10 +9,17 @@ on:
...
+      - .github/workflows/rtd-pr-preview.yml
       - .readthedocs.yaml
-      - requirements_docs.txt
+      - .vale.ini
+      - CHANGES.rst
+      - Makefile
+      - pyproject.toml
+      - "!src/icalendar/fuzzing"
+      - "!src/icalendar/tests"
+      - "!src/icalendar/timezone"
```
SashankBhamidi: GitHub Actions path filters need a positive pattern that matches the path before a `!` exclusion takes effect. The only positive pattern matching `src/icalendar/...` here is `src/icalendar/*.py` (top-level only). So `!src/icalendar/fuzzing` etc. don't exclude anything, those paths weren't going to match anyway.

Once line 12 becomes `src/icalendar/**/*.py` (see the other comment), these need `/**` too:

```suggestion
      - "!src/icalendar/fuzzing/**"
      - "!src/icalendar/tests/**"
      - "!src/icalendar/timezone/**"
```

.github/workflows/rtd-pr-preview.yml
```diff
@@ -8,10 +9,17 @@ on:
     # documentation files.
     paths:
       - "docs/**"
-      - "*.rst"
       - "src/icalendar/*.py"
+      - .github/workflows/documentation.yml
+      - .github/workflows/rtd-pr-preview.yml
       - .readthedocs.yaml
```
SashankBhamidi: The actual file is `.readthedocs.yml` (no `a`). Both this workflow and `.readthedocs.yml` itself (line 27) reference `.readthedocs.yaml`. This pre-existed, but since the PR is about aligning these files it's a good spot to fix.

```suggestion
      - .readthedocs.yml
```

(And the same fix in `.readthedocs.yml` line 27, if you want to do it here.)

SashankBhamidi (changes requested):
Thanks! Couple of things:

.github/workflows/rtd-pr-preview.yml
stevepiercy: Well spotted!

.github/workflows/rtd-pr-preview.yml
stevepiercy: Done in 585a3a2b74ba64a8380c16d14d6eb661609ae2d9

.github/workflows/rtd-pr-preview.yml
stevepiercy: Found another typo and fixed in 5173f38b3e1396ea353365e1ba769e81bd62ab7e

SashankBhamidi (dismissed):
@stevepiercy,

LGTM on the workflow changes!

One pre-existing thing I noticed while diffing, `.readthedocs.yml` line 25 (`.github/workflows/documentation.yml`) is missing the trailing `\`. I tested locally and the resulting shell script breaks: bash evaluates the `if` against `documentation.yml` only, then tries to execute `.github/workflows/rtd-pr-preview.yml` as a standalone command. The `exit 183` never runs.

This might actually explain the "sometimes fires when it shouldn't" symptom you mentioned in the PR description.

Out of scope for this PR, but worth a follow-up. Approving

stevepiercy:
@SashankBhamidi fixed in d75486ebc38d6f3ef630cfc4d04d157cb3916ab6. Thank you!

PR #1410, opened 2026-05-24
Added instructions to Contribute for how to avoid a filename conflict…

stevepiercy:
… when adding a change log entry

- See https://github.com/collective/icalendar/pull/1369#discussion_r3294330498

----
📚 Documentation preview 📚: https://icalendar--1410.org.readthedocs.build/en/1410/

### DIFF
--- a/docs/contribute/index.rst
+++ b/docs/contribute/index.rst
@@ -163,6 +163,13 @@ The change log entry's format must be ``#.type``, where ``#`` is the referenced 
 ``chore``
     For routine tasks that shouldn't be published, but will satisfy the checker for the presence of a change log entry.
 
+To avoid a conflict with another pull request for the same issue number, append an integer to the filename.
+Don't edit a change log entry from another pull request.
+
+..  code-block:: text
+
+    1158.documentation.1
+
 For orphan change log entries—that is, those that don't need to be linked to any issue ID or other identifier—start the file name with ``+``.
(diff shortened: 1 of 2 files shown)


docs/contribute/index.rst
```diff
@@ -163,6 +163,13 @@ The change log entry's format must be ``#.type``, where ``#`` is the referenced
 ``chore``
     For routine tasks that shouldn't be published, but will satisfy the checker for the presence of a change log entry.
 
+To avoid a conflict with another pull request for the same issue number, append an integer to the filename.
+Don't edit a change log entry from another pull request.
+
+..  code-block:: text
+
+    1158.documentation.1
```
SashankBhamidi, line 171: "Append an integer" doesn't say where to start. If PR A has already merged `1158.documentation` and PR B is the second one for the same issue, does B use `1158.documentation.1`? Or do both need numbered suffixes? Worth a one-liner so contributors don't have to guess.

SashankBhamidi (commented):
Tested locally, towncrier picks up `1158.documentation.1` and `1158.documentation.2` as separate entries. The pattern works!

One thing though:

docs/contribute/index.rst
stevepiercy: It doesn't matter where to start or end. It could be `.666`. The point is "to avoid a conflict".

SashankBhamidi (approved):

PR #1417, opened 2026-05-25
Add Alarm.new_display(), new_audio(), and new_email() factory methods

### LINKED ISSUE #865: VALARM custom new() methods
VALARM can be 

- email
- display
- sound

according to RFC 5545.

Please create specific `new_...()` methods to verify all fields are set properly for these cases.

- [ ] create a nice doc string for them
- [ ] document parameters
- [ ] link to RFC
- [ ] give an example of how to use the class method adding this to an Event with `Event.new()`

klouds27:
Closes #865.

## What

Adds three class methods to `Alarm` for creating the three standard VALARM action types defined in RFC 5545 §3.6.6:

- `Alarm.new_display(description, trigger, ...)` — `ACTION:DISPLAY`
- `Alarm.new_audio(trigger, attach=None, ...)` — `ACTION:AUDIO`
- `Alarm.new_email(summary, description, trigger, attendees, ...)` — `ACTION:EMAIL`

Each method sets `ACTION` and `TRIGGER` (both required by RFC 5545 for every alarm type) and enforces the action-specific required fields, raising `InvalidCalendar` on missing or invalid inputs. `DURATION` and `REPEAT` are accepted as optional paired parameters by all three methods and raise `InvalidCalendar` when only one is supplied.

## Example

```python
from datetime import timedelta
from icalendar import Alarm, vCalAddress

# Display alarm — 15 minutes before event start
alarm = Alarm.new_display(
    description="Team meeting in 15 minutes",
    trigger=timedelta(minutes=-15),
)

# Audio alarm with a custom sound file
alarm = Alarm.new_audio(
    trigger=timedelta(minutes=-5),
    attach="ftp://example.com/pub/sounds/bell-01.aud",
)

# Email alarm sent to two recipients
alarm = Alarm.new_email(
    summary="Meeting reminder",
    description="Your meeting starts in 30 minutes.",
    trigger=timedelta(minutes=-30),
    attendees=[
        vCalAddress("mailto:a@example.com"),
        vCalAddress("mailto:b@example.com"),
    ],
)
```

## Tests

25 unit tests in `src/icalendar/tests/attr/test_alarm_factory_methods.py` covering:
- Correct `ACTION` value for each method
- Required property setters (description, summary, trigger, attendees)
- Optional `ATTACH` and `DURATION`/`REPEAT` pairing
- `InvalidCalendar` raised for each missing required field
- `InvalidCalendar` raised when `DURATION` and `REPEAT` are not supplied together

----
📚 Documentation preview 📚: https://icalendar--1417.org.readthedocs.build/en/1417/


news/865.feature
```diff
@@ -0,0 +1 @@
+Add :meth:`Alarm.new_display() <icalendar.cal.alarm.Alarm.new_display>`, :meth:`Alarm.new_audio() <icalendar.cal.alarm.Alarm.new_audio>`, and :meth:`Alarm.new_email() <icalendar.cal.alarm.Alarm.new_email>` factory methods for creating RFC 5545 VALARM components. @klouds27
```
stevepiercy:
```suggestion
Add :meth:`Alarm.new_display() <icalendar.cal.alarm.Alarm.new_display>`, :meth:`Alarm.new_audio() <icalendar.cal.alarm.Alarm.new_audio>`, and :meth:`Alarm.new_email() <icalendar.cal.alarm.Alarm.new_email>` factory methods for creating :rfc:`5545#section-3.6.6` VALARM components. @klouds27
```

stevepiercy:
@klouds27 would you please respond to the comments from copilot that you requested, and update your branch to `main`? I'll review separately. Thank you!

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
...
+            trigger: When the alarm fires, as a :class:`~datetime.timedelta`
+                relative to the event start (negative means before) or as an
+                absolute UTC :class:`~datetime.datetime`.
+            duration: Gap between repeated triggers. Must be paired with
+                ``repeat``. Corresponds to the :attr:`DURATION` property.
+            repeat: Number of *additional* times to fire after the initial
+                trigger. Must be paired with ``duration``.
+                Corresponds to the :attr:`REPEAT` property.
+            uid: Unique identifier for the alarm. Generated automatically
+                when ``None``.
```
stevepiercy: It doesn't get automatically generated.

```suggestion
            uid: Unique identifier for the alarm or ``None``.
```

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
...
+        Parameters:
+            summary: Subject line of the email.
+                Corresponds to the :attr:`summary` property.
+            description: Body of the email.
+                Corresponds to the :attr:`description` property.
+            trigger: When the alarm fires, as a :class:`~datetime.timedelta`
+                relative to the event start (negative means before) or as an
+                absolute :class:`~datetime.datetime` (recommend UTC-aware).
+            attendees: One or more recipient addresses as
+                :class:`~icalendar.vCalAddress` instances, e.g.
```
stevepiercy: - Use proper reStructuredText syntax to link.
- Don't use Latin abbreviations.
- Rewrap as needed for lint.

```suggestion
                :class:`~icalendar.prop.cal_address.vCalAddress` instances, such as
```

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
...
+                ``[vCalAddress("mailto:user@example.com")]``.
+                At least one address is required.
+            attachments: Optional list of URIs to attach to the email.
+            duration: Gap between repeated triggers. Must be paired with
+                ``repeat``. Corresponds to the :attr:`DURATION` property.
+            repeat: Number of *additional* times to fire after the initial
+                trigger. Must be paired with ``duration``.
+                Corresponds to the :attr:`REPEAT` property.
+            uid: Unique identifier for the alarm. Generated automatically
+                when ``None``.
```
stevepiercy: It's not generated automatically.

```suggestion
            uid: Unique identifier for the alarm or ``None``.
```

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
...
+        ``attach`` URI points to the audio file to play; when omitted,
+        the client uses its default alert sound.
+
+        Conforms to :rfc:`5545#section-3.6.6`.
+
+        Parameters:
+            trigger: When the alarm fires, as a :class:`~datetime.timedelta`
+                relative to the event start (negative means before) or as an
+                absolute :class:`~datetime.datetime` (recommend UTC-aware).
+            attach: Optional URI of the audio file to play, e.g.
```
stevepiercy: Don't use Latin abbreviations.

```suggestion
            attach: Optional URI of the audio file to play, such as
```

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
...
+                ``"ftp://example.com/pub/sounds/bell.aud"``. When ``None``
+                the client uses its default sound.
+            duration: Gap between repeated triggers. Must be paired with
+                ``repeat``. Corresponds to the :attr:`DURATION` property.
+            repeat: Number of *additional* times to fire after the initial
+                trigger. Must be paired with ``duration``.
+                Corresponds to the :attr:`REPEAT` property.
+            uid: Unique identifier for the alarm. Generated automatically
+                when ``None``.
```
stevepiercy: It's not automatically generated.

```suggestion
            uid: Unique identifier for the alarm or ``None``.
```

stevepiercy (changes requested):
This is excellent work. Thank you for improving icalendar! I have a few suggestions for you to consider. Would you please take care?

I also would like a second pair of eyes from a maintainer. @SashankBhamidi @niccokunzmann @angatha

klouds27:
Addressed all review comments:

- Removed the positional-only `/` separator from `new_display`, `new_audio`, and `new_email` — all parameters are now explicitly keyword-passable
- Updated `news/865.feature` to use `:rfc:\`5545#section-3.6.6\`` markup
- Replaced `e.g.` / `i.e.` with `such as` in `attach` and `attendees` docstrings
- Fixed `vCalAddress` cross-reference to use the full module path `icalendar.prop.cal_address.vCalAddress`
- Simplified all three `uid` docstring lines to `or \`\`None\`\``
- The UTC trigger wording was already updated in the previous commit (changed to "recommend UTC-aware")
- The `repeat`-without-`duration` tests were already added in the previous commit
- Rebased onto current `main`

stevepiercy (dismissed):
This looks good to me. Since this is a new feature, I'd like to get a maintainer to review as well. @niccokunzmann @SashankBhamidi @angatha would you please do the honors?

Thank you!

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,247 @@ def new(
...
+    @classmethod
+    def new_display(
+        cls,
+        description: str,
+        trigger: timedelta | datetime,
+        duration: timedelta | None = None,
+        repeat: int | None = None,
+        uid: str | uuid.UUID | None = None,
+    ) -> Alarm:
+        """Create a new DISPLAY alarm that shows a text reminder.
```
SashankBhamidi, line 387: Issue #865's checklist explicitly asks for an example of "adding this to an Event with `Event.new()`". The current pycon blocks all show standalone alarm creation. Worth adding one example showing the integration:

```python
event = Event.new(summary="Meeting", start=..., end=...)
event.add_component(Alarm.new_display(description="...", trigger=timedelta(minutes=-15)))
```

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,247 @@ def new(
...
+                ACTION:DISPLAY
+                DESCRIPTION:Team meeting in 15 minutes
+                TRIGGER:-PT15M
+                END:VALARM
+        """
+        if not description:
+            raise InvalidCalendar("DISPLAY alarm requires a description")
+        if trigger is None:
+            raise InvalidCalendar("DISPLAY alarm requires a trigger")
+        alarm: Alarm = cls()
```
SashankBhamidi: All three factories (here, line 486, and line 578) call `cls()` directly. That diverges from how `Event.new()` and the existing `Alarm.new()` are structured (both go through `super().new(...)`). It also means Component-level params already supported by `Alarm.new()`: `concepts`, `links`, `refids`, `related_to`, are inaccessible through the new factories. A user wanting a DISPLAY alarm with refids has to mutate it after the fact.

The cleaner shape would be to delegate to `Alarm.new()` for the common plumbing and only add action-specific setup on top.

This is a design call I'd like @niccokunzmann to weigh in on. He's back mid-June.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,247 @@ def new(
...
+        alarm.description = description
+        alarm.TRIGGER = trigger
+        alarm.uid = uid
+        if duration is not None or repeat is not None:
+            if duration is None or repeat is None:
+                raise InvalidCalendar(
+                    "DURATION and REPEAT must be set together or not at all"
+                )
+            alarm.DURATION = duration
+            alarm.REPEAT = repeat
```
SashankBhamidi: The same 7-line block appears here and at 492-498 and 588-594. A small private helper would centralize the logic and reduce drift if the rules change.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,247 @@ def new(
...
+                TRIGGER:-PT5M
+                END:VALARM
+        """
+        if trigger is None:
+            raise InvalidCalendar("AUDIO alarm requires a trigger")
+        alarm: Alarm = cls()
+        alarm.add("ACTION", "AUDIO")
+        alarm.TRIGGER = trigger
+        alarm.uid = uid
+        if attach is not None:
```
SashankBhamidi:
```python
Alarm.new_audio(trigger=..., attach="")
```

produces `ATTACH:` with no value, which isn't valid ICS. Either reject empty strings or treat empty/None identically.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,247 @@ def new(
...
+        if not description:
+            raise InvalidCalendar("DISPLAY alarm requires a description")
+        if trigger is None:
+            raise InvalidCalendar("DISPLAY alarm requires a trigger")
+        alarm: Alarm = cls()
+        alarm.add("ACTION", "DISPLAY")
+        alarm.description = description
+        alarm.TRIGGER = trigger
+        alarm.uid = uid
+        if duration is not None or repeat is not None:
```
SashankBhamidi: The `is not None` check (also at 492 and 588) treats `0` as set, so `new_*(..., repeat=0)` raises "DURATION and REPEAT must be set together". RFC-correct in spirit (REPEAT=0 with DURATION makes no sense), but probably surprising. Either document the gotcha or coerce `0` and `None` together.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,247 @@ def new(
...
+        This is the most common alarm type used by calendar clients.
+
+        Conforms to :rfc:`5545#section-3.6.6`.
+
+        Parameters:
+            description: The text to display when the alarm fires.
+                Corresponds to the :attr:`description` property.
+            trigger: When the alarm fires, as a :class:`~datetime.timedelta`
+                relative to the event start (negative means before) or as an
+                absolute UTC :class:`~datetime.datetime`.
```
SashankBhamidi: `new_display` still says "absolute UTC :class:`~datetime.datetime`" (line 377), while `new_audio` (line 448) and `new_email` (line 527) were softened to "(recommend UTC-aware)". Looks like the softening from Copilot's earlier feedback was applied to two of three methods. Either align all three on one phrasing or, better, tighten the implementation to actually enforce UTC for absolute datetimes (RFC 5545 §3.8.6.3 makes this a MUST). Naive datetimes currently slip through silently on all three.

SashankBhamidi (dismissed):
Nice work. The factory methods are RFC-correct.

A few things to address before merge. Findings 1, 3, 4, 6 are mechanical. Finding 2 is a design question I'd like to revisit with @niccokunzmann when he's back mid-June, @angatha curious if you have thoughts in the meantime.

SashankBhamidi:
@klouds27 When addressing each finding, please mark the conversation as resolved with a short reply, either confirming the fix in a specific commit, or noting why it's being deferred/skipped if applicable. 

Same goes for the Copilot review you requested earlier. Helps keep the review state clean. Thanks!

klouds27:
Addressed the mechanical findings: extracted repeated duration/repeat validation to a `_apply_duration_repeat` helper (finding 3), `new_audio` now rejects empty string `attach` the same as `None` (finding 4), aligned `new_display` trigger docstring to "recommend UTC-aware" to match the other two methods (finding 6), added an `Event.new()` integration example to `new_display` (finding 1).

Finding 2 (factory delegation to `Alarm.new()`) noted, happy to revisit when @niccokunzmann is back. Finding 5 (`repeat=0` behavior) not yet addressed, let me know if you'd like it coerced or just documented.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
         alarm.attendees = attendees
         return alarm
 
+    @classmethod
+    def new_display(
+        cls,
+        /,
+        description: str,
+        trigger: timedelta | datetime,
```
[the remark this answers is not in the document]
angatha: @klouds27 this was fine. There was only one positional-only parameter: `cls`. `description` and `trigger` (as well as `duration`, `repeat` and `uid`) where positional-or-keyword parameters.

https://peps.python.org/pep-0570/#specification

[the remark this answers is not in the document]
angatha: Please add thouse again to be consistent with existing code.

angatha:
About finding 2, personally I would add the parameters from `Alarm.new` as additional Parameters with the same default values and call `Alarm.new` instead of `cls()` and would make a comment at `Alarm.new` that the factories need to be updated as well. The factories get corresponding notes as well. This gives a complete API and everyone touching the API knows what also needs to be updated.

src/icalendar/cal/alarm.py
angatha: Should we prevent negative `repeat` and negative `duration`?

baptx:
A property to get the action value should also be added:
https://github.com/collective/icalendar/issues/1421

src/icalendar/cal/alarm.py
klouds27: Added in commit `2f4d232` — new_display docstring now includes a full Event.new() + alarm.new_display() example.

src/icalendar/cal/alarm.py
klouds27: angatha's suggestion of mirroring Alarm.new parameters and calling it internally makes sense. Happy to implement once niccokunzmann is back.

src/icalendar/cal/alarm.py
klouds27: Extracted to `_apply_duration_repeat` classmethod in commit `2f4d232`. All three factory methods now delegate to it.

src/icalendar/cal/alarm.py
klouds27: Changed to `if attach:` in commit `2f4d232`. Empty string is now rejected the same as None. Added a test for this case.

src/icalendar/cal/alarm.py
klouds27: Let me know if you'd prefer it coerced to None, validated, or just documented.

src/icalendar/cal/alarm.py
klouds27: Updated in commit `2f4d232` — now reads "absolute (recommend UTC-aware)" matching new_audio and new_email.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,261 @@ def new(
         alarm.attendees = attendees
         return alarm
 
+    @classmethod
```
niccokunzmann: This looks like it should be a method.

src/icalendar/cal/alarm.py
niccokunzmann: Yes, having it based on Alarm.new sounds good.
Having the additional checks inside the new_ is great.

Alarm.new is allowed to check for correctness.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,261 @@ def new(
...
+        alarm.TRIGGER = trigger
+        alarm.uid = uid
+        cls._apply_duration_repeat(alarm, duration, repeat)
+        return alarm
+
+    @classmethod
+    def new_audio(
+        cls,
+        trigger: timedelta | datetime,
+        attach: str | None = None,
```
niccokunzmann: I wonder: attach could also be vBinary? In that case it should also allow bytes.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,261 @@ def new(
...
+        cls._apply_duration_repeat(alarm, duration, repeat)
+        return alarm
+
+    @classmethod
+    def new_email(
+        cls,
+        summary: str,
+        description: str,
+        trigger: timedelta | datetime,
+        attendees: list[vCalAddress],
```
niccokunzmann: Does str also work? It would be nice to also just pass a single one. And use Sequence instead of list.

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,231 @@
...
+        Alarm.new_email(
+            summary="S",
+            description="D",
+            trigger=timedelta(minutes=-30),
+            attendees=[vCalAddress("mailto:user@example.com")],
+            duration=timedelta(minutes=5),
+        )
+
+
+def test_new_email_repeat_without_duration_raises():
```
niccokunzmann: Using pytest.mark.parametrize, this file can be reduced in size and it is easier to spot patterns. Not a blocker for merge.

niccokunzmann (changes requested):
Thank you so much for all that work! Let me know if you need anything. I added a few ideas to fit it into the existing structure of the package.

klouds27:
addressed all three. `attach` on `new_audio` now accepts `bytes` and wraps them as `vBinary`. `attendees` and `attachments` on `new_email` now take a single value or a sequence, normalized at entry. test file refactored with `parametrize` for the duration/repeat validation cases.

stevepiercy:
@klouds27 would you please update your branch to `main`? I've also requested another round of reviews from those who've submitted comments. Thank you!

src/icalendar/cal/alarm.py
niccokunzmann: Please make this a method. It is not a class method if it only operates in an alarm object.

niccokunzmann:
Sashank and I both left comments which we would like to see addressed. It is already good and a nice contribution. Thanks so much! Let us know if you don't like to address them or if you plan to! Thanks!

klouds27:
on it — converted _apply_duration_repeat to an instance method in the latest commit, call sites updated to alarm._apply_duration_repeat(...). let me know if there is anything else.

niccokunzmann:
@klouds27 Thanks. I am trying to pick this up... When you go through though this, is there anything that seems like it needs doing before I review? Let me know when you are done and I will have a look! Sorry if it was not clear that I am waiting for your confirmation that you are ready for review!

klouds27:
done — all three factories now route through cls.new() instead of cls(), picking up links, related_to, refids, and concepts. ready for review.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +355,295 @@ def new(
...
+        cls,
+        description: str,
+        trigger: timedelta | datetime,
+        duration: timedelta | None = None,
+        repeat: int | None = None,
+        uid: str | uuid.UUID | None = None,
+        links: LINKS_TYPE_SETTER = None,
+        related_to: RELATED_TO_TYPE_SETTER = None,
+        refids: list[str] | str | None = None,
+        concepts: CONCEPTS_TYPE_SETTER = None,
```
SashankBhamidi, line 385: `links`, `related_to`, `refids`, and `concepts` are accepted by all three factory methods and forwarded to `cls.new()`, but none of the three docstrings mention them in the Parameters section. Same gap exists in `new_audio` (lines 467-474) and `new_email` (lines 551-557). 

A reader who finds `new_display()` through autocomplete will not know these exist. Worth adding a brief entry for each, pointing at the same attrs as `Alarm.new()` does.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +355,295 @@ def new(
...
+                )
+                event.add_component(Alarm.new_display(
+                    description="Team meeting in 15 minutes",
+                    trigger=timedelta(minutes=-15),
+                ))
+        """
+        if not description:
+            raise InvalidCalendar("DISPLAY alarm requires a description")
+        if trigger is None:
+            raise InvalidCalendar("DISPLAY alarm requires a trigger")
```
SashankBhamidi, line 456: `if not description` rejects empty string, which is correct. But it also rejects any falsy value, including `0` or whitespace-only strings (which `bool(" ")` is `True`, so that is fine actually), and `None`. The type annotation says `str`, so `None` should not arrive here, but since `trigger is None` is checked explicitly one line below, there is a consistency question: why guard `description` with truthiness but `trigger` with an explicit `None` check? Using `if not description` is actually the right call here since an empty description is semantically meaningless for a DISPLAY alarm, and the same pattern is used in `new_email`. I am just asking whether this was a deliberate choice, because mixing the two styles in adjacent guards looks accidental.

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,260 @@
...
+
+import pytest
+
+from icalendar.cal.alarm import Alarm
+from icalendar.error import InvalidCalendar
+from icalendar.prop import vCalAddress
+
+# ---------------------------------------------------------------------------
+# new_display
+# ---------------------------------------------------------------------------
```
SashankBhamidi: The section comments (`# ---------------------------------------------------------------------------`) are narrating structure, not explaining why. The test file already groups tests by function name prefix, and pytest output shows the group clearly. These delimiters add noise without value. Not a blocker, just a style note.

SashankBhamidi (commented):
Thank you for seeing this through across multiple rounds of feedback. The factory methods are RFC-correct, the delegation through `cls.new()` is the right architecture, and the `_apply_duration_repeat` instance method cleans up the repeated validation nicely. The doctest examples in all three methods produce exactly the output shown.

A few things I want a second set of eyes on before this merges. The `links`, `related_to`, `refids`, and `concepts` parameters appear in the signature of all three factory methods but are not documented in any of their Parameters sections. That is a gap worth filling since users discovering `new_display()` in isolation will not know these exist. The other items below are inline.

Deferring the merge decision to @niccokunzmann.

klouds27:
documented links, related_to, refids, and concepts in all three docstrings. on the guard style: if not description is deliberate — an empty string is semantically invalid for a description, so truthiness is the right check. trigger is None is explicit because None is the only invalid value there, a zero-length timedelta would be valid.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +355,307 @@ def new(
...
+        alarm: Alarm = cls.new(
+            description=description,
+            uid=uid,
+            links=links,
+            related_to=related_to,
+            refids=refids,
+            concepts=concepts,
+        )
+        alarm.add("ACTION", "DISPLAY")
```
niccokunzmann, line 465: If Alarm.ACTION does not exist, yet. We should open an issue after merge.

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def test_new_audio_with_attach():
+    uri = "ftp://example.com/pub/sounds/bell-01.aud"
+    alarm = Alarm.new_audio(timedelta(minutes=-5), attach=uri)
+    assert str(alarm["ATTACH"]) == uri
+
+
+def test_new_audio_with_empty_attach_ignored():
+    alarm = Alarm.new_audio(timedelta(minutes=-5), attach="")
```
niccokunzmann, line 118: I think ghat this is a useful test. It allows people to use type str always and convey emptyness.

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,260 @@
...
+            attendees=[],
+        )
+
+
+def test_new_audio_with_bytes_attach():
+    data = b"\x00\x01\x02\x03"
+    alarm = Alarm.new_audio(timedelta(minutes=-5), attach=data)
+    from icalendar.prop.binary import vBinary
+
+    assert isinstance(alarm["ATTACH"], vBinary)
```
niccokunzmann: This test should check for the actual value, not just type.

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def test_new_email_single_attendee():
+    alarm = Alarm.new_email(
+        summary="S",
+        description="D",
+        trigger=timedelta(minutes=-30),
+        attendees=vCalAddress("mailto:user@example.com"),
+    )
+    assert len(alarm.attendees) == 1
```
niccokunzmann, line 238: Check the value.

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,260 @@
...
+
+def test_new_email_single_attachment():
+    alarm = Alarm.new_email(
+        summary="S",
+        description="D",
+        trigger=timedelta(minutes=-30),
+        attendees=[vCalAddress("mailto:user@example.com")],
+        attachments="https://example.com/file.pdf",
+    )
+    assert "ATTACH" in alarm
```
niccokunzmann, line 250: Again, yhd value should be checked to cover functionality of the edge case.

src/icalendar/cal/alarm.py
niccokunzmann: Another follow-up issue for a new PR: attendees are always a vCalAddress. As such, we should be able to pass str.

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +355,307 @@ def new(
...
+                Corresponds to the :attr:`description` property.
+            trigger: When the alarm fires, as a :class:`~datetime.timedelta`
+                relative to the event start (negative means before) or as an
+                absolute :class:`~datetime.datetime` (recommend UTC-aware).
+            duration: Gap between repeated triggers. Must be paired with
+                ``repeat``. Corresponds to the :attr:`DURATION` property.
+            repeat: Number of *additional* times to fire after the initial
+                trigger. Must be paired with ``duration``.
+                Corresponds to the :attr:`REPEAT` property.
+            uid: Unique identifier for the alarm or ``None``.
```
niccokunzmann, line 405: uid, links, concepts and related_to should be covered by tests (did I miss one?). That can be a follow-up issue from my point of view. (TODO)

niccokunzmann (changes requested):
Hi, thanks for the PR! This looks really good to me. The only thing I would request is that the tests at the bottom of the file check the values, so the edge case is properly covered.

klouds27:
added value assertions to all five tests that were checking type or presence only. the three follow-up items you flagged (ACTION attribute, attendees accepting str, uid/links/concepts coverage) can be opened as issues after merge.

niccokunzmann (dismissed):

niccokunzmann:
@angatha @SashankBhamidi please.have another look and choose which chnages are essential.

niccokunzmann:
TODO as follow-up:

- [x] https://github.com/collective/icalendar/pull/1417#discussion_r3537179281
- [x] https://github.com/collective/icalendar/pull/1417#discussion_r3537153742
- [x] https://github.com/collective/icalendar/pull/1417#discussion_r3537077472
- [ ] validate repeat >= 0 https://github.com/collective/icalendar/pull/1417#issuecomment-4559192392
- [x] Create an issue that the attendees can also be set to a list/SEQUENCE of strings with and without `mailto:`, `new()` methods need to be adjusted.
- [ ] Complete the TODOs in https://github.com/collective/icalendar/issues/1421#issuecomment-5017802289.

stevepiercy:
@niccokunzmann @SashankBhamidi @angatha please resolve unresolved conversations that should be resolved in this issue to allow auto-merge when enabled.

stevepiercy (dismissed):

src/icalendar/cal/alarm.py
```diff
@@ -352,6 +353,253 @@ def new(
...
+        This is the most common alarm type used by calendar clients.
+
+        Conforms to :rfc:`5545#section-3.6.6`.
+
+        Parameters:
+            description: The text to display when the alarm fires.
+                Corresponds to the :attr:`description` property.
+            trigger: When the alarm fires, as a :class:`~datetime.timedelta`
+                relative to the event start (negative means before) or as an
+                absolute UTC :class:`~datetime.datetime`.
```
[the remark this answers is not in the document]
niccokunzmann: Resolving. TRIGGER is supposed to handle that.

src/icalendar/cal/alarm.py
niccokunzmann: Thanks for implementing this!

src/icalendar/cal/alarm.py
niccokunzmann: ; 'duration' and 'repeat' are both OPTIONAL,
                  ; and MUST NOT occur more than once each;
                  ; but if one occurs, so MUST the other.

So, the check for None is correct. This does not block merging. If anyone would like this more handy, they can open a separate issue.

Resolving this.

SashankBhamidi:
@klouds27,
One last thing before I re-approve. 
Can you remove the section divider comments from the test file? The test names already make the grouping clear.

If anyone else gets to this before me, feel free to merge without my approval once @klouds27 has addressed this.

angatha (dismissed):

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,265 @@
...
+
+from icalendar.cal.alarm import Alarm
+from icalendar.error import InvalidCalendar
+from icalendar.prop import vCalAddress
+from icalendar.prop.binary import vBinary
+
+# ---------------------------------------------------------------------------
+# new_display
+# ---------------------------------------------------------------------------
+
```
niccokunzmann:
```suggestion
```

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,265 @@
...
+)
+def test_repeat_without_duration_raises(alarm_fn):
+    with pytest.raises(InvalidCalendar, match="DURATION and REPEAT"):
+        alarm_fn()
+
+
+# ---------------------------------------------------------------------------
+# new_audio
+# ---------------------------------------------------------------------------
+
```
niccokunzmann:
```suggestion
```

src/icalendar/tests/attr/test_alarm_factory_methods.py
```diff
@@ -0,0 +1,265 @@
...
+    data = b"\x00\x01\x02\x03"
+    alarm = Alarm.new_audio(timedelta(minutes=-5), attach=data)
+    assert isinstance(alarm["ATTACH"], vBinary)
+    assert alarm["ATTACH"] == vBinary(data)
+
+
+# ---------------------------------------------------------------------------
+# new_email
+# ---------------------------------------------------------------------------
+
```
niccokunzmann:
```suggestion
```

niccokunzmann (commented):
Removing the comments in the test files.

src/icalendar/cal/alarm.py
niccokunzmann: I will resolve this now. The PR improves icalendar massively. If you feel precious about it, open an issue. It is probably not totally wrong to do this. Also - we can trust that if this is a small mistake, this is a chance for the community to give feedback.

niccokunzmann (approved):

src/icalendar/cal/alarm.py
niccokunzmann: Tracked in https://github.com/collective/icalendar/issues/1591

PR #1423, opened 2026-05-28
docs: replace print() with direct expressions in documentationFix/use to unicode in escape char

JuanEspinosa-design:
- [x] See #1257

## Description
Replaced `print()` calls with direct expressions in documentation examples:
- `docs/how-to/usage.rst`
- `docs/explanation/api-design.rst`

In Python doctests, values are displayed automatically without needing `print()`. 
Using `print()` is unnecessary and inconsistent with Python doctest conventions.

----
📚 Documentation preview 📚: https://icalendar--1423.org.readthedocs.build/en/1423/

### DIFF
--- a/docs/explanation/api-design.rst
+++ b/docs/explanation/api-design.rst
@@ -138,7 +138,7 @@ They are automatically added.
     ...     summary="New Year's Day Celebration",
     ...     start=date(2022, 1, 1)
     ... )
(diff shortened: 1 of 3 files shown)


niccokunzmann:
https://github.com/collective/icalendar/issues/1257#issuecomment-4565469750

PR #1424, opened 2026-05-28
Reorganize change log entry format section

stevepiercy:
- Contributors either look on the "stable" branch, where this is not clarified, or think it's OK to edit an existing news item. This PR attempts to direct readers to avoid making that mistake.
- The change log entry doesn't link to an issue, and it won't appear in the change log, as it would just create noise, since there hasn't been a recent release to include the first revision.

After approval and merge, it would be good to have another minor release to get this published on "stable".

----
📚 Documentation preview 📚: https://icalendar--1424.org.readthedocs.build/en/1424/

### DIFF
--- a/docs/contribute/index.rst
+++ b/docs/contribute/index.rst
@@ -137,7 +137,37 @@ Change log entry format
 
 To create a change log entry or news item, create a file in the :file:`news` directory, located in the root of the package.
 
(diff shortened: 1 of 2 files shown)


niccokunzmann (approved):

niccokunzmann:
Thanks for introducing towncryer! Looking forward to using it!

PR #1429, opened 2026-06-01
Fix broken Calendar.from_ical links in custom components guide

aniketaggarwal:
## Linked issue

- [x] See #1158

## Description

This pull request corrects the broken Python object links in the custom components how-to guide (`docs/how-to/custom-components.rst`). The references to `Calendar.from_ical` have been updated to point to `Component.from_ical`, which is the class where the method is defined. This resolves the Sphinx warnings generated during the build.

AI disclosure: I used Gemini 3.5 Flash to scan documentation warnings, locate the correct target definition, and draft the fix.

## Checklist

## Additional information

- Built and validated the documentation locally using `make html`.
- Verified that all warnings for `docs/how-to/custom-components.rst` are successfully resolved.

----
📚 Documentation preview 📚: https://icalendar--1429.org.readthedocs.build/en/1429/

### DIFF
--- a/docs/how-to/custom-components.rst
+++ b/docs/how-to/custom-components.rst
@@ -20,7 +20,7 @@ icalendar preserves all custom components through dynamic component creation usi
 Parse custom components
 =======================
 
-Parse custom components using either :meth:`Component.from_ical() <icalendar.cal.component.Component.from_ical>` or :meth:`Calendar.from_ical() <icalendar.cal.calendar.Calendar.from_ical>`.
+Parse custom components using either :meth:`Component.from_ical() <icalendar.cal.component.Component.from_ical>` or :meth:`Calendar.from_ical() <icalendar.cal.component.Component.from_ical>`.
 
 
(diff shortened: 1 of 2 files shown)


stevepiercy (changes requested):
You successfully found an issue. However, this pull request doesn't fix the original issue because it links to the wrong Python object.

The issue is that `icalendar.cal.calendar.Calendar.from_ical` is not a documented method, so it can't be linked.

It should be documented, but I'm not sure how to proceed. It could be something with overloading and how Sphinx has trouble with that. I'll need help from a maintainer to figure it out, unless you can propose a solution. Please let me know. Thank you!

stevepiercy:
I created a new issue to track what you found. Thanks for bringing this up and contributing your finding. However, I would prefer to keep this link broken until it is documented, and then it will link correctly. I'm closing this PR since it doesn't fix the underlying issue.

If you'd like to work on other issues, please do so!

PR #1435, opened 2026-06-03
feat: let Alarm.uid fall back to vendor UID properties (X-EVOLUTION-ALARM-UID)

### LINKED ISSUE #1421: Make all properties available to avoid additional loop
## Describe the bug

On the example https://icalendar.readthedocs.io/en/latest/how-to/usage.html#read-calendar-from-file, we can see see some properties are available, for example the summary that can be accessed like that:
`event.get("SUMMARY")`
And the trigger property that can be accessed like that (https://icalendar.readthedocs.io/en/latest/reference/api/icalendar.cal.event.html#icalendar.cal.event.Event.alarms):
`event.alarms.times[0].trigger`
However it looks like we cannot access the ACTION property to check for example if the value is DISPLAY or AUDIO. If I understand correctly, this would require an additional loop using the function property_items.

## Expected behavior

To avoid unnecessary processing and battery usage, especially on Linux smartphones that don't have a long autonomy like the PinePhone, it should be possible to make all properties available or at least additional properties available on demand the first time the ICS file is parsed in the function from_ical so we don't have to create an additional loop. This could also improve performance by avoiding duplicate work.

## Environment

 - Operating system:         Mobian (Debian 13) with Phosh on the PinePhone
 - Python version:          Python 3.13.5
 - `icalendar` version:      7.1.2

## Additional context

I tested with the ICS file generated by Gnome Calendar.

mvanhorn:
## Summary

When a `VALARM` exported by Evolution / GNOME Calendar has no canonical `UID`, its only stable identifier is the vendor property `X-EVOLUTION-ALARM-UID`, so `alarm.uid` previously returned the empty-string default. This PR lets `Alarm.uid` fall back to that vendor key.

- `single_string_property` now accepts `other_name: str | list[str] | None`, so a property can fall back across an ordered list of alias keys. A single-string argument keeps working unchanged (it is normalized to a one-element list internally), and `fdel` pops every alias so deletion stays consistent.
- `Alarm.uid` now passes `["X-ALARMUID", "X-EVOLUTION-ALARM-UID"]`. Canonical `UID` still wins; then the existing `X-ALARMUID` fallback; then the new `X-EVOLUTION-ALARM-UID`. The list makes further vendor UIDs (e.g. Apple's) easy to append later.
- The shared `uid_property` used by non-alarm components is intentionally left unchanged, since `X-EVOLUTION-ALARM-UID` is alarm-specific.

## Why this matters

In #1421, Nicco Kunzmann gave a concrete directive (2026-05-28): "the lowercase UID property should consider also the evolution UID and other UIDs in the future that e.g. apple and other calendars add." The issue's sample `.ics` shows alarms whose only identifier is `X-EVOLUTION-ALARM-UID`, which `alarm.uid` could not surface. This implements that fallback while keeping the door open for additional vendor aliases.

## Testing

- Added a `TestUidVendorFallback` test class in `test_consolidation_attr_functions.py` covering: the Evolution fallback, canonical `UID` winning over the vendor key, the empty-string default when nothing is set, deletion (`del alarm.uid` and `alarm.uid = None`) removing both canonical and vendor keys, and a regression that the existing `X-ALARMUID` fallback still works.
- `python3 -m pytest src/icalendar/tests/test_consolidation_attr_functions.py` — 24 passed.
- Ran the UID, alarm, RFC 9074, and equality suites locally; all pass. `ruff format --check` and `ruff check` are clean on the changed files.

Fixes #1421

AI was used for assistance.

----
📚 Documentation preview 📚: https://icalendar--1435.org.readthedocs.build/en/1435/


news/1421.feature
```diff
@@ -0,0 +1 @@
+The :attr:`Alarm.uid <icalendar.cal.alarm.Alarm.uid>` accessor now falls back to vendor-specific UID properties (the existing ``X-ALARMUID`` and the newly added ``X-EVOLUTION-ALARM-UID``) when no canonical ``UID`` is present, so alarms exported by Evolution/GNOME Calendar expose a usable identifier. ``single_string_property`` now accepts an ordered list of fallback keys so further vendor aliases can be added later. AI disclosure: I used GPT-5 Codex (via the Codex CLI) to draft and refine this change and its tests; I reviewed and validated the output locally. @mvanhorn (`Issue #1421 <https://github.com/collective/icalendar/issues/1421>`_)
```
stevepiercy: With towncrier, by virtue of the filename, a link is generated to the issue automatically.

```suggestion
The :attr:`Alarm.uid <icalendar.cal.alarm.Alarm.uid>` accessor now falls back to vendor-specific UID properties (the existing ``X-ALARMUID`` and the newly added ``X-EVOLUTION-ALARM-UID``) when no canonical ``UID`` is present, so alarms exported by Evolution/GNOME Calendar expose a usable identifier. ``single_string_property`` now accepts an ordered list of fallback keys so further vendor aliases can be added later. AI disclosure: I used GPT-5 Codex (via the Codex CLI) to draft and refine this change and its tests; I reviewed and validated the output locally. @mvanhorn
```

stevepiercy (changes requested):
@mvanhorn thank you for your contribution. This looks excellent, with only a minor change needed on the change log entry.

I'd like to get one other review from a maintainer, since this fixes a bug.

mvanhorn:
Applied your suggestion to the change log entry - thanks for the towncrier pointer.

stevepiercy (approved):
This LGTM. Could we get a second review from a maintainer? @niccokunzmann @angatha @SashankBhamidi whenever you have some free cycles. Thank you!

niccokunzmann (approved):

mvanhorn:
Much appreciated @niccokunzmann! Glad Alarm.uid can now fall back to the vendor UID properties.

mvanhorn:
Appreciate you taking this, @niccokunzmann. The vendor UID fallback should make Alarm.uid much more dependable across producers.

PR #1436, opened 2026-06-03
Deprecate implicit string file-path reading in Component.from_ical

uwezkhan:
Component.from_ical now warns when you hand it a file path as a string. It still reads the file as before, but this is deprecated: in icalendar 8 a string will always be parsed as calendar data, never treated as a path. If you want to read from a file, pass a pathlib.Path.

----
📚 Documentation preview 📚: https://icalendar--1436.org.readthedocs.build/en/1436/

### DIFF
--- a/news/1362.removal
+++ b/news/1362.removal
@@ -1,0 +1,1 @@
+``Component.from_ical`` now warns when you hand it a file path as a string. It still reads the file as before, but this is deprecated: in icalendar 8 a string will always be parsed as calendar data, never treated as a path. If you want to read from a file, pass a :class:`pathlib.Path`. @uwezkhan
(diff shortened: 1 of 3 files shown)


stevepiercy:
@uwezkhan thanks for leading the charge. However, there needs to be a discussion about whether to deprecate a feature before it's implemented. I propose that you create a new issue referencing this PR, as well as the related discussion in https://github.com/collective/icalendar/pull/1362/, where the maintainers can discuss and evaluate the proposed deprecation. Would that work for you? Please let me know. Thank you!

uwezkhan:
Thanks for the suggestion @stevepiercy . That sounds reasonable to me.

I will open an issue to discuss the idea separately and reference both this PR and the discussion in #1362 so the maintainers can decide whether this is something worth pursuing before any implementation changes.

PR #1442, opened 2026-06-08
Accept the (dt, None) period form when adding RDATE/EXDATE (closes #1439)

### LINKED ISSUE #1439: [Bug] event.add('RDATE', (dt, None)) raises exception
## Describe the bug

Calling `EVENT.add` with `'RDATE'` as first argument and a tuple as second argument where the first element is a `datetime` and the second id `None` raises an exception.

## To reproduce

```python
from datetime import datetime
from icalendar import Event
event = Event.new()
event.add('RDATE', (datetime(2025, 4, 28, 16, 5), None))
```

Output:

```pycon
  File "/home/ke/code/icalfilter/mwe.py", line 4, in <module>
    event.add('RDATE', (datetime(2025, 4, 28, 16, 5), None))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ke/code/icalfilter/.venv/lib/python3.14/site-packages/icalendar/cal/component.py", line 327, in add
    value = self._encode(name, value, parameters, encode)
  File "/home/ke/code/icalfilter/.venv/lib/python3.14/site-packages/icalendar/cal/component.py", line 270, in _encode
    obj = klass(value)
  File "/home/ke/code/icalfilter/.venv/lib/python3.14/site-packages/icalendar/prop/dt/list.py", line 28, in __init__
    dt = vDDDTypes(dt_l) if not isinstance(dt_l, vDDDTypes) else dt_l
         ~~~~~~~~~^^^^^^
  File "/home/ke/code/icalfilter/.venv/lib/python3.14/site-packages/icalendar/prop/dt/types.py", line 47, in __init__
    raise TypeError(
        "You must use datetime, date, timedelta, time or tuple (for periods)"
    )
TypeError: You must use datetime, date, timedelta, time or tuple (for periods)
```

## Expected behavior

This should behave the same as passing just the datetime instead of the tuple.

## Environment

 - [X] Operating system:         Ubuntu 24.04
 - [X] Python version:           Python 3.14
 - [X] `icalendar` version:      7.1.2

## Additional context

The `(dt, None)` format should be accepted because it is the format that `Event.rdates` uses.

To be sure, it is easy to work around in client code, but it shouldn't be necessary.

- [X] I tested it with the latest version. For example, I used the command `pip3 install --upgrade https://github.com/collective/icalendar.git`.
- [X] I attached the `.ics` sample file, or there is no `.ics` sample file.

gaoflow:
Closes #1439.

### Problem

`Event.rdates` / `Event.exdates` return each value as a `(dt, end)` tuple, where a plain date is `(dt, None)`. But feeding that form back into `add()` raised:

```python
event.add('RDATE', (datetime(2025, 4, 28, 16, 5), None))
# TypeError: You must use datetime, date, timedelta, time or tuple (for periods)
```

Two layers were involved:

- `vDDDLists.__init__` treats any iterable as a *list of values*, so the bare `(dt, None)` tuple was iterated element-wise into `dt` and `None`, and `vDDDTypes(None)` raised.
- The list form `[(dt, None)]` (exactly what `rdates` returns) got as far as a `vDDDTypes` with `VALUE=PERIOD` holding `(dt, None)`, which then failed when serialised (`end_or_duration MUST be a datetime, date or timedelta instance`).

As discussed in the issue, `(dt, None)` is a period with no end, i.e. just the date, and @stevepiercy / @niccokunzmann agreed it should be supported so the value `rdates` produces round-trips.

### Fix

Treat a 2-tuple whose second element is `None` as the bare first value, in both places:

- `vDDDTypes.__init__` normalises `(dt, None)` → `dt` (so it is a plain DATE/DATE-TIME, not an unserialisable PERIOD).
- `vDDDLists.__init__` no longer iterates a bare `(dt, None)` tuple element-wise, wrapping it as a single value instead.

The guard is specifically `len == 2 and [1] is None`, so the existing meaning of `(dt1, dt2)` (two separate values) and `[(dt1, dt2)]` (a period) is unchanged — only the previously-erroring `None` case is affected.

### Verification

- New tests in `tests/attr/test_rdate.py` (parametrised over Event/Todo/Journal/Timezone components): `(dt, None)` and `[(dt, None)]` now serialise identically to the bare date; `event.add('RDATE', event.rdates)` round-trips; and `(dt1, dt2)` still yields two values. The first two fail on `main`.
- Full suite green: 9697 passed, 735 skipped. `ruff check` clean.
- EXDATE (same machinery) is fixed too.

Changelog: `news/1439.bugfix`.

----
📚 Documentation preview 📚: https://icalendar--1442.org.readthedocs.build/en/1442/


src/icalendar/prop/dt/list.py
```diff
@@ -20,7 +20,11 @@ class vDDDLists:
     def __init__(self, dt_list, params: dict[str, Any] | None = None):
         if params is None:
             params = {}
-        if not hasattr(dt_list, "__iter__"):
+        if not hasattr(dt_list, "__iter__") or (
+            isinstance(dt_list, tuple) and len(dt_list) == 2 and dt_list[1] is None
```
niccokunzmann: Does it work without the check for tuple? Would sequence be ok? That also allows lists.

src/icalendar/prop/dt/list.py
```diff
@@ -20,7 +20,11 @@ class vDDDLists:
...
+        ):
+            # A bare ``(dt, None)`` tuple is a single date (the rdates form), not a
+            # list of two values, so don't iterate it element-wise (#1439).
             dt_list = [dt_list]
```
niccokunzmann, line 29: Hm. Should there be dt_list[0] somewhere?

src/icalendar/prop/dt/types.py
```diff
@@ -43,6 +43,10 @@ class vDDDTypes(TimeBase):
     def __init__(self, dt, params: dict[str, Any] | None = None):
         if params is None:
             params = {}
+        if isinstance(dt, tuple) and len(dt) == 2 and dt[1] is None:
```
niccokunzmann: Again, check for sequence, please.

src/icalendar/tests/attr/test_rdate.py
```diff
@@ -133,3 +133,36 @@ def test_get_example_4(calendars, tzp):
...
     assert event.rdates == RDATES_4
+
+
+def test_add_rdate_with_none_end_is_a_single_date(c_rdate):
+    """``add('RDATE', (dt, None))`` is the rdates form of a single date and must
+    behave like adding the bare date, not raise (#1439)."""
+    dt = datetime(2025, 4, 28, 16, 5)
+    bare = c_rdate.__class__()
+    bare.add("RDATE", dt)
+    for value in ((dt, None), [(dt, None)]):
```
niccokunzmann: Please use pytest.parametrize and split up the test, so only assert is at the end.

src/icalendar/tests/attr/test_rdate.py
```diff
@@ -133,3 +133,36 @@ def test_get_example_4(calendars, tzp):
...
     pprint(RDATES_4)
     assert event.rdates == RDATES_4
+
+
+def test_add_rdate_with_none_end_is_a_single_date(c_rdate):
+    """``add('RDATE', (dt, None))`` is the rdates form of a single date and must
+    behave like adding the bare date, not raise (#1439)."""
+    dt = datetime(2025, 4, 28, 16, 5)
+    bare = c_rdate.__class__()
+    bare.add("RDATE", dt)
```
niccokunzmann: It would be good to check setting rdates, too.

Also, tests should be added for the change in the vDDDLists and vDDDTypes.

niccokunzmann (changes requested):
Thanks for taking this on! I added some comments to make sure, we test this in a way that it does not get broken in the future. It is a little complex.

One could test for Sequence instead of tuple but it is probably best to use tuple for now as you suggest. It makes it very clean and clear.

src/icalendar/prop/dt/list.py
binggao1230: Without the tuple check it breaks: `(dt, None)` is iterable, so vDDDLists walks it element-wise and hits `vDDDTypes(None)`, which raises — that's the original bug. I kept it `tuple` on purpose: `(dt, None)` is exactly the form `.rdates`/`.exdates` return for a single date, whereas a plain list `[dt, None]` would be ambiguous with an ordinary two-value list. That matches your summary — tuple keeps it unambiguous.

src/icalendar/prop/dt/list.py
binggao1230: The `[0]` effectively happens one level down: this wraps to `[(dt, None)]`, and each element then goes through `vDDDTypes`, which now unwraps `(dt, None)` -> `dt` (verified: `vDDDLists((dt, None)).dts == [dt]`, no `None` leaks). I kept that unwrap in vDDDTypes only, so the list path and a direct `vDDDTypes((dt, None))` stay consistent instead of duplicating the check — the two new unit tests cover both.

src/icalendar/prop/dt/types.py
binggao1230: Kept `tuple` here too, per your summary ("probably best to use tuple for now"). It's tuple-specific on purpose: `(dt, None)` is the period-with-no-end form `.rdates` emits; widening to any `Sequence` would also catch `[dt, None]`, which collides with a normal two-value list. Happy to switch to `Sequence` if you'd rather.

src/icalendar/tests/attr/test_rdate.py
binggao1230: Done in 669ed492 — parametrized the `(dt, None)` / `[(dt, None)]` forms and split them into one-assert tests.

src/icalendar/tests/attr/test_rdate.py
binggao1230: Done in 669ed492: added `test_vdddtypes_treats_none_end_tuple_as_a_bare_date` and `test_vdddlists_treats_none_end_tuple_as_a_single_date` covering the two changes directly, plus the single-date tests assert `.rdates` after adding. Your point about testing it robustly was spot on — the `.rdates`-only assertions pass even on the buggy code (the failure only surfaces at `to_ical()`), so the real guards are the serialization and unit-level checks.

src/icalendar/prop/dt/list.py
binggao1230: Switched to `collections.abc.Sequence` in eaefb3d7, so a plain `[dt, None]` list now works the same as the tuple. The `len == 2 and [1] is None` guard keeps strings and genuine two-value pairs out.

src/icalendar/prop/dt/list.py
binggao1230: Still handled one level down: the pair wraps to `[(dt, None)]` and `vDDDTypes` unwraps it to `dt`, so no `None` leaks. Keeping the unwrap in `vDDDTypes` only avoids duplicating it here — the parametrized unit tests now cover both the tuple and list forms.

src/icalendar/prop/dt/types.py
binggao1230: Done in eaefb3d7 — `vDDDTypes` checks `Sequence` now, so `[dt, None]` unwraps just like `(dt, None)`.

niccokunzmann (changes requested):
Thanks for addressing this so cleanly!

For completeness, since the issue is for setting .rdates, please add a test that checks (1) that setting rdates still works - with a list that contains all the different types of values that we can have in there.
(2) in a second part of that test, that setting `component.rdates = component.rdates[:]` does not change the value, length and content remain correct.

binggao1230:
All review points are addressed (Sequence support, parametrized single-assert tests) and squashed into the disclosed commit — ready for another look.

binggao1230:
Added that test — `test_set_all_rdate_value_types_then_round_trip`, parametrized over all five RDATE components. Part 1 sets one of every accepted value type (bare date; naive, UTC and zoned datetime; a period; a period given as `(start, duration)`; and the `(dt, None)` single-date form) and checks the resulting `.rdates`. Part 2 re-adds `c.rdates[:]` and asserts value, length and content are unchanged.

One note on the wording: `.rdates` is a read-only property (its docstring points to `.add`), so a literal `component.rdates = component.rdates[:]` raises `AttributeError`. I tested the same round-trip through `.add`, which is the supported way to set RDATE — happy to add an actual `.rdates`/`.exdates` setter if you want one, but that felt like a separate change from #1439.

Also rebased onto `main`. (AI disclosure as before: produced with Claude Opus 4 under my account and responsibility; I validate the result before submitting.)

binggao1230:
Added `test_set_all_rdate_value_types_then_round_trip` (in 695b4754) covering both parts you asked for:

1. Sets `RDATE` with one of every accepted value type in a single list — bare date, naive datetime, UTC datetime, zoned datetime, `(start, end)` period, `(start, duration)` period, and the `(dt, None)` single-date form — and asserts the full `.rdates` output.
2. Feeds that output straight back (`restored.add("RDATE", c_rdate.rdates[:])`) and asserts value, length and content are unchanged.

One note on the exact form: `.rdates` is a read-only property ("You cannot modify the RDATE value by modifying the result. Use `Component.add`"), so part (2) round-trips via `.add(...)` rather than `component.rdates = component.rdates[:]`. If you'd like assignment to work too, I'm happy to add a real `rdates` setter as a follow-up.

binggao1230:
Thanks for the review! One clarification before I add this.

`.rdates` (and `.exdates`) are currently read-only properties (getter only), and their docstring directs callers to `Component.add(...)`, so `component.rdates = component.rdates[:]` is not possible yet on `main`.

The PR already includes `test_set_all_rdate_value_types_then_round_trip`, which covers both halves you describe, but expressed via `add`: it sets one of every accepted value type (bare date, naive / UTC / zoned datetime, a period, a period given as `(start, duration)`, and the `(dt, None)` single-date form) in a single list, asserts the resulting `.rdates`, then feeds `component.rdates[:]` straight back in and asserts value, length, and content are unchanged.

To match your exact `component.rdates = ...` form there are two options:

- **A.** Add a `.rdates` / `.exdates` setter with replace semantics (mirroring the existing `categories` property), so `component.rdates = component.rdates[:]` works literally. This is a small public-API addition on top of the bug fix, and is the natural capstone of making `.rdates` output round-trippable.
- **B.** Keep RDATE add-only and write the round trip as `component.add("RDATE", component.rdates[:])`, as the current test does.

I lean toward A, but since it widens this PR's scope from a pure bug fix to a new setter I wanted your call first. Happy to push A here if you prefer.

niccokunzmann:
> This is a small public-API addition on top of the bug fix, and is the natural capstone of making .rdates output round-trippable

Hi. Ah. I did not realize we did not have this, yet. Yes, this would be woulderful and very useful. Are you up for that? del should be added, too, e.g. just setting it it []. Would you like to work on that? Maybe, this is good for a new PR? As you like!

binggao1230:
Yes, done. I added writable `rdates` and `exdates` in f7ffb9f: assigning a list replaces the values, `del event.rdates` (or assigning `[]` / `None`) clears them, and assigning the getter's own output round-trips. Tests cover set, round-trip across every value type, del, and the empty cases for both properties.

I folded it into this PR since it is the natural capstone of the round-trip work here, but I am happy to split it into a separate PR if you would rather keep the bug fix isolated. Just let me know.

binggao1230:
Added the two-part test you asked for in deaa88b5:

- \`test_set_rdates_via_property_all_value_types\` sets \`.rdates\` through the property with a list holding one of every accepted value type (bare date, naive/UTC/zoned datetimes, a period, a (start, duration) period, and the (dt, None) single-date form) and asserts they all come back correctly.
- \`test_set_rdates_via_property_its_own_output_round_trips\` then does \`component.rdates = component.rdates[:]\` and asserts value, length and content are unchanged.

Both run across Event, Todo, Journal and the two timezone components via the existing fixture.

niccokunzmann (commented):
Perfect! Thanks!

niccokunzmann (commented):
Perfect! Thanks!

niccokunzmann (approved):

PR #1443, opened 2026-06-08
reject non-finite float values in vFloat and vGeo from_ical

alhudz:
Repro: parse any calendar carrying `GEO:1e999;0.0` (or `FLOAT:nan`), then call `to_json()`.
Expected: the malformed value is rejected at parse time, same as `1361.bugfix`/`1349.bugfix` do for date-times and binary.
Actual: `float()` happily returns `inf`/`nan`, which `Component.to_json()` emits as bare `Infinity`/`NaN` tokens that no conformant JSON parser accepts.
Fix: reject non-finite results inside both `from_ical` callees so the parse fails instead of producing an out-of-range value.

Covers both float-parsing sites (`vFloat` and `vGeo`); the same input reaches `vFloat.from_ical` for plain `FLOAT` properties and `vGeo.from_ical` for `GEO`.

----
📚 Documentation preview 📚: https://icalendar--1443.org.readthedocs.build/en/1443/


src/icalendar/prop/geo.py
```diff
@@ -104,7 +105,10 @@ def ical_value(self) -> tuple[float, float]:
     def from_ical(ical: str) -> tuple[float, float]:
         try:
             latitude, longitude = ical.split(";")
-            return (float(latitude), float(longitude))
+            latitude, longitude = float(latitude), float(longitude)
+            if not (math.isfinite(latitude) and math.isfinite(longitude)):
+                raise ValueError
```
niccokunzmann: Please add an error message.

niccokunzmann (commented):
Thanks! I have one suggestion.

src/icalendar/prop/geo.py
alhudz: Done. The inner raise now carries `Expected finite 'float;float', got: {ical}`, which shows up as the cause of the outer `ValueError`.

src/icalendar/prop/float.py
```diff
@@ -69,9 +70,12 @@ def to_ical(self) -> bytes:
     @classmethod
     def from_ical(cls, ical: str | float) -> Self:
         try:
-            return cls(ical)
+            self = cls(ical)
         except Exception as e:
             raise ValueError(f"Expected float value, got: {ical}") from e
+        if not math.isfinite(self):
+            raise ValueError(f"Expected float value, got: {ical}")
```
angatha: Can you please include `"finite"` in the error message.

alhudz: Done, it now reads `Expected finite float value, got: {ical}`, matching the `vGeo` message.

angatha (dismissed):

stevepiercy:
@alhudz would you please address the CI failure for ruff?

https://github.com/collective/icalendar/actions/runs/27183154608/job/80246389627?pr=1443

alhudz:
Done. ruff was flagging `TRY300`/`TRY301` on the new `raise`/`return` sitting inside the `try` in `vGeo.from_ical`. Moved the finiteness check out of the `try`, same shape as `vFloat.from_ical`, so the `finite` message is now raised directly rather than wrapped. `ruff check` and `ruff format .` are both clean locally.

stevepiercy (approved):
Ruff is happy now, and so am I for your contribution. Thank you!

alhudz:
Glad to hear it! Thanks for the review and for merging the contribution.

PR #1444, opened 2026-06-08
fix(from_ical): treat unusable path strings as calendar data, not OS …

uwezkhan:
Component.from_ical() probes a string as a file path when it has no line breaks. The probe only caught OSError, so a newline-free string the OS rejects as a path — one with an embedded null byte, or one that's too long — leaked a raw OSError (or a low-level "embedded null byte" ValueError) instead of being parsed as calendar data.

This widens the probe to also swallow ValueError, so any string that can't be a path on the current platform falls through to the parser and produces a consistent ValueError everywhere (Windows/macOS/Linux). Behavior for real calendar data and real file paths is unchanged.

Adds tests pinning the str-vs-path boundary, and a comment explaining why the newline check is safe (valid iCalendar is always CRLF-folded per RFC 5545, so calendar content never reaches the file-probe branch):

a valid calendar string never touches the filesystem (is_file/read_bytes are never called),
invalid / non-existent string paths raise ValueError, never OSError,
a missing pathlib.Path raises FileNotFoundError — proving a Path is always treated as a path.
Closes #1436. Related: #756, #1362.

----
📚 Documentation preview 📚: https://icalendar--1444.org.readthedocs.build/en/1444/


angatha (dismissed):
LGTM

news/1436.bugfix
```diff
@@ -0,0 +1 @@
+Treat a newline-free string that the operating system cannot use as a file path (for example, one containing an embedded null byte or one that is too long) as calendar data rather than letting an :exc:`OSError` propagate from :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`. Such input now raises a consistent :exc:`ValueError` across platforms, and the str-vs-path boundary is covered by tests. @uwezkhan
```
stevepiercy: This flows a little better, placing the subject before the verb.

```suggestion
For a newline-free string that the operating system can't use as a file path—for example, one containing an embedded null byte or one that is too long—treat it as calendar data, instead of propagating an :exc:`OSError` from :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`. Such input now raises a consistent :exc:`ValueError` across platforms. The string-versus-path boundary is now covered by tests. @uwezkhan
```

stevepiercy:
> Closes #1436. Related: #756, #1362.

@uwezkhan when we merge this PR, GitHub will automatically close your old PR #1436. Is that what you want?

If you want to keep #1436 open, then please change the description to `See #1436`. Please let us know, either way. Thank you!

stevepiercy (changes requested):
A minor wording tweak on the news fragment. Otherwise good to merge. Thank you for your contribution!

uwezkhan:
Yes, that's what I intended. #1444 addresses the follow-up work that came out of the discussion in #1436, so I'm happy for GitHub to close #1436 automatically when this PR is merged.

stevepiercy (approved):
Thank you for improving icalendar! Will merge when CI passes.

PR #1446, opened 2026-06-09
strictly validate vDate and vTime values in from_ical

alhudz:
## Linked issue

No associated issue.

## Description

`Repro:` parse `X-PROP;VALUE=TIME:120000GARBAGE`, or `DTSTART;VALUE=DATE:19_70714`.
`Expected:` a `ValueError` for the malformed value.
`Actual:` `vTime.from_ical` returns `time(12, 0)` and `vDate.from_ical` returns `date(197, 7, 14)` — trailing data is dropped and `int()` happily accepts underscores, whitespace and signs.
`Cause:` both `from_ical` callees slice straight into `int()` with no length or format check. `VALUE=TIME` maps directly to `vTime` in the type factory, and 8-character malformed `DATE` values clear the `vDDDTypes` length guard and reach `vDate`.
`Fix:` require the exact RFC digit count (8 for `DATE`, 6 for `TIME` after the optional `Z`) and reject anything that isn't ASCII digits, matching the strictness `vDatetime.from_ical` already applies.

## Checklist

## Additional information

Both forms reproduce through the public `Component.from_ical` path; `vUTCOffset` is left untouched as it keeps a deliberately looser contract (unsigned offsets like `0000` are accepted and tested).

----
📚 Documentation preview 📚: https://icalendar--1446.org.readthedocs.build/en/1446/


stevepiercy:
@alhudz would you please?

- Resolve the CI failure with Ruff
- Update your branch against `main`
- Review our [AI policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) and disclose your usage?

Then we can begin a review. Thank you!

alhudz:
Done on all three:

- Ruff `TRY301` is resolved. The check in `vTime.from_ical` was raising inside the `try`; I hoisted the length/ASCII-digit guard out in front of it, matching how `vDate.from_ical` is already structured. `ruff check` is clean locally.
- Branch is updated against `main`.
- Read the AI policy. To disclose: yes, I used AI on this ( Claude Code) to dig through the RFC 5545 date/time grammar and to draft the validation and regression tests. I reviewed everything and ran the suite locally before pushing. I've added the disclosure to the commit message and the change log entry as the policy asks.

Happy to adjust the wording of the disclosure if you'd prefer it in a different form.

src/icalendar/prop/dt/date.py
```diff
@@ -76,6 +77,11 @@ def to_ical(self):
 
     @staticmethod
     def from_ical(ical):
+        ical = to_unicode(ical)
```
stevepiercy: Why is `to_unicode` and its import necessary?

src/icalendar/prop/dt/time.py
```diff
@@ -173,12 +173,16 @@ def from_ical(ical: str, timezone: str | None | tzinfo = None) -> time:
...
 
+        if isinstance(ical, bytes):
+            ical = ical.decode()
+        utc = ical.endswith("Z")
+        if utc:
+            ical = ical[:-1]
+        # time = time-hour time-minute time-second [time-utc], six digits,
+        # per https://datatracker.ietf.org/doc/html/rfc5545#section-3.3.12
+        if len(ical) != 6 or not ical.isascii() or not ical.isdigit():
+            raise ValueError(f"Expected time, got: {ical}")
```
stevepiercy, line 187: This will raise a ValueError for the valid Form #3, per https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.12:

```
TZID=America/New_York:083000
```

Would you please account for that third form? I recall we did something for the datetime type to handle this as well.

stevepiercy (changes requested):
Thanks for starting work on this improvement! There are a couple of things that I'd like you to address. Would you please take care?

src/icalendar/prop/dt/date.py
alhudz: You're right, it isn't. The original `vDate.from_ical` never decoded, and the `len`/`isascii`/`isdigit`/`int` guard all accept `bytes` anyway, so it was redundant. Dropped the call and the import.

src/icalendar/prop/dt/time.py
alhudz: Good catch. `vTime.from_ical` now does `ical.rpartition(":")[2]` before the length check, the same way `vDatetime.from_ical` pulls the value out of Form #3, so `TZID=America/New_York:083000` parses to `time(8, 30)` instead of raising. Added a regression test for it.

src/icalendar/prop/dt/time.py
```diff
@@ -173,12 +173,20 @@ def from_ical(ical: str, timezone: str | None | tzinfo = None) -> time:
...
             tzinfo = timezone
 
+        if isinstance(ical, bytes):
+            ical = ical.decode()
+        # Extract the value part if parameters are present per
+        # https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.5
+        # Form #3: TZID=America/New_York:083000
+        ical = ical.rpartition(":")[2]
+        utc = ical.endswith("Z")
+        if utc:
```
stevepiercy: These two steps can be combined into one to simplify.

stevepiercy (changes requested):
One small optimization, then I approve. Would you please take care of this?

I'd like to get a second approval from a maintainer, too. Thank you!

src/icalendar/prop/dt/time.py
alhudz: Done, combined into `if utc := ical.endswith("Z")`. Tests and ruff are clean locally.

src/icalendar/prop/dt/date.py
```diff
@@ -76,6 +76,10 @@ def to_ical(self):
 
     @staticmethod
     def from_ical(ical):
+        # date-value = 4DIGIT 2DIGIT 2DIGIT, no separators,
+        # per https://datatracker.ietf.org/doc/html/rfc5545#section-3.3.4
+        if len(ical) != 8 or not ical.isascii() or not ical.isdigit():
+            raise ValueError(f"Wrong date format {ical}")
```
stevepiercy, line 82: After further review, I think these lines aren't necessary. The next try/except block tries to return a date, and if it's not a date, then it will raise the exact same `ValueError` and message. Try removing these additions to see if the tests you added in `src/icalendar/tests/prop/test_unit.py` still pass.

stevepiercy: Also if it turns out that these lines aren't needed, would you please remove them and update your change log entry accordingly? The tests you added should remain, though. Thank you!

stevepiercy (changes requested):

src/icalendar/prop/dt/date.py
alhudz: Tried it. With those four lines removed, `test_prop_vDate` fails on the first new assertion:

```
>       self.assertRaises(ValueError, vDate.from_ical, "20010102T")
E       AssertionError: ValueError not raised by from_ical
```

The try/except only raises when `int()` itself raises, and `int()` is more forgiving than the RFC grammar: it strips whitespace, accepts a leading sign, and allows underscores between digits, while the `[:4]`/`[4:6]`/`[6:8]` slicing silently drops anything past index 8. So most malformed values convert cleanly and never reach the `except`:

- `'20010102T'` -> `date(2001, 1, 2)` (trailing data sliced off)
- `'2001 102'` -> `date(2001, 1, 2)` (whitespace stripped by `int()`)
- `'+0010102'` -> `date(1, 1, 2)` (leading sign accepted)
- `'19_70714'` -> `date(197, 7, 14)` (underscore between digits)

That silent acceptance is the bug this PR fixes, so the guard and the change log entry stay as they are.

stevepiercy: Thanks for taking care. Let's keep it as is.

stevepiercy (approved):
LGTM. Thank you for improving icalendar!

Let's get a second approval from a maintainer.

niccokunzmann (approved):
Thanks for the PR!

PR #1449, opened 2026-06-10
fix quadratic content-line unfolding and whitespace stripping

alhudz:
## Linked issue

No linked issue.

## Description

`Repro:` parse a tiny calendar carrying a long run of bare line breaks, or a parameter section padded with spaces around `;`/`=`, e.g. `Calendar.from_ical("BEGIN:VCALENDAR\r\n" + "\n"*40000 + "END:VCALENDAR\r\n")`.

`Cause:` `UFOLD = (\r?\n)+[ \t]` and `OWS_AROUND_DELIMITERS_RE = [ \t]*([;=])[ \t]*` are applied with `re.sub`, so the greedy run is re-tried at every position of the run. A failed attempt rescans the whole remaining run, which is `O(n^2)`. ~32 KB of newlines already takes ~12 s in `Contentline`/`Contentlines.from_ical`, and ~64 KB of spaces takes ~16 s in `_strip_ows_around_delimiters`. Both sit on the `Calendar.from_ical` path, so they are reachable from untrusted iCalendar input.

`Fix:` pin each match to the start of its run. `UFOLD` becomes a lookbehind-anchored equivalent; the whitespace stripper splits into two anchored passes (leading then trailing). Both match exactly the same strings as before, checked over 500k–1M random inputs with zero differences, and now run in linear time (the 40 KB cases above drop to ~3 ms).

Benchmark of the unfold regex on a run of bare newlines:

```
n        before     after
8000     0.76 s     <0.01 s
16000    3.0 s      <0.01 s
32000    12.2 s     <0.01 s
64000    ~49 s      <0.01 s
```

## Checklist

## Additional information

The two replacement regexes were validated for exact equivalence against the originals over random fuzzed inputs; the new regression test also asserts an adversarial input parses well under the old quadratic time.

----
📚 Documentation preview 📚: https://icalendar--1449.org.readthedocs.build/en/1449/


stevepiercy:
@alhudz would you please resolve the CI failure for ruff?

https://github.com/collective/icalendar/actions/runs/27255823045/job/80489800379?pr=1449

news/+linear-content-line-unfolding.bugfix
```diff
@@ -0,0 +1 @@
+Parse content lines in linear time. The unfolding and whitespace-stripping regexes in :mod:`icalendar.parser.content_line` re-tried a greedy run at every position, so a small calendar carrying a long run of bare line breaks or of spaces around a parameter delimiter made :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>` take seconds. The patched regexes match the same strings without the quadratic blow-up. @alhudz
```
stevepiercy: Log entries should be brief. They should address users of the software so they can understand the problem and its solution. Please see our guide [Write a good change log entry](https://icalendar.readthedocs.io/en/stable/contribute/index.html#write-a-good-change-log-entry).

```suggestion
For iCalendar files that have thousands of bare line breaks, reduce the content line parse time by a quadratic order of magnitude when reading a component with :meth:`Component.from_ical <icalendar.cal.component.Component.from_ical>`. @alhudz
```

stevepiercy (changes requested):
Thank you for your contribution! This looks good, but the change log entry is a wee bit too nerdy. Would you please see my suggestion?

news/+linear-content-line-unfolding.bugfix
alhudz: Makes sense, applied your wording as-is.

alhudz:
Done. `ruff format` wanted the test file's long `ics` string wrapped; fixed that and took your change log wording. `ruff format --check` and `ruff check` both pass locally now.

stevepiercy (approved):
Thank you for updating the news fragement.

I'd like a second technical review from a maintainer before merging.

I also wonder whether this should have gone through a security process. See https://github.com/collective/icalendar/blob/main/SECURITY.md.

stevepiercy:
@angatha @niccokunzmann @SashankBhamidi when you have a moment, would you please take a look at this PR? Thank you!

niccokunzmann (approved):
The regex and functionality are equivalent. I trust that the speedup happens. I understand what it does. Thanks! That is a learning!

PR #1450, opened 2026-06-10
Preserve RFC 7265 unknown property values verbatim

### LINKED ISSUE #1445: [Bug] `unknown` value type is escaped/unescaped instead of kept verbatim, breaking round-trips
## Describe the bug

When a property's value type is `unknown` (unrecognized properties, including
`X-` properties without a `VALUE` parameter), the value seems to be run through
RFC 5545 TEXT escaping rules instead of being kept as written. As I read
[RFC 7265](https://datatracker.ietf.org/doc/html/rfc7265.html#autoid-32) §5.1/§5.2, an `unknown` value should be passed through "without
processing."

The problem does not look limited to jCal. In my testing, the value changes in
three places:

1. **Just reading it** — the value I get back from `from_ical` is already
   different from the source, before I write anything out.
2. **iCal → iCal round-trip** — iCalendar in, iCalendar out, no jCal involved;
   backslashes are added or removed.
3. **iCal → jCal** — the case RFC 7265 §5.1 warns "breaks round-tripping
   values."

I think the jCal case (3) is a clear RFC 7265 violation. I am less sure about
the plain-iCal cases (1, 2).

## To reproduce

```python

orig = r"""BEGIN:VEVENT
UID:1
X-MYSTERY:a;b\,c\:d\ne
END:VEVENT
"""
ev = Component.from_ical(orig)

print(repr(str(ev["X-MYSTERY"])))      # just reading it
print(ev.to_ical().decode())           # iCal -> iCal
print(ev.to_jcal())                     # iCal -> jCal
```

Output:

```pycon
'a;b,c:d\ne'             # read:  \, -> ,   \: -> :   \n -> real newline
BEGIN:VEVENT
UID:1
X-MYSTERY:a\;b\,c:d\ne   # iCal:  ; -> \;   \, kept   \: -> :
END:VEVENT

['vevent', [['uid', {}, 'text', '1'], ['x-mystery', {}, 'unknown', 'a;b,c:d\ne']], []]
# jCal:  \, -> ,  (kept in iCal above, lost here -> breaks round-tripping, RFC 7265 §5.3)
```

The source value `a;b\,c\:d\ne` is preserved by none of the three. The escaped
`\,` is the clearest case — it survives the iCal → iCal round-trip but loses its
backslash in jCal. Across my tests the affected characters were `,`, `;`, `\`,
`:`, and newlines; others came through unchanged, though I have not tried every
input.

## Expected behavior

As I read RFC 7265 §5.1/§5.2 and the §5.3 example, an `unknown` value should
come through exactly as written, in both directions. For the input above that
would mean:

```pycon
>>> str(ev["X-MYSTERY"])
'a;b\\,c\\:d\\ne'
>>> print(ev.to_ical().decode())
...
X-MYSTERY:a;b\,c\:d\ne
...
>>> ev.to_jcal()
['vevent', [['uid', {}, 'text', '1'], ['x-mystery', {}, 'unknown', 'a;b\\,c\\:d\\ne']], []]
```

i.e. no RFC 5545 escaping or unescaping for `unknown` values either way (only
JSON string escaping, which the JSON layer does on its own).

The RFC 7265 §5.3 example shows the same idea — the decoded jCal value matches
the iCalendar value exactly:

| | value |
|---|---|
| iCalendar | `Stenophylla;Guinea\,Africa` |
| jCal (decoded) | `Stenophylla;Guinea\,Africa` |

## Environment

 - [x] Operating system:         Ubuntu (WSL2)
 - [x] `icalendar` version:      7.1.2

## Additional context

One case I have not worked out: a jCal `unknown` string could hold a character
(such as a raw line break) that RFC 5545 may not be able to put in a bare
value, so passing it through unchanged could clash with producing valid
iCalendar. The RFC examples do not seem to cover this.

- [x] I tested it with the latest version (current `main`).
- [x] I attached the `.ics` sample file, or there is no `.ics` sample file. (Reproducer is inline above.)

lcampanella98:
## Linked issue

- [x] Closes #1445

## Description

Values of unrecognized properties — and `X-` properties without a `VALUE`
parameter — were run through RFC 5545 TEXT escaping, so they did not round-trip
unchanged. RFC 7265 reserves the `unknown` value type precisely for these cases
and requires the value to be carried "without processing" (§5.1/§5.2). Treating
it as TEXT re-escapes it and, as the RFC warns, "breaks round-tripping values".

Example (before this change) — in iCal the `;` gains a backslash that wasn't there; in jCal the `\,` loses its backslash, where RFC 7265 [§5.3](https://datatracker.ietf.org/doc/html/rfc7265.html#section-5.3) gives the value as `"Stenophylla;Guinea\\,Africa"`, preserved exactly:

```pycon
>>> from icalendar.cal import Component
>>> orig = r"""BEGIN:VEVENT
UID:1
X-COFFEE-DATA:Stenophylla;Guinea\,Africa
END:VEVENT
"""
>>> ev = Component.from_ical(orig)
>>> print(ev.to_ical().decode())
BEGIN:VEVENT
UID:1
X-COFFEE-DATA:Stenophylla\;Guinea\,Africa
END:VEVENT
>>> ev.to_jcal()
['vevent', [['uid', {}, 'text', '1'], ['x-coffee-data', {}, 'unknown', 'Stenophylla;Guinea,Africa']], []]
```

After this change the value is preserved exactly, matching the RFC 7265 §5.3 example:

```pycon
>>> print(ev.to_ical().decode())
BEGIN:VEVENT
UID:1
X-COFFEE-DATA:Stenophylla;Guinea\,Africa
END:VEVENT
>>> ev.to_jcal()
['vevent', [['uid', {}, 'text', '1'], ['x-coffee-data', {}, 'unknown', 'Stenophylla;Guinea\\,Africa']], []]
```

### What this changes

- `unknown` values are now preserved **verbatim** when parsing, serializing, and
  converting to/from jCal. The fix skips backslash unescaping on parse and
  emits the stored value unchanged on output.
- Known types are untouched: an explicit `VALUE` (e.g. `VALUE=TEXT`) still gets
  normal TEXT escaping, and property *parameters* are still decoded as before.
  An unrecognized `VALUE` type keeps its type name so it round-trips losslessly.
- `vUnknown` no longer inherits from `vText`, so the two cannot silently share
  escaping behaviour again (per maintainer review).
- `PROXIMITY` ([RFC 9074](https://datatracker.ietf.org/doc/html/rfc9074#name-proximity-property)) is now recognized as `TEXT` rather than falling back to
  the unknown type — a latent gap this change surfaced. It had passed
  [`test_calendar_types`](https://github.com/collective/icalendar/blob/8f1c77376fa8c65eac1c53ecf29c372062b96fd9/src/icalendar/tests/test_rfc_9074.py#L18)
  only because `vUnknown` used to subclass `vText`.

### Known limitation

A jCal `unknown` value containing a raw line break cannot be represented as a
bare iCal content line (RFC 7265 is silent here, and RFC 5545 makes it
impossible). This is only reachable from jCal input and fails loudly rather than
emitting invalid iCalendar. If a value genuinely needs a newline, it should be
declared `VALUE=TEXT`.

## Checklist

## Additional information

Tests are parametrized over the affected characters and conversion directions,
and cover the `VALUE` boundary, unrecognized parameters, `IMAGE`, and the
`vText`/`vUnknown` overlap. `tox -e nopytz` passes.

AI disclosure: I used Claude Code (Anthropic's Claude Opus) to investigate the
bug, draft this change and its tests, and write this description. I reviewed the
output, made changes as necessary, and validated everything locally.

----
📚 Documentation preview 📚: https://icalendar--1450.org.readthedocs.build/en/1450/


news/1445.bugfix.rst
```diff
@@ -0,0 +1 @@
+Values of unrecognized properties -- and ``X-`` properties without a ``VALUE`` parameter -- were altered by escaping when parsed, serialized, or converted to and from jCal, so they did not round-trip unchanged (:rfc:`7265`). These values are now preserved verbatim. The ``PROXIMITY`` property (:rfc:`9074`) was treated as an unknown value type instead of ``TEXT`` and is now recognized correctly. AI disclosure: I used Claude Code (Anthropic's Claude Opus) to draft and refine this change and its tests; I reviewed the output and validated the change locally. @lcampanella98
```
stevepiercy:
```suggestion
Values of unrecognized properties and ``X-`` properties without a ``VALUE`` parameter were altered by escaping when parsed, serialized, or converted to and from jCal, so they did not round-trip unchanged as :rfc:`7265` specifies. These values are now preserved verbatim. Additionally, the ``PROXIMITY`` property (:rfc:`9074`) was treated as an unknown value type instead of ``TEXT`` and is now recognized correctly. AI disclosure: I used Claude Code (Anthropic's Claude Opus) to draft and refine this change and its tests; I reviewed the output and validated the change locally. @lcampanella98
```

src/icalendar/parser/content_line.py
```diff
@@ -168,6 +168,15 @@ def parts(self) -> tuple[str, Parameters, str]:
 
             DESCRIPTION:The Fall'98 Wild
         """
+        return self._parts()
+
+    def _parts(self, should_unescape: bool = True) -> tuple[str, Parameters, str]:
+        """Implementation of :meth:`parts`.
```
stevepiercy: This should probably be removed, assuming that this private method gets merged into its caller `parts()`. If not, then change as shown.

```suggestion
        """Private implementation of :meth:`parts` to unescape or not the values of ``TEXT`` or ``UNKNOWN`` properties, respectively.
```

src/icalendar/parser/content_line.py
```diff
@@ -168,6 +168,15 @@ def parts(self) -> tuple[str, Parameters, str]:
...
+
+        When ``should_unescape`` is ``False`` the value is returned raw, without
+        backslash unescaping. This is used for :rfc:`7265` ``unknown`` values,
+        which must be preserved verbatim.
```
stevepiercy: This should probably be merged into the docstring of `parts()`, per above comments.

```suggestion
        When ``should_unescape`` is ``True``, unescape backslashes. This is used
        for the values of ``TEXT`` properties.

        When ``should_unescape`` is ``False`` the value is returned raw, without
        backslash unescaping. This is used for :rfc:`7265` ``UNKNOWN`` values,
        which must be preserved verbatim.

        The default is ``True``.
```

src/icalendar/parser/content_line.py
```diff
@@ -168,6 +168,15 @@ def parts(self) -> tuple[str, Parameters, str]:
 
             DESCRIPTION:The Fall'98 Wild
         """
+        return self._parts()
+
+    def _parts(self, should_unescape: bool = True) -> tuple[str, Parameters, str]:
```
stevepiercy: I don't understand why you created a private implementation of `parts`, as the public API for it now merely calls `_parts`. Can you elaborate?

I think that it's OK to alter the original `parts` signature with ``should_unescape``, and update its docstrings with these suggestions. I expect code and tests to continue to work, since it's an optional parameter with a default of ``True``. What do you think?

src/icalendar/parser/content_line.py
```diff
@@ -217,7 +226,10 @@ def parts(self) -> tuple[str, Parameters, str]:
                 for key, value in iter(params.items())
             )
             # Unescape backslash sequences in values but preserve URL encoding
```
stevepiercy:
```suggestion
```

src/icalendar/parser/content_line.py
```diff
@@ -217,7 +226,10 @@ def parts(self) -> tuple[str, Parameters, str]:
...
-            values = unescape_backslash(self[value_split + 1 :])
+            if should_unescape:
```
stevepiercy:
```suggestion
            if should_unescape:
                # Unescape backslash sequences in TEXT values, while preserving URL encoding
```

src/icalendar/parser/content_line.py
```diff
@@ -217,7 +226,10 @@ def parts(self) -> tuple[str, Parameters, str]:
...
+                values = unescape_backslash(self[value_split + 1 :])
+            else:
+                values = self[value_split + 1 :]
```
stevepiercy:
```suggestion
                # Preserve both backslash sequences and URL encoding in UNKNOWN values
                values = self[value_split + 1 :]
```

src/icalendar/prop/text.py
```diff
@@ -27,6 +27,12 @@ class vText(str):
     characters are escaped or changed.
     These characters include the COMMA, SEMICOLON, BACKSLASH, and line breaks.
 
+    Contrast with :class:`~icalendar.prop.unknown.vUnknown`, the :rfc:`7265`
+    ``unknown`` type: that class does **not** apply this escaping and preserves
+    its value verbatim, because the escaping rules of an unrecognized value type
+    are not known. ``vUnknown`` deliberately does not inherit from ``vText`` so
+    the two cannot silently share escaping behaviour.
```
stevepiercy: There's some RFC and Python implementation syntax that we should follow, but I admit we haven't done so consistently in our documentation. For value data types, the RFCs upper-case the name without quotes or markup. `v*` is a Python class implementation to handle the corresponding RFC value data types.

```suggestion
    Contrast TEXT with the UNKNOWN value data type specified in :rfc:`7265#section-5`.
    UNKNOWN is implemented in the Python class :class:`~icalendar.prop.unknown.vUnknown`,
    which does **not** apply this escaping and preserves its value verbatim,
    because the escaping rules of an unrecognized value type are not known.
    ``vUnknown`` deliberately does not inherit from ``vText``, so the two
    don't share escaping behavior.
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
-from typing import ClassVar
+:class:`vUnknown` is a deliberate near-duplicate of
+:class:`~icalendar.prop.text.vText`. It does **not** inherit from ``vText`` even
+though most of the plumbing is identical: the one method that differs,
+:meth:`vUnknown.to_ical`, is the entire reason the class exists, and inheriting
+from ``vText`` is what made the value get re-escaped (see :rfc:`7265#section-5.1`).
+Keeping them as separate classes stops that from silently coming back.
+"""
+
+from typing import Any, ClassVar
```
stevepiercy: Please revert these changes. The docstring at the top is deliberately brief, and this change is redundant to that in the class.

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
-from icalendar.prop.text import vText
+from icalendar.error import JCalParsingError
+from icalendar.parser import Parameters
+from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE, to_unicode
 
 
-class vUnknown(vText):
-    """This is text but the VALUE parameter is unknown.
+class vUnknown(str):
+    r"""A property value of the :rfc:`7265` reserved ``unknown`` type.
```
stevepiercy: `r` is not necessary in a docstring when there is nothing to escape.

```suggestion
    """A property value of the :rfc:`7265#section-5` reserved UNKNOWN value data type.
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
 
-    Since :rfc:`7265`, it is important to record if values are unknown.
-    For :rfc:`5545`, we could just assume TEXT.
+    Unlike :class:`~icalendar.prop.text.vText`, the value is preserved
+    **verbatim** -- taken into and out of iCalendar without :rfc:`5545` escaping
+    or unescaping. The real value type of an unrecognized property is not known,
+    so no escaping rules can be applied and the value must round-trip
+    byte-for-byte (:rfc:`7265#section-5.1`).
```
stevepiercy: We'll need to check this version upon release.

```suggestion
    ..  versionchanged:: 7.1.4
    
        Previously ``vUnknown`` inherited from ``vText``, which unescapes values.
        Now ``vUnknown`` doesn't unescape its values, which is the correct behavior.

    Unlike :class:`~icalendar.prop.text.vText`, the value is preserved verbatim
    when imported from and exported to iCalendar data, without :rfc:`5545` escaping
    or unescaping. When the value type of an unrecognized property is not known,
    then no escaping rules can be applied, and the value must be preserved as is
    round-trip.

    See also:

        :rfc:`7265#section-5.1`
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
+    def __repr__(self) -> str:
+        return f"vUnknown({self.to_ical()!r})"
+
+    def to_ical(self) -> bytes:
+        r"""Return the value verbatim, without :rfc:`5545` escaping (:rfc:`7265#section-5.2`).
+
+        This is the one place ``vUnknown`` deliberately differs from
+        :class:`~icalendar.prop.text.vText`, whose ``to_ical`` escapes ``;`` ``,``
+        ``\\`` and newlines.
+        """
```
stevepiercy: Let's check how it renders at https://icalendar--1450.org.readthedocs.build/en/1450/reference/api/icalendar.prop.unknown.html#icalendar.prop.unknown.vUnknown.to_ical.

I made several subtle changes. Would you please check my accuracy?

It would also be good to have an "Example" section before the "See also" section in the docstring, for example:

```python

        Example:
        
            ..  code-block:: pycon
            
                >>> Python console example 1
                broken doctest
```


```suggestion
        r"""Return the value verbatim, without :rfc:`5545` escaping.

        This method's implementation is different from that in
        :class:`~icalendar.prop.text.vText`, whose
        :meth:`~icalendar.prop.text.vText.to_ical` method escapes ``;``, ``,``,
        ``\``, and newlines.

        See also:

            :rfc:`7265#section-5.2`

        """
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
+
+    @property
+    def ical_value(self) -> str:
+        """The string value of the property."""
+        return str(self)
+
+    from icalendar.param import ALTREP, GAP, LANGUAGE, RELTYPE, VALUE
+
+    def to_jcal(self, name: str) -> list:
+        """The jCal representation of this property according to :rfc:`7265`.
```
stevepiercy:
```suggestion
        """The jCal representation of this property, according to :rfc:`7265#section-5.1`.
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
+    def from_ical(cls, ical: ICAL_TYPE) -> Self:
+        """Take the value verbatim, without unescaping."""
+        return cls(ical)
+
+    @property
+    def ical_value(self) -> str:
+        """The string value of the property."""
+        return str(self)
+
+    from icalendar.param import ALTREP, GAP, LANGUAGE, RELTYPE, VALUE
```
stevepiercy, line 105: Imports should be at the top of the module. I'm surprised ruff didn't complain.

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,109 @@
...
+
+    def to_jcal(self, name: str) -> list:
+        """The jCal representation of this property according to :rfc:`7265`.
+
+        The value is passed through unchanged. The type field is the lowercased
+        ``VALUE`` -- ``"unknown"`` by default, or a preserved unrecognized value
+        type if one was set.
```
stevepiercy: This didn't make sense to me. I rewrote it, based on my interpretation of the RFC per https://datatracker.ietf.org/doc/html/rfc7265.html#section-5 and its subsections. Please double-check my work.

```suggestion
        If the property doesn't include a VALUE property parameter and its value type is not known, then its value type is set to ``"unknown"``.
        Else the property's value type is converted to lowercase.
        
        The property's value is the unprocessed value text, aside from standard
        JSON string escaping.
```

src/icalendar/cal/component.py
```diff
@@ -367,8 +367,9 @@ def _decode(self, name: str, value: VPROPERTY):
             return value
         decoded = self.types_factory.from_ical(name, value)
         # TODO: remove when proper decoded is implemented in every prop.* class
-        # Workaround to decode vText properly
-        if isinstance(decoded, vText):
+        # Workaround to decode vText properly. vUnknown is no longer a vText
```
```suggestion
        # Workaround to decode vText properly. vUnknown is not a vText
```

stevepiercy (changes requested):
This is exceptional work. Thank you for your effort!

Most of my suggestions are minor tweaks and clarifications, with one concern about creating a new private method instead of changing the existing one. Would you please take a look? Thank you!

src/icalendar/parser/content_line.py
lcampanella98: I originally created the private method to avoid changing the public method's signature. But I agree it's better to just alter `parts` signature. 

`should_unescape` is now an optional param on `parts()` and the
private `_parts` is gone. It also let me drop the `# noqa: SLF001`.

src/icalendar/parser/content_line.py
lcampanella98: removed as the private method is now merged into `parts()`

src/icalendar/prop/unknown.py
lcampanella98: Reverted

src/icalendar/prop/unknown.py
lcampanella98: I'd keep this. The names become class-level param descriptors, so they live in
the class body — same pattern as `vText` (`prop/text.py:106`) and ~24 other
prop modules, which is why ruff is quiet. Since the goal was to mirror `vText`,
this is part of that. Can revisit library-wide if you prefer, but feels out of
scope here.

src/icalendar/prop/unknown.py
lcampanella98: Applied. I set it to **7.2.0**, not 7.1.4: latest tag is v7.1.2 and there are feature
fragments queued in `news/`, so the next release looks like a minor bump.
Happy to correct if that's wrong.

src/icalendar/prop/unknown.py
lcampanella98: Took your rewrite; accuracy checks out. Added an Example with `vUnknown("a;b")`
rather than a backslash value — a backslash makes the doctest output ambiguous,
and `a;b` already shows the `;` kept verbatim. Runs clean as a doctest. Checked the render at https://icalendar--1450.org.readthedocs.build/en/1450/reference/api/icalendar.prop.unknown.html#icalendar.prop.unknown.vUnknown.to_ical

src/icalendar/prop/unknown.py
lcampanella98: Used your wording. The suggestion had half-merged with the old text, so I
removed the leftover sentence. Confirmed against the code: `self.VALUE.lower()`
is `"unknown"` with no VALUE param, the lowercased type otherwise.

lcampanella98:
Thanks for the thorough review! Addressed your comments. There are 2 remaining open threads on the `versionchanged` and the class-level imports. Details in the threads.

src/icalendar/prop/unknown.py
stevepiercy: Let's leave this conversation unresolved until we're ready to cut the next release. I expect a 7.1.3 release without this PR, but I don't know whether we'll do a subsequent minor or patch release to bump the version number yet again.

src/icalendar/prop/unknown.py
stevepiercy: Ah, OK. That makes sense.

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,130 @@
...
+        ``\``, and newlines.
+
+        Example:
+
+            The semicolon is kept verbatim, unlike a TEXT value which would
+            escape it as ``\;``.
+
+            .. code-block:: pycon
+
+                >>> from icalendar.prop import vUnknown
```
stevepiercy: Let's compare them in the example.

```suggestion
                >>> from icalendar.prop import vText, vUnknown
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,130 @@
...
+                >>> vUnknown("a;b").to_ical()
+                b'a;b'
```
stevepiercy, line 72:
```suggestion
                b'a;b'
                >>> vText("a;b").to_ical()
                b'a\\;b'
```

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,130 @@
...
+
+        This method's implementation is different from that in
+        :class:`~icalendar.prop.text.vText`, whose
+        :meth:`~icalendar.prop.text.vText.to_ical` method escapes ``;``, ``,``,
+        ``\``, and newlines.
+
+        Example:
+
+            The semicolon is kept verbatim, unlike a TEXT value which would
+            escape it as ``\;``.
```
stevepiercy: Let's make sure this renders correctly and aligns with the updated example.

```suggestion
            The semicolon is kept verbatim for UNKNOWN, unlike a TEXT value
            which would escape it as ``\\;``.
```

stevepiercy (changes requested):
Thanks for making the changes, explaining choices made, and adding an example.

I'd like to add to the example, as shown, and it surfaced an issue. Would you please take a look? Thank you!

src/icalendar/prop/unknown.py
lcampanella98: Got it. will leave unresolved

src/icalendar/prop/unknown.py
lcampanella98: Good catch. Verified it looks good in the [docs render](https://icalendar--1450.org.readthedocs.build/en/1450/reference/api/icalendar.prop.unknown.html#icalendar.prop.unknown.vUnknown.to_ical). 

One thing to flag: `\\;` matches the repr shown in the example,
which is what a reader sees, so I think it's the right call. The actual byte
sequence written to an `.ics` file is a single backslash (`a\;b`) — the doubling
is just Python's repr. Noting it here so the repr-vs-wire distinction is on
record.

lcampanella98:
Applied the three Example suggestions (the `vText`/`vUnknown` comparison) in one commit. I checked the combined doctest locally and it passes. Verified docs render as well.

src/icalendar/prop/unknown.py
stevepiercy: Yes. Two steps. For TEXT, (1) create the Python object and escape specific characters, and (2) `to_ical()` to convert the Python object to iCalendar format. Before you raised the issue, I didn't know it was two steps. TIL.

stevepiercy (dismissed):

stevepiercy:
@lcampanella98 thank you for your contribution to improve icalendar. It's a pleasure to collaborate with you.

> [!IMPORTANT]  
> For maintainers, please note that the version number in the docstring in the [hidden and unresolved conversation](https://github.com/collective/icalendar/pull/1450#discussion_r3406818767) is currently `7.2.0`. Please leave this PR open until release, and update if necessary. Thank you!

lcampanella98:
Happy to contribute and collaborate. Thanks for the review!

src/icalendar/parser/content_line.py
```diff
@@ -160,12 +160,21 @@ def from_parts(
                 return cls(f"{name};{params}:{values}")
         return cls(f"{name}:{values}")
 
-    def parts(self) -> tuple[str, Parameters, str]:
+    def parts(self, should_unescape: bool = True) -> tuple[str, Parameters, str]:
```
niccokunzmann: Could you split this method up so it is parts() and raw_parts() which is used by parts()? Thus would make is clearer architecture wise.

src/icalendar/parser/ical/component.py
```diff
@@ -203,6 +204,11 @@ def handle_property(
         elif name == "RDATE" and vals == "":
             vals_list = []
         else:
+            factory = self.get_factory_for_property(name, params)
+            if factory is vUnknown:
```
niccokunzmann: This has several OOP violations.

1. Avoid `is`  as we encourage polymorphy
2. Ask the object what it wants.

I suggest this: add vUnknown.initialize_with_raw_content_line
Then use getattr to get the value.

We might be able to use this as optimization in other places, too.

src/icalendar/prop/text.py
```diff
@@ -27,6 +27,13 @@ class vText(str):
     characters are escaped or changed.
     These characters include the COMMA, SEMICOLON, BACKSLASH, and line breaks.
 
+    Contrast TEXT with the UNKNOWN value data type specified in :rfc:`7265#section-5`.
+    UNKNOWN is implemented in the Python class :class:`~icalendar.prop.unknown.vUnknown`,
+    which does **not** apply this escaping and preserves its value verbatim,
+    because the escaping rules of an unrecognized value type are not known.
+    ``vUnknown`` deliberately does not inherit from ``vText``, so the two
```
niccokunzmann:
```suggestion
    :class:`~icalendar.prop.unknow.vUnknown` deliberately does not inherit from ``vText``, so the two
```

Please reference classes.

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,132 @@
...
+        params: dict[str, Any] | None = None,
+    ) -> Self:
+        value = to_unicode(value, encoding=encoding)
+        self = super().__new__(cls, value)
+        self.encoding = encoding
+        self.params = Parameters(params)
+        return self
+
+    def __repr__(self) -> str:
+        return f"vUnknown({self.to_ical()!r})"
```
niccokunzmann:
```suggestion
        return f"{self.__class__._name__}({self.to_ical()!r})"
```

src/icalendar/tests/test_issue_1445.py
```diff
@@ -0,0 +1,224 @@
...
+    assert ev.to_jcal()[1][1] == [
+        "x-foo",
+        {"x-roast": "a,b;c", "x-note": "line1\nline2"},
+        "unknown",
+        "val;ue",
+    ]
+
+
+def test_image_value_is_verbatim():
+    """IMAGE always resolves to the unknown type and is preserved verbatim."""
```
niccokunzmann: Why is this always tne case? I thought, we have IMAGE covered...
Investigating... I think it might be because it required an explicit VALUE parameter.
The reason should be documemted for future reference in case someone thinks differently.

src/icalendar/tests/test_issue_1445.py
```diff
@@ -0,0 +1,224 @@
...
+    assert isinstance(ev["SUMMARY"], vText)
+    assert str(ev["SUMMARY"]) == "a;b,c"  # TEXT is unescaped on read
+    assert _line(ev, "SUMMARY") == r"SUMMARY:a\;b\,c"
+
+
+def test_jcal_unknown_value_with_raw_newline_fails_safely():
+    """A jCal unknown value with a raw newline cannot be a bare iCal value.
+
+    This is only reachable from jCal (an iCal content line cannot carry a raw
+    newline). It fails loudly rather than emitting invalid iCalendar.
```
niccokunzmann, line 206: Nice catch of an edge case. In the future, we might change this in case it is not practical for jCal2iCal conversion.

niccokunzmann (changes requested):
Thanks! I have some thoughts on code quality. Let me know what you think. Nice PR!

niccokunzmann:
Sorry it took me sol long to get back to you! I appreciate the PR and if you can continue.

src/icalendar/parser/content_line.py
lcampanella98: Done. `raw_parts()` holds the logic and returns values verbatim; `parts()` is
`raw_parts()` plus unescaping. This also drops the `should_unescape` flag, so
`parts()` is back to its original signature.

I kept the examples and the full explanation on `parts()`, since that's the one
most callers use, and gave the two distinct summary lines so they're easy to
tell apart in the API docs.

src/icalendar/parser/ical/component.py
lcampanella98: Done. `vUnknown` provides `initialize_with_raw_content_line(line)` and the parser
looks it up with `getattr`. The parser no longer imports `vUnknown` at all.

I made it a classmethod returning the value, to pair with `raw_parts()`. Happy to
switch to a flag if that's what you meant.

src/icalendar/prop/text.py
lcampanella98: Applied, with a fix: the suggestion had `icalendar.prop.unknow` (missing the
final "n"), which doesn't resolve.

src/icalendar/prop/unknown.py
lcampanella98: Applied the idea, with a fix: `_name__` (single underscore) raises
`AttributeError` on every `repr()`. Used `self.__class__.__name__`. Unchanged for
`vUnknown`; subclasses now repr under their own name.

`vText.__repr__` still hardcodes its name — happy to make it dynamic too.

src/icalendar/tests/test_issue_1445.py
lcampanella98: Documented in the test. It's not about the VALUE parameter: IMAGE is
special-cased in `TypesFactory.for_property` to always return the unknown type,
even for `VALUE=URI`/`BINARY`. From commit [`aac97fa9`](https://github.com/collective/icalendar/commit/aac97fa9b3371a2201c7b0cb1e86e9fbfea5ba0b) ("Fixed Image data corruption 
problem") — the verbatim type is what stops image data being corrupted.

Parametrized over no VALUE / `URI` / `BINARY`. One nuance: the class is always
`vUnknown`, but an explicit VALUE is still preserved as the jCal type field, so
the test asserts that too.

src/icalendar/tests/test_issue_1445.py
lcampanella98: Thanks! Agreed — happy to revisit if it's impractical for jCal-to-iCal.

lcampanella98:
Thanks for the review — the OOP points made this cleaner. All five are addressed, and I merged in the latest main (the only conflict was `value_separator_index` and my new `parts()` landing in the same spot; kept both). 

Two open questions where I picked a direction and would rather you confirm, see comment threads above.

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,147 @@
...
+        Parameters:
+            jcal_property: The jCal property to parse.
+
+        Raises:
+            ~error.JCalParsingError: If the provided jCal is invalid.
+        """
+        JCalParsingError.validate_property(jcal_property, cls)
+        string = jcal_property[3]
+        JCalParsingError.validate_value_type(string, str, cls, 3)
+        return cls(
```
niccokunzmann, line 137: @gaoflow Could you have a check if jCal roundtrip serilaizatikn works with the changes to vUnknown?

See also #1553

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,147 @@
...
+        """
+        return self.encode(self.encoding)
+
+    @classmethod
+    def from_ical(cls, ical: ICAL_TYPE) -> Self:
+        """Take the value verbatim, without unescaping."""
+        return cls(ical)
+
+    @classmethod
+    def initialize_with_raw_content_line(cls, line: "Contentline") -> str:
```
niccokunzmann:
```suggestion
    def get_value_from_content_line(cls, line:  Contentline) -> str:
```

I think, that is be nicer as it speaks of the contract without giving away the implementation. Thanks!
The ContentLine can be imported in the TYPE_CHECKING block.

src/icalendar/prop/unknown.py
```diff
@@ -1,26 +1,147 @@
...
+        value = to_unicode(value, encoding=encoding)
+        self = super().__new__(cls, value)
+        self.encoding = encoding
+        self.params = Parameters(params)
+        return self
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self.to_ical()!r})"
+
+    def to_ical(self) -> bytes:
```
niccokunzmann, line 55: @alhudz You contributed a lot concerning injected line breaks. Could you give this a review as this is not inheriting from vText any more?

niccokunzmann (changes requested):
Thanks for your changes! I asked two previous contributors to give their opinion to make sure, we do not introduce errors or exploits.

src/icalendar/parser/ical/component.py
niccokunzmann: This is perfect!

src/icalendar/prop/unknown.py
niccokunzmann: I think, making repr(vText) dynamic would be beneficial. Thanks!

src/icalendar/tests/test_issue_1445.py
niccokunzmann: Thanks for that! Image should obey the value parameter. I am on my phone and cannot check that. If this is really the case that IMAGE ignores the VALUE parameter, could you open an issue? I think it is either vUri or vBinaryif VALUE is given. Since it got introduced later, it should only default to vUnknown.

lcampanella98: Verified IMAGE does not currently honor VALUE, and opened #1561 to track the fix (resolve to vUri/vBinary when VALUE given, vUnknown as the default). Repro steps in the issue.

src/icalendar/prop/unknown.py
lcampanella98: Done, renamed to `get_value_from_content_line`. `Contentline` is imported under
`TYPE_CHECKING`, and I added `from __future__ import annotations` so the
annotation can stay unquoted ([`icalendar.cal.component.py`](https://github.com/collective/icalendar/blob/bd02053f4c8d65e957d839415c7bc4debac34c63/src/icalendar/cal/component.py#L3) does this too). 
The parser looks the method up by the new name.

With deferred annotations, `Self` is now only needed for type checking, so I also
moved its import into the `TYPE_CHECKING` block (ruff TC001 flagged it once the
`__future__` import made the annotations lazy).

src/icalendar/prop/unknown.py
lcampanella98: Done. `vText.__repr__` now derives the name from `self.__class__.__name__`, so
subclasses repr under their own name instead of always `vText(...)`.

src/icalendar/parser/content_line.py
```diff
@@ -160,23 +160,15 @@ def from_parts(
...
-
-        Example with parameter:
-
-        .. code-block:: ics
-
-            DESCRIPTION;ALTREP="cid:part1.0001@example.org":The Fall'98 Wild
+        This is :meth:`parts` without the unescaping: the values are returned
+        verbatim, preserving both backslash sequences and URL encoding. It is
+        used for :rfc:`7265` ``UNKNOWN`` values, whose real value type -- and
+        therefore whose escaping rules -- are not known.
```
stevepiercy: Use em-dash.

```suggestion
        used for :rfc:`7265` ``UNKNOWN`` values, whose real value type—and
        therefore whose escaping rules—are not known.
```

src/icalendar/parser/content_line.py
```diff
@@ -226,14 +218,38 @@ def parts(self) -> tuple[str, Parameters, str]:
...
+
+        .. code-block:: ics
+
+            DESCRIPTION;ALTREP="cid:part1.0001@example.org":The Fall'98 Wild
+
+        Example without parameters:
+
+        .. code-block:: ics
+
+            DESCRIPTION:The Fall'98 Wild
```
stevepiercy: Let's use a section heading.

```suggestion
        Examples:
        
            With parameter:

            ..  code-block:: ics

                DESCRIPTION;ALTREP="cid:part1.0001@example.org":The Fall'98 Wild

            Without parameters:

            ..  code-block:: ics

                DESCRIPTION:The Fall'98 Wild
```

stevepiercy (changes requested):
Minor changes to the docstrings. Otherwise LGTM! Thank you!

lcampanella98:
Applied the docstring suggestions, thanks!

stevepiercy (approved):
Docs LGTM.

- https://icalendar--1450.org.readthedocs.build/en/1450/reference/api/icalendar.parser.content_line.html#icalendar.parser.content_line.Contentline.parts
- https://icalendar--1450.org.readthedocs.build/en/1450/reference/api/icalendar.parser.content_line.html#icalendar.parser.content_line.Contentline.raw_parts

Thank you!

niccokunzmann:
Thanks! I will wait 2 days for the others to respond. Then, create issues from the comments or check it myself.

src/icalendar/prop/unknown.py
niccokunzmann: TImed out. https://github.com/collective/icalendar/pull/1450#issuecomment-5049400632

src/icalendar/prop/unknown.py
niccokunzmann: Timed out. https://github.com/collective/icalendar/pull/1450#issuecomment-5049400632

niccokunzmann (approved):
Thanks for the PR! Ready for merge!
Sorry that it takes so long.

src/icalendar/prop/unknown.py
niccokunzmann: This seems to be the only open comment. I will create an issue for the next release.

niccokunzmann: #1595

niccokunzmann:
@stevepiercy, I just saw this: https://github.com/collective/icalendar/pull/1450#issuecomment-4702719308

I created the issue #1595 so that we can go through that. But I would like to merge this PR. Do you see a problem with this approach?

Ooopsi.... Auto merge already did it...

stevepiercy:
@niccokunzmann as long as the create version number goes into the admonition before a release is made, it's fine. I added an item to the checklist in #1595.

lcampanella98:
Thanks for merging, and for the thorough review from everyone, it's a better change than what I started with.

PR #1458, opened 2026-06-15
fix: clean up the auto-generated "Accepted values" docstring section

### LINKED ISSUE #1259: Accepted values for RECURRENCE_ID property looks strange
When reviewing PR #1236, I found a weird docstring format.

https://icalendar--1236.org.readthedocs.build/en/1236/reference/api/icalendar.cal.event.html#icalendar.cal.event.Event.RECURRENCE_ID

Image

Observe that the property has type hints that hyperlink, but the "Accepted values:" section heading is a wall of text. It's also not a standard section heading.

It's generated in this method.

https://github.com/collective/icalendar/blob/bb480a92f5f56fe7b088b8d4cb9aaf0d7f39b14f/src/icalendar/attr.py#L954-L958

With the following snippet, it appears somewhat better.

```python
To delete the value, either use ``del`` or set it to ``None``.

Returns:
    If the value is absent, return ``None``.

Raises:
    :exc:`~icalendar.error.InvalidCalendar`
        If the attribute has invalid values.
```

Image

I think that "Accepted values" is redundant to the type hints, so I'd remove it.

I'd keep the usage text, however, and append it to the description. This could be a problem, though, if there are other heading sections in the original docstring after the description.

@niccokunzmann what do you think?

mvanhorn:
## Closes issue

- [x] Closes #1259

## Description

The property docstrings generated by `create_single_property` in `src/icalendar/attr.py` rendered an `Accepted values:` line as a wall of text with manually inlined type names. As @stevepiercy and @niccokunzmann noted on the issue, that line predates the project's current Sphinx typing and linking conventions and is redundant with the type-hinted, cross-referenced annotations the property already exposes.

This removes the `Accepted values:` line and reformats the return, error, and deletion notes into standard Sphinx `Returns:` and `Raises:` sections, matching the snippet approved in the issue:

```
To delete the value, either use ``del`` or set it to ``None``.

Returns:
    If the value is absent, return ``None``.

Raises:
    :exc:`~icalendar.error.InvalidCalendar`
        If the attribute has invalid values.
```

The change applies to every property built through the factory, including `Event.RECURRENCE_ID`.

## Checklist

## Additional information

Added `news/1259.documentation` and a regression test in `test_consolidation_attr_functions.py` that asserts the new sectioned form is present on `Event.RECURRENCE_ID.__doc__` and that `Accepted values` is gone. Ran `pytest src/icalendar/tests/test_consolidation_attr_functions.py` (25 passed) plus `ruff check` and `ruff format --check` on the touched files.

----
📚 Documentation preview 📚: https://icalendar--1458.org.readthedocs.build/en/1458/


news/1259.documentation
```diff
@@ -0,0 +1 @@
+Tidied the auto-generated property docstrings by removing the redundant accepted-values sentence and formatting return and error details as Sphinx sections. @mvanhorn
```
stevepiercy:
```suggestion
Tidied the auto-generated property docstrings by removing the redundant accepted values sentence and formatting return and error details as Sphinx sections. @mvanhorn
```

src/icalendar/tests/test_consolidation_attr_functions.py
stevepiercy, line 1: I think this test is too much. I wouldn't test the detailed content of docstrings, as it may be necessary to change them, and it doesn't test the functionality of icalendar. The tests, after all, are called "functional tests." Would you please remove this test?

However, this gave me another idea, specifically how to test the structure of docstrings. For example, there were docstrings that used `Args` instead of `Parameters`, all of which have been since fixed. We still have a lot of copy-paste from RFCs as docstrings, per https://github.com/collective/icalendar/issues/1244. Some of those docstrings use sections not in the set defined in https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure, plus "See also". I also don't care too much about docstrings in private methods since they don't get published into documentation, but there are a few exceptions where they're helpful. Anyway, a test for docstrings that flags incompatible section headings would be nifty.

stevepiercy (changes requested):
A minor grammar tweak on the news fragment, and a removal, as well as a ramble about possible docstring structure testing, which is out of scope for this PR.

src/icalendar/tests/test_consolidation_attr_functions.py
niccokunzmann: Hi Steve, would it be ok to add add a new file for just testing docs? This test here can be changed and it is easy to change. If all documentation tests go in one file, that might be the correct way to go. We can e.g. open a new issue for docs compliance tests that yield good error messages. E.g. a test for wrong sections like 'Args:'. Also, they can be parametrized to run on every method or class or component, ... With this, we can assure a common format across the documentation.

If you agree, you can open an issue for that, describing how docs should be structured or link other issues. Then, we can extract the subset that is testable. Then, the community can chip in.

This is just a thought here. Let me know what you think, we can also discuss this in an issue as it is not required for this PR but the PR really points something out to think about.
Tests can reduce the amount of review youneed to do to point people in the right direction.

Thanks for the PR!

niccokunzmann (commented):
Thanks!

mvanhorn:
Thanks both. Done:

- Removed `test_single_property_docstring_uses_sphinx_sections` (the docstring-content assertion @stevepiercy flagged). The functional getter/setter tests in that file are untouched - 24 still pass locally.
- Applied @stevepiercy's `news/1259.documentation` wording.

Pushed in 08b7135c.

On the doc-structure tests idea: happy to open a separate issue for it (flagging incompatible section headings, `Args:` vs `Parameters:`, parametrized across methods, tied to #1244 and the docstring style guide) so it can be discussed and the community can chip in, keeping it out of this PR as @niccokunzmann suggested. Want me to open it?

stevepiercy (approved):
Thank you!

stevepiercy:
> On the doc-structure tests idea: happy to open a separate issue for it (flagging incompatible section headings, `Args:` vs `Parameters:`, parametrized across methods, tied to #1244 and the docstring style guide) so it can be discussed and the community can chip in, keeping it out of this PR as @niccokunzmann suggested. Want me to open it?

Yes, would you please? I'd appreciate it. Thank you!

src/icalendar/tests/test_consolidation_attr_functions.py
stevepiercy: @mvanhorn offered to create the issue. I think it's worth pursuing.

mvanhorn:
Thanks @stevepiercy for the quick merge. Cleaning up the auto-generated Accepted values docstrings makes the property docs much more readable.

mvanhorn:
Thank you for the merge, @stevepiercy. Cleaning up the auto-generated Accepted values section makes those property docstrings a lot more readable.

PR #1465, opened 2026-06-16
reject non-ASCII digits in vMonth from_ical

alhudz:
## Linked issue

No related issue; reporting directly.

## Description

`Repro:` parse a calendar containing `RRULE:FREQ=MONTHLY;BYMONTH=١٢` (Arabic-Indic digits), or call `vMonth("١٢")` / `vMonth("²")` directly.
`Cause:` `vMonth.__new__` gates its numeric parse on `str.isdigit()`, which is `True` for non-ASCII digits. `int("١٢")` happily parses those, so the malformed `BYMONTH` is accepted and silently normalised to month `12`; for code points where `str.isdigit()` is `True` but `int()` is not (`²`, `⑧`) the same guard is cleared and a raw `invalid literal for int()` `ValueError` leaks instead of the intended `Invalid month` error.
`Fix:` gate both the plain and `L`-suffixed branches on `str.isascii()` as well as `str.isdigit()` so the check matches what `int()` accepts, and the non-conforming values are rejected.

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1465.org.readthedocs.build/en/1465/


niccokunzmann (dismissed):
Thanks for the fix!

niccokunzmann:
Oh. The code style checks fail. Could you correct them?

alhudz:
Sorted - ran `ruff format` over the multi-line guard, pushed. Format and check both pass locally now.

news/+reject-non-ascii-month-digits.bugfix
```diff
@@ -0,0 +1 @@
+Reject non-ASCII digits in :class:`~icalendar.prop.recur.month.vMonth` (``BYMONTH``). ``str.isdigit`` is ``True`` for digits like the Arabic-Indic ``١٢``, which was silently accepted and normalised to month ``12``, while characters such as ``²`` cleared the same guard and then leaked a raw ``int()`` error. :meth:`vMonth.from_ical <icalendar.prop.recur.month.vMonth.from_ical>` now rejects both. @alhudz
```
stevepiercy: Per https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#spelling-and-grammar

```suggestion
Reject non-ASCII digits in :class:`~icalendar.prop.recur.month.vMonth` (``BYMONTH``). ``str.isdigit`` is ``True`` for digits like the Arabic-Indic ``١٢``, which was silently accepted and normalized to month ``12``, while characters such as ``²`` cleared the same guard and then leaked a raw ``int()`` error. :meth:`vMonth.from_ical <icalendar.prop.recur.month.vMonth.from_ical>` now rejects both. @alhudz
```

stevepiercy (changes requested):

stevepiercy (approved):

Issue #1370, opened 2026-05-08
Unbounded nested components can trigger `RecursionError` in `repr()` / `str()`


gistrec:
## Summary

Deeply nested iCalendar components can be parsed successfully, but formatting the resulting object with `repr()`, `str()`, or an f-string can raise `RecursionError`.

The recursive path appears to be `Component.__repr__()`:

```python
# src/icalendar/cal/component.py
def __repr__(self):
    """String representation of class with all of it's subcomponents."""
    subs = ", ".join(str(it) for it in self.subcomponents)
    return f"{self.name or type(self).__name__}({dict(self)}{', ' + subs if subs else ''})"
```

This can cause a denial-of-service in applications that parse untrusted `.ics` files and later log or format the parsed calendar object.

## Reproducer

```python
import icalendar

ics = (
    "BEGIN:VCALENDAR\r\n"
    "VERSION:2.0\r\n"
    "PRODID:-//x//x//EN\r\n"
    + "BEGIN:VEVENT\r\n" * 500
    + "UID:t@e\r\n"
    "DTSTAMP:20260101T000000Z\r\n"
    + "END:VEVENT\r\n" * 500
    + "END:VCALENDAR\r\n"
)

cal = icalendar.Calendar.from_ical(ics)  # succeeds
str(cal)  # RecursionError
```

## Observed behavior

| Operation | Result |
|---|---|
| `Calendar.from_ical(ics)` | succeeds |
| `repr(cal)` / `str(cal)` / `f"{cal}"` | raises `RecursionError` |
| `cal.to_ical()` | succeeds |
| `cal.walk()` | succeeds |

The payload is small, around 13 KB.

## Impact

This is a denial-of-service / robustness issue. Example scenarios:

- a CalDAV server or calendar import service logs a parsed calendar object;
- an email invitation parser includes the object in diagnostics;
- error reporting formats local variables containing the parsed calendar.

The crash is not in parsing itself, but in later diagnostic/logging paths.

## Possible fixes

Any of these should mitigate the issue:

1. Reject excessive nesting during parsing, for example with a component nesting-depth limit.
2. Make `__repr__` iterative or depth-limited.
3. Truncate subcomponent output in `__repr__` after a safe depth.

A parser-level limit would prevent this class of issue more generally, but a bounded `__repr__` may be less invasive if compatibility is a concern.

gistrec:
The impact appears limited to denial-of-service in applications that format parsed calendar objects from untrusted input

[...]

PR #1466, opened 2026-06-16
validate property, parameter and recur names in from_jcal

alhudz:
## Linked issue

No separate issue; reported and fixed here.

## Description

Repro: convert untrusted jCal where a property name, parameter name or `RRULE` part name carries a lone `\r` (or a `:`/`;`) into iCalendar, for example via `Calendar.from_jcal(jcal).to_ical()`.
Cause: `Component.from_jcal`, `Parameters.from_jcal` and `vRecur.from_jcal` keep the jCal name verbatim and emit it raw into the content line, whereas `Parameters.from_ical` already runs names through `validate_token`. A `:` or `;` rewrites the property or parameter structure on the same line; a lone `\r` slips past the `\n`-only assert in `Contentline` and turns up as a new content line for consumers that treat a bare carriage return as a break.
Fix: run each jCal name through `validate_token` at parse time, raising `JCalParsingError` for names that are not valid iCalendar tokens.

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1466.org.readthedocs.build/en/1466/


src/icalendar/cal/component.py
```diff
@@ -1008,6 +1009,15 @@ def from_jcal(cls, jcal: str | list) -> Component:
         for i, prop in enumerate(properties):
             JCalParsingError.validate_property(prop, cls, path=[1, i])
             prop_name = prop[0]
+            try:
+                validate_token(prop_name)
```
angatha: Validation can be even stricter since [RFC 7265 3.3](https://datatracker.ietf.org/doc/html/rfc7265#section-3.3) requires lower case. Havent checked usage of `JCalParsingError.validate_property` but maybe the validation can be moved there?

src/icalendar/parser/parameter.py
```diff
@@ -482,6 +482,12 @@ def from_jcal(cls, jcal: dict[str : str | list[str]]):
                 raise JCalParsingError(
                     "All parameter names must be strings.", cls, value=name
                 )
+            try:
+                validate_token(name)
```
angatha: `name` must also be lower case as per [section 3.5](https://datatracker.ietf.org/doc/html/rfc7265#section-3.5)

src/icalendar/prop/recur/recur.py
```diff
@@ -260,6 +260,15 @@ def from_jcal(cls, jcal_property: list) -> Self:
             )
         recur = {}
         for key, value in jcal_property[3].items():
+            try:
+                validate_token(key)
```
angatha: Must be lower case as per [3.6.10](https://datatracker.ietf.org/doc/html/rfc7265#section-3.6.10)

angatha:
Nice work. I checked the lower case part and linked the sections. Hope I did not make a mistake. Can you please add the lower case validation as well?

src/icalendar/cal/component.py
alhudz: Added a lower-case check after the token check here. I looked at folding it into `validate_property`, but that only sees the property list (not parameter names or `RRULE` part names), and `vRecur.from_jcal` calls it on the `rrule` property itself rather than the part keys, so I kept the check at each of the three parse sites to keep the messages specific. One small thing: property names are 3.4, 3.3 is components, so I cited 3.4 in the news entry.

src/icalendar/parser/parameter.py
alhudz: Done, `name` now has to be lower case here too, otherwise it raises `JCalParsingError`. 3.5 was right.

src/icalendar/prop/recur/recur.py
alhudz: Done, same lower-case check on the part name. 3.6.10 was right.

alhudz:
Pushed the lower-case validation for all three names, with tests and an updated news entry. A few existing recur tests passed upper-case keys like `FREQ`, so I switched those to lower case as well. Your section links were right, only nudged the property one from 3.3 to 3.4 (3.3 is components).

alhudz:
any update?

SashankBhamidi:
Hi, could you resolve the conflicts, I'll review this right after you do this. Thanks for being patient.

angatha:
Can you please disclose if any generative AI was involved in creating this PR wither by following the [artificial intelligence policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) or replying that no generative AI was involved.

alhudz:
@SashankBhamidi conflicts resolved and `main` merged in. #1370 moved the property loop out of `from_jcal` into the new `_node_from_jcal` helper, so the token and lower-case checks moved there with it. The parameter and recur name checks were untouched, and the jCal suite is still green.

@angatha on your question: I write the patches and repros myself, but I do use AI a bit on the side for sanity-checking and finding my way around the RFC sections, nothing heavy.

angatha (dismissed):

SashankBhamidi (dismissed):

alhudz:
Thanks for the approval. I already covered this above in reply to @angatha: I write the patches and repros myself and only lean on AI a little on the side for sanity-checking and finding my way around the RFC sections, nothing that goes straight into the diff.

SashankBhamidi:
@alhudz,

Ok, one thing before I merge, can you write out what a malicious jCal payload would look like and show me the exact content line the old code would have produced from it?

alhudz:
Sure, here is the parameter-name case, which is the clearest.

Payload (jCal parameter name carrying a lone `\r`):

```python
Calendar.from_jcal([
    "vcalendar",
    [["version", {}, "text", "2.0"], ["prodid", {}, "text", "x"]],
    [["vevent", [
        ["uid", {}, "text", "x"],
        ["summary", {"x\rrrule": "v"}, "text", "hi"],
    ], []]],
]).to_ical()
```

Old code kept the name verbatim, so `to_ical()` produced this `VEVENT` block:

```
BEGIN:VEVENT
SUMMARY;X\rRRULE=v:hi
UID:x
END:VEVENT
```

The `\r` in `X\rRRULE` is a bare CR, not `\r\n`, so it slips past the `\n`-only assert in `Contentline`. A consumer that treats a lone CR as a line break splits that one line in two and sees a forged content line:

```
SUMMARY;X
RRULE=v:hi   <- injected
```

The property-name case is the same shape with a `:`: `["summary:INJECT", {}, "text", "hi"]` serialised to `SUMMARY:INJECT;VALUE=TEXT:hi`, so the `:` rewrites the name/value split on the line. Recur part names behave the same way (`RRULE:FREQ\rINJ=DAILY`).

After the patch all three raise `JCalParsingError` at parse time, e.g. for the payload above:

```
JCalParsingError: [2][0][1][1][1] in Parameters: The parameter name is not a valid token. Got value: 'x\rrrule'
```

so the bad name never reaches serialisation, and legitimate jCal still round-trips.

stevepiercy:
@angatha is it OK to resolve the conversations you opened? Unresolved conversations block auto-merge, but can be overridden by privileged users.

news/+validate-jcal-name-tokens.bugfix
```diff
@@ -0,0 +1 @@
+jCal property names, parameter names and ``RRULE`` part names are now validated as lower-case iCalendar tokens when parsing jCal (RFC 7265 sections 3.4, 3.5 and 3.6.10). They were kept verbatim and re-emitted into the content line on serialisation, so a name containing ``:``/``;`` or a lone carriage return could inject parameters or a new content line for consumers that treat a bare ``\r`` as a break. @alhudz
```
stevepiercy: The Oxford comma, link to each RFC section, and American English spelling.

```suggestion
jCal property names, parameter names, and ``RRULE`` part names are now validated as lower-case iCalendar tokens when parsing jCal (RFC 7265 sections :rfc:`7265#section-3.4`, :rfc:`7265#section-3.5`, and :rfc:`7265#section-3.6.10`). They were kept verbatim and re-emitted into the content line on serialization, so a name containing ``:``, ``;``, or a lone carriage return could inject parameters or a new content line for consumers that treat a bare ``\r`` as a break. @alhudz
```

src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py
```diff
@@ -453,14 +453,85 @@ def test_skip():
...
 
 
+INVALID_NAME_TOKENS = ["x-foo:INJECT", "x-foo\rRRULE", "a;b", "a=b", "a,b", "a b", ""]
+
+
+@pytest.mark.parametrize("name", INVALID_NAME_TOKENS)
+def test_invalid_property_name_token(name):
+    """A jCal property name that is not a valid token is rejected.
+
+    Otherwise the name is re-emitted verbatim into the content line, so a ``:``
```
stevepiercy: Also `;`? See change log entry.

src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py
```diff
@@ -453,14 +453,85 @@ def test_skip():
...
+    or a lone carriage return injects structure on serialisation.
```
stevepiercy:
```suggestion
    or a lone carriage return injects structure on serialization.
```

stevepiercy (changes requested):

angatha:
> @angatha is it OK to resolve the conversations you opened? Unresolved conversations block auto-merge, but can be overridden by privileged users.

Yes. Hope I did not miss any.

news/+validate-jcal-name-tokens.bugfix
alhudz: Applied, thanks. News entry now has the Oxford commas, the `:rfc:` role links per section, and the US spelling.

src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py
alhudz: Good catch. The docstring now lists `:`, `;`, and the lone CR to match the change log, and switched to `serialization`.

src/icalendar/cal/component.py
```diff
@@ -1134,6 +1135,22 @@ class for the root, the parent's resolved class for a child).
     for i, prop in enumerate(properties):
         JCalParsingError.validate_property(prop, component_cls, path=[1, i])
         prop_name = prop[0]
+        try:
```
niccokunzmann: This looks like code duplication. Could you refactor both occurrences into validate_jcal_token? Extract method.

niccokunzmann (changes requested):
Hi, thanks for the PR!
The same codeis used in 3 places. Please add a function that does the job and use it in these 3 places.

stevepiercy (dismissed):
Thanks for the changes. I approve, pending the resolution of @niccokunzmann's feedback.

niccokunzmann (changes requested):
Hi, just a little refactoring should do it.

src/icalendar/cal/component.py
alhudz: Done. Pulled the token and lower-case checks into `JCalParsingError.validate_jcal_token` and call it from all three sites: property names in `_node_from_jcal`, parameter names in `Parameters.from_jcal`, and `RRULE` part names in `vRecur.from_jcal`. Same messages and paths as before, jCal suite still green.

src/icalendar/error.py
```diff
@@ -231,6 +231,35 @@ def validate_list_type(
...
+        kind: str,
+        parser: str | type = "",
+        path: list[str | int] | None | str | int = None,
+    ) -> None:
+        """Validate a jCal ``name`` as a lower-case iCalendar token.
+
+        ``kind`` names the token in the error message, e.g. ``"property name"``.
+        A name is kept verbatim and re-emitted into the content line on
+        serialization, so a ``:``, ``;``, or lone carriage return in it could
+        inject parameters or a new content line.
```
stevepiercy: I'm not clear whether this is a description or an attempt to define the parameters. I think it would be more clear to define the parameters, as described in https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure.

I also think a description would be appropriate, especially what are the validation rules.

The docstring should also include a `See also` section heading with a link to ``:meth:`~icalendar.parser.string.validate_token`.``

Finally, please don't use `e.g.` or any other Latin abbreviation. Instead use "for example, EXAMPLE HERE".

src/icalendar/error.py
```diff
@@ -231,6 +231,35 @@ def validate_list_type(
...
+
+        Raises:
+            ~error.JCalParsingError: if ``name`` is not a valid lower-case token.
```
stevepiercy:
```suggestion
            ~error.JCalParsingError: If ``name`` is not a valid lower-case iCalendar token.
```

src/icalendar/error.py
```diff
@@ -231,6 +231,35 @@ def validate_list_type(
...
+        """
+        from icalendar.parser.string import validate_token
+
+        try:
+            validate_token(name)
+        except ValueError:
+            raise cls(
+                f"The {kind} is not a valid token.", parser, path, value=name
```
stevepiercy: Error messages that say how to fix a problem are more helpful.

```suggestion
                f"The {kind} must be valid iCalendar token, matching the regular expression pattern `[\w.-]+`.", parser, path, value=name
```

stevepiercy (changes requested):
Now that the code is in good shape, I made a pass at docstrings and error messages. Would you please take a look, and let me know what you think? Thank you!

src/icalendar/cal/component.py
```diff
@@ -1134,6 +1134,9 @@ class for the root, the parent's resolved class for a child).
     for i, prop in enumerate(properties):
         JCalParsingError.validate_property(prop, component_cls, path=[1, i])
         prop_name = prop[0]
+        JCalParsingError.validate_jcal_token(
```
niccokunzmann: Would this not be handier if put into validate_property? I am a bit confused.

As a human you would probably recognise this. AI usually fails to capture intention.

So: What happened here? Why did you decide on this path? What is your take on this?

niccokunzmann (changes requested):
Please help me understand the decision you made here.

src/icalendar/cal/component.py
alhudz: `validate_property` only sees the property list `[name, params, type, value]` and checks its shape: a list of at least four items, name a string, params a mapping, type a string. The token rule has to cover three different names though: the property name here, the parameter names in `Parameters.from_jcal`, and the `RRULE` part names in `vRecur.from_jcal`. Only the property name goes through `validate_property`; the other two are dict keys that never reach it.

So folding the check into `validate_property` would only guard property names, and I'd still need the same check in the other two places, which is the duplication you asked me to pull out earlier. Keeping it in the standalone `validate_jcal_token` lets all three sites share one check with one message. Happy to inline it into `validate_property` instead if you'd prefer, but the parameter and recur sites would still need their own call.

src/icalendar/error.py
alhudz: Reworked the docstring to the style guide layout: a `Parameters:` entry for each argument, a description of the two rules it enforces (the `[\w.-]+` token pattern and the lower-case requirement from RFC 7265), a `See also:` link to `validate_token`, and swapped the `e.g.` for `for example,`.

src/icalendar/error.py
alhudz: Applied. Had to wrap it onto a second line since the full sentence runs past the 88-char limit for this file, the wording is unchanged.

src/icalendar/error.py
alhudz: Done, the token error now names the `[\w.-]+` pattern. Updated the three test matchers that were keyed to the old `is not a valid token` wording.

src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py
```diff
@@ -453,14 +453,85 @@ def test_skip():
...
+@pytest.mark.parametrize("name", INVALID_NAME_TOKENS)
+def test_invalid_property_name_token(name):
+    """A jCal property name that is not a valid token is rejected.
+
+    Otherwise the name is re-emitted verbatim into the content line, so a ``:``,
+    ``;``, or a lone carriage return injects structure on serialization.
+    """
+    with pytest.raises(
+        JCalParsingError,
+        match=r"The property name must be valid iCalendar token",
```
stevepiercy: The error message and tests are missing a `a `.

```suggestion
        match=r"The property name must be a valid iCalendar token",
```

src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py
```diff
@@ -453,14 +453,85 @@ def test_skip():
...
+    ):
+        Component.from_jcal(["vcalendar", [[name, {}, "text", "x"]], []])
+
+
+@pytest.mark.parametrize("name", INVALID_NAME_TOKENS)
+def test_invalid_parameter_name_token(name):
+    """A jCal parameter name that is not a valid token is rejected."""
+    with pytest.raises(
+        JCalParsingError,
+        match=r"The parameter name must be valid iCalendar token",
```
stevepiercy:
```suggestion
        match=r"The parameter name must be a valid iCalendar token",
```

src/icalendar/tests/rfc_7265_jcal/test_invalid_jcal.py
```diff
@@ -453,14 +453,85 @@ def test_skip():
...
+    ):
+        Parameters.from_jcal_property(["x-prop", {name: "v"}, "text", "x"])
+
+
+@pytest.mark.parametrize("key", ["FREQ:INJECT", "FREQ\rINJECT", "a;b", "a=b", ""])
+def test_invalid_recur_part_name_token(key):
+    """A jCal RRULE part name that is not a valid token is rejected."""
+    with pytest.raises(
+        JCalParsingError,
+        match=r"The recurrence rule part name must be valid iCalendar token",
```
stevepiercy:
```suggestion
        match=r"The recurrence rule part name must be a valid iCalendar token",
```

src/icalendar/error.py
```diff
@@ -231,6 +231,51 @@ def validate_list_type(
...
+        verbatim and re-emits it into the content line on serialization, so a
+        ``:``, ``;``, or lone carriage return in the name could inject
+        parameters or a new content line. A valid name matches the iCalendar
+        token pattern ``[\w.-]+`` and, per :rfc:`7265`, must be lower case.
+
+        Parameters:
+            name: The jCal name to validate.
+            kind: Names the token in the error message, for example,
+                ``"property name"``.
+            parser: The parser or component the name belongs to.
```
stevepiercy:
```suggestion
            parser: The parser or component to which the name belongs.
```

src/icalendar/error.py
```diff
@@ -231,6 +231,51 @@ def validate_list_type(
...
+        See also:
+            :meth:`~icalendar.parser.string.validate_token`
+        """
+        from icalendar.parser.string import validate_token
+
+        try:
+            validate_token(name)
+        except ValueError:
+            raise cls(
+                rf"The {kind} must be valid iCalendar token, matching the "
```
stevepiercy:
```suggestion
                rf"The {kind} must be a valid iCalendar token, matching the "
```

src/icalendar/error.py
```diff
@@ -231,6 +231,51 @@ def validate_list_type(
...
+                rf"regular expression pattern `[\w.-]+`.",
+                parser,
+                path,
+                value=name,
+            ) from None
+        if name != name.lower():
+            raise cls(f"The {kind} must be lower case.", parser, path, value=name)
```
stevepiercy: Per https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/l/lowercase.

```suggestion
            raise cls(f"The {kind} must be lowercase.", parser, path, value=name)
```

stevepiercy (changes requested):
Per https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/l/lowercase, would you please normalize the use of both "lower-case", "lower_case", and "lower case" to "lowercase"? This also goes for test names and any other Python objects. I only made one suggestion for over a dozen instances.

src/icalendar/cal/component.py
niccokunzmann: It is not an instead of duplication. When the property is validated, if could also now validate its property name in all places. That could be an improvement. So inlining has possible value in other places, too. What do you think? Do you agree?

alhudz: I traced where the property name actually reaches serialisation before answering. It is only `_node_from_jcal`, through `component.add(prop_name, v_prop)`. That is the single place the untrusted name becomes a content-line key.

The per-type `from_jcal` methods that call `validate_property` read `jcal_property[0]` only to dispatch (`if name == "categories"` and the like) and then drop it. They return the value object and the caller supplies the name, so `vText.from_jcal(["summary:INJECT", {}, "text", "hi"])` gives back just `hi` with no params and the bad name never lands in output.

So folding the token and lowercase check into `validate_property` would not cover an extra serialisation path. On the parse path it doubles the check already in `_node_from_jcal`, and on the direct per-type path it starts rejecting names those methods currently ignore, which is a behaviour change across every property type rather than a tighter guard.

I do agree it reads cleaner to validate the name wherever a property is validated. If you want that consistency in this PR I can move the call into `validate_property` and drop the explicit one in `_node_from_jcal`, keeping the standalone `validate_jcal_token` for the parameter and recur part names, since those are dict keys that never pass through `validate_property`. Happy either way, just say which you prefer.

alhudz:
@stevepiercy done. `lower-case`/`lower case` is now `lowercase` across the error message, the `validate_jcal_token` docstring, the three test names with their docstrings and matchers, and the news entry. Also applied your inline suggestions: the `a valid iCalendar token` wording in the error and the three test matchers, and the `parser` docstring line. jCal suite is still green.

stevepiercy (dismissed):
Docstrings and error messages LGTM now. Thank you for the quick response!

src/icalendar/cal/component.py
alhudz: Agreed, and done in the latest push. Moved the token and lowercase check into `validate_property`, so the property name is now validated wherever a property is validated (all the per-type `from_jcal` methods, not just `_node_from_jcal`), and dropped the now-redundant call there. The message and path for the parse path are unchanged. `validate_jcal_token` stays as the shared helper and is still called directly for parameter names and `RRULE` part names, since those are dict keys that never pass through `validate_property`. Full jCal suite still green (1555 passed).

stevepiercy: @niccokunzmann OK to merge?

niccokunzmann: @alhudz If this is not your writing but an AI, I would value not reading any of that AI written text but your own summary.

niccokunzmann (approved):

niccokunzmann:
Ready for merge. Thanks for the contribution!

stevepiercy:
@niccokunzmann merging is blocked, all comments must be resolved.

PR #1467, opened 2026-06-16
validate input type in vDDDTypes.from_ical before calling .upper()

HrachShah:
vDDDTypes.from_ical started with `ical.upper()` without guarding against non-string input, so passing `None`, an `int`, a `float`, a `datetime` instance, a `list`, or a `dict` surfaced a confusing `AttributeError: 'NoneType' object has no attribute 'upper'`. The new path wraps the `.upper()` call in a `try/except (AttributeError, TypeError)` and translates both into a clear `ValueError` that names the offending type and repr, matching the wrapping pattern used by the sibling `vDate`, `vDatetime`, `vTime`, `vPeriod`, and `vDuration` `from_ical` helpers (e.g. `vDatetime.from_ical`'s `raise ValueError(f"Wrong datetime format: {ical}") from e`).

Added a parametrized test in `test_vDDDTypes` covering `None`, `int`, `float`, `datetime`, `date`, `list`, and `dict` inputs and asserting the new clear `ValueError`. All 14,682 existing tests still pass; the new tests fail on the unpatched code with the original `AttributeError`.

----
📚 Documentation preview 📚: https://icalendar--1467.org.readthedocs.build/en/1467/

### DIFF
--- a/src/icalendar/prop/dt/types.py
+++ b/src/icalendar/prop/dt/types.py
@@ -87,7 +87,18 @@ class vDDDTypes(TimeBase):
     def from_ical(cls, ical, timezone=None):
         if isinstance(ical, cls):
             return ical.dt
-        u = ical.upper()
+        try:
+            u = ical.upper()
+        except (AttributeError, TypeError) as e:
+            # `ical` is not a str/bytes (e.g. int, None, datetime) or is bytes
+            # whose .upper() returns bytes that can't be matched against the str
+            # prefixes/tuples below. Surface a clear ValueError naming the
+            # offending input's type, matching the wrapping pattern used by
+            # vDate / vDatetime / vTime / vPeriod / vDuration.from_ical.
+            raise ValueError(
+                f"Expected datetime, date, or time as str or bytes, "
+                f"got {type(ical).__name__}: {ical!r}"
+            ) from e
         if u.startswith(("P", "-P", "+P")):
             return vDuration.from_ical(ical)
         if "/" in u:
--- a/src/icalendar/cal/component.py
+++ b/src/icalendar/cal/component.py
@@ -389,7 +389,23 @@ class Component(CaselessDict):
     # Handling of components
 
     def add_component(self, component: Component) -> None:
-        """Add a subcomponent to this component."""
+        """Add a subcomponent to this component.
+
+        Raises:
+            TypeError: If ``component`` is not an instance of :class:`Component`.
+                The subcomponents list is type-hinted as ``list[Component]`` and
+                ``property_items``/``to_ical``/``walk`` all call methods on
+                whatever is stored there, so a stray non-Component entry
+                (string, dict, None) used to crash the next serialization
+                with ``AttributeError: 'str' object has no attribute
+                'property_items'``. The type check here turns that into a
+                clear, immediate TypeError at the call site.
+        """
+        if not isinstance(component, Component):
+            raise TypeError(
+                f"add_component requires a Component instance, got "
+                f"{type(component).__name__}: {component!r}"
+            )
         self.subcomponents.append(component)
 
     def _walk(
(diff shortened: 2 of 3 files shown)


src/icalendar/prop/dt/types.py
```diff
@@ -92,7 +87,18 @@ def to_ical(self) -> bytes:
     def from_ical(cls, ical, timezone=None):
         if isinstance(ical, cls):
             return ical.dt
-        u = ical.upper()
+        try:
```
angatha, line 90: A type check should be prefered over exceptions. I tried to derive the possible type (also as verification) and found `vTime.from_ical`, which is wrongly annotated as just `str` despite it handling `bytes` explicitly.

`ical` in `vDuration.from_ical` may be annotated with `str` since `Pattern.match` takes just `str`. Since that, it may call `to_unicode` or this method should, which makes the remaining `bytes` handling easyer. You may also decide to not support `bytes` as that is currently not the case.

`vPeriod.from_ical` has no annotated type. It may be just `str` or depending on the annotation of this argument `ical`, it may be `bytes` but the implementation needs to be adjusted so `ical.split("/")` is done for `str` and `ical.split(b"/")` for a `bytes` argument

src/icalendar/prop/dt/types.py
```diff
@@ -92,7 +87,18 @@ def to_ical(self) -> bytes:
     def from_ical(cls, ical, timezone=None):
```
angatha, line 87:
```suggestion
    def from_ical(cls, ical: self | str | bytes, timezone=None):
```

src/icalendar/prop/dt/types.py
```diff
@@ -92,7 +87,18 @@ def to_ical(self) -> bytes:
...
+            # `ical` is not a str/bytes (e.g. int, None, datetime) or is bytes
+            # whose .upper() returns bytes that can't be matched against the str
+            # prefixes/tuples below. Surface a clear ValueError naming the
+            # offending input's type, matching the wrapping pattern used by
+            # vDate / vDatetime / vTime / vPeriod / vDuration.from_ical.
+            raise ValueError(
+                f"Expected datetime, date, or time as str or bytes, "
+                f"got {type(ical).__name__}: {ical!r}"
+            ) from e
         if u.startswith(("P", "-P", "+P")):
```
angatha, line 102: Will fail for `bytes`. `bytes may be supported by a type check and then using `.startswith((b"P", b"-P", b"+P"))`

src/icalendar/prop/dt/types.py
```diff
@@ -92,7 +87,18 @@ def to_ical(self) -> bytes:
...
             return vDuration.from_ical(ical)
         if "/" in u:
```
angatha, line 104: Also a type error. Would need a `b"/" in u` instead

angatha (changes requested):
Thank you for your contribution. The first type check is fine but the second one with `try-except` is to vague. Please repalce it with a type check: `isinstance(ical, (str, bytes))` or just `str` depending on your decision.

binggao1230:
The `ruff` job is failing because `ruff format .` reformats `src/icalendar/tests/prop/test_vDDDTypes.py`. No semantic change seems needed; this is the local formatting diff:

```diff
diff --git a/src/icalendar/tests/prop/test_vDDDTypes.py b/src/icalendar/tests/prop/test_vDDDTypes.py
@@
 @pytest.mark.parametrize(
     "bad_input",
-    [None, 42, 3.14, datetime(2025, 1, 1), date(2025, 1, 1), ["20250101"], {"dt": "20250101"}],
+    [
+        None,
+        42,
+        3.14,
+        datetime(2025, 1, 1),
+        date(2025, 1, 1),
+        ["20250101"],
+        {"dt": "20250101"},
+    ],
     ids=["None", "int", "float", "datetime", "date", "list", "dict"],
 )
@@
-    with pytest.raises(ValueError, match="Expected datetime, date, or time as str or bytes"):
+    with pytest.raises(
+        ValueError, match="Expected datetime, date, or time as str or bytes"
+    ):
         vDDDTypes.from_ical(bad_input)
```

I verified locally on the PR head with:

```bash
ruff format src/icalendar/tests/prop/test_vDDDTypes.py
ruff check src/icalendar/tests/prop/test_vDDDTypes.py
python -m pytest -q src/icalendar/tests/prop/test_vDDDTypes.py
```

The test file passed: `14 passed`.

stevepiercy:

-   create a new issue first
-   update branch to main
-   pass CI checks
-   resolve merge conflicts

PR #1468, opened 2026-06-17
reject CR and LF in vUri, vCalAddress and vInline values

alhudz:
## Description

`vUri`, `vCalAddress` and `vInline` are `str` subclasses whose `to_ical()` writes the value verbatim, with none of the escaping `vText` applies. A value carrying a lone `\r` therefore lands raw in the content line, where the only guard is `Contentline.__new__`'s `assert "\n" not in value`, which never looks at `\r`. Consumers that treat a bare CR as a line break read the tail as a separate property.

`repro`: `Calendar.from_jcal([..., ["url", {}, "uri", "http://e.com/a\rX-EVIL:1"], ...]).to_ical()`
`expected`: the malformed value is rejected
`actual`: the output contains `URL:http://e.com/a\rX-EVIL:1`, an injected `X-EVIL` content line on a bare-CR consumer

These value types cannot encode a newline the way `TEXT` does, so the guard rejects raw `CR`/`LF` in `__new__` rather than escaping. `vXmlReference` inherits the check from `vUri`.

## Checklist

----
📚 Documentation preview 📚: https://icalendar--1468.org.readthedocs.build/en/1468/


stevepiercy:
@alhudz please see my comment regarding the treatment of a lone `\r` in https://github.com/collective/icalendar/pull/1462#issuecomment-4728005548. Would you please defer any further PRs on how to treat a lone `\r` until we definitively settle the matter?

niccokunzmann:
Yes, I think, it is good to put our heads together in an issue that lists the way to go forward and also all the places that need touching. That is to keep a record of the decisions. This is a really good find. We should give it proper attention and the documentation that it needs, so we can be sure people will understand the code and decisions in the years to come. Thank @alhudz would you like to open the issue? As a maintainer, I would chip in and do some edits, with a checklist of what I would like to see done to really assure this is thoroughly handled everywhere. Right now, splitting decision making across PRs calls in inconsistency over the long term. Thanks for the work so far!

Note: For documentation:

- [ ] link to the issue
- [ ] quote RFC5545 and 6868 on line breaks
- [ ] outline how the RFC handles line breaks - actually platform dependent!

TODO:

- [ ] So, this is also why I wanted the `ical_value` property - that can take care of converting the text into a platform dependent version. e.g. a solution vText can always just contain `\n`, `vText.ical_value` does the conversion to the platform.
- [ ] list all the places that are affected by \r and platform-dependent line breaks
- [ ] decide on a strategy
- [ ] implement the change consistently
- [ ] do not touch vUnkown (check with jCal RFC)

@alhudz You see, the issue is quite big and also, to solve this you do not need to do all the work. Could you open an issue to start us off?

alhudz:
Opened #1471 to gather the line-break treatment in one place, with the RFC quotes, the affected paths and a checklist to start from. Left the structure loose so you can drop in the items you want tracked.

Happy to hold off on further lone `\r` PRs until the strategy's settled there, and I'll pick up the implementation once it is.

alhudz:
any update?

niccokunzmann:
I left a comment on the other PR. Cannot find it on mobile. But yeah. It progresses.

alhudz:
Found it, the note on #1462 about re-using and parametrising the test code. Done here: the class list and the injection payloads are shared constants now, and the jCal test re-uses the same payloads across `URL`, `ATTENDEE` and `ORGANIZER` rather than one hard-coded `URL` case. 32 parametrised cases, all green locally.

src/icalendar/prop/uri.py
```diff
@@ -66,6 +66,10 @@ def __new__(
         params: dict[str, Any] | None = None,
     ) -> Self:
         value = to_unicode(value, encoding=encoding)
+        if "\r" in value or "\n" in value:
```
niccokunzmann, line 69: I think the ICalParsingError is now in main.
Could you ha e a look at it and at the issue and give your insight in case we cod improve the roadmap? This is one of the examples where a wrong value is sometimes passed when parsing and sometimes when creating or adding values. Curious about your take on how to offer a nice API.

niccokunzmann (dismissed):
This is ready to merge.

niccokunzmann:
@alhudz I would like your input here:
https://github.com/collective/icalendar/pull/1468#discussion_r3589647355

src/icalendar/tests/prop/test_crlf_in_values.py
```diff
@@ -0,0 +1,64 @@
+r"""``URI``, ``CAL-ADDRESS`` and inline values must reject raw CR and LF.
+
+Unlike ``TEXT`` these value types are not escaped on serialisation, so a raw
```
stevepiercy:
```suggestion
Unlike ``TEXT``, these value types are not escaped on serialization, so a raw
```

src/icalendar/tests/prop/test_crlf_in_values.py
```diff
@@ -0,0 +1,64 @@
...
+
+@pytest.mark.parametrize("cls", VERBATIM_VALUE_TYPES)
+def test_valid_values_still_parse(cls):
+    assert cls.from_ical("http://example.com/my-report.txt")
+
+
+@pytest.mark.parametrize(("prop", "jcal_type"), JCAL_PROPERTIES)
+@pytest.mark.parametrize("value", CRLF_INJECTION_VALUES)
+def test_jcal_value_with_cr_or_lf_is_rejected(prop, jcal_type, value):
+    """A CR or LF from untrusted jCal must not reach the serialised output."""
```
stevepiercy:
```suggestion
    """A CR or LF from untrusted jCal must not reach the serialized output."""
```

news/+reject-crlf-in-uri-values.bugfix
```diff
@@ -0,0 +1 @@
+Reject raw CR and LF in :class:`~icalendar.prop.uri.vUri`, :class:`~icalendar.prop.cal_address.vCalAddress` and :class:`~icalendar.prop.inline.vInline` (and so :class:`~icalendar.prop.xml_reference.vXmlReference`). These value types are not escaped on serialisation, so a lone ``\r`` in the value was written straight into the content line, slipping past the newline-only assertion in ``Contentline`` and reparsing as a separate property. Reachable from untrusted jCal via ``URL``/``ATTENDEE``/``ORGANIZER``. @alhudz
```
stevepiercy: Although you say that `vXmlReference` rejects CR and LF, I don't see where you've actually done what you say. I also don't see where in the existing module that it rejects it. Can you please explain? Thank you!

```suggestion
Reject raw CR and LF in :class:`~icalendar.prop.uri.vUri`, :class:`~icalendar.prop.cal_address.vCalAddress`, :class:`~icalendar.prop.inline.vInline`, and :class:`~icalendar.prop.xml_reference.vXmlReference`. These value types are not escaped on serialization, so a lone ``\r`` in the value was written straight into the content line, slipping past the newline-only assertion in ``Contentline`` and reparsing as a separate property. This was reachable from untrusted jCal via ``URL``, ``ATTENDEE``, and ``ORGANIZER``. @alhudz
```

stevepiercy (changes requested):
Some grammar and American English spellings, and a question about vXmlReference. Would you please take care? Thank you!

news/+reject-crlf-in-uri-values.bugfix
alhudz: Pushed all three suggestions, thanks for the wording.

On `vXmlReference`: it subclasses `vUri` (`src/icalendar/prop/xml_reference.py`) and doesn't override `__new__`, so the new check in `vUri.__new__` runs for it too; there's no separate code in the xml_reference module and no pre-existing check there either. It is tested directly though: `vXmlReference` is in `VERBATIM_VALUE_TYPES` in the test file, so all four injection payloads run against it.

src/icalendar/prop/uri.py
alhudz: Had a look at `error.py` on main. My take: the value types should stay context-free. `vUri.__new__` can't know whether it's being called from the ical parser, `from_jcal` or user code building a calendar, so it should raise one plain error about the value itself. The context lives in the parsers, so the parsing entry points catch at the boundary and wrap: `ICalParsingError` with `line`/`line_number` for ical, `JCalParsingError` with the `path` for jCal, using `raise ... from` so the original is chained.

That gives users one rule: parsing raises a `*ParsingError` that says where, direct construction raises the plain error that says what. And since both sit under `InvalidCalendar(ValueError)`, `except ValueError` keeps working everywhere.

For this PR that means the guard could raise `InvalidCalendar` (or a dedicated subclass for invalid values) instead of bare `ValueError`. The tests here only assert `ValueError`, so tightening it later is non-breaking; happy to switch it in this PR or in the follow-up, whichever you prefer. Can copy this into #1471 if you want it on the roadmap.

news/+reject-crlf-in-uri-values.bugfix
stevepiercy: Thanks for the explanation. I appreciate it.

stevepiercy (approved):

PR #1349, opened 2026-04-30
Strictly validate Base64 binary values

uwezkhan:
This PR makes vBinary reject malformed Base64 input instead of silently ignoring invalid characters during decoding.

PR #1472, opened 2026-06-17
split categories value at the first colon outside parameters

alhudz:
## Linked issue

No existing issue; found while reading the parser.

## Description

`Repro:` `CATEGORIES:CONFIDENTIAL,http://example.com/tag` parses to `['//example.com/tag']`.
`Cause:` `handle_categories` locates the value with `line.rfind(":")`, the last colon on the line. `TEXT` values do not escape `:`, so a colon inside the value (a URI here) is taken as the separator and the real categories are dropped.
`Fix:` find the value separator as the first colon outside a quoted parameter section, the same boundary `Contentline.parts()` already uses. A colon inside a quoted parameter (`ALTREP="http://x"`) is still skipped.

## Checklist

## Additional information

The new `Contentline.value_separator_index` mirrors the scan in `parts()`, so a category value containing a colon round-trips and a quoted parameter colon is not mistaken for the separator.

----
📚 Documentation preview 📚: https://icalendar--1472.org.readthedocs.build/en/1472/


src/icalendar/parser/content_line.py
```diff
@@ -234,6 +234,28 @@ def parts(self) -> tuple[str, Parameters, str]:
...
+        quoted parameter value (for example ``ALTREP="http://x"``) is skipped,
+        and a colon that belongs to the value (``TEXT`` does not escape ``:``)
+        is not mistaken for the separator. Returns ``-1`` if there is none.
+        """
+        in_quotes = False
+        escaped = False
+        for i, ch in enumerate(self):
+            if ch == '"' and not escaped:
+                in_quotes = not in_quotes
+            elif ch == "\\" and not in_quotes:
```
angatha: In which section does the spec define escaping?

alhudz: It isn't, for this region. Backslash escaping is only defined for the `TEXT` value type in RFC 5545 §3.3.11 (`ESCAPED-CHAR = "\\" / "\;" / "\," / "\N" / "\n"`), which is the value after the colon. The name and parameter grammar in §3.1 uses `paramtext` / `quoted-string`, where `\` is just a `SAFE-CHAR` with no special meaning.

So the `escaped` branch isn't implementing spec escaping. It's only there so this scan returns the exact same colon `parts()` treats as `value_split`, since `parts()` already carries that branch (the loop just above, ~L187-201). The two have to agree, otherwise `handle_categories` would split at a different colon than the parameter parser. Quick check on `CATEGORIES;X=a\":b:VALUE`: both `parts()` and `value_separator_index()` land on index 16; drop the branch and the scan returns `-1` instead.

Happy to pull it out if you'd rather not mirror `parts()`, but I kept them in lockstep on purpose.

angatha: See the following example: `CATEGORIES;X-A=\::`. The category is `:`, just the last character. The colon separating the value is at index 16, the first colon, but this algorithm returned 17. I did not find any comment in the commit introducing that in `parts()`, which may explain it. I would classify it as a bug. Here is how I derived the example. Maybe I did some mistake:

```
categories
"CATEGORIES" catparam ":" text CRLF
"CATEGORIES" ";" other-param ":" text CRLF
"CATEGORIES" ";" x-param ":" text CRLF
"CATEGORIES" ";" x-name "=" param-value *("," param-value) ":" text CRLF
"CATEGORIES" ";" "X-" [vendorid "-"] 1*(ALPHA / DIGIT / "-") "=" param-value *("," param-value) ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" param-value *("," param-value) ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" param-value ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" paramtext ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" *SAFE-CHAR ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" %x5C ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" "\" ":" text CRLF
"CATEGORIES" ";" "X-" "A" "=" "\" ":" *(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR) CRLF
"CATEGORIES" ";" "X-" "A" "=" "\" ":" ":" CRLF
"CATEGORIES;X-A=\::" CRLF
```

based on the definitions from [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545).

```
categories = "CATEGORIES" catparam ":" text *("," text) CRLF

catparam   = *(
                  ;
                  ; The following is OPTIONAL,
                  ; but MUST NOT occur more than once.
                  ;
                  (";" languageparam ) /
                  ;
                  ; The following is OPTIONAL,
                  ; and MAY occur more than once.
                  ;
                  (";" other-param)
                  ;
                  )
other-param   = (iana-param / x-param)

x-param     = x-name "=" param-value *("," param-value)

x-name        = "X-" [vendorid "-"] 1*(ALPHA / DIGIT / "-")

param-value   = paramtext / quoted-string

paramtext     = *SAFE-CHAR

SAFE-CHAR     = WSP / %x21 / %x23-2B / %x2D-39 / %x3C-7E
                   / NON-US-ASCII

text       = *(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)
```

alhudz: Good catch, you're right. Dropped the `escaped` branch from `value_separator_index`, so it now returns the first colon outside a quoted section. For `CATEGORIES;X-A=\::` it lands on index 16 and the category is `:`, as your derivation shows. Added a regression test for it.

You're also right that `parts()` carries the same branch and has the same off-by-one. That one predates this PR (it came in with #1349), so I've left it out of scope here to keep the diff focused. Happy to fix it in a separate PR, or fold it in here if you'd rather have both done together.

angatha (approved):

src/icalendar/tests/test_categories_value_with_colon.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_value_separator_index_skips_quoted_colon():
+    line = Contentline('CATEGORIES;ALTREP="http://p":A,B')
+    assert line.value_separator_index() == line.index(":A,B")
+
+
+def test_value_separator_index_does_not_treat_backslash_as_escape():
+    # Backslash has no special meaning in the parameter grammar (RFC 5545
+    # §3.1), so the first colon after ``X-A=\`` is the value separator and
+    # the category is the single ``:``.
```
niccokunzmann, line 62: Yeah, good test!

niccokunzmann (approved):
Nice! Thanks for the thorough work and understanding of the grammar!

src/icalendar/parser/content_line.py
niccokunzmann: Yes, please open a separate PR for that. I also think, it is useful to open an issue for that to describe the general bug that we might find in several places. Then, we can use that issue to

- ask others where they suppose that bug can occur
- make them aware of the specifics of the grammar (prevents later mistakes)
- link all PRs to the issue (helps keep problem description separate from solutions and discussions about diffs)

Would you be up for creating an issue?

I think, this PR can be merged. I would really value an issue to describe the problem.

niccokunzmann:
@angatha Thanks for the review! That really speeded it up for me!
@alhudz Thanks for finding these very specific bugs! Much appreciated :pray:

src/icalendar/parser/content_line.py
```diff
@@ -234,6 +234,25 @@ def parts(self) -> tuple[str, Parameters, str]:
...
+    def value_separator_index(self) -> int:
+        r"""Return the index of the colon that separates the value.
+
+        This is the first colon that is not inside a quoted parameter section.
+        A colon inside a quoted parameter value (for example
+        ``ALTREP="http://x"``) is skipped, and a colon that belongs to the
+        value (``TEXT`` does not escape ``:``) is not mistaken for the
+        separator. Backslash has no special meaning in the parameter grammar
+        (RFC 5545 §3.1), so it is treated as an ordinary character. Returns
+        ``-1`` if there is none.
```
stevepiercy, line 246: @alhudz I'm late to the game, but would you please amend this docstring with a link to the RFC and a Returns section heading, per https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#general-cross-references and https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure?

@niccokunzmann whenever there's a docstring involved, please tag me or request a review before merging to save effort of a follow up PR. Thank you!

```suggestion
        :rfc:`5545#section3.1`, so it is treated as an ordinary character.

        Returns:
            An integer representing the index position of the separator,
            or ``-1`` if there is none.
```

stevepiercy (commented):
@alhudz @niccokunzmann please see my review. Thank you!

src/icalendar/parser/content_line.py
niccokunzmann: Ah. Alright! Sure ... 
@alhudz Would you be up for creating a new PR with the docstring change?

Sorry for the confusion :|

alhudz: This was already merged so I couldn't amend it here. Opened #1498 with the change: `:rfc:` role for the §3.1 reference and a `Returns` section. I used the `#section-3.1` anchor to match the existing `:rfc:` usages in the tree.

src/icalendar/parser/content_line.py
alhudz: Opened #1497 to track the general problem and list the spots (`parts()`, `_strip_ows_around_delimiters()`, and `value_separator_index()` which #1472 already changed).

On the separate PR: I had a go at dropping the `escaped` branch from `parts()` to match `value_separator_index()`, but it regresses the lenient backslash escaping of parameter values that's intentional here. `CN=Society\: 2014` and `ALTREP=http\://x` (the latter from #1480) both stop round-tripping. So it's a policy call rather than a one-liner, which is why I put the trade-off in the issue. Happy to send the PR once we've settled which way to go.

PR #1479, opened 2026-06-18
Resolve globally unique TZIDs with a vendor prefix (#313)

### LINKED ISSUE #313: parsing uncommon timezone formats
Hi,

I was reported an [issue where the time zone format look different](https://github.com/niccokunzmann/python-recurring-ical-events/issues/27).

`DTSTART;TZID=/freeassociation.sourceforge.net/Europe/Berlin: 20200426T140000`
is the format
it is parsed as UTC where the time zone `Europe/Berlin` would be expected.
Did you encounter such time zone formats anywhere else? Should this be covered by this module?

vjsai:
## Linked issue

- [x] Closes #313

## Description

A *globally unique* `TZID` (per [RFC 5545 §3.2.19](https://www.rfc-editor.org/rfc/rfc5545#section-3.2.19)) begins with a `/` and is emitted by clients such as libical and Evolution in the form `/<vendor>/<Olson/Name>` — for example:

```
DTSTART;TZID=/freeassociation.sourceforge.net/Europe/Berlin:20200426T140000
```

`TZP.timezone()` did not recognise this form. `clean_timezone_id()` only strips leading/trailing slashes, so `/freeassociation.sourceforge.net/Europe/Berlin` became `freeassociation.sourceforge.net/Europe/Berlin`, which matches no provider zone, no `WINDOWS_TO_OLSON` alias and no cached component. `timezone()` therefore returned `None`, and the datetime was parsed as **timezone-naive** — silently dropping the offset (the original bug report in #313).

### Fix

When the existing lookups fail and the id begins with `/`, strip the vendor prefix and resolve the trailing Olson identifier, trying the **longest suffix first** so multi-part names like `America/Argentina/Buenos_Aires` still match. Unknown ids continue to return `None`, so there are no false positives, and non-prefixed ids are untouched.

Before / after for the reported value:

| | `dt.tzinfo` | `dt.utcoffset()` |
|---|---|---|
| before | `None` | `None` |
| after | `Europe/Berlin` | `2:00:00` (CEST) |

### Tests

New regression test `test_issue_313_globally_unique_tzid.py`, parametrised over **both** the `zoneinfo` and `pytz` providers, covering: the reported value parses tz-aware to `Europe/Berlin`; an extra vendor path component (`/citadel.org/20190914_1/Europe/Berlin`); a multi-part Olson name behind a prefix; and an unknown suffix still resolving to `None`. The full timezone/parsing suite passes (`8172 passed`), and `ruff check` / `ruff format` are clean.

## Checklist

- [x] I've added a change log entry to `/news` (`news/313.bugfix`).
- [x] I've run and ensured all tests pass locally.
- [x] I've added or edited documentation, both as docstrings and narrative documentation, as necessary (the new resolution path is documented in the method docstring; no narrative-docs change needed for this behavioural fix).

## Additional information

This change was prepared with AI assistance (Claude Code, `claude-opus-4-8`); the news fragment notes this in line with the project's existing convention. I have reviewed and understand the change and how `TZP.timezone()` feeds the datetime/time/date parsing paths.


src/icalendar/timezone/tzp.py
```diff
@@ -145,7 +145,29 @@ def timezone(self, tz_id: str) -> datetime.tzinfo | None:
...
+    def _timezone_from_global_id(self, tz_id: str) -> datetime.tzinfo | None:
+        """Resolve a globally unique TZID (RFC 5545, section 3.2.19).
+
+        A TZID beginning with a ``/`` is a "globally unique" identifier, which
+        clients such as libical/Evolution emit as ``/<vendor>/<Olson/Name>``
+        (e.g. ``/freeassociation.sourceforge.net/Europe/Berlin``). The vendor
+        prefix is not a real timezone, so strip it and resolve the trailing
+        Olson identifier. Try the longest suffix first so multi-part names such
+        as ``America/Argentina/Buenos_Aires`` still match.
+        """
```
angatha: Is it convention for vendors to only assign such suffixes when tey mean the same timezone as everyone else? I would prefer to emmit a warning, when we assume one timezone to be another. Maybe in a follow up PR, we can introduce a machanism for users to remove the warning by letting them define such mappings themself.

vjsai: You're right, its not guaranteed the globally unique TZID is opaque (RFC 5545 3.2.19) the real rules are in the VTIMEZONE, and the Europe/Berlin tail is just convention.In its favor, this only fires as a last resort once an embedded VTIMEZONE, the cache, and normal lookups all come up empty but it's still a guess, so a warning makes sense So I'll add one.I'll use a small UserWarning subclass so people can filter just this case instead of all UserWarnings. For user defined mappings I will raise a followup PR. Is it ok ?

niccokunzmann: We have this quote:

> Note: This document does not define a naming convention for
         time zone identifiers.  Implementers may want to use the naming
         conventions defined in existing time zone specifications such
         as the public-domain TZ database [[TZDB](https://www.rfc-editor.org/rfc/rfc5545.html#ref-TZDB)].  The specification of
         globally unique time zone identifiers is not addressed by this
         document and is left for future study.
         
Also, as far as I was informed, the VTIMEZONE is mostly irgnored now-adays.
So, what would be really good is to do this:

1. Try to resolve the timezone without the vendor prefix
2. If we can, then we use that
3. If we cannot resolve it, we parse the VTIMEZONE

I don't think that we need a warning for that. It is alright. If someone uses a valid name, we can assume that they mean that timezone.

Thanks for the discussion... I will have a look now..

src/icalendar/tests/test_issue_313_globally_unique_tzid.py
```diff
@@ -0,0 +1,61 @@
...
+    assert dt.tzinfo is not None
+    assert dt.utcoffset() == timedelta(hours=2)  # CEST on 2020-04-26
+    assert tzid_from_tzinfo(dt.tzinfo) == "Europe/Berlin"
+
+
+@pytest.mark.parametrize(
+    "tz_id",
+    [
+        "/freeassociation.sourceforge.net/Europe/Berlin",
+        "/citadel.org/20190914_1/Europe/Berlin",  # extra vendor path component
```
niccokunzmann: Could you please have a look online if you find other examples?

Also, please do not just test Europe/Berlin.
We have other timezones, even without / and also with more /. So they should also work. Thanks for parametrizing the test.

src/icalendar/tests/test_issue_313_globally_unique_tzid.py
```diff
@@ -0,0 +1,61 @@
...
+
+import pytest
+
+from icalendar import Calendar
+from icalendar.timezone import tzid_from_tzinfo
+
+GLOBAL_BERLIN = "/freeassociation.sourceforge.net/Europe/Berlin"
+
+
+def _event_ics(tzid: str) -> str:
```
niccokunzmann: Please use a .ics file for this.

Also, use 3 events with different timezones of different formats.

src/icalendar/tests/test_issue_313_globally_unique_tzid.py
```diff
@@ -0,0 +1,61 @@
...
+    """The vendor-prefixed TZID resolves to Europe/Berlin, not a naive datetime."""
+    calendar = Calendar.from_ical(_event_ics(GLOBAL_BERLIN))
+    dt = calendar.walk("VEVENT")[0]["DTSTART"].dt
+    assert dt.tzinfo is not None
+    assert dt.utcoffset() == timedelta(hours=2)  # CEST on 2020-04-26
+    assert tzid_from_tzinfo(dt.tzinfo) == "Europe/Berlin"
+
+
+@pytest.mark.parametrize(
+    "tz_id",
```
niccokunzmann: `("tzid", "resolved_tzid")`

Please use that and merge the all the following tests into this one.

src/icalendar/timezone/tzp.py
niccokunzmann: This algorithm basically is a multi-step approach that could be made nicer and more transparent. It just grew with time.

It would be nice if we can document it and make sure we continue to understand it.

1. we test for a cleaned timezone id
2. we test for OLSON
3. we test for unclean
4. we test for cache
5. we test your algorithm

This is all a bit weird. I feel like yours could be much earlier!Also, your result should be in the cache.

So, to clean that, it would be good change the algorithm:

1. create a generator of possible names to look up, yours goes before the unclean name. just a function with different yield statements
2. they all check the cache first
3. then we see if we get a valid timezone
4. we cache the result under the first name, so that we get a fast look-up next time
5. the result is returned.

That would clean it up. Would you be up for the change? Let me know if I request too much! Or if anything is unclear!

niccokunzmann (changes requested):
Thanks for the PR! This is really nice. I suggested some changes and you seem to touch code that has grown over time and can use a bit of clean up so that your fix is made best use of!

niccokunzmann (dismissed):
Yay! I was looking a while through the interaction of caching and this method. It is really nice. I love it. It seems to work and as far as I see, there is no issue.

stevepiercy:
@niccokunzmann auto-merge is blocked because you left unresolved conversations. OK to merge?

src/icalendar/timezone/tzp.py
```diff
@@ -137,15 +139,49 @@ def clean_timezone_id(self, tzid: str) -> str:
...
-        """Return a timezone with an id or None if we cannot find it."""
-        _unclean_id = tz_id
-        tz_id = self.clean_timezone_id(tz_id)
-        tz = self.__provider.timezone(tz_id)
-        if tz is not None:
-            return tz
-        if tz_id in WINDOWS_TO_OLSON:
-            tz = self.__provider.timezone(WINDOWS_TO_OLSON[tz_id])
-        return tz or self.__provider.timezone(_unclean_id) or self.__tz_cache.get(tz_id)
+        """Return a timezone with an id or None if we cannot find it.
```
stevepiercy:
```suggestion
        """Return a timezone with an ID or ``None`` if we can't find it.
```

src/icalendar/timezone/tzp.py
```diff
@@ -137,15 +139,49 @@ def clean_timezone_id(self, tzid: str) -> str:
...
+
+        ``tz_id`` may be a plain Olson name (``Europe/Berlin``), a Windows
+        timezone name, or a "globally unique" identifier
+        (:rfc:`5545#section-3.2.19`) such as
+        ``/freeassociation.sourceforge.net/Europe/Berlin``. We try the
+        candidate ids from :meth:`_lookup_ids` in order, checking the cache
```
stevepiercy:
```suggestion
        candidate IDs from :meth:`_lookup_ids` in order, checking the cache
```

src/icalendar/timezone/tzp.py
```diff
@@ -137,15 +139,49 @@ def clean_timezone_id(self, tzid: str) -> str:
...
+        before the provider for each one, and cache the first match under the
+        primary id so the next lookup is fast.
```
stevepiercy:
```suggestion
        primary ID so the next lookup is fast.
```

src/icalendar/timezone/tzp.py
```diff
@@ -137,15 +139,49 @@ def clean_timezone_id(self, tzid: str) -> str:
...
+            if primary is None:
+                primary = lookup_id
+            tz = self.__tz_cache.get(lookup_id) or self.__provider.timezone(lookup_id)
+            if tz is not None:
+                self.__tz_cache[primary] = tz
+                return tz
+        return None
+
+    def _lookup_ids(self, tz_id: str) -> Iterator[str]:
+        """Yield the ids to try when resolving ``tz_id``, best match first.
```
stevepiercy:
```suggestion
        """Yield the IDs to try when resolving ``tz_id``, best match first.
```

src/icalendar/timezone/tzp.py
```diff
@@ -137,15 +139,49 @@ def clean_timezone_id(self, tzid: str) -> str:
...
+        1.  the cleaned id, without any surrounding ``/``
+        2.  the Olson name of a Windows timezone (e.g.
+            ``W. Europe Standard Time`` -> ``Europe/Berlin``)
+        3.  for a "globally unique" TZID (:rfc:`5545#section-3.2.19`) of the
+            form ``/<vendor>/<Olson/Name>`` -- emitted by clients such as
+            libical, Evolution and Mozilla Lightning -- the trailing Olson
+            identifier, dropping vendor path components from the front. The
+            longest suffix is tried first so multi-part names such as
+            ``America/Argentina/Buenos_Aires`` still match.
+        4.  the original, unmodified id
```
stevepiercy:
```suggestion
        1.  The cleaned ID, without any surrounding ``/``.
        2.  The Olson name of a Windows timezone (for example,
            ``W. Europe Standard Time`` -> ``Europe/Berlin``).
        3.  For a "globally unique" TZID (:rfc:`5545#section-3.2.19`) of the
            form ``/<vendor>/<Olson/Name>``—emitted by clients such as
            libical, Evolution and Mozilla Lightning—the trailing Olson
            identifier, dropping vendor path components from the front. The
            longest suffix is tried first, so multi-part names such as
            ``America/Argentina/Buenos_Aires`` still match.
        4.  The original, unmodified ID.
```

news/313.bugfix
```diff
@@ -0,0 +1 @@
+Resolve the timezone of a "globally unique" :rfc:`5545#section-3.2.19` ``TZID`` such as ``/freeassociation.sourceforge.net/Europe/Berlin`` (emitted by clients like libical and Evolution). :meth:`TZP.timezone <icalendar.timezone.tzp.TZP.timezone>` now strips the leading vendor prefix and resolves the trailing Olson identifier, so the value is parsed as timezone-aware instead of being silently treated as naive. Prepared with AI assistance. @vjsai
```
stevepiercy:
```suggestion
Resolve the timezone of a "globally unique" :rfc:`5545#section-3.2.19` ``TZID`` such as ``/freeassociation.sourceforge.net/Europe/Berlin``, which is emitted by some clients, including libical and Evolution. :meth:`TZP.timezone <icalendar.timezone.tzp.TZP.timezone>` now strips the leading vendor prefix and resolves the trailing Olson identifier, so the value is parsed as timezone-aware instead of being silently treated as naive. Prepared with AI assistance. @vjsai
```

stevepiercy (changes requested):
Some grammar and spelling tweaks. ID is capitalized, whereas id is a Latin term used in psychology.

niccokunzmann:
@vjsai Thanks for your contribution! It is technically sound and we would like to merge it. @stevepiercy has suggested edits to the documentation. When they are merged, we will merge this. Could you have a look?

stevepiercy:
@niccokunzmann feel free to apply the suggestions in a batch change, since it's been 10 days since I made the suggestions. Depending on your default GitHub UI choice, you can go to the `Files changed` tab and click `Add suggestion to batch` for each item, then merge them all in one commit. See https://docs.github.com/en/pull-requests/how-tos/review-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

SashankBhamidi:
Hey @vjsai, picked this up due to inactivity.

Applied Steve's pending suggestions and added the `GloballyUniqueTZIDGuessed` warning you committed to in the thread. 

@stevepiercy if everything looks good to you, feel free to merge.

SashankBhamidi (approved):
LGTM. Thanks!

stevepiercy (approved):

PR #1480, opened 2026-06-20
preserve literal percent escapes in parameter values

alhudz:
## Linked issue

No existing issue; found while round-tripping URI-valued parameters.

## Description

`Repro:` parse a content line whose parameter value contains a literal percent escape, e.g. `X-P;ALTREP="http://x/a%2Cb%3Ac%3Bd%5Ce":v`.
`Expected:` `ALTREP` parses back as `http://x/a%2Cb%3Ac%3Bd%5Ce`.
`Actual:` it comes back as `http://x/a,b:c;d\e`.
`Cause:` `Contentline.parts()` shields backslash-escaped delimiters from the parameter splitter by rewriting `\,` `\:` `\;` `\\` to the markers `%2C` `%3A` `%3B` `%5C` via `_escape_string`, then reverses that with `_unescape_string`. The reverse step can't tell a marker it produced from one already present in the value, so any real percent escape in the value is decoded too. Percent encoding is normal in `ALTREP`/`DELEGATED-FROM`/`MEMBER`/`SENT-BY` URIs, and `%3A` -> `:` can move a URI's scheme boundary.
`Fix:` escape the escape character. `_escape_string` maps a literal `%` to `%25` first and `_unescape_string` restores `%25` last, so the transport encoding is a bijection and pre-existing percent sequences pass through untouched. Backslash-escaped delimiters still decode exactly as before.

## Checklist

## Additional information

New regression in `src/icalendar/tests/test_parsing.py` fails on the current parser and passes with the fix.

----
📚 Documentation preview 📚: https://icalendar--1480.org.readthedocs.build/en/1480/


src/icalendar/tests/test_parsing.py
```diff
@@ -45,6 +45,24 @@ def test_content_lines_parsed_properly(raw_content_line, expected_output):
...
+@pytest.mark.parametrize(
+    ("raw_content_line", "expected_altrep"),
+    [
+        # Literal percent escapes in a parameter value (common in URIs) must
+        # survive parsing. They collided with the parser's internal
+        # ``\\,`` -> ``%2C`` transport encoding and were silently decoded.
+        ('X-P;ALTREP="http://x/a%2Cb%3Ac%3Bd%5Ce":v', "http://x/a%2Cb%3Ac%3Bd%5Ce"),
+        ('X-P;ALTREP="http://x/%2520":v', "http://x/%2520"),
+        # A backslash-escaped delimiter still decodes to the bare character.
+        ("X-P;ALTREP=a\\,b:v", "a,b"),
```
angatha, line 60: Can you please include a Test like `"X-P;ALTREP=http\\://x/a%2Cb%3Ac%3Bd%5Ce:v"` to check that percent encoding is preserved when the parameter is not quoted. Besides that, the code loogs good. Thank you for your contribution.

angatha (changes requested):

src/icalendar/tests/test_parsing.py
alhudz: Added that case. The unquoted `http\://x/a%2Cb%3Ac%3Bd%5Ce:v` parses back to `http://x/a%2Cb%3Ac%3Bd%5Ce`, so the escaped colon decodes while the percent escapes survive. All four cases pass.

angatha (approved):

niccokunzmann:
@alhudz Thanks for your contribution! @angatha Thanks for the review!

PR #1485, opened 2026-06-21
Add gh-profiler CI workflow

### LINKED ISSUE #1484: Add gh-profiler to CI to help triage low-quality contributions
So around the 7.1.3 release we got hit with a bunch of AI-generated PRs and we were talking about it in the group chat. @stevepiercy brought up https://github.com/ehmatthes/gh-profiler and I think it's worth adding here.

Basically when a PR or issue opens it runs a quick check on the contributor: account age, how many of their recent external PRs actually got merged, whether they've been spamming the same issue title across a bunch of repos. Posts a little summary so you have some context before you even look at the PR.

One YAML file to set up, uses the default GITHUB_TOKEN. And it's not going to cry wolf on a genuine first-time contributor who just hasn't filled their profile out yet, the heuristics are smart enough to handle that. We still decide what to do, it doesn't auto-close anything.

Eric has two variants: one posts the output inline as a comment, the other just links to the action log. I'd go with the link-only one so it doesn't clutter up every PR thread.

Worth trying out.

SashankBhamidi:
Closes #1484

Adds the link-only variant of the gh-profiler workflow so we get a contributor profile link on every new PR and issue.

### DIFF
--- a/news/1484.internal
+++ b/news/1484.internal
@@ -1,0 +1,1 @@
+Add gh-profiler workflow to help triage AI-generated contributions.
--- a/news/1484.internal
+++ b/news/1484.internal
@@ -1,1 +1,1 @@
-Add gh-profiler workflow to help triage AI-generated contributions.
+Add gh-profiler workflow to help triage AI-generated contributions. @SashankBhamidi
--- a/.github/workflows/profile-contributors.yml
+++ b/.github/workflows/profile-contributors.yml
@@ -1,0 +1,73 @@
+# Run gh-profiler against the author of each new issue and PR.
+# See https://github.com/ehmatthes/gh-profiler
+
+name: Profile contributors
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  profile-author:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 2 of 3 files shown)


SashankBhamidi:
The workflow triggers on `issues: opened` and `pull_request_target: opened`, so it'll fire on the next PR or issue opened after this one is merged.

Let's open a dummy issue on the repo after merging, it'll trigger immediately.

niccokunzmann (dismissed):
Hi, this seems nice.

I would like there to be a configurable action step, so that there is lesscopiesy & paste for standard reporting bahaviour. Can be an issue.

niccokunzmann:
I can merge this. Does it need documenting somewhere? In this case, we can follow up with a PR or issue.

news/1484.internal
```diff
@@ -0,0 +1 @@
+Add gh-profiler workflow to help triage AI-generated contributions.
```
stevepiercy:
```suggestion
Add gh-profiler workflow to help triage AI-generated contributions. @SashankBhamidi
```

stevepiercy (changes requested):
LGTM, but giving you credit.

stevepiercy:
> I can merge this. Does it need documenting somewhere? In this case, we can follow up with a PR or issue.

Docs added in https://github.com/collective/icalendar/pull/1485/commits/fcaa4e580d4d366c0a9d2339e0d4a59c06ab81a6. I wanted to keep it vague, while adding that we'll investigate, which was previously omitted.

niccokunzmann (approved):

stevepiercy (approved):

PR #1488, opened 2026-06-21
Reject non-ASCII digits in vWeekday ordwk (BYDAY/BYWEEKDAY/WKST)

Labib-Bin-Salam:
## Linked issue

No related issue; reporting directly (same as #1465).

## Description

`vWeekday` silently accepts non-ASCII digits in the `ordwk` (relative) part of a `BYDAY` / `BYWEEKDAY` / `WKST` value.

**Repro:** `vWeekday("١٢MO")` (Arabic-Indic "12"), or parse a calendar containing `RRULE:FREQ=MONTHLY;BYDAY=١٢MO`.

```pycon
>>> from icalendar import vWeekday
>>> w = vWeekday("١٢MO")   # Arabic-Indic digits
>>> w.relative
12
```

**Cause:** `WEEKDAY_RULE` matches the relative part with `(?P<relative>[\d]{0,2})`. In Python's default Unicode mode, `\d` is `True` for non-ASCII decimal digits (Arabic-Indic `١٢`, extended-Arabic `۱۲`, Devanagari, …). `int()` then parses those, so a malformed `ordwk` is silently accepted and normalized to a valid `relative` number. [RFC 5545 §3.3.10](https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10) allows only ASCII `DIGIT` here (`ordwk = 1*2DIGIT`).

**Fix:** match `[0-9]` instead of `\d` so the grammar agrees with what `int()` accepts. After the fix, `vWeekday("١٢MO")` raises `ValueError` like any other malformed weekday, while every valid ASCII value (`MO`, `2MO`, `-1SU`, `+3WE`, `53SU`) is unchanged.

This is the direct sibling of the recently merged `vMonth` fix (#1465), which fixed the same non-ASCII-digit class in `BYMONTH`.

## Checklist

- [ ] I've added or edited documentation, both as docstrings to be rendered in the API documentation and narrative documentation, as necessary. *(N/A — behaviour fix; added an explanatory code comment only.)*

## Additional information

- Added regression test `test_non_ascii_ordwk_digits_rejected` (Arabic-Indic and extended-Arabic digits, single- and double-digit) next to the existing `vWeekday` tests; it fails before the change and passes after.
- `ruff check` and `ruff format --check` are clean on the changed files; the targeted `prop/` (887) and recurrence/`RRULE` (444) suites pass locally. (The unrelated `test_timezone_identification` / `test_with_doctest` failures reproduce identically on a clean checkout in this environment and are not affected by this change.)
- **AI usage disclosure** (per the "Responsible AI use" policy): this change was drafted with the assistance of Claude Opus 4.8 (`claude-opus-4-8`), run via Claude Code — used to locate this sibling of the `vMonth` fix and to draft the regex change, the regression test, and the change log entry. All output was reviewed, executed, and verified by me, and I take responsibility for it. The model/version and usage are also recorded in the commit message and the disclosure is included in the change log entry.

----
📚 Documentation preview 📚: https://icalendar--1488.org.readthedocs.build/en/1488/


src/icalendar/tests/prop/test_vWeekday.py
```diff
@@ -27,6 +27,21 @@ def test_error():
         vWeekday.from_ical("-100MO")
 
 
+@pytest.mark.parametrize("value", ["١٢MO", "٥SU", "۱۲FR", "١MO"])
+def test_non_ascii_ordwk_digits_rejected(value):
+    r"""Non-ASCII digits must not be accepted in the ``ordwk`` (relative) part.
+
+    ``\d`` is ``True`` for non-ASCII digits (e.g. Arabic-Indic ones), and
+    ``int`` parses them, so an Arabic-Indic ``"12MO"`` was silently accepted
```
angatha: `"12MO"` is the valid version, did you mean `"١٢MO"`?
```suggestion
    ``int`` parses them, so an Arabic-Indic ``"١٢MO"`` was silently accepted
```

angatha: @stevepiercy Is there anything you want to change with the comments/documentation? Otherwise, I would merge this.

news/+reject-non-ascii-weekday-digits.bugfix
```diff
@@ -0,0 +1 @@
+Reject non-ASCII digits in the ``ordwk`` part of :class:`~icalendar.prop.recur.weekday.vWeekday` (``BYDAY``, ``BYWEEKDAY``, and ``WKST``). The value validation regular expression matched ``\d``, which is ``True`` for digits such as the Arabic-Indic ``١٢``; ``int`` then parsed them, so a value such as ``١٢MO`` was silently accepted as ``relative == 12`` instead of being rejected as required by :rfc:`5545#section-3.3.10` (``ordwk = 1*2DIGIT``). This mirrors the earlier :class:`~icalendar.prop.recur.month.vMonth` fix. Drafted with AI assistance (Claude Opus 4.8); reviewed, tested, and verified by the author. @Labib-Bin-Salam
```
stevepiercy: This was difficult to parse. Shorter sentences are easier to read. Also ``int`` is a type, not a parser, and a few grammar fixes.

```suggestion
Reject non-ASCII digits in the ``ordwk`` part of :class:`~icalendar.prop.recur.weekday.vWeekday` (``BYDAY``, ``BYWEEKDAY``, and ``WKST``). The value validation regular expression ``\d`` matches non-ASCII digits, such as the Arabic-Indic ``١٢``. A value such as ``١٢MO`` was silently accepted as ``relative == 12``, instead of being rejected, as required by :rfc:`5545#section-3.3.10` (``ordwk = 1*2DIGIT``). This mirrors the earlier :class:`~icalendar.prop.recur.month.vMonth` fix. Drafted with AI assistance (Claude Opus 4.8); reviewed, tested, and verified by the author. @Labib-Bin-Salam
```

src/icalendar/prop/recur/weekday.py
```diff
@@ -9,8 +9,12 @@
 from icalendar.parser import Parameters
 from icalendar.parser_tools import DEFAULT_ENCODING, to_unicode
 
+# ``\d`` matches non-ASCII digits (e.g. the Arabic-Indic "١٢"), which ``int``
+# then parses below, silently turning a malformed ``ordwk`` into a valid
+# ``relative`` number. RFC 5545, section 3.3.10 allows only ASCII ``DIGIT``
+# here, so match ``[0-9]`` to stay consistent with what ``int`` accepts.
```
stevepiercy: This comment won't make sense after it's merged because `\d` won't appear except in the git history.

```suggestion
# Use ``[0-9]`` to explicitly specify that only ASCII digits should be
# matched, instead of ``\d``. In Python, ``\d`` matches a digit zero through
# nine in any script except ideographic scripts, and is equivalent to
# ``\p{Nd}``. For example, the Arabic-Indic "١٢" would match. Here, a
# malformed ``ordwk`` would get parsed into a valid ``relative`` number.
# RFC 5545, section 3.3.10, allows only ASCII characters per its definition
# of ``DIGIT``.
```

src/icalendar/tests/prop/test_vWeekday.py
```diff
@@ -27,6 +27,21 @@ def test_error():
...
+    as ``relative == 12``, contrary to :rfc:`5545#section-3.3.10`, which allows
+    only ASCII ``DIGIT`` (``ordwk = 1*2DIGIT``).
```
stevepiercy: I'd reuse my comment from above, as there are multiple problems with this docstring. `int` is a type, not a parser. You caught another issue. Tests aren't rendered in docs to links.


```suggestion
    Use ``[0-9]`` to explicitly specify that only ASCII digits should be
    matched, instead of ``\d``. In Python, ``\d`` matches a digit zero through
    nine in any script except ideographic scripts, and is equivalent to
    ``\p{Nd}``. For example, the Arabic-Indic "١٢" would match. Here, a
    malformed ``ordwk`` would get parsed into a valid ``relative`` number.
    RFC 5545, section 3.3.10, allows only ASCII characters per its definition
    of ``DIGIT``.
```

stevepiercy (changes requested):

SashankBhamidi (approved):
Hey @Labib-Bin-Salam, picked this up due to inactivity.

Applied Steve's and Angatha's suggestions. Note: angatha's docstring suggestion contained Arabic-Indic digits which triggered a ruff RUF002 lint error, so I dropped the inline example from the docstring instead (the parametrize values above already illustrate it).

@stevepiercy @angatha re-requesting your review. This LGTM and is ready for merge.

angatha (approved):

stevepiercy (approved):

PR #1491, opened 2026-06-21
skip None values in Parameters.to_ical instead of crashing

HrachShah:
Parameters.to_ical() dereferenced the value without checking for None.

The `single_string_parameter` setter (cn, sent-by, etc.) removes the key when assigned None, but direct dict-style assignment (`params['CN'] = None`) goes through CaselessDict.__setitem__ and stores the literal None. Any later to_ical call then crashed:

- For a key in `always_quoted` (ALTREP, DIR, MEMBER, SENT-BY, X-ADDRESS, X-TITLE, LINKREL, DELEGATED-FROM, DELEGATED-TO) `param_value()` did `value.to_ical().decode(...)` on None and raised AttributeError.
- For CN (the `quote_also` entry) the generator expression `any(c in value for c in check_quoteable_characters)` raised `TypeError: argument of type 'NoneType' is not iterable`.
- For a sequence value with a None inside (e.g. `params['MEMBER'] = ['mailto:a@x.com', None, 'mailto:b@x.com']`) `dquote()` raised `TypeError: expected string or bytes-like object, got 'NoneType'` on the None entry.

The realistic path that hits this: build a vCalAddress, attach it to an event, set its CN (or DIR, or MEMBER) to None via dict-style assignment (which is exactly what data round-tripped through `from_ical` with a missing value looks like), then call `event.to_ical()`. Before the fix, this raised mid-serialization and the caller had no clean way to opt out — the documented setter path does the right thing (`fset` deletes on None) but only for the typed property descriptors, not for direct `params[key] = ...` writes.

This PR treats None as 'absent' for scalar values (drop the key entirely) and drops individual None entries from list/tuple values; if a sequence ends up empty after filtering, the whole key is dropped. That matches what the documented property setters do, and it lets the realistic 'ORGANIZER without CN' calendar round-trip cleanly instead of raising mid-serialization. Three new tests cover the always_quoted keys, the quote_also key, sequence values with None entries, and the realistic ORGANIZER-without-CN round trip. All 14,678 existing tests still pass.

----
📚 Documentation preview 📚: https://icalendar--1491.org.readthedocs.build/en/1491/

### DIFF
--- a/src/icalendar/cal/component.py
+++ b/src/icalendar/cal/component.py
@@ -389,7 +389,23 @@ class Component(CaselessDict):
     # Handling of components
 
     def add_component(self, component: Component) -> None:
-        """Add a subcomponent to this component."""
+        """Add a subcomponent to this component.
+
+        Raises:
+            TypeError: If ``component`` is not an instance of :class:`Component`.
+                The subcomponents list is type-hinted as ``list[Component]`` and
+                ``property_items``/``to_ical``/``walk`` all call methods on
+                whatever is stored there, so a stray non-Component entry
+                (string, dict, None) used to crash the next serialization
+                with ``AttributeError: 'str' object has no attribute
+                'property_items'``. The type check here turns that into a
+                clear, immediate TypeError at the call site.
+        """
+        if not isinstance(component, Component):
+            raise TypeError(
+                f"add_component requires a Component instance, got "
+                f"{type(component).__name__}: {component!r}"
+            )
         self.subcomponents.append(component)
 
     def _walk(
--- a/src/icalendar/cal/journal.py
+++ b/src/icalendar/cal/journal.py
@@ -265,7 +265,6 @@ class Journal(Component):
         journal.url = url
         journal.organizer = organizer
         journal.contacts = contacts
-        journal.start = start
         journal.status = status
         journal.attendees = attendees
         journal.RECURRENCE_ID = recurrence_id
--- a/src/icalendar/parser/parameter.py
+++ b/src/icalendar/parser/parameter.py
@@ -323,6 +323,13 @@ class Parameters(CaselessDict):
             items.sort()
 
         for key, value in items:
+            if value is None:
+                # single_string_parameter setters delete the key when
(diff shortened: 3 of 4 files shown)


src/icalendar/cal/component.py
```diff
@@ -430,20 +389,34 @@ def set_inline(self, name, values, encode=1):
     # Handling of components
 
     def add_component(self, component: Component) -> None:
-        """Add a subcomponent to this component."""
+        """Add a subcomponent to this component.
```
niccokunzmann, line 392: Hi, this has nothing to do with setting the value to None. Please open separate pull requests for separate issues. Branch off the latest `main` commit and do not mix contributions.

niccokunzmann (changes requested):
Hi, I do find contributions in here that do not fit in with the description.
Could you mention an issue or open one to discuss this problem before contributing?

Also, please follow our AI guideline: https://icalendar.readthedocs.io/en/stable/contribute/index.html#artificial-intelligence-policy

src/icalendar/cal/component.py
```diff
@@ -430,20 +389,34 @@ def set_inline(self, name, values, encode=1):
...
+
+        Raises:
+            TypeError: If ``component`` is not an instance of :class:`Component`.
+                The subcomponents list is type-hinted as ``list[Component]`` and
+                ``property_items``/``to_ical``/``walk`` all call methods on
+                whatever is stored there, so a stray non-Component entry
+                (string, dict, None) used to crash the next serialization
+                with ``AttributeError: 'str' object has no attribute
+                'property_items'``. The type check here turns that into a
+                clear, immediate TypeError at the call site.
```
angatha, line 402: @stevepiercy do this information belong here?

stevepiercy: No, that looks more like a description to me. I'd move it above `Raises:` and dedent and rewrap it so it becomes the description.

stevepiercy:

-   create a new issue first
-   update branch to main
-   pass CI checks
-   resolve merge conflicts

PR #1502, opened 2026-06-23
Fix line folding across escape sequences

### LINKED ISSUE #1501: [Bug] Description wrapping can create an invalid event
## Describe the bug

When a field is wrapped, it doesn't treat an escaped line break within the text as a single character. When you import a calendar with that event into Google Calendar, the event will not show (some calendars can handle this, e.g. Thunderbird).

## To reproduce

Make a description longer than 75 characters, make the 75th character a \ and the 76th character an n, and try to import a calendar with that event into Google Calendar, and it will not show.

This code will make an event that fails:
```python
    event = Event()
    title = 'Test event'
    event.add('uid', '123456789')
    event.add('dtstamp', datetime.now())
    event.add('summary', 'Test event')
    event['location'] = vText('Place')
    description = 'This is a description that will be wrapped, and it also has a\nnew line in it right where it wraps. It doesn't display in Google'
    event.add('description', description)
    event.add('dtstart', event_start)
    event.add('dtend', event_end)
    cal.add_component(event)
```

Output:

The resulting description is:
```pycon
DESCRIPTION:This is a description that will be wrapped, and it also has a\
 nnew line in it right where it wraps.
```
In my case, the description was a url, followed by a line break, then followed by some text.

## Expected behavior

It should see that the text has \n where it is going to wrap and if so wrap one character earlier, e.g. 
```pycon
DESCRIPTION:This is a description that will be wrapped, and it also has a
 \nnew line in it right where it wraps.
```
## Environment

 - [ ] Operating system:  Windows 11
 - [ ] Python version:  Python 3.14
 - [ ] `icalendar` version:   icalendar 7.1.1

## Additional context

Put this event in an ics file, and try to import it into Google Maps, and it never displays
```text
BEGIN:VEVENT
SUMMARY:Test event
DTSTART;VALUE=DATE:20260718
DTEND;VALUE=DATE:20260720
DTSTAMP:20260614T144333Z
UID:123TESTEVENT456
DESCRIPTION:This is a description that will be wrapped, and it also has a\
 nnew line in it right where it wraps.
LOCATION:Denmark
END:VEVENT
```

(sorry I just saved the description that failed, not the full event, and then made a mock event, since I don't know if I can share the actual event)

kingrubic:
## Summary
- fixes #1501 by avoiding folded line breaks between escape prefixes and their escaped characters
- applies to TEXT backslash escapes (`\n`, `\,`, etc.) and RFC 6868 parameter escapes (`^n`, `^^`, `^\'`)
- adds a regression test covering both escape families
- adds a towncrier bugfix fragment

## Verification
- `uv run pytest src/icalendar/tests/test_parsing.py -k '1501'`
- `uv run pytest src/icalendar/tests/test_parsing.py`
- `uv run pytest` (16090 passed, 28 skipped)
- `uv run ruff check src/icalendar/parser/string.py src/icalendar/tests/test_parsing.py`
- `uv run ruff format --check src/icalendar/parser/string.py src/icalendar/tests/test_parsing.py`

### DIFF
--- a/src/icalendar/parser/string.py
+++ b/src/icalendar/parser/string.py
@@ -129,27 +129,31 @@ def _foldline(line: str, limit: int = 75, fold_sep: str = "\r\n ") -> str:
     assert isinstance(line, str)
     assert "\n" not in line
 
-    # Use a fast and simple variant for the common case that line is all ASCII.
-    try:
-        line.encode("ascii")
-    except (UnicodeEncodeError, UnicodeDecodeError):
-        pass
-    else:
-        return fold_sep.join(
-            line[i : i + limit - 1] for i in range(0, len(line), limit - 1)
-        )
-
-    ret_chars: list[str] = []
+    folded_lines: list[str] = []
+    current_chars: list[str] = []
     byte_count = 0
     for char in line:
         char_byte_len = len(char.encode(DEFAULT_ENCODING))
+        if current_chars and byte_count + char_byte_len >= limit:
+            # For compatibility with existing clients, avoid splitting escaped
+            # values such as TEXT backslash escapes or RFC 6868 parameter
+            # escapes across a folded line boundary. See issue #1501.
+            if len(current_chars) > 1 and current_chars[-1] in r"\^":
+                escaped_prefix = current_chars.pop()
+                folded_lines.append("".join(current_chars))
+                current_chars = [escaped_prefix]
+                byte_count = len(escaped_prefix.encode(DEFAULT_ENCODING))
+            else:
+                folded_lines.append("".join(current_chars))
+                current_chars = []
+                byte_count = 0
+        current_chars.append(char)
         byte_count += char_byte_len
-        if byte_count >= limit:
-            ret_chars.append(fold_sep)
-            byte_count = char_byte_len
-        ret_chars.append(char)
 
-    return "".join(ret_chars)
+    if current_chars:
+        folded_lines.append("".join(current_chars))
+
+    return fold_sep.join(folded_lines)
 
 
 foldline = deprecate_for_version_8(_foldline)
--- a/src/icalendar/tests/test_parsing.py
+++ b/src/icalendar/tests/test_parsing.py
@@ -9,7 +9,13 @@ from icalendar import vBinary, vRecur
 from icalendar.cal.calendar import Calendar
 from icalendar.cal.component_factory import ComponentFactory
 from icalendar.cal.event import Event
-from icalendar.parser import Contentline, Parameters, _escape_char, _unescape_char
+from icalendar.parser import (
+    Contentline,
+    Parameters,
+    _escape_char,
+    _foldline,
+    _unescape_char,
+)
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


src/icalendar/parser/string.py
```diff
@@ -121,27 +121,30 @@ def _foldline(line: str, limit: int = 75, fold_sep: str = "\r\n ") -> str:
...
-
-    ret_chars: list[str] = []
+    folded_lines: list[str] = []
+    current_chars: list[str] = []
     byte_count = 0
     for char in line:
         char_byte_len = len(char.encode(DEFAULT_ENCODING))
+        if current_chars and byte_count + char_byte_len >= limit:
+            # Avoid splitting escaped values such as TEXT backslash escapes
+            # or RFC 6868 parameter escapes across a folded line boundary.
```
angatha: I commented #1501 and asked for a newer RFC version for content line folding. If no one could provide a version explicitly forbidding splitting thouse sequence, can you please include the reason (compatibility) why we aren't following the SHOULD from RFC 5545 or at least link the issue so everyone seeing that code knows that this is intentional.

Besides that, the code looks great.

I would wait for an answer on the issue before accepting that.

angatha (changes requested):

niccokunzmann:
Hi, if you used AI, please see our AI policy and disclose the model.

src/icalendar/tests/test_parsing.py
```diff
@@ -123,6 +123,29 @@ def test_description_parsed_properly_issue_53(events):
...
+        r"\;",
+        r"\\",
+        "^n",
+        "^^",
+        "^'",
+    ],
+)
+def test_foldline_does_not_split_escape_sequences_issue_1501(escape_sequence):
+    """Issue #1501 - line folding must not split escape sequences."""
+    prefix = "DESCRIPTION:" + ("a" * (74 - len("DESCRIPTION:") - 1))
```
niccokunzmann: Please use pytest.mark.parametrize to check that this breaking of escape characters does not happen in the first line with length 60-90 and also not for more than one line. We could fix the test right now by just breaking at 80 chars for example.

Also add a test that checks that line folding takes place.

niccokunzmann (changes requested):
Thanks for the work. Please follow our AI policay.

kingrubic:
Thanks for the review. I updated the PR to address both requests:

- expanded the regression test with `pytest.mark.parametrize` across fold limits 60-90
- added a multi-line folding case so the escape-aware behavior is covered after the first folded line too
- added an assertion that folding still happens, so the test cannot pass by avoiding folding
- added a compatibility comment in `_foldline` that points future readers to issue #1501
- added the AI disclosure to the changelog fragment and amended the commit message with the model/workflow summary

Validated locally with:

- `uv run pytest src/icalendar/tests/test_parsing.py -k '1501'`\n- `uv run ruff check src/icalendar/parser/string.py src/icalendar/tests/test_parsing.py`\n- `uv run ruff format --check src/icalendar/parser/string.py src/icalendar/tests/test_parsing.py`\n- `git diff --check`

niccokunzmann (dismissed):
Thanks! I think it does the job and is ready for merge!

kingrubic:
Thank you for the approval. I rebased the branch onto current `main`, resolved the test import conflict while preserving both upstream coverage and this regression test, and force-pushed with lease. Focused tests, Ruff checks, formatting, and `git diff --check` pass locally; CI is rerunning now.

angatha (approved):

src/icalendar/tests/test_parsing.py
angatha: Close conversation since that was done and you said it is ready for mearge

Issue #1473, opened 2026-06-17
Clarify auto-generated properties for `new` class methods


stevepiercy:
This is a good issue for new-to-icalendar contributors. Help is welcome!

In #1459, we realized that the initial choice of wording in the docstrings for the `new()` class methods might not be accurate in regards to whether properties are required, optional, or recommended, per the [RFC specification](https://datatracker.ietf.org/doc/html/rfc5545.html#section-2) of these terms.

To work on this issue, select one class, then:

- Compare the usage of its current docstring grammar against their Python implementation and their RFC specification, and determine the next step.
- Either add a comment to this issue stating that you have verified that the docstring is correct as is, or create a pull request where you adjust the class's docstring to align with both the Python implementation and RFC.
- For a pull request, follow our [contributor guide](https://icalendar.readthedocs.io/en/latest/contribute/index.html) and its relevant sections.

## Files to check

- [ ] src/icalendar/cal/alarm.py
- [ ] src/icalendar/cal/availability.py
- [ ] src/icalendar/cal/available.py
- [ ] src/icalendar/cal/event.py
- [ ] src/icalendar/cal/free_busy.py
- [ ] src/icalendar/cal/journal.py
- [ ] src/icalendar/cal/todo.py

_Originally posted by @scop in https://github.com/collective/icalendar/issues/1459#issuecomment-4734519544_

andycheng2018:
Hi! I'm a student looking to make my first contribution. I'd like to take src/icalendar/cal/free_busy.py: its new() docstring says "Create a new alarm...", which looks like it should reference FreeBusy. I'll fix that and check the property wording against RFC 5545.

w3lld1:
I’ll check `src/icalendar/cal/journal.py`. The current `Journal.new()` docstring documents an `end` parameter that is not present in the method signature; I’ll keep the change focused on that API-doc mismatch and validate the documentation build.

niccokunzmann:
Thanks for volunteering! Please check if there are PRs already for this and if thisis the case, consider giving them a review.

sandeep780049:
I opened #1800 covering the remaining unchecked files: availability.py (wrong "Create a new event" wording + undocumented `priority`/`components`), todo.py (missing `priority`), available.py (undocumented required `start`) and alarm.py (docstring claimed "all required properties" while ACTION/TRIGGER are not set). event.py was audited and is correct as is.

PR #1459, opened 2026-06-15
Clarify Calendar.new auto generated properties

scop:
## Linked issue

Replace `ISSUE_NUMBER` with the issue number that your pull request addresses. This links the pull request to the related issue. Choose the appropriate form.

If you completely resolve the issue, then use `"Closes"`.
GitHub will automatically close the related issue when the PR gets merged.

- [ ] Closes #ISSUE_NUMBER

If you resolve only a part of the issue, then use `"See"`.
This form keeps the issue open for future work. 

- [ ] See #ISSUE_NUMBER

## Description

new() generates a UID which is not required, but recommended.

https://www.rfc-editor.org/info/rfc7986/#section-5.3

## Checklist

- [ ] I've added a change log entry to `/news`, following the instructions in [Change log entry format](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-entry-format).
- [ ] I've added or updated tests if applicable.
- [ ] I've run and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [ ] I've added or edited documentation, both as docstrings to be rendered in the API documentation and narrative documentation, as necessary.

## Additional information

Upload screenshots, videos, links to documentation, or any other relevant information.

----
📚 Documentation preview 📚: https://icalendar--1459.org.readthedocs.build/en/1459/

PR #1503, opened 2026-06-24
Fix FreeBusy.new() docstring referencing alarm instead of free/busy

andycheng2018:
## Linked issue

- [x] See #1473

## Description

The `FreeBusy.new()` docstring summary was copied from the Alarm class and
incorrectly described creating an alarm. This corrects it to reference the
free/busy component.

I also verified the property required/optional structure for VFREEBUSY against
RFC 5545 §3.6.4, and the implementation already matches the specification, so
only the docstring summary text needed correcting.

This addresses the `src/icalendar/cal/free_busy.py` item from the checklist in #1473.

## Checklist

- [x] I've added a change log entry to `/news`, following the instructions in Change log entry format.
- [ ] I've added or updated tests if applicable. (Not applicable: this is a docstring text correction with no behavioral change.)
- [x] I've run and ensured all tests pass locally.
- [x] I've added or edited documentation (the corrected docstring renders in the API documentation).

## Additional information

AI assistance: Claude Opus 4.5 (Anthropic), used via the web chat interface to
help locate the issue, verify the property structure against RFC 5545, and draft
the docstring correction. I reviewed and validated the change and take
responsibility for it. This is also disclosed in the commit message and the
changelog entry.

PR #1504, opened 2026-06-25
docs: clarify Event.new() optional and auto-generated properties

checkzhao8888:
## Linked issue

- [x] See #1473

## Description

Clarifies the `Event.new()` docstring to align with RFC terminology used elsewhere in the project (for example, #1459 for `Calendar.new()`).

Changes:
- Replace the misleading summary "with all required properties" with "Create a new event."
- Mark optional parameters explicitly.
- Document auto-generated `UID` and `DTSTAMP` values when arguments are `None`.

AI disclosure: Grok assisted with reviewing RFC terminology, comparing existing `new()` docstring patterns, and drafting the update.

## Checklist


src/icalendar/cal/event.py
```diff
@@ -459,40 +459,40 @@ def new(
...
+            sequence: Optional. The :attr:`sequence` of the event.
             stamp: The :attr:`~icalendar.Component.stamp` of the event.
-                If None, this is set to the current time.
-            start: The :attr:`start` of the event.
-            status: The :attr:`status` of the event.
-            subcomponents: The subcomponents of the event.
-            summary: The :attr:`summary` of the event.
-            transparency: The :attr:`transparency` of the event.
+                If ``None``, this is set to the current UTC time.
+            start: Optional. The :attr:`start` of the event.
```
angatha: Misleading since it is not true in all cases. Please manually check the [RFC](https://datatracker.ietf.org/doc/html/rfc5545#section-3.6.1).

angatha (changes requested):

src/icalendar/cal/event.py
niccokunzmann: Also, do not write Optional. That is covered by typing.

niccokunzmann:
@angatha, I will assign this to you. I cannot wrap my head around wether this is useful. Feel free to call in steve or close it if there is no response.

src/icalendar/cal/event.py
angatha: Good point. @checkzhao8888, please remove the Optional. Then this should be a small change. If you want, you can include the condition when start is required but please word it so that users understand, that they can specify it later. You may also add that this validates `start <= end` iff. both where specified.

If you do not want to continue with this, I will close this PR in one week.

angatha (changes requested):

src/icalendar/cal/event.py
stevepiercy: I agree with @niccokunzmann to remove all the "Optional. "s.

src/icalendar/cal/event.py
```diff
@@ -459,40 +459,40 @@ def new(
         uid: str | uuid.UUID | None = None,
         url: str | None = None,
     ):
-        """Create a new event with all required properties.
+        """Create a new event.
```
stevepiercy: I often wonder what are the required properties, so this suggestion would help stop me from wondering. The properties are the Python representations of DTSTAMP and UID.

```suggestion
        """Create a new event with the required properties of ``stamp`` and ``uid``.
```

src/icalendar/cal/event.py
```diff
@@ -459,40 +459,40 @@ def new(
...
 
         This creates a new Event in accordance with :rfc:`5545`.
```
stevepiercy:
```suggestion
        This creates a new ``Event`` in accordance with :rfc:`5545#section-3.6.1`.
```

stevepiercy (changes requested):
With the suggested changes, I will approve. Thank you!

angatha (dismissed):

news/1473.documentation
stevepiercy, line 1: Revert this deletion. Never alter another contributor's news fragment.

stevepiercy (changes requested):

news/1473.documentation
SashankBhamidi: Well, I only renamed the file. 

image

See that the original fragment was from another user and not this PR author.

Also see the commit message regarding Grok usage which also is there in the news fragment, but the 1473.documentation has claude and some other user.

stevepiercy: I don't understand what you're saying.

https://github.com/collective/icalendar/pull/1503/changes#diff-cdc008713fb6266bef6cd9f32991e798e40fb8c4169fced15031fd44cb9e51f0

This PR deletes that news fragment `news/1473.documentation`. It shouldn't.

SashankBhamidi: That's weird, I didn't do it though. Must've been the author.

angatha (approved):

niccokunzmann (approved):

stevepiercy (approved):
Thank you for your contribution. Will auto-merge after CI passes.

PR #1510, opened 2026-06-26
Consolidate maintainers list with an include

stevepiercy:


SashankBhamidi (approved):
Nice! Didn't verify RTD render.

PR #1514, opened 2026-06-27
Add commit-check workflow and configuration

stevepiercy:
This is a test to see whether we can better detect commits from AI bots in CI.

See #1508.

### DIFF
--- a/.github/commit-check.toml
+++ b/.github/commit-check.toml
@@ -1,0 +1,4 @@
+# Check for bots on each commit.
+# https://commit-check.github.io/
+[commit]
+ignore_authors = ["dependabot[bot]"]
--- a/.github/workflows/commit-check.yml
+++ b/.github/workflows/commit-check.yml
@@ -1,5 +1,6 @@
 # Run commit-check against the author of each commit in a PR.
-# https://commit-check.github.io/
+# https://github.com/commit-check/commit-check-action
+# Based on  https://commit-check.github.io/
 name: Commit check
 
 on:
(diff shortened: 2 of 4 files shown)


stevepiercy:
I give up. I thought I had something that would check each commit for a bot, but I just can't figure out how to get it to work.

If someone else wants to take over this PR:

https://github.com/commit-check/commit-check

https://commit-check.github.io/commit-check/example.html#ci-cd-integration

I also tried:

https://github.com/commit-check/commit-check-action

stevepiercy:
Although it's not as robust as I would like to configure bot detection, I think the GitHub Action version of this tool is better than nothing.

The command line version that I wasted hours on seems to return a false positive in that it passes silently. It ought to return some message, such as "Commit appears to be from a human."

stevepiercy:
This GitHub action checks whether the user's last commit appears to be a bot or AI agent. That could be a problem, as we want to check all commits, not just the last one. That's why I tried to use the command line version instead of the action.

As configured, it inserts a comment into the PR only on branches in this repo, not from forks. See https://github.com/commit-check/commit-check-action#fork-pr-comments for explanation. We could enable it from forks. They provide sample workflows at https://github.com/commit-check/commit-check-action/blob/main/docs/fork-pr-comments.md.

@niccokunzmann @angatha @SashankBhamidi what do you think?

angatha (approved):
Tool seems quite naiv but the usage is fine.

niccokunzmann:
I would say, it is ok to try it out and improve, ignore or remove it later. A squash commit makes that easier to do manually.
Just go for it! Let's have a look. We should point from the discussion to all the things we try out, so that affected people can have one place to read up on what we do and we can keep track ourselves.

Merge your best guess. Thanks for working on it.

PR #1516, opened 2026-06-30
Contentlines.from_ical: drop the dead try/except that masked nothing

HrachShah:
The try/except in Contentlines.from_ical wrapped a UFOLD.sub, a list comprehension over NEWLINE.split, and a Contentline(line) constructor, and converted any Exception into ValueError('Expected StringType with content lines').

None of those steps can actually raise in the current code:
- to_unicode catches UnicodeDecodeError internally and returns a string
- UFOLD.sub uses module-level constants; no exception path
- NEWLINE.split consumes the newline, so the lines it produces never contain a newline; the assert "\n" not in value in Contentline.__new__ is therefore unreachable from this code path
- Contentline(line) inherits from str; str() does not raise on any input

That meant the try/except could only ever be triggered by a non-Exception (e.g. BaseException, KeyboardInterrupt, SystemExit) being re-raised, and converting it into a misleading ValueError. A bare except Exception also shadowed the actual call site if a future change ever started raising something real.

The block is replaced with a plain function body and a comment that explains why nothing inside can raise, so the next reader does not re-add the try/except.

Added two regression tests in tests/test_parsing.py:
- test_contentlines_from_ical_unfolded_round_trip verifies that an unfolded input round-trips through from_ical and to_ical unchanged
- test_contentlines_from_ical_handles_lf_only_line_endings verifies that LF-only line endings are accepted (a behaviour that would be broken by any future change that treats non-CRLF as malformed)

### DIFF
--- a/src/icalendar/cal/component.py
+++ b/src/icalendar/cal/component.py
@@ -389,7 +389,23 @@ class Component(CaselessDict):
     # Handling of components
 
     def add_component(self, component: Component) -> None:
-        """Add a subcomponent to this component."""
+        """Add a subcomponent to this component.
+
+        Raises:
+            TypeError: If ``component`` is not an instance of :class:`Component`.
+                The subcomponents list is type-hinted as ``list[Component]`` and
+                ``property_items``/``to_ical``/``walk`` all call methods on
+                whatever is stored there, so a stray non-Component entry
+                (string, dict, None) used to crash the next serialization
+                with ``AttributeError: 'str' object has no attribute
+                'property_items'``. The type check here turns that into a
+                clear, immediate TypeError at the call site.
+        """
+        if not isinstance(component, Component):
+            raise TypeError(
+                f"add_component requires a Component instance, got "
+                f"{type(component).__name__}: {component!r}"
+            )
         self.subcomponents.append(component)
 
     def _walk(
--- a/src/icalendar/cal/journal.py
+++ b/src/icalendar/cal/journal.py
@@ -265,7 +265,6 @@ class Journal(Component):
         journal.url = url
         journal.organizer = organizer
         journal.contacts = contacts
-        journal.start = start
         journal.status = status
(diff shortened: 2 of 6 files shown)


angatha (changes requested):
Thank you for your contribution. Since this seems to be your first pull request, note that we (as many projects) have a [templte for pull requests](https://github.com/collective/icalendar/blob/main/.github/pull_request_template.md) as well as a [contribution guide](https://github.com/collective/icalendar/blob/main/CONTRIBUTING.md).

Please read the guide and adjust your PR accordingly.

src/icalendar/cal/component.py
```diff
@@ -430,20 +389,34 @@ def set_inline(self, name, values, encode=1):
     # Handling of components
 
     def add_component(self, component: Component) -> None:
-        """Add a subcomponent to this component."""
+        """Add a subcomponent to this component.
```
angatha, line 392: #1491, #1489 and #1467 already include this non-related code and you where already informed, that you should not mix contributions:

1. https://github.com/collective/icalendar/pull/1491/changes#r3449129379
2. https://github.com/collective/icalendar/pull/1489/changes#r3449131941
3. https://github.com/collective/icalendar/pull/1467/changes#diff-2d53f355ae69458d028bde3263d693d0d3ae583870105f7861ccbb0c70a0ae1cR392

Please follow such requests (especially since it is good practis and unrelated to this project). Repeating this just drains resources from maintainers.

Aditionally please address the problems with your other PRs first before opening any new ones. Spamming may get you suspended [as per our contribution guide](https://icalendar.readthedocs.io/en/latest/contribute/index.html#pull-request-requirements).

Ones your PRs where accepted, you can create one with this change. You may also include the already addressed concerns.

stevepiercy:

-   create a new issue first
-   update branch to main
-   pass CI checks
-   resolve merge conflicts

PR #1520, opened 2026-07-01
promote date-typed alarm triggers to datetime before applying local timezone

HrachShah:
`Alarms._alarm_time` previously called `trigger.replace(tzinfo=self._local_tzinfo)` on whatever `_add` returned. But `_add` can return a pure `datetime.date` when the start/end is a DATE-typed property and the trigger delta has no time component (e.g. `-P2D`). A `date` instance has no `tzinfo` attribute and does not accept the `tzinfo` keyword in `replace()`, so any alarm on a DATE-typed DTSTART/DTEND with a local timezone set crashed with a `TypeError` when `Alarms.times` was accessed. The downstream `is_active()` then compounded this by reading `trigger.tzinfo` directly, raising `AttributeError` on a date.

The fix in `_alarm_time` promotes the trigger to a midnight datetime via `to_datetime()` when it is a pure date and a local timezone has been set, so the rest of the code can treat the trigger as a tz-aware datetime. The existing `test_start_as_date_with_delta_as_date_stays_date` still passes because the no-local-tz path is unchanged (the trigger stays a date, since there is no timezone to apply).

New tests cover:
1. trigger becomes a tz-aware datetime when local timezone is set
2. acknowledged alarm with date trigger + local timezone no longer raises
3. unacknowledged date trigger with local timezone reports active

All 14,683 existing tests still pass; the three new tests fail on the pre-fix code (raise `TypeError` / `AttributeError`) and pass with the fix.

### DIFF
--- a/src/icalendar/alarms.py
+++ b/src/icalendar/alarms.py
@@ -319,6 +319,11 @@ class Alarms:
     def _alarm_time(self, alarm: Alarm, trigger: date):
         """Create an alarm time with the additional attributes."""
         if getattr(trigger, "tzinfo", None) is None and self._local_tzinfo is not None:
+            # A pure date has no tzinfo attribute and does not accept
+            # ``tzinfo=`` in ``replace()``. Promote it to a midnight
+            # datetime first so the local timezone can be applied.
+            if is_date(trigger):
+                trigger = to_datetime(trigger)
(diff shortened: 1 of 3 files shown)


stevepiercy:

- create a new issue first
- use our PR template for the description
- add a link to an open issue in the description
- disclose responsible AI use
- add a news fragment for the change log
- update branch to `main`
- pass CI checks
- resolve merge conflicts

PR #1526, opened 2026-07-04
Preserve URI parameters on jCal round-trip

gaoflow:
`vUri.from_jcal` reconstructs a `vUri` from a jCal property and passed the parameters positionally:

```python
return cls(
    jcal_property[3],
    Parameters.from_jcal_property(jcal_property),
)
```

But `vUri.__new__(cls, value, encoding=DEFAULT_ENCODING, /, params=None)` has `encoding` as its second positional argument (positional-only, before the `/`). So the `Parameters` object was consumed as the `encoding` argument and the parameters were dropped, while `encoding` silently received a non-string. A `URI` value's parameters were therefore lost on a jCal round-trip, e.g. `ATTACH;FMTTYPE=text/plain` or `CONFERENCE;FEATURE=AUDIO`.

The fix passes them as the `params` keyword, matching every other `from_jcal` that targets an `encoding`-taking type (e.g. `vText`, `vInline`):

```python
return cls(
    jcal_property[3],
    params=Parameters.from_jcal_property(jcal_property),
)
```

`vUri` is the only `from_jcal` that had this mismatch: the other positional callers (`vCategory`, `vGeo`, adr, org) take `params` as their second positional, so they were unaffected.

`test_uri_parameters_round_trip.py` covers a single parameter, multiple parameters, `ATTACH`/`CONFERENCE` real-world cases, and jCal round-trip idempotence. Full `rfc_7265_jcal` suite passes (1213).

Prepared with the assistance of an AI coding agent (Anthropic's Claude Opus 4.8); I reviewed and validated the change locally.


angatha:
Thank you for your contribution. First please follow our contributing guide. That includes, that you disclose AI use not just in the PR but also in each commit message:

> In your git commit messages, you must specify both (1) which AI model and version you used, and (2) how you used it, by either including the prompts and interactions you used or summarizing them. You can automate this by using the ai-prompt-auto-commit pre-commit hook. See [AI prompt automation](https://icalendar.readthedocs.io/en/latest/contribute/development.html#pre-commit-ai-prompts) for setup instructions.

https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use third to last bullet point.

Note that we are working on an update of the the guideline. The relevant PR is already linked. Please have a look when it was accepted.

angatha:
LGTM

news/1526.bugfix
```diff
@@ -0,0 +1 @@
+Preserve ``URI`` value parameters on a jCal round-trip. :meth:`vUri.from_jcal <icalendar.prop.uri.vUri.from_jcal>` passed the parameters positionally, but ``vUri.__new__`` takes ``encoding`` as its second positional argument, so the parameters were consumed as the encoding and dropped (for example ``ATTACH;FMTTYPE=...`` or ``CONFERENCE;FEATURE=...`` lost their parameters). They are now passed as the ``params`` keyword. Prepared with the assistance of an AI coding agent (Anthropic's Claude Opus 4.8). @gaoflow
```
stevepiercy: "parameter" is the variable in the signature, whereas an "argument" is the actual value passed into the parameter. Also made a couple of changes to make it easier to understand.

```suggestion
Preserve ``URI`` value parameters on a jCal round-trip. :meth:`vUri.from_jcal <icalendar.prop.uri.vUri.from_jcal>` previously passed the parameters positionally, but ``vUri.__new__`` takes ``encoding`` as its second positional parameter. As a result, the ``URI`` value parameters were consumed as ``encoding`` and dropped. For example, ``ATTACH;FMTTYPE=...`` or ``CONFERENCE;FEATURE=...`` lost their parameters. They are now passed as the ``params`` keyword. Prepared with the assistance of an AI coding agent (Anthropic's Claude Opus 4.8). @gaoflow
```

stevepiercy (changes requested):

binggao1230:
Applied your suggested wording for the changelog — thanks for the parameter/argument correction and the clearer phrasing.

src/icalendar/prop/uri.py
```diff
@@ -102,7 +102,7 @@ def from_jcal(cls, jcal_property: list) -> Self:
         JCalParsingError.validate_property(jcal_property, cls)
         return cls(
             jcal_property[3],
-            Parameters.from_jcal_property(jcal_property),
+            params=Parameters.from_jcal_property(jcal_property),
```
niccokunzmann, line 109: I think, we might have other vProperty types that also take an encoding or maybe a second argument. I think it is worth checking if these need testing, too. This is not a requirement for this PR, just a TODO for me.

niccokunzmann (approved):

stevepiercy (approved):
Thank you for your contribution! Will automerge after CI checks pass.

niccokunzmann:

Issue #1448, opened 2026-06-10
from_ical() should only raise ICalParsingError


niccokunzmann:
We have `from_jcal()` which provides a really nice interface to help us understand where we hit a problem within an JSON structure. In order to reach consistency and remove inconsistencies of the past, `from_ical()` should only raise `ICalParsingError`, a subclass of `ValueError`.

Implementation:

- [x] create `ICalParsingError` in `icalendar.errors` (Subclass of InvalidCalendar)
- [x] Consistency: make JCalParsingError a subclass of InvalidCalendar
- [ ] When the error is raised, it should mention the line it is in (the line of the property) - optional, can be unknown
- [ ] The error should contain the wrong value, e.g. a malformed date - as an argument and an attribute - consistent formatting ensures that this is mentioned in the same place always and never required to be part of the description. E.g. `raise ICalParsingError("Malformed date", value=dt_string, ...)` yields: `ICalParsingError: Malformed date: 20240399`
- [ ] line AND line number should be properties of the error
- [ ] The line MUST be filled when we operate on passed data - optional only e.g. in vProperty types
- [ ] Make sure the line also works after a property value that goes over several lines 
- [ ] move all fuzzer catches over to only match ICalParsingError
- [ ] document this as `Raises:` in the docstring of all such implementations
- [ ] test all implementations of `from_ical` with all kinds of errors that are there
- catch/check/replace the current places where the errors are raised (TypeError, ValueError)
  - [ ] Create tests for the errors or modify existing tests
  - [ ] create a list of places that need working on
  - [ ] use the with statement style of JCalParsingError in order to add the line while parsing
  - [ ] test that line numbers are not set in certain cases (create a wrapper for that test to avoid duplication)
  - [ ] test the attronites of the error message (consistently with a function to avoid duplication and assure nothing is forgotten during development)
  - [ ] use test parametrization where possible `pytest.mark.parametrize`
- [ ] The error message should contain the line number only if given (pytest this unit)
- [ ] Unittest how the arguments create the error message
- [ ] clearly disclose AI usage and make sure before submission that we do not have to look through low quality code, unneeded edits and code duplication

## Notes

This is a big issue. Before you start, discuss your strategy of how to go about it to make it work. Invite others for collaboration.

## Suggestion

Some places can be started with already:

1. Create a list of all `raise` places that can occur during parsing.
2. sort: which ones raise ValueError - move them over first, e.g. in several PRs (The change does not break the old interface)
3. discuss how to go about this further.

[...]

PR #1536, opened 2026-07-10
Add ICalParsingError exception

GAUTAMSANSWAL:
This PR introduces ICalParsingError as a subclass of InvalidCalendar and exports it through the public API.

It also adds an initial test for the new exception.

This is intended as the minimal programmatic interface discussed in #1448 and provides a foundation for follow-up work.


angatha:
Thank you for the first PR for that huge issue. We have a [CONTRIBUTING.md](https://github.com/collective/icalendar/blob/main/CONTRIBUTING.md) which you should follow. We are currently working on our [PR template](https://github.com/collective/icalendar/blob/main/.github/pull_request_template.md) which contains a checklist supporting the contributing guide. Please use that for future PRs.

Please add a changelog entry as discribed in the contributing guide and disclose AI use, if applicable.

src/icalendar/tests/test_ical_parsing_error.py
```diff
@@ -0,0 +1,19 @@
...
+    assert issubclass(ICalParsingError, InvalidCalendar)
+
+
+def test_ical_parsing_error_stores_attributes():
+    error = ICalParsingError(
+        message="Invalid property",
+        line="SUMMARY:Meeting",
+        line_number=42,
+        value="SUMMARY:Meeting",
+    )
```
angatha, line 14: Can you please add a test that verifies that parameter order is as expected and one with only requried parameters as well?

GAUTAMSANSWAL: Thanks for the review! I've added a changelog entry following the contributing guide and added the requested tests for required parameters and positional argument order.

news/1536.feature
```diff
@@ -0,0 +1,5 @@
+Added the ``ICalParsingError`` exception as a public API for programmatic handling of iCalendar parsing failures.
+
+AI assistance: ChatGPT was used to help debug terminal errors encountered during development.
+
+@GAUTAMSANSWAL
```
stevepiercy: This is good, but all this should be on one line, and use reStructuredText to link to the new exception.

```suggestion
Added the :exc:`~icalendar.error.ICalParsingError` exception as a public API for programmatic handling of iCalendar parsing failures. AI assistance: ChatGPT was used to help debug terminal errors encountered during development. @GAUTAMSANSWAL
```

src/icalendar/error.py
```diff
@@ -16,6 +16,23 @@ class InvalidCalendar(ValueError):
...
 
 
+class ICalParsingError(InvalidCalendar):
+    """Could not parse an iCalendar."""
+
+    def __init__(
+        self,
+        message: str,
+        line: str | None = None,
+        line_number: int | None = None,
```
stevepiercy, line 26: Neither of these attributes are used anywhere else in this module. What is their purpose? I assume that the default Python stack trace returns this information anyway without them.

stevepiercy (changes requested):

src/icalendar/error.py
GAUTAMSANSWAL: Thanks for the question! The idea was to keep parser-specific context (the offending line and its number) with the exception so callers can use it for better error reporting. This PR only introduces the exception as the minimal API, so those fields aren't populated yet, they're intended for the follow-up work.

GAUTAMSANSWAL: If you want, I can remove them now and introduce them in the later PRs.

stevepiercy: Thanks for the explanation. Please keep them. I checked the original issue, and these lines are part of the spec.

stevepiercy (dismissed):
Thank you for your contribution on this huge issue.

stevepiercy:
@angatha would you please do one more review on the changes you requested? If it looks good, you can resolve the conversation so that you can merge the PR. Thank you!

src/icalendar/tests/test_ical_parsing_error.py
```diff
@@ -17,3 +17,26 @@ def test_ical_parsing_error_stores_attributes():
...
+    assert error.line_number is None
+    assert error.value is None
+
+
+def test_ical_parsing_error_parameter_order():
+    error = ICalParsingError(
+        "Invalid property",
+        "SUMMARY:Meeting",
+        42,
+        "SUMMARY:Meeting",
```
angatha: Two times the same parameter is not good for this test case. It can not verify that the parameters aren't mixed up. Just imamgine if the implementation mixed the second and fourth parameter. No test would caught this but this test is supposed to do so.

angatha: Same applies to the other test as well now that I#m thinking about it.

GAUTAMSANSWAL: Thanks! Good catch. I've updated both tests to use distinct values so they verify the positional argument order correctly and would fail if the parameters were mixed up.

angatha (approved):

sin-strokes (commented):
Commit

niccokunzmann:

PR #1542, opened 2026-07-12
Replace parser assertions with runtime validation

esbreeb:
## Linked issue

- Closes #1532

## Description

Replace four runtime parser assertions with explicit `ValueError` checks. Python removes assertions when running with `-O`, which previously allowed embedded newlines through `Contentline` and `_foldline`, and caused inconsistent failures for invalid input types.

The change also updates `_unescape_char`'s return annotation and docstrings to document the new validation behavior. Existing tests now assert the stable exception type and cover every replacement branch.

## Validation

- `python -m pytest src/icalendar/tests/test_icalendar.py -q`
- Full test suite under branch coverage: 15,612 tests collected, successful
- Overall coverage: 97%
- `content_line.py` coverage: 97%
- `string.py` coverage: 97%
- `python -m ruff format --check ...`
- `python -m ruff check ...`
- `python -O` manual regression covering all five invalid input cases

## Checklist

- [x] I added a change log entry, following the instructions in [Change log entry format](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-entry-format).
- [x] I followed icalendar's [Artificial intelligence policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) and disclosed my [Responsible AI use](https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use) in the commit message and change log entry.
- [x] I added or updated tests.
- [x] I ran and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [x] I updated the affected parser docstrings and return annotation.

## AI disclosure

OpenAI Codex (GPT-5, 2026-07-12) assisted with issue analysis, drafting the implementation and tests, and running validation. The use is disclosed in commit `9ada5da1` and `news/1532.bugfix` as required by the project policy.

PR #1543, opened 2026-07-12
fix: replace parser assert guards with explicit raises (#1532)

### LINKED ISSUE #1532: `assert` statements in the parser produce invalid output when Python runs with `-O`

```python
```

```
$ python -O -c "
print(cl.to_ical())
"
b'SUMMARY:Test\nINJECTED:Bad'
```

gitcommit90:
Fixes #1532

Replace `assert` statements in the parsing pipeline with explicit `raise` statements so validation guards still run under `python -O`.

## Changes
- `Contentline.__new__`: newline guard now raises `ValueError`
- `_escape_char`, `_unescape_char`: type guards now raise `TypeError`
- `_foldline`: type and newline guards now raise `TypeError`/`ValueError`
- Updated existing tests that expected `AssertionError`
- Added regression tests in `test_issue_1532_assert_to_valueerror.py`
- Added towncrier news fragment

## Testing
- Focused tests: 12 passed
- Full suite: 15586 passed, 30 skipped
- Verified `-O` now raises `ValueError` for newline injection
- ruff clean on all touched files

### DIFF
--- a/news/1532.bugfix
+++ b/news/1532.bugfix
@@ -1,0 +1,1 @@
+Replace parser ``assert`` guards with explicit raises so newline (``ValueError``) and type (``TypeError``) checks still run under ``python -O``. @gitcommit90
--- a/src/icalendar/parser/content_line.py
+++ b/src/icalendar/parser/content_line.py
@@ -123,9 +123,10 @@ class Contentline(str):
 
     def __new__(cls, value, strict=False, encoding=DEFAULT_ENCODING):
         value = to_unicode(value, encoding=encoding)
-        assert "\n" not in value, (
-            "Content line can not contain unescaped new line characters."
-        )
+        if "\n" in value:
+            raise ValueError(
+                "Content line can not contain unescaped new line characters."
+            )
         self = super().__new__(cls, value)
         self.strict = strict
         return self
--- a/src/icalendar/parser/string.py
+++ b/src/icalendar/parser/string.py
@@ -37,7 +37,8 @@ def _escape_char(text: str | bytes) -> str:
         not part of :rfc:`5545`, which only defines ``\n`` or ``\N`` for an
         intentional line break, and doesn't give an escape form for a lone ``\r``.
     """
-    assert isinstance(text, (str, bytes))
+    if not isinstance(text, (str, bytes)):
+        raise TypeError("text must be str or bytes")
     text = to_unicode(text)
     # NOTE: ORDER MATTERS!
     return (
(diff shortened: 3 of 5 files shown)


niccokunzmann:
Hi, please coordinate with PR #1542 and you both review the changes and agree on the best version, then ping us!

esbreeb (commented):
Thanks for coordinating. I compared #1542 and this PR in detail.

I think #1543 should be the base for the final version:

- `TypeError` for invalid input types is more idiomatic than `ValueError` and satisfies Ruff's `TRY004` rule without suppressions.
- The dedicated five-case regression test module is easier to review and keeps the issue-specific coverage together.
- `ValueError` remains appropriate for embedded newlines.

I suggest incorporating two contract/documentation fixes from #1542 before we close the duplicate:

1. Change `_unescape_char`'s return annotation from `str | bytes | None` to `str | bytes`, and remove the now-unreachable `return None` path. Invalid types raise before that path, so the current annotation and docstring are stale.
2. Add `Raises` documentation to `_escape_char`, `_unescape_char`, and `_foldline`, since these parser docstrings are rendered in the API documentation.

Although issue #1532 says `ValueError` for all four assertions, I agree that the repository's Ruff rule and normal Python exception semantics are stronger reasons to use `TypeError` for the type guards.

If this plan works for you, please add those two pieces here. Once they land and CI passes, I will close #1542 and ping @niccokunzmann that we agreed on this version.

stevepiercy:
> Add `Raises` documentation to `_escape_char`, `_unescape_char`, and `_foldline`, since these parser docstrings are rendered in the API documentation.

@esbreeb where? This smells like AI slop. Private functions never appear in the API Reference documentation. See:

- https://icalendar.readthedocs.io/en/stable/contribute/index.html#responsible-ai-use
- https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.parser.string.html
- https://icalendar.readthedocs.io/en/latest/contribute/development.html#internal-use-only

I'm inclined to block AI slop users from the organization.

angatha:
@stevepiercy maybe you are right but I agree with the suggestions. So please keep the comment.

SashankBhamidi:
@stevepiercy @angatha,

>If this plan works for you, please add those two pieces here. Once they land and CI passes, I will close https://github.com/collective/icalendar/pull/1542

Authors of both the PRs are different, I believe @esbreeb is asking us to add the commits "two pieces" from their PR into this PR? _I haven't reviewed either PRs._

angatha:
Yes this is how I understand it.

niccokunzmann:
I asked the two who created both PRs almost at the same time to collaborate and ping us when they are done evaluating what works best. So, for now, we can wait until they ask us :)

esbreeb:
It looks like he may have missed these

niccokunzmann:
@esbreeb Since this PR creator is not responsive, how would you like to proceed? What do you like to take from the ideas of this PR?

I close this in favor of the other PR.

PR #1545, opened 2026-07-13
Improve ICalParsingError message formatting

GAUTAMSANSWAL:
## Linked issue

- See #1448

## Description

This PR improves `ICalParsingError` by including the optional `value`, `line`, and `line_number` in the exception message when available.

It also adds unit tests covering the different message formatting cases.

## Checklist

- I added a change log entry, following the instructions in [Change log entry format]
- I followed icalendar's [Artificial intelligence policy]
- I added tests, where applicable.
- I ran and ensured all tests pass locally

## Additional information

None.


GAUTAMSANSWAL:
Since the roadmap specifies what information should be included in the exception message but doesn't define the exact format, I made a consistent formatting for the message in this PR. If you have a preferred format, I'm happy to update it.

stevepiercy:
@GAUTAMSANSWAL FYI, the PR Quality job checks for the presence of ticked checkboxes, such as `- [x]`. If you remove them or alter their text, the check will fail. See https://github.com/collective/icalendar/actions/runs/29249051340?pr=1545. It will also fail if you alter or remove headings, except for "Additional information" which is optional. You don't need to change your PR description here, as I already reviewed what it was complaining about and it's OK, but please keep this in mind for future contributions. It will save us the trouble of investigating. Thank you!

stevepiercy (dismissed):
Excellent work! Let's get one more review, then merge. Thank you, @GAUTAMSANSWAL.

GAUTAMSANSWAL:
> @GAUTAMSANSWAL FYI, the PR Quality job checks for the presence of ticked checkboxes, such as `- [x]`. If you remove them or alter their text, the check will fail. See https://github.com/collective/icalendar/actions/runs/29249051340?pr=1545. It will also fail if you alter or remove headings, except for "Additional information" which is optional. You don't need to change your PR description here, as I already reviewed what it was complaining about and it's OK, but please keep this in mind for future contributions. It will save us the trouble of investigating. Thank you!

I'll keep this in mind, x seemed like a cross so I got bit confused

src/icalendar/error.py
```diff
@@ -30,7 +32,19 @@ def __init__(
...
+        full_message = message
+
+        if value is not None:
+            full_message += f": {value}"
+        if line_number is not None and line is not None:
+            full_message += f" (line {line_number}: {line})"
+        elif line_number is not None:
+            full_message += f" (line {line_number})"
+        elif line is not None:
+            full_message += f" ({line})"
```
angatha: I would prefere the representation of the string instead of the raw value. That makes non printable characters visible.
```suggestion
            full_message += f": {value!r}"
        if line_number is not None and line is not None:
            full_message += f" (line {line_number}: {line!r})"
        elif line_number is not None:
            full_message += f" (line {line_number})"
        elif line is not None:
            full_message += f" ({line!r})"
```

angatha (changes requested):

src/icalendar/error.py
GAUTAMSANSWAL: Applied, thanks! I updated the error message to use the representation of value and line so that non-printable characters are visible.

angatha (approved):

PR #1547, opened 2026-07-13
Update Mountain Standard Time (Mexico) mapping

RobHannay:
## Linked issue

- Closes #1546

## Description

The bundled CLDR 29 table maps `Mountain Standard Time (Mexico)` to `America/Chihuahua`, but current CLDR maps both the global and Mexico entries to `America/Mazatlan`. Those zones diverged after Mexico's 2022 timezone changes, so dates such as 6 July 2026 currently resolve at UTC-6 instead of UTC-7.

This updates the mapping to `America/Mazatlan` and adds a focused behavior regression for the July 2026 offset. It deliberately does not change `TZP.timezone()` lookup precedence: the latest revision of #1479 already checks the embedded timezone cache before provider fallbacks and would overlap with that work.

## Checklist

- [x] I added a change log entry, following the instructions in [Change log entry format](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-entry-format).
- [x] I followed icalendar's [Artificial intelligence policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) and disclosed my [Responsible AI use](https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use) in my commit message and news fragment.
- [x] I checked the applicable AI terms of use, reviewed and understand the change, take responsibility for it, and confirmed the contribution is original rather than reconstructed from copyrighted material.
- [x] I added a regression test covering both configured timezone providers.
- [x] I ran and ensured all tests pass locally.
- [x] I reviewed the documentation impact; no API or narrative documentation changes are necessary for this data correction.

## Verification

- `uv run pytest -q`: 15,586 passed, 30 skipped
- `uv run pytest src/icalendar/tests/prop/test_windows_to_olson_mapping.py -q`: 216 passed
- `uv run ruff check ...`: passed
- `uv run ruff format --check ...`: passed
- `uv run towncrier build --draft`: rendered `news/1546.bugfix` successfully

---
Slack thread: https://metaviewapp.slack.com/archives/C0A3T96UMT2/p1783959558578489
Requested by: Rob

Rook session: https://rook.is/cooking/dfe003fd-a945-4099-acc2-c923b94a8b95

PR #1554, opened 2026-07-17
Created a script to automatically generate the windows_to_olson module

### LINKED ISSUE #1548: Audit latest CLDR timezone mapping of Windows to Olson
We should create a tool to automatically audit and build the data in the timezone mapping of Windows to Olson in https://github.com/collective/icalendar/blob/main/src/icalendar/timezone/windows_to_olson.py against the latest CLDR data.

Additionally, the mapping uses Region to distinguish specific time zone mappings. Should we consider Region as a new parameter? If so, how would that work?

See:

- https://www.unicode.org/cldr/cldr-aux/charts/29/supplemental/zone_tzid.html
- https://github.com/unicode-org/cldr/blob/main/common/supplemental/windowsZones.xml
- https://github.com/collective/icalendar/issues/1546

h/t @RobHannay

Note that work on this issue should wait until the questions have been discussed and resolved.

stevepiercy:
## Linked issue

Closes #1548.

## Description

- Created a script to automatically generate the :mod:`~icalendar.timezone.windows_to_olson` module from the Unicode Consortium's CLDR GitHub repository.
- Added lxml as a new development dependency.
- Added the target ``wo`` to the ``Makefile`` to generate the file from the root of the repository with `make wo`.

## Checklist

Please let me know if there is anything further I should do. I'm not sure what I would test.

We don't need to run `make wo` more than twice per year. I wouldn't put it in CI. I could add some documentation, perhaps in https://icalendar.readthedocs.io/en/stable/contribute/maintenance.html#release-versions as an item to run with each release. What do y'all think?


SashankBhamidi:
> We don't need to run make wo more than twice per year. I wouldn't put it in CI. I could add some documentation, perhaps in https://icalendar.readthedocs.io/en/stable/contribute/maintenance.html#release-versions as an item to run with each release. What do y'all think?

Release docs sounds perfect! Maybe do it in this PR itself? I'll review altogether.

stevepiercy:
> Release docs sounds perfect! Maybe do it in this PR itself? I'll review altogether.

Done in https://github.com/collective/icalendar/pull/1554/commits/86846f8264e2464131a37ae62250950085fd51b2

SashankBhamidi:
Let me know when rfr

stevepiercy:
Sorry for all the noise. All green now. Ready for review.

angatha (dismissed):

src/icalendar/timezone/windows_to_olson.py
```diff
@@ -1,87 +1,109 @@
-"""This module contains mappings from Windows timezone identifiers to
+"""
+This module contains mappings from Windows timezone identifiers to
 Olson timezone identifiers.
 
-The data is taken from the unicode consortium [0], the proposal and rationale
-for this mapping is also available at the unicode consortium [1].
+This file is automatically generated by generate-windows-olson-mapping.py.
```
niccokunzmann:
```suggestion
This file is automatically generated by generate_windows_olson_mapping.py.
```

generate_windows_to_olson_mapping.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Parse the XML file and extract the timezone mapping.
+result = {}
+tree = et.fromstring(xml_content)
+for zone in tree.xpath("//mapZone"):
+    attrib = zone.attrib
+    if attrib["territory"] == "001":
+        result[attrib["other"]] = attrib["type"]
+
+path = "src/icalendar/timezone"
```
niccokunzmann: I would recommend using a path relative to `__file__`. This way, the script works independent of the current working directory.

niccokunzmann (changes requested):
Some hints here.

Open questions:
- Does this not break the fix from the other issue that was created?
- Is the mapping stable or is there a problem of its time of when it is applied? E.g. if the mapping changes in the future, does this mean that retrospectively, the timezone also has changed?

stevepiercy:
>     * Does this not break the fix from the other issue that was created?

It does not break it.

- https://github.com/collective/icalendar/pull/1547/changes#diff-2ce3dc83b0a8890adff14b143c162b955846aed09c734977dbdbaf85e32a2cd5
- https://github.com/collective/icalendar/pull/1554/changes#diff-51b8d8422dc6abb8c57bc3584adce6e9e1b20e9c37496ef96b39fa8e0d1df191R89

>     * Is the mapping stable or is there a problem of its time of when it is applied? E.g. if the mapping changes in the future, does this mean that retrospectively, the timezone also has changed?

No. The [CLDR explicitly states](https://cldr.unicode.org/index/downloads):

> Each release of the Unicode CLDR is a stable release and may be used as reference material or cited as a normative reference by other specifications. _Each version, once published, is absolutely stable and will never change._ [emphasis added]

generate_windows_to_olson_mapping.py
stevepiercy: I don't understand. Can you elaborate?

The script is called only from the root of the repo via `make wo`.

niccokunzmann: Oh. That is a principle of mine, especially when they write files, to make sure they are always writing them into the right location.

The intention is to write to a file relative to the location of this file, not in the location relative to the invoker. The invoker is implicitly limited to make. However, the intention is not reflected in the code.

I do not see this as a blocker. It is my recommendation to use 

```
HERE = Path(__file__).parent
```

And make it relative when that is the intention.

src/icalendar/timezone/windows_to_olson.py
```diff
@@ -1,87 +1,109 @@
...
     "Bangladesh Standard Time": "Asia/Dhaka",
     "Belarus Standard Time": "Europe/Minsk",
+    "Bougainville Standard Time": "Pacific/Bougainville",
     "Canada Central Standard Time": "America/Regina",
     "Cape Verde Standard Time": "Atlantic/Cape_Verde",
     "Caucasus Standard Time": "Asia/Yerevan",
     "Cen. Australia Standard Time": "Australia/Adelaide",
     "Central America Standard Time": "America/Guatemala",
-    "Central Asia Standard Time": "Asia/Almaty",
+    "Central Asia Standard Time": "Asia/Bishkek",
```
niccokunzmann, line 43: I am asking about the changes like this one. 

E.g. 15 years ago, one was valid, now the other one is.
Does that mean that dates 15 years ago should use the old definition? Is this a simple mapping or is there a time validity to it?

I will have a look.

src/icalendar/timezone/windows_to_olson.py
```diff
@@ -1,87 +1,109 @@
...
     "E. Africa Standard Time": "Africa/Nairobi",
     "E. Australia Standard Time": "Australia/Brisbane",
     "E. Europe Standard Time": "Europe/Chisinau",
     "E. South America Standard Time": "America/Sao_Paulo",
+    "Easter Island Standard Time": "Pacific/Easter",
     "Eastern Standard Time": "America/New_York",
     "Eastern Standard Time (Mexico)": "America/Cancun",
     "Egypt Standard Time": "Africa/Cairo",
     "Ekaterinburg Standard Time": "Asia/Yekaterinburg",
-    "FLE Standard Time": "Europe/Kyiv",
```
niccokunzmann, line 49: With these changes, we should also add a test that makes sure that the right side of the mapping can actually be resolved by zoneinfo.

I would make this a closing requirement for the issue, not a merge requirement.

generate_windows_to_olson_mapping.py
```diff
@@ -0,0 +1,57 @@
...
+url = "https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml"
+with req.urlopen(url) as response:  # noqa: S310
+    xml_content = response.read()
+
+# Parse the XML file and extract the timezone mapping.
+result = {}
+tree = et.fromstring(xml_content)
+for zone in tree.xpath("//mapZone"):
+    attrib = zone.attrib
+    if attrib["territory"] == "001":
```
niccokunzmann, line 29: I am curious about the decision to check for  territory here. This again is not a blocker for merge but good to understand. It changes and reduces the content of the mapping.

SashankBhamidi:
Hi, please don't merge, I want to review this :)

src/icalendar/timezone/windows_to_olson.py
stevepiercy: I don't know.

However, why does it matter for icalendar? Calendaring apps are concerned with the future, not history. I can't imagine having [49+ CLDR Windows to Olson mapping versions](https://cldr.unicode.org/index/downloads) to have any value to icalendar users.

generate_windows_to_olson_mapping.py
stevepiercy: tl;dr: `001` is the default.

See https://cldr.unicode.org/development/development-process/design-proposals/extended-windows-olson-zid-mapping#default-mapping.

icalendar currently doesn't support a territory or region.

See also the source file that gets pulled from GitHub at https://github.com/unicode-org/cldr/blob/main/common/supplemental/windowsZones.xml.

src/icalendar/timezone/windows_to_olson.py
stevepiercy: That's a new separate issue. See https://github.com/collective/icalendar/issues/1562.

generate_windows_to_olson_mapping.py
stevepiercy: Done in c43596a1d6174546ba9ef41a69072206be4390ba

generate_windows_to_olson_mapping.py
niccokunzmann: Thanks!

src/icalendar/timezone/windows_to_olson.py
niccokunzmann: Good. That is alright. I cannot find anything like that on their website either. I just wondered because updates happen to tzdata that change a timezone offset even in the past.

niccokunzmann (dismissed):

stevepiercy:
@SashankBhamidi last call for review! I'll merge by EOD Monday local time.

generate_windows_to_olson_mapping.py
```diff
@@ -0,0 +1,59 @@
...
+DEST = HERE / "src" / "icalendar" / "timezone"
+file = "windows_to_olson.py"
+filepath = Path(DEST, file)
+print(f"Writing {filepath}")  # noqa: T201
+with Path.open(filepath, "w") as f:
+    f.write("""\"\"\"
+This module contains mappings from Windows timezone identifiers to
+Olson timezone identifiers.
+
+This file is automatically generated by generate_windows_olson_mapping.py.
```
SashankBhamidi: The filename in the generated docstring is `generate_windows_olson_mapping.py`, missing `_to_`. 
Every `make wo` run bakes that wrong name into `windows_to_olson.py` line 5.

```suggestion
This file is automatically generated by generate_windows_to_olson_mapping.py.
```

generate_windows_to_olson_mapping.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Source - https://stackoverflow.com/a/10469748
+# Posted by mzjn, modified by community. See post 'Timeline' for change history
+# Retrieved 2026-07-16, License - CC BY-SA 3.0
+
+import json
+import urllib.request as req
+from pathlib import Path
+
+import lxml.etree as et
```
SashankBhamidi: `lxml` is only doing `fromstring()` and `xpath("//mapZone")` here. stdlib `xml.etree.ElementTree` handles both with `.findall('.//mapZone')` and produces identical output. 

Swapping it out removes the new dev dependency entirely. 

The `>=6.1.1` pin clears the XXE CVE from earlier this year so it is safe either way, just something to consider.

generate_windows_to_olson_mapping.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Get the XML data from the CLDR on GitHub and save to a file.
+# Unfortunately, the CLDR does not have a latest or stable tag
+# that points to the stable version.
+# Use the `main` branch, which is the development branch.
+url = "https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml"
```
SashankBhamidi, line 20: The `$Revision$` in the CLDR XML is an unexpanded placeholder, so the generated file has no record of which commit its data came from. Two runs a month apart against different `main` commits produce different output with nothing in the file to show for it. 

Pulling the latest commit SHA for `windowsZones.xml` from the GitHub API and writing it into the docstring would give us something to point at when someone asks which version of CLDR this reflects.

What do you think?

SashankBhamidi (changes requested):
Really glad this exists :)

Worth addressing Nicco's stability question properly. Eight of the nine changed values are just Olson renames, same data different label, no behaviour change. The one that looks alarming is `Central Asia Standard Time` going from `Asia/Almaty` to `Asia/Bishkek`, because those are genuinely different UTC offsets. Kazakhstan moved to UTC+5 permanently in 2018, CLDR reassigned the "Central Asia" Windows key to Kyrgyzstan, and Almaty ended up under a different key. Our old value was the stale one.

I think this PR is quietly fixing a real bug!

Two non-blocking thoughts before merge, lmk what you think:

generate_windows_to_olson_mapping.py
stevepiercy: Maybe in a new issue for improvement? I don't have enough information to give a qualified answer. If the endpoint requires authentication, then that could be a problem.

Keep in mind that this script is only run by release managers. I suppose it's possible that someone who monitors the latest release of the CLDR and is on Windows and wants to update the Windows to Olson mapping in a PR could also run it.

Given that no one has complained for 10 years about using CLDR v29, it's an extremely low priority for me.

generate_windows_to_olson_mapping.py
stevepiercy: Nice catch!

generate_windows_to_olson_mapping.py
stevepiercy: Swap out what? The suggestion is not complete. I tried but couldn't figure out your intent.

Also, FWIW, I originally tried with `xml` only, but couldn't figure it out, as it had issues with passing content from a URL as bytes. That's the advantage of lxml over xml, as far as I can tell.

If you have a complete replacement, then please make the suggestion, or push a commit that works.

SashankBhamidi: `ET.fromstring()` actually takes bytes just fine, so the fetch stays exactly as is. Only two lines change:

```
import xml.etree.ElementTree as ET
```

and then swap `et.fromstring` for `ET.fromstring` and `tree.xpath("//mapZone")` for `tree.findall(".//mapZone")`. 

I tested it locally, 139 entries, output is identical.

generate_windows_to_olson_mapping.py
SashankBhamidi: New issue works.

generate_windows_to_olson_mapping.py
SashankBhamidi:
```
import urllib.request as req
import xml.etree.ElementTree as ET

url = "https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml"
with req.urlopen(url) as response:
    xml_content = response.read()

tree = ET.fromstring(xml_content)
for zone in tree.findall(".//mapZone"):
    if zone.attrib["territory"] == "001":
        print(zone.attrib["other"], "->", zone.attrib["type"])
```

SashankBhamidi (dismissed):
Perfect! Thank you!

generate_windows_to_olson_mapping.py
stevepiercy: Argh, I missed the leading `.`. Commit incoming.

stevepiercy: Done in https://github.com/collective/icalendar/pull/1554/commits/d84269caa44ec92fb7547b0b8824d01acde15502.

stevepiercy: Also see commit message in https://github.com/collective/icalendar/pull/1554/commits/50880d399e3ffe31b89f331f45a470f5f591fd2f.

stevepiercy:
@SashankBhamidi one final pass, for good measure, if you please? Thank you!

SashankBhamidi (approved):
Looks good, thank you! 

- [x] Todo file issue for https://github.com/collective/icalendar/pull/1554#discussion_r3612371896

stevepiercy:
Issue created from comment: https://github.com/collective/icalendar/issues/1575

PR #1559, opened 2026-07-18
fix: replace assert statements in parser with explicit raises (#1532)

### LINKED ISSUE #1532: `assert` statements in the parser produce invalid output when Python runs with `-O`

gagana2023:
Closes #1532

  `assert` is stripped by Python's `-O` optimization flag, which let 5
  runtime-validation checks in the parser silently no-op under `-O`:

  - `Contentline.__new__` — allowed embedded newlines through, producing
    `to_ical()` output that violates RFC 5545 §3.1's content line structure.
  - `_foldline` — allowed embedded newlines and non-`str` input through,
    bypassing the 75-octet line folding requirement.
  - `_escape_char` / `_unescape_char` — allowed non-`str`/`bytes` input through.

  This replaces all five with explicit `if`/`raise` checks (`ValueError` for
  invalid newline content, `TypeError` for invalid argument types) that hold
  regardless of interpreter optimization level. Verified manually with
  `python -O` both before (bug reproduces) and after (guards hold) the fix.

  Also fixes `_unescape_char`'s return type annotation (`str | bytes | None`
  → `str | bytes`) and drops its now-dead `None` return path, addressing the
  review feedback that was outstanding on the related #1543.

  Two legacy tests in `test_icalendar.py` that asserted the old
  `AssertionError` behavior are updated to expect the new exception types.
  New parametrized tests cover every raise path in `test_parsing.py`.

  Full test suite run against this branch shows no new failures or errors
  versus unmodified `main`.

  AI disclosure: I used Anthropic Claude (Sonnet 5, via Claude Code) to
  research the issue and prior competing PRs, draft this change and its
  tests, and verify the fix under `python -O`. I reviewed and validated the
  output locally.

  ## Test plan
  - [x] `pytest src/icalendar/tests/test_parsing.py src/icalendar/tests/test_icalendar.py` — all pass
  - [x] Full suite (`pytest src/icalendar`) — failure/error count identical to unmodified `main` baseline
  - [x] `ruff check` on all changed files — clean
  - [x] Manual `python -O` reproduction: confirmed bug exists on `main`, confirmed fixed on this branch


niccokunzmann (dismissed):
Thanks!

Are you confident that you found all assert statements?

gagana2023:
Good catch to double-check — I had scoped this to the 4 locations the
  issue cited, but you're right to ask. I just grepped every non-test
  `.py` file in `src/icalendar` for `assert ` and found one more:

  `Contentline.from_parts` (content_line.py) was type-checking `params`
  with `assert isinstance(params, Parameters)`, same failure mode as the
  others — silently stripped under `-O`, which would let a bad `params`
  argument fall through to an unguarded `AttributeError` in
  `params.to_ical()` instead of a clear error at the boundary.

  Pushed a follow-up commit replacing it with an explicit `raise
  TypeError` plus a test. I also confirmed it holds under `python -O`.

  The only `assert` left anywhere in non-test source now is
  `cal/lazy.py:297`, which is inside an `if __name__ == "__main__":`
  benchmark script — never reachable from library code, so I left that
  one alone as out of scope. Full test suite still shows zero new
  failures vs. `main`.

niccokunzmann:
This looks like an AI answer.

@gagana2023 Are you there? How are you doing?

angatha:
@stevepiercy should the documentation now include a `Raises` part?

niccokunzmann:
Yes, the docs should reflect the errors.

angatha (changes requested):
Can you please update the relevant documentation to include a `Raises` section.

src/icalendar/parser/string.py
stevepiercy, line 25:
```suggestion
    Raises:
        TypeError: If ``text`` is neither ``str`` nor ``bytes``.

    Note:
```

src/icalendar/parser/string.py
```diff
@@ -71,7 +72,7 @@ def _unescape_char(text: str | bytes) -> str | bytes | None:
         text: The escaped text.
 
     Returns:
-        The unescaped text, or ``None`` if ``text`` is neither ``str`` nor ``bytes``.
+        The unescaped text.
 
     Note:
```
stevepiercy, line 83:
```suggestion
    Raises:
        TypeError: If ``text`` is neither ``str`` nor ``bytes``.

    Note:
```

src/icalendar/parser/string.py
```diff
@@ -126,8 +126,10 @@ def _foldline(line: str, limit: int = 75, fold_sep: str = "\r\n ") -> str:
     immediately followed by a single linear white-space character (i.e.,
     SPACE or HTAB).
     """
```
stevepiercy, line 139:
```suggestion

    Raises:
        TypeError: If ``line`` is not a ``str``.
        ValueError: If ``line`` contains ``\n``.

    """
```

src/icalendar/parser/string.py
stevepiercy:
```suggestion
    r"""Make a string folded as defined in RFC5545.
```

src/icalendar/parser/content_line.py
stevepiercy:
```suggestion
r"""A content line is basically a string that can be folded and parsed into
   parts.

    Raises:
        ValueError: If the value contains ``\n``.

```

src/icalendar/parser/content_line.py
```diff
@@ -139,7 +140,8 @@ def from_parts(
         sorted: bool = True,  # noqa: A002
     ):
         """Turn a parts into a content line."""
```
stevepiercy:
```suggestion
        r"""Turn a parts into a content line.
        
        Raises:
            TypeError: If ``params`` is not type ``Parameters``.

        """
```

stevepiercy (changes requested):
Double check that all these proposed changes render properly in the pull request preview.

- https://icalendar--1559.org.readthedocs.build/en/1559/reference/api/icalendar.parser.content_line.html#icalendar.parser.content_line.Contentline
- https://icalendar--1559.org.readthedocs.build/en/1559/reference/api/icalendar.parser.content_line.html#icalendar.parser.content_line.Contentline.parts
- https://icalendar--1559.org.readthedocs.build/en/1559/reference/api/icalendar.parser.string.html#icalendar.parser.string.unescape_char

The remaining docstrings won't render because they're in private methods, but that's OK, as long as the docs build without errors.

stevepiercy:
This PR and #1588 may conflict. There have been 5 PRs opened for this issue, 3 closed, and 2 remain open. We should always check which PR is best before merging.

niccokunzmann:
@gagana2023 Thanks for the PR. This is the last remaining PR to solve this issue. Could you have a look at the suggestions and apply them?

If we do not get a reply within a week, we either apply them ourselves and merge or close the PR.
Thanks for the work so far!

angatha (dismissed):

angatha:
@stevepiercy can this be merged?

src/icalendar/parser/string.py
```diff
@@ -117,17 +123,24 @@ def _unescape_char(text: str | bytes) -> str | bytes | None:
 
 
 def _foldline(line: str, limit: int = 75, fold_sep: str = "\r\n ") -> str:
-    """Make a string folded as defined in RFC5545.
+    r"""Make a string folded as defined in RFC5545.
```
stevepiercy:
```suggestion
    r"""Make a string folded as defined in :rfc:`5545#section-3.1`.
```

src/icalendar/parser/content_line.py
```diff
@@ -115,17 +115,22 @@ def flush_pending() -> None:
 
 
 class Contentline(str):
-    """A content line is basically a string that can be folded and parsed into
+    r"""A content line is basically a string that can be folded and parsed into
     parts.
```
stevepiercy:
```suggestion
    r"""A content line is a string that can be folded and parsed into parts.
```

src/icalendar/parser/string.py
```diff
@@ -117,17 +123,24 @@ def _unescape_char(text: str | bytes) -> str | bytes | None:
...
+        TypeError: If ``line`` is not a ``str``.
+        ValueError: If ``line`` contains ``\n``.
+
     """
-    assert isinstance(line, str)
-    assert "\n" not in line
+    if not isinstance(line, str):
+        raise TypeError(f"Expected str, got {type(line).__name__}")
+    if "\n" in line:
+        raise ValueError("line must not contain unescaped new line characters")
```
stevepiercy:
```suggestion
        raise ValueError("Line must not contain unescaped new line characters.")
```

stevepiercy (changes requested):
I'll merge these three nitpicks.

stevepiercy (approved):
Thank you! Will auto-merge.

Issue #1550, opened 2026-07-15
Add a property for the Base64 string view in vBinary


niccokunzmann:
vBinary is stored as base64 encoded data.
It would be very handy to have `vBinary.base64data` as a property with getter and setter.

- [x] Wait for #1356 to be merged.
- [x] add a getter based on the bytes attribute from #1356.
- [x] add a setter which takes str only and validates the input, raises a ValueError if not base64
- [x] nice docstrings, following our [style guide](https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure)
- [x] test all edge cases
- [x] replace all manual base64 encode/decode behaviour in our code base where possible.
 such as https://github.com/collective/icalendar/pull/1356#discussion_r3421606404_

PR #1560, opened 2026-07-18
Add vBinary.base64data property

gagana2023:
## Summary

- Adds `vBinary.base64data`, a getter/setter for the Base64 string view of a `vBinary` value, so callers don't have to call `base64.b64encode`/`base64.b64decode` by hand.
- The getter mirrors `to_ical()` (as `str` instead of `bytes`); the setter reuses the existing `from_ical` validation, so it raises the same `ValueError("Not valid base 64 encoding.")` on invalid input.
- Adds tests covering the getter, setter, round-trip, empty value, invalid Base64 input, and non-`str` input.
- Adds a `news/1550.feature` fragment.

Addresses #1550.

## Note on scope / #1356

Issue #1550's checklist asks to build `base64data` on top of the `bytes` attribute from #1356, and to wait for #1356 to merge first. Since #1356 is still open, this PR implements `base64data` independently against `vBinary`'s current `obj` (`str`) storage on `main`, so the feature isn't blocked.

One consequence: the checklist's last item ("replace all manual base64 encode/decode behaviour in our code base where possible", e.g. `image.py`) is **not** done here. I checked `src/icalendar/prop/image.py:59` (`params["b64data"] = value.obj`) and confirmed it currently only works because the existing test happens to pass an already-Base64-encoded string in as `obj` — `vBinary.obj` is ambiguous on `main` (sometimes raw text, sometimes pre-encoded Base64), which is exactly what #1356's `.bytes` attribute fixes. Swapping `value.obj` for `value.base64data` now would double-encode in that case and break the existing test.

**Follow-up once #1356 merges:** rebase `base64data` to read/write `self.bytes` instead of `self.obj`, then replace `value.obj` with `value.base64data` in `image.py` (the discussion this issue references: https://github.com/collective/icalendar/pull/1356#discussion_r3421606404).

## Test plan

- [x] `pytest src/icalendar/tests/prop/test_vBinary.py -v` — 22 passed
- [x] `pytest` full suite — 1556 passed, no regressions (including `test_image.py`, jCal tests)
- [x] `ruff check` / `ruff format --check` — clean
- [x] `mypy src/icalendar/prop/binary.py` — no new errors (one pre-existing, unrelated error in `__repr__` on `main`, confirmed by running mypy against the unmodified file)
- [x] Manual REPL check of getter/setter/error paths

### DIFF
--- a/news/1550.feature
+++ b/news/1550.feature
@@ -1,1 +1,1 @@
-Added :attr:`icalendar.prop.binary.vBinary.base64data`, a getter/setter for the Base64 string view of a BINARY property's value, so callers no longer need to call :func:`base64.b64encode`/:func:`base64.b64decode` manually. @gagana2023
+Added :attr:`vBinary.base64data <icalendar.prop.binary.vBinary.base64data>`, a getter/setter for the Base64 string view of a BINARY property's value, so callers no longer need to call :func:`base64.b64encode` or :func:`base64.b64decode` manually. @gagana2023
--- a/news/1550.feature
+++ b/news/1550.feature
@@ -1,0 +1,1 @@
+Added :attr:`icalendar.prop.binary.vBinary.base64data`, a getter/setter for the Base64 string view of a BINARY property's value, so callers no longer need to call :func:`base64.b64encode`/:func:`base64.b64decode` manually. @gagana2023
--- a/src/icalendar/prop/binary.py
+++ b/src/icalendar/prop/binary.py
@@ -36,6 +36,31 @@ class vBinary:
         except (binascii.Error, ValueError) as e:
             raise ValueError("Not valid base 64 encoding.") from e
 
+    @property
+    def base64data(self) -> str:
+        """The Base64-encoded string view of this value.
+
+        This is the same string that :meth:`to_ical` produces, exposed as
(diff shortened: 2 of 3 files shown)


news/1550.feature
```diff
@@ -0,0 +1 @@
+Added :attr:`icalendar.prop.binary.vBinary.base64data`, a getter/setter for the Base64 string view of a BINARY property's value, so callers no longer need to call :func:`base64.b64encode`/:func:`base64.b64decode` manually. @gagana2023
```
stevepiercy: Let's display only the last two segments of the path when rendered to HTML.

```suggestion
Added :attr:`vBinary.base64data <icalendar.prop.binary.vBinary.base64data>`, a getter/setter for the Base64 string view of a BINARY property's value, so callers no longer need to call :func:`base64.b64encode` or :func:`base64.b64decode` manually. @gagana2023
```

stevepiercy (changes requested):
Just a minor tweak to the news fragment. Thank you!

stevepiercy (approved):
Thank you! Let's get a second technical review.

angatha (approved):

PR #1566, opened 2026-07-19
Test Windows mapping targets via zoneinfo tzdata fallback

### LINKED ISSUE #1562: Create test to resolve timezones in Windows to Olson mapping
Add a test that ensures the right side of the Windows to Olson mapping can actually be resolved by zoneinfo.

- [ ] The test should check all the values (right side) in the file [`src/icalendar/timezone/windows_to_olson.py`](https://github.com/collective/icalendar/blob/main/src/icalendar/timezone/windows_to_olson.py) actually resolve using zoneinfo.

See https://github.com/collective/icalendar/pull/1554#discussion_r3608376188

floze-the-genius:
## Linked issue

- Closes #1562

## Description

The existing mapping test resolved Olson identifiers through the project timezone-provider abstraction, which could use either `pytz` or `zoneinfo`. This change checks every value in `WINDOWS_TO_OLSON` directly with `zoneinfo.ZoneInfo`.

The test temporarily empties `zoneinfo.TZPATH` and clears the cache so resolution uses the declared `tzdata` dependency, matching the fallback path used on supported Windows installations. The original path and cache state are restored after every case.

## Checklist

- [x] I added `news/1562.internal` following the change log format.
- [x] I followed the artificial intelligence policy and disclosed OpenAI Codex with GPT-5 usage in the commit and news fragment.
- [x] I added direct coverage for all mapped Olson identifiers.
- [x] I ran the full test suite locally.
- [x] I reviewed documentation impact; no public API or narrative documentation changes are needed.

## Verification

- `uv run pytest -q`: 15,775 passed, 28 skipped
- Python 3.10 focused test: 110 passed
- Python 3.13 focused test: 110 passed
- Python 3.14 focused test: 110 passed
- `uv run ruff check .`: passed
- `uv run ruff format --check .`: 247 files already formatted
- `uv run towncrier build --draft`: rendered `news/1562.internal` successfully

## Additional information

I reviewed the generated change against the project data and test conventions. The contribution was created specifically for this issue and was not reconstructed from copyrighted source material.


news/1562.internal
```diff
@@ -0,0 +1 @@
+Ensured every Olson timezone identifier in the Windows mapping resolved through Python's :class:`zoneinfo.ZoneInfo` using the ``tzdata`` fallback required on Windows. I used OpenAI Codex with GPT-5 to assist with this change. @floze-the-genius
```
stevepiercy: Let's fix the link.

```suggestion
Ensured every Olson timezone identifier in the Windows mapping resolved through Python's :class:`zoneinfo.ZoneInfo <icalendar.timezone.zoneinfo.ZoneInfo>` using the ``tzdata`` fallback required on Windows. I used OpenAI Codex with GPT-5 to assist with this change. @floze-the-genius
```

src/icalendar/tests/prop/test_windows_to_olson_mapping.py
```diff
@@ -25,6 +26,13 @@ def test_mountain_standard_time_mexico(tzp):
...
+def test_olson_names_resolve_with_zoneinfo(olson_id):
+    """Test that all mappings resolve through zoneinfo's tzdata fallback."""
+    original_tzpath = zoneinfo.TZPATH
+    try:
+        zoneinfo.reset_tzpath(())
+        zoneinfo.ZoneInfo.clear_cache()
+        assert zoneinfo.ZoneInfo(olson_id).key == olson_id
+    finally:
+        zoneinfo.reset_tzpath(original_tzpath)
+        zoneinfo.ZoneInfo.clear_cache()
```
stevepiercy: Will this actually catch a bad Windows to Olson timezone mapping? Please demonstrate that it will perform as you intend.

stevepiercy (changes requested):

news/1562.internal
floze-the-genius: Applied the suggested cross-reference.

src/icalendar/tests/prop/test_windows_to_olson_mapping.py
floze-the-genius: Added a negative regression that passes an invalid Olson identifier through the same validation helper and asserts `ZoneInfoNotFoundError`. The focused file now passes all 111 tests.

stevepiercy (dismissed):
Thanks for the changes. Let's get another review.

niccokunzmann:
Hm. I would expect this to test our implementation at icalendar.timezone. For some reason, this only goes to zoneinfo. Does that mean that we can skip the OLSON lookup entirely for zoneinfo and it might only be needed for pytz?

angatha:
Do you plan to add a similar test for pytz?

src/icalendar/tests/prop/test_windows_to_olson_mapping.py
```diff
@@ -25,6 +32,19 @@ def test_mountain_standard_time_mexico(tzp):
 
 
 @pytest.mark.parametrize("olson_id", WINDOWS_TO_OLSON.values())
-def test_olson_names(tzp, olson_id):
-    """test if all mappings actually map to valid tzids"""
-    assert tzp.timezone(olson_id) is not None
```
niccokunzmann, line 30: Please undo all the changes and just this line needs to be improved to test that the id is correct.

Actually the tests which the issue requests are already in place.

niccokunzmann (changes requested):

floze-the-genius:
Reworked this to the narrow change requested in review: no production-code rewrite or new test module, just the existing assertion now verifies the requested Olson ID is among `tzids_from_tzinfo(tzp.timezone(olson_id))`. The required one-line news fragment remains for the changelog check. The focused file has 216 passing tests and the full PR CI is green; ready for re-review.

stevepiercy (approved):
@niccokunzmann this LGTM. Would you please verify? Thank you!

src/icalendar/tests/prop/test_windows_to_olson_mapping.py
floze-the-genius: Done in the current head: I reverted the broader changes and kept the existing assertion, now verifying that the requested Olson ID is present in `tzids_from_tzinfo(tzp.timezone(olson_id))`. The required one-line news fragment is the only other PR change.

floze-the-genius:
@angatha Not in this PR. The follow-up review clarified that the requested coverage is already present and asked me to keep this change to strengthening the existing assertion only, so the current head follows that narrower scope.

niccokunzmann (approved):

niccokunzmann:
Thanks, this works.

Issue #1570, opened 2026-07-20
Add `security` as a towncrier fragment type


SashankBhamidi:
Currently there's no `security` fragment type in the towncrier config, so security fixes can't be included in the changelog via `make changes`.

Add a `security` directory under `news/` as a registered towncrier type (e.g. "Security fixes") in `pyproject.toml` so future security releases go through the normal changelog flow.

niccokunzmann:
Will the past 2 security fixes be renamed?

PR #1571, opened 2026-07-20
Port GHSA-qjcq-q7h7-r74v fix to main

SashankBhamidi:
Port the GHSA-qjcq-q7h7-r74v security fix from `7.x` to `main`.

Cherry-picked 19 commits covering the fix, tests, docs, and follow-up cleanup. Already shipped as `7.2.2`.


SashankBhamidi:
Removed the news fragment since `security` isn't a registered towncrier type yet (see #1570). 

The changelog entry already landed in `7.x` via 7.2.2.

PR #1576, opened 2026-07-20
Fix #1561: Respect VALUE parameter for IMAGE properties, fix vBinary …

### LINKED ISSUE #1561: [Bug] IMAGE ignores the VALUE parameter and always resolves to vUnknown

KartavyaDikshit:
This PR addresses the issue where `IMAGE` parameters hardcoded the property as unknown even when a `VALUE=URI` or `VALUE=BINARY` parameter was present. 
  It properly maps `VALUE=URI` to `vUri` and `VALUE=BINARY` to `vBinary`, while retaining the fallback `vUnknown` behavior for when `VALUE` is not specified.
                                                                                                                                                            
    Additionally, this PR fixes a core bug discovered in `vBinary` where binary data was being corrupted via `to_unicode()` on parsing. `vBinary` has been  
  refactored to natively store and manipulate raw `bytes`. This prevents round-trip data corruption. The `base64data` property setter has been adapted      
  accordingly.                                                                                                                                              
                                                                                                                                                            
    Tests across the suite (`test_vBinary.py`, `test_image.py`, etc.) have been updated to reflect the fixed behavior of `vBinary` taking raw bytes, and a  
  new test has been added specifically for `IMAGE` properties.                                                                                              
                                                                                                                                                            
    Checklist                                                                                                                                               
                                                                                                                                                            
    - [x] I added a change log entry, following the instructions in Change log entry format.                                                                
    - [x] I followed icalendar's Artificial intelligence policy and disclosed my Responsible AI use in my commit messages, if applicable.                   
    - [x] I ran and ensured all tests pass locally by following Run tests.                                                                                  
    - [x] I added or edited documentation as necessary, both as docstrings to be rendered in the API documentation and narrative documentation, following   
  the Style guide.                                                                                                                                          
                                                                                                                                                            
    Additional information                                                                                                                                  
                                                                                                                                                            
    N/A

### DIFF
--- a/news/1561.bugfix
+++ b/news/1561.bugfix
@@ -1,0 +1,1 @@
+Fixed `IMAGE` property parsing to correctly respect the `VALUE` parameter (`VALUE=URI` or `VALUE=BINARY`), and resolved a data corruption issue in `vBinary` where binary bytes were improperly treated as Unicode strings.
--- a/src/icalendar/prop/binary.py
+++ b/src/icalendar/prop/binary.py
@@ -15,10 +15,13 @@ class vBinary:
 
     default_value: ClassVar[str] = "BINARY"
     params: Parameters
-    obj: str
+    obj: bytes
 
     def __init__(self, obj: str | bytes, params: dict[str, str] | None = None) -> None:
-        self.obj = to_unicode(obj)
+        if isinstance(obj, str):
+            self.obj = obj.encode("utf-8")
+        else:
+            self.obj = obj
         self.params = Parameters(encoding="BASE64", value="BINARY")
         if params:
             self.params.update(params)
@@ -27,7 +30,7 @@ class vBinary:
         return f"vBinary({self.to_ical()})"
 
     def to_ical(self) -> bytes:
-        return binascii.b2a_base64(self.obj.encode("utf-8"))[:-1]
+        return binascii.b2a_base64(self.obj)[:-1]
 
     @staticmethod
     def from_ical(ical: str | bytes) -> bytes:
@@ -82,12 +85,12 @@ class vBinary:
         if params.get("encoding") == "BASE64":
             # BASE64 is the only allowed encoding
             del params["encoding"]
-        return [name, params, self.VALUE.lower(), self.obj]
+        return [name, params, self.VALUE.lower(), binascii.b2a_base64(self.obj).decode("ascii")[:-1]]
 
     @property
     def ical_value(self) -> bytes:
         """The bytes value of the BINARY property."""
-        return self.from_ical(self.obj)
+        return self.obj
 
     @classmethod
     def from_jcal(cls, jcal_property: list) -> Self:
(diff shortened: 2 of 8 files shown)


stevepiercy:
@KartavyaDikshit there are several open PRs for #1561 and an ongoing discussion about them. Always check the open issue for linked PRs, review them, and join the conversion on the original issue or in the PRs to discuss what improvements you would suggest. The icalendar team especially likes it when contributors review each others' work.

Also please note that when you alter the PR template so that the headings are not identifiable, you omit a link to an open issue, or alter the checkboxes, then the PR Quality bot will flag your pull request and delay its review.

PR #1581, opened 2026-07-21
Add CLDR source version to Windows mapping

### LINKED ISSUE #1575: Investigate getting the latest commit SHA for `windowsZones.xml` from the GitHub API
The `$Revision$` in the CLDR XML is an unexpanded placeholder, so the generated file has no record of which commit its data came from. Two runs a month apart against different `main` commits produce different output with nothing in the file to show for it. 

Pulling the latest commit SHA for `windowsZones.xml` from the GitHub API and writing it into the docstring would give us something to point at when someone asks which version of CLDR this reflects.

What do you think?

_Originally posted by @SashankBhamidi in https://github.com/collective/icalendar/pull/1554#discussion_r3612371896_

patrickswedish:
## Linked issue

- Closes #1575

## Description

Resolve the latest CLDR commit for `windowsZones.xml` through the GitHub commits API, fetch the XML from that exact revision, and expose the source SHA as `version` in the generated mapping module. This keeps generated data and its recorded provenance aligned. Generator and generated-output regression tests cover the behavior.

## Checklist

- [x] I followed icalendar's [Artificial intelligence policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) and disclosed my [Responsible AI use](https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use) in my commit messages.
- [x] I added or updated tests.
- [x] I ran all tests locally: 15,984 passed and 30 skipped.
- [x] I updated the generated module docstring to identify the pinned source revision.

## Additional information

Local validation also passed `ruff check` and `towncrier check`.


generate_windows_to_olson_mapping.py
```diff
@@ -13,11 +13,17 @@
...
 from pathlib import Path
 
-# Get the XML data from the CLDR on GitHub and save to a file.
-# Unfortunately, the CLDR does not have a latest or stable tag
-# that points to the stable version.
-# Use the `main` branch, which is the development branch.
-url = "https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml"
+CLDR_PATH = "common/supplemental/windowsZones.xml"
+commits_url = (
+    f"https://api.github.com/repos/unicode-org/cldr/commits?path={CLDR_PATH}&per_page=1"
```
angatha: Does this still target only the main branch?

news/1575.feature
```diff
@@ -0,0 +1 @@
+Record the exact CLDR commit used to generate the Windows-to-Olson timezone mapping and fetch the mapping from that pinned revision. I used AI to assist me with this change.
```
angatha: Can you please add your `@github_username` as described in [Write a good change log entry](https://icalendar.readthedocs.io/en/latest/contribute/index.html#write-a-good-change-log-entry)

angatha (changes requested):

generate_windows_to_olson_mapping.py
patrickswedish: Yes. Without an explicit `sha`, the commits endpoint used the repository default branch, which is currently `main`; the XML fetch was then pinned to the exact SHA returned. I have now added `sha=main` explicitly (and updated the regression test) so this remains unambiguous even if the repository default branch changes.

news/1575.feature
patrickswedish: Added `@patrickswedish` to the changelog entry as requested. I also applied the repository Ruff formatting to the explicit `sha=main` URL and verified the focused tests locally (`2 passed`).

generate_windows_to_olson_mapping.py
angatha: Thank you for clarifying that. Does the API guarantee that the first entry is the latest one? I did not find that in https://docs.github.com/en/rest/commits/commits?apiVersion=2026-03-10.

generate_windows_to_olson_mapping.py
```diff
@@ -35,24 +39,27 @@
...
 Unicode Consortium [1].
 
-[0] https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml
+[0] https://github.com/unicode-org/cldr/blob/{version}/{CLDR_PATH}
 [1] https://cldr.unicode.org/development/development-process/design-proposals/extended-windows-olson-zid-mapping
 
-\"\"\"
+"""
+
+version = "{version}"
```
niccokunzmann: I
```suggestion
version = {json.dumps(version)}
```

This prevents code injection. Better maybe to check version for conformity.

src/icalendar/tests/test_windows_to_olson_version.py
```diff
@@ -0,0 +1,11 @@
+import re
+
+from icalendar.timezone import windows_to_olson
+
+
+def test_windows_to_olson_records_cldr_source_version():
+    """The generated mapping identifies the exact CLDR source commit."""
+    assert re.fullmatch(r"[0-9a-f]{40}", windows_to_olson.version)
```
niccokunzmann, line 8: Here is the check. That should be in the file.

generate_windows_to_olson_mapping.py
```diff
@@ -13,11 +13,15 @@
...
 
-# Get the XML data from the CLDR on GitHub and save to a file.
-# Unfortunately, the CLDR does not have a latest or stable tag
-# that points to the stable version.
-# Use the `main` branch, which is the development branch.
-url = "https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml"
+CLDR_PATH = "common/supplemental/windowsZones.xml"
+commits_url = f"https://api.github.com/repos/unicode-org/cldr/commits?path={CLDR_PATH}&sha=main&per_page=1"
+with req.urlopen(commits_url) as response:  # noqa: S310
+    version = json.load(response)[0]["sha"]
```
niccokunzmann: If you got the data, it would be good to see the date in the file, too. It is easier to say we want to update it, then :)

niccokunzmann (changes requested):

niccokunzmann:
@stevepiercy I think, the windows to olson mapping could be its very own little package. Then, it can be updated, versioned and also someone might come by and suggest how to do the regions.

stevepiercy:
> @stevepiercy I think, the windows to olson mapping could be its very own little package. Then, it can be updated, versioned and also someone might come by and suggest how to do the regions.

I don't have any interest in doing that, since I don't use Windows. Maybe @SashankBhamidi would, since he uses Windows. Else it could be an issue or a discussion, and see who nibbles.

I don't have a clue about "how to do the regions" would work. My only experience with Windows timezones is with MS SQL, which doesn't use regions AFAICT https://learn.microsoft.com/en-us/sql/t-sql/queries/at-time-zone-transact-sql?view=sql-server-ver17. There is a list of timezones available at https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/default-time-zones?view=windows-11.

generate_windows_to_olson_mapping.py
SashankBhamidi: This writes `version` straight into Python source. `json.dumps` handles the quoting safely and signals it is intentional. No surrounding quotes needed since `json.dumps` produces them.

```suggestion
version = {json.dumps(version)}
```

generate_windows_to_olson_mapping.py
```diff
@@ -13,11 +13,15 @@
 import xml.etree.ElementTree as ET
 from pathlib import Path
 
-# Get the XML data from the CLDR on GitHub and save to a file.
-# Unfortunately, the CLDR does not have a latest or stable tag
-# that points to the stable version.
-# Use the `main` branch, which is the development branch.
-url = "https://raw.githubusercontent.com/unicode-org/cldr/refs/heads/main/common/supplemental/windowsZones.xml"
+CLDR_PATH = "common/supplemental/windowsZones.xml"
+commits_url = f"https://api.github.com/repos/unicode-org/cldr/commits?path={CLDR_PATH}&sha=main&per_page=1"
```
SashankBhamidi, line 20: The API does return newest-first but doesn't document that guarantee. Worth a comment so the next person doesn't have to go verify it.

```suggestion
CLDR_PATH = "common/supplemental/windowsZones.xml"
# GitHub returns commits newest-first by default; [0] is the latest commit touching this file.
commits_url = f"https://api.github.com/repos/unicode-org/cldr/commits?path={CLDR_PATH}&sha=main&per_page=1"
```

SashankBhamidi (changes requested):
I ran the script end-to-end and got 139 entries with the SHA written consistently into both `version` and the docstring URL. The two-request dance (commits API first, then fetch by pinned SHA) is exactly right: you can't get a version/data mismatch because both requests hit the same commit.

Two small things left from Nicco and angatha's threads, see inline.

stevepiercy (dismissed):
I didn't test this out, but it looks like a very nice improvement. Thank you for your work! I requested the three other reviewers to take another look.

SashankBhamidi (dismissed):
Good work closing out all the threads @patrickswedish!

The new injection-attempt test is a nice touch.

patrickswedish:
Thank you @stevepiercy and @SashankBhamidi — I really appreciate the thoughtful reviews and the follow-up. The earlier detailed feedback from @angatha and @niccokunzmann also helped tighten the pinned-SHA handling and strengthen the safety coverage. Thanks for taking another look!

generate_windows_to_olson_mapping.py
```diff
@@ -9,15 +9,28 @@
...
+# GitHub returns commits newest-first by default; [0] is the latest commit
+# touching this file.
+commits_url = f"https://api.github.com/repos/unicode-org/cldr/commits?path={CLDR_PATH}&sha=main&per_page=1"
+with req.urlopen(commits_url) as response:  # noqa: S310
+    commit = json.load(response)[0]
+version = commit["sha"]
+version_date = commit["commit"]["committer"]["date"][:10]
+if re.fullmatch(r"[0-9a-f]{40}", version) is None:
+    raise ValueError(f"Invalid CLDR commit SHA: {version!r}")
+if re.fullmatch(r"\d{4}-\d{2}-\d{2}", version_date) is None:
```
angatha: `\d` matches any unicode decimal digit. Please use either `[0-9]` or pass the [ASCII flag](https://docs.python.org/3/library/re.html#re.ASCII).

src/icalendar/tests/test_windows_to_olson_version.py
```diff
@@ -0,0 +1,12 @@
...
+    assert re.fullmatch(r"\d{4}-\d{2}-\d{2}", windows_to_olson.version_date)
```
angatha: Same unicode digit thingy here

angatha (changes requested):

generate_windows_to_olson_mapping.py
patrickswedish: Fixed in a26edc43 by replacing the date pattern with explicit ASCII `[0-9]` ranges. The focused regression test passes and Ruff is clean. Thanks for catching the Unicode behavior.

src/icalendar/tests/test_windows_to_olson_version.py
patrickswedish: Fixed in a26edc43 as well—the regression assertion now uses the same ASCII-only `[0-9]` date pattern. Focused test: 1 passed; Ruff: clean.

generate_windows_to_olson_mapping.py
angatha: This is fine but I prefere `repr(version)` since this returns the python represenation. `json.dumps` is still valid.

angatha (approved):

stevepiercy:
@niccokunzmann have all of your concerns been addressed? I think this is ready to merge.

stevepiercy (approved):

SashankBhamidi (approved):

niccokunzmann (approved):

niccokunzmann:
Thanks! Yes. This is perfect! Merging...

PR #1583, opened 2026-07-22
Add example test for vBinary attachment round-trip (#1549)

### LINKED ISSUE #1549: Add test code with example for vBinary
vBinary can use some more testing. 
An example was shown here:

https://github.com/collective/icalendar/pull/1356#issuecomment-4938460560

To solve this issue:

- [x] Wait for #1356 to be merged.
- [ ] setup the repo for development
- [ ] familiarize yourself with how we test examples through pytest fixtures
- [ ] choose the right PR template
- [ ] create a test case that shows that the example works

adityaanikam:
## Linked issue

- Closes #1549

## Description

Adds an example calendar (`issue_1549_binary_attachment.ics`) containing a
binary PNG attachment, and a test confirming `ATTACH.bytes` preserves the
raw PNG bytes exactly through a full parse via `Calendar.from_ical()`.

This is a regression test for the binary-data corruption fixed in #1356.
It exercises the fix through the public parsing API and the example-fixture
system (`calendars` fixture), rather than unit-testing `vBinary` in
isolation — #1356 already added unit-level coverage for `vBinary` itself,
so this fills in the "does the documented example actually work end to
end" layer that #1549 asked for.

## Checklist

## Additional information

Verified locally: `uv run pytest` — 16012 passed, 30 skipped.

The new test (`test_attach_example_preserves_binary_data`) runs across all
4 parametrize combinations exposed by the existing `calendars` fixture
(`Calendar`/`LazyCalendar` × `pytz`/`zoneinfo`), confirming the PNG's magic
bytes (`\x89PNG\r\n`) survive the round trip in every parsing mode.


stevepiercy:
> This pull request did not pass quality checks and AI use is suspected. Please review [Contribute](https://icalendar.readthedocs.io/en/latest/contribute/index.html) and make any necessary amendments.

Failed due to https://github.com/collective/icalendar/actions/runs/29944438776?pr=1583

news/1549.chore
stevepiercy, line 1: Needs content.

src/icalendar/tests/calendars/issue_1549_binary_attachment.ics
stevepiercy, line 1: Add new line at EOF.

stevepiercy (changes requested):

news/1549.chore
adityaanikam: Fixed — added the content in b633032f.

src/icalendar/tests/calendars/issue_1549_binary_attachment.ics
adityaanikam: Fixed — added the trailing newline in b633032f.

news/1549.chore
```diff
@@ -0,0 +1,4 @@
+Added an example calendar with a binary PNG attachment and a regression
+test confirming it round-trips losslessly through `Calendar.from_ical()`,
+per #1549. Prepared with the assistance of Claude (via the Claude Code
+VS Code extension). @adityaanikam
```
stevepiercy: Make it link using reStructuredText markup, and place all content on a single line.

```suggestion
Added an example calendar with a binary PNG attachment and a regression test confirming it round-trips losslessly through :func:`Calendar.from_ical <icalendar.cal.calendar.Calendar.from_ical>`, per #1549. Prepared with the assistance of Claude (via the Claude Code VS Code extension). @adityaanikam
```

stevepiercy (changes requested):
Minor tweak to the news fragment. Needs a technical review.

news/1549.chore
adityaanikam: Fixed used the :func: cross reference and put it all on one line

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -180,3 +180,15 @@ def test_base64data_setter_stores_raw_bytes():
...
     assert obj.base64data == encoded
+
+
+def test_attach_example_preserves_binary_data(calendars):
+    """The example calendar's PNG attachment round-trips without corruption.
+
+    Regression test for the corruption fixed in #1356. See #1549.
+    """
+    calendar = calendars.issue_1549_binary_attachment
+    event = calendar.subcomponents[0]
```
stevepiercy: I wonder if there is there another way to get the event component from a Calender object? If so, that should be tested, too.

src/icalendar/tests/prop/test_vBinary.py
```diff
@@ -180,3 +180,15 @@ def test_base64data_setter_stores_raw_bytes():
...
+    attach = event["ATTACH"]
+
+    assert attach.bytes[:6] == b"\x89PNG\r\n"
```
stevepiercy: This test feels brittle. It doesn't use the icalendar API to get an attachment. Can you find a way to do that instead?

stevepiercy (commented):
Please see my comments. This PR needs a further technical review.

src/icalendar/tests/prop/test_vBinary.py
adityaanikam: You are right, `decoded()` does it. Switched to `event.decoded("ATTACH")`. That is actually a better regression test here than reading `.bytes` off the property, because `decoded()` routes through `vBinary.ical_value`, which is the exact thing PR #1356 changed (its docstring notes the 7.1.3 change to return the raw stored bytes). So the test now goes through the same path a user would.

 I also tightened the assertion. The old one only checked a 6 byte prefix, so it would have passed even if the rest of the payload were corrupted. It now checks the full 8 byte PNG signature, the IEND trailer and the total length, and I added a second test that round-trips through `to_ical()` and `from_ical()` and compares the decoded bytes.

src/icalendar/tests/prop/test_vBinary.py
adityaanikam: Good catch. Switched to `calendar.events`, which is the documented property (it just calls `walk("VEVENT")` internally). I used `(event,) = calendar.events` so the test also asserts there is exactly one event, rather than silently assuming the VEVENT happens to be first in `subcomponents`.

niccokunzmann (approved):
This is sound. Thanks!

angatha (approved):

stevepiercy (approved):

stevepiercy:
Thank you! Will automerge when CI passes.

PR #1587, opened 2026-07-23
docs: convert vUri docstring to Pythonic style (#1244)

SpiliosDimakopoulos:
Drafted with assistance from Claude (Anthropic). I reviewed the change, ran the full test suite (10358 passed) and doctest, and confirmed ruff formatting/linting pass before submitting

### DIFF
--- a/src/icalendar/prop/uri.py
+++ b/src/icalendar/prop/uri.py
@@ -9,35 +9,15 @@ from icalendar.parser_tools import DEFAULT_ENCODING, to_unicode
 
 
 class vUri(str):
-    """URI
+    """A value that identifies a resource with a URI, per :rfc:`5545#section-3.3.13`.
 
-    Value Name:
-        URI
+    This value type is used to reference values that are large, binary, or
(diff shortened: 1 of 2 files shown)


src/icalendar/prop/uri.py
```diff
@@ -9,35 +9,15 @@
...
-    """URI
+    """A value that identifies a resource with a URI, per :rfc:`5545#section-3.3.13`.
 
-    Value Name:
-        URI
+    This value type is used to reference values that are large, binary, or
+    otherwise undesirable to include directly in the iCalendar object, such
+    as a network file. Property values with this value type must follow the
+    generic URI syntax defined in :rfc:`3986`. When a property *parameter*
+    value is a URI, it must be specified as a quoted-string value.
```
angatha, line 18: @stevepiercy Does the last sentence not belong here since it is part of the spec and has nothing to do with this python API?

stevepiercy: @angatha yes, the last sentence is incorrect because it describes the RFC spec, and not the Python implementation.

PR #1588, opened 2026-07-23
fix: replace assert statements with explicit exceptions in parser

### LINKED ISSUE #1532: `assert` statements in the parser produce invalid output when Python runs with `-O`

Madhavkalia076:
Fixes #1532.

Four `assert` statements in the parsing pipeline guarded things we care about at runtime (embedded newlines, wrong types), but Python's `-O` flag strips all assertions silently. Ruff rule B011, already referenced in pyproject.toml, flags exactly this pattern.

Without these checks under `-O`:
- `Contentline` accepted embedded newlines, so `to_ical()` could emit invalid RFC 5545 output.
- `_foldline()` passed multiline strings through unchanged, violating the 75-octet line length rule.
- `_escape_char()` / `_unescape_char()` silently accepted wrong types instead of failing fast.

**Changes:**
- `Contentline.__new__` now raises `ValueError` for embedded newlines.
- `_escape_char` / `_unescape_char` now raise `TypeError` for non-str/bytes input.
- `_foldline` now raises `TypeError`/`ValueError` for bad type/embedded newlines.
- Updated two existing tests that asserted the old `AssertionError` behavior.
- Added a `news/1532.bugfix` changelog fragment.

**Verified:**
- Reproduced the exact bug from the issue with `python -O`; confirmed the fix resolves it.
- Full test suite passes: 15984 passed, 30 skipped.
- `ruff check` / `ruff format --check` (pinned version 0.15.2) both pass clean.


niccokunzmann:
Hi, is there a PR already open for this? In this case, please give it a review.

angatha:
Jup, please review #1543 and add your opinion to is. Thank you very much for trying to contribute.

SashankBhamidi:
Picked this up since it's been sitting for a bit. 

The previous attempts (#1542 and #1543) were both closed without merging, so this fix still needs to land. The code is correct. 

I added `test_issue_1532_assert_to_explicit_errors.py` with a parametrized test covering each guard, which was the one thing missing from the original.

@niccokunzmann @angatha ready for your review.

news/1532.bugfix
```diff
@@ -0,0 +1 @@
+Replaced four internal ``assert`` statements in the parsing pipeline with explicit :class:`TypeError`/:class:`ValueError` checks. Python's ``-O`` flag strips all ``assert`` statements, which silently disabled validation in :class:`~icalendar.parser.content_line.Contentline` (embedded newlines) and in ``_escape_char``, ``_unescape_char``, and ``_foldline`` (type checks and embedded newlines). Under ``-O``, this allowed invalid content lines to be produced instead of raising an error. These checks are now unconditional regardless of the ``-O`` flag.
```
angatha, line 1: Attribution is missing

src/icalendar/parser/content_line.py
angatha, line 120: This should now mention the possible exception? @stevepiercy

angatha (changes requested):

news/1532.bugfix
stevepiercy: It's also too verbose AI slop. Come on, @Madhavkalia076, don't copy-paste from AI, and instead summarize and be human. No one wants to read how sausage is made. Change log details must be concise without all the implementation details.

src/icalendar/parser/content_line.py
stevepiercy: Correct. The class should now have a `Raises` section heading.

src/icalendar/parser/string.py
```diff
@@ -19,6 +19,9 @@ def _escape_char(text: str | bytes) -> str:
     Returns:
         The escaped text with special characters escaped.
 
+    Raises:
+        TypeError: If ``text`` is not ``str`` or ``bytes``.
```
stevepiercy, line 23:
```suggestion
        TypeError: If ``text`` is neither ``str`` nor ``bytes``.
```

src/icalendar/parser/string.py
```diff
@@ -71,7 +75,10 @@ def _unescape_char(text: str | bytes) -> str | bytes | None:
         text: The escaped text.
 
     Returns:
-        The unescaped text, or ``None`` if ``text`` is neither ``str`` nor ``bytes``.
+        The unescaped text.
+
+    Raises:
+        TypeError: If ``text`` is not ``str`` or ``bytes``.
```
stevepiercy, line 81:
```suggestion
        TypeError: If ``text`` is neither ``str`` nor ``bytes``.
```

src/icalendar/parser/string.py
stevepiercy, line 14:
```suggestion
    The order of replacements matters to avoid double-escaping.
    
    ..  versionchanged:: 7.2.3
        Raise a ``TypeError`` if ``text`` is neither ``str`` nor ``bytes``.
```

stevepiercy, line 72:
```suggestion
    :rfc:`5545#section-3.3.11` TEXT escaping rules.
    
    ..  versionchanged:: 7.2.3
        Raise a ``TypeError`` if ``text`` is neither ``str`` nor ``bytes``.
```

stevepiercy (changes requested):

stevepiercy:
This PR and #1559 may conflict. There have been 5 PRs opened for this issue, 3 closed, and 2 remain open. We should always check which PR is best before merging.

niccokunzmann:
@Madhavkalia076 thanks for opening the PR. We had no feedback or reply from you. So, I will close this now so that we do not spend more time on it. If you feel like you would continue, no problem! Just comment and we can proceed.
Best,
Nicco

Issue #1594, opened 2026-07-25
Validate Alarm repeat when setting


niccokunzmann:
Alarm.new* methods can set the repeat argument.
That should be validated to be an integer >= 0.

- [ ] introduce a `min_value` to `single_int_property`
- [ ] Test setting a positive value
- [ ] Test setting a negative value
- [ ] Test setting 0
- [ ] Test the boundary
- [ ] Check which **lower case** attributes actually require a boundary (sequence, repeat, ...?):
    - [ ] make sure they use as lowercase, not uppercase attributes
- [ ] use `repeat` in alarm.py as a setter for the new methods
- [ ] raise the correct error
- [ ] document the errors and values, consider complying with https://github.com/collective/icalendar/issues/1244
- [ ] Add `versionadded` admonition to the docstrings, anticipating the next release version number.

Related:

- https://www.rfc-editor.org/info/rfc5545/#section-3.8.6.2
- https://github.com/collective/icalendar/pull/1417#issuecomment-4559192392
- https://github.com/collective/icalendar/blob/main/src/icalendar/attr.py
- https://github.com/collective/icalendar/blob/main/src/icalendar/cal/alarm.py

stevepiercy:
@niccokunzmann I think this validation should also comply with https://www.rfc-editor.org/info/rfc5545/#section-3.8.6.2, specifically, "If the alarm triggers more than once, then this property MUST be specified along with the "DURATION" property."

shiva-kantamneni:
@niccokunzmann 
can you assign it to me i am interested to solve  this issue

stevepiercy:
@shiva-kantamneni please read and follow [Contribute](https://icalendar.readthedocs.io/en/stable/contribute/index.html).

niccokunzmann:
This has been released with v7.3.0.

PR #1596, opened 2026-07-25
Validate Alarm REPEAT when setting: reject negative values

pratyushsinghal7:
## Linked issue

- See #1594

## Description

Setting a negative value on the alarm REPEAT property used to be stored silently, even though RFC 5545 defines REPEAT as a non-negative count. This adds a `min_value` argument to the int property helper and uses it for REPEAT, so a negative value now raises a ValueError. A failed set leaves any existing value untouched. The alarm factory methods assign through the validated setter, so they get the same check.

This covers the setter-validation part of the issue; auditing other lowercase int attributes for boundaries is left for follow-up, hence "See" rather than "Closes".

## Checklist

- [x] I added a change log entry.
- [x] I followed icalendar's AI policy and disclosed my AI use in the commit message and change log entry.
- [x] I added tests for positive, zero, and negative values.
- [x] I ran all tests locally.
- [x] I updated the docstrings.

### DIFF
--- a/news/1594.feature
+++ b/news/1594.feature
@@ -1,0 +1,1 @@
+Validate the ``REPEAT`` property of :class:`~icalendar.cal.alarm.Alarm` when setting it: negative values now raise a :class:`ValueError`. :func:`icalendar.attr.single_int_property` accepts a new ``min_value`` argument for this. AI disclosure: I used Devin (Cognition) to draft this change and its tests; I reviewed and validated the result. @pratyushsinghal7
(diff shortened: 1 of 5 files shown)


stevepiercy:
@pratyushsinghal7 it's always best to have a discussion in the original issue about unclear things before starting work. See https://github.com/collective/icalendar/issues/1594#issuecomment-5081419732. If you had bothered to check the RFC, you would have realized the omission. There's no points awarded for being first with a PR containing AI slop, and actually puts your reputation at risk, as well as getting your account blocked from the organization and getting reported to GitHub. I'm inclined to do exactly that because you did not follow our AI policy.

PR #1598, opened 2026-07-26
fix: validate non-negative integer values for repeat, sequence, priority

### LINKED ISSUE #1594: Validate Alarm repeat when setting

heyncth:
## Summary

Fixes #1594. Setting a negative or non-integer value on a `VALARM` repeat
used to be silently accepted and stored. Those values are later used to
expand alarm triggers, so a bad value caused confusing behavior downstream.

This PR lets `single_int_property()` enforce a minimum value and applies it
where RFC 5545 requires non-negative integers:

- `Alarm.REPEAT` and the lowercase `repeat` attribute (used by the
  `Alarm.new_display()` / `new_audio()` / `new_email()` factories)
- `sequence` on events, todos, journals, and availability components
- `priority` on events and todos

Invalid values now raise `InvalidCalendar` with a message that names the
property, the expected minimum, and the offending value.

## Changes

- `single_int_property()` accepts an optional keyword-only `min_value`
  parameter. The setter validates with an `isinstance(value, int)` check
  followed by a `value >= min_value` comparison — no implicit conversion,
  no `try/except`.
- `sequence`, `priority`, and the lowercase `repeat` property use
  `min_value=0`.
- The alarm factory methods now assign through the validated lowercase
  `repeat` setter.
- Added `src/icalendar/tests/attr/test_issue_1594.py` covering positive
  values, zero, negative values, non-integers, and custom minimum
  boundaries, asserting the exact error messages.
- `versionadded` notes for 7.2.3.

## Test plan

```
python -m pytest src/icalendar/tests/attr/test_issue_1594.py \
    src/icalendar/tests/attr/test_alarm.py -q
```


stevepiercy:
Another slop AI PR for #1594.

heyncth:
> Another slop AI PR for #1594.

Hey thats not really cool. Just because I made a new account and Im starting from zero does NOT mean my work is AI slop. Judge the PR on its own merits, not the age of my account

stevepiercy:
> Judge the PR on its own merits

I did. It doesn't address the discussion in the original issue.

The CI check also found faults, per https://github.com/collective/icalendar/pull/1598#issuecomment-5081524330, in https://github.com/collective/icalendar/actions/runs/30183837818?pr=1598.

news/1594.bugfix
```diff
@@ -0,0 +1,3 @@
+Validate Alarm repeat value is non-negative.
+Alarm.REPEAT now validates that the repeat value is an integer >= 0,
+raising InvalidCalendar if a negative value or non-integer is provided.
```
niccokunzmann: The file must only have one line.

src/icalendar/attr.py
```diff
@@ -421,6 +424,14 @@ def fget(self: Component) -> int:
 
     def fset(self: Component, value: int | None):
         """Set the property."""
+        if value is not None:
+            try:
+                if min_value is not None and int(value) < min_value:
+                    raise InvalidCalendar(f"{prop} must be >= {min_value}, got {value}")
```
niccokunzmann: Hm. Is invalidCalendar a ValueError?

src/icalendar/attr.py
```diff
@@ -421,6 +424,14 @@ def fget(self: Component) -> int:
...
+            except (TypeError, ValueError):
```
niccokunzmann: It is unusual to catch what is raised inside. Can you guard just the statement that has the exception?

src/icalendar/attr.py
```diff
@@ -421,6 +424,14 @@ def fget(self: Component) -> int:
 
     def fset(self: Component, value: int | None):
         """Set the property."""
+        if value is not None:
+            try:
+                if min_value is not None and int(value) < min_value:
```
niccokunzmann: Hm. The type hint says int. I think, we can expect int here. It would minimalize the code. I wonder what the is in the tests.

src/icalendar/attr.py
```diff
@@ -421,6 +424,14 @@ def fget(self: Component) -> int:
...
+                    raise InvalidCalendar(f"{prop} must be >= {min_value}, got {value}")
+            except (TypeError, ValueError):
+                raise InvalidCalendar(
```
niccokunzmann: We usually would use a isinstance check to raise that.

niccokunzmann (changes requested):
Thanks for opening the PR!

In icalendar, we use tests to make sure that added functionality works over a long time with many people editing the code. Could you add a test file in pytest style that checks different minimal values and that the errors have the correct messages?

niccokunzmann:
It does not look like the usual AI generated PR to me. But if you used an AI, please disclose it.

We got a lot of PRs coming in and some people just post AI output directly without thinking. That puts a strain on the team to keep up with reviews and sort through.

heyncth:
Thanks for the thorough review — that all made sense, and I've reworked the PR to address each point.

1. **Tests** — I added `src/icalendar/tests/attr/test_issue_1594.py` in pytest style. It checks different minimal values (including a custom `min_value` built directly with `single_int_property`), the 0 boundary, positive and negative values, non-integers, and asserts the exact error messages for each case.

2. **News file** — collapsed `news/1594.bugfix` to a single line.

3. **InvalidCalendar being a ValueError** — you were right, it is (`class InvalidCalendar(ValueError)`), and that's exactly what made my original `try/except` wrong: the `except (TypeError, ValueError)` was catching the very `InvalidCalendar` raised inside the `try`, so a negative value ended up with the wrong message ("must be an int" instead of "must be >= 0"). I removed the `try/except` entirely.

4. **Type hint / isinstance** — since the parameter is typed `int | None`, I now use `isinstance(value, int)` to reject non-integers, then a plain `value < min_value` comparison for the boundary. No conversion, nothing to catch.

While in there, I also followed the issue's checklist: the lowercase `repeat` attribute (used as the setter by `Alarm.new_display()`, `new_audio()`, and `new_email()`) now validates with `min_value=0`, and so do `sequence` and `priority`, which RFC 5545 also requires to be non-negative. The factory methods now assign through that validated `repeat` setter.

The relevant test suites pass locally (`2077 passed` across the attr tests and the sequence/priority/new-method suites). Let me know if you'd like anything else adjusted.

src/icalendar/attr.py
```diff
@@ -421,6 +431,11 @@ def fget(self: Component) -> int:
 
     def fset(self: Component, value: int | None):
         """Set the property."""
+        if value is not None:
+            if not isinstance(value, int):
+                raise InvalidCalendar(f"{prop} must be an int, got {value}")
```
SashankBhamidi: Wrong exception type here. 

The codebase uses `TypeError` when a caller passes the wrong Python type to a setter, and `InvalidCalendar` for RFC constraint violations. 

See `single_utc_property` (line 485) and `_single_component_property` (line 1006) for examples. 

Passing a string or float to `REPEAT` is a programming mistake, not invalid calendar data. 

The `min_value` check on line 438 is correctly `InvalidCalendar`.

```suggestion
                raise TypeError(f"{prop} must be an int, got {value!r}")
```

src/icalendar/attr.py
```diff
@@ -421,6 +431,11 @@ def fget(self: Component) -> int:
 
     def fset(self: Component, value: int | None):
         """Set the property."""
+        if value is not None:
+            if not isinstance(value, int):
```
SashankBhamidi: `isinstance(value, int)` accepts `bool` since `bool` is a subclass of `int`. 

`alarm.REPEAT = True` currently sets REPEAT to 1 silently. 

That may be fine, but it should be a deliberate choice. If you want to exclude it:

```suggestion
            if not isinstance(value, int) or isinstance(value, bool):
```

Otherwise just note it in the docstring.

SashankBhamidi (changes requested):
Thanks for working through Nicco's feedback and addressing all the points from the first round. 

The structure of the fix is right: `min_value` goes into `single_int_property`, validation happens at the setter, and all three factory methods flow through `_apply_duration_repeat` which now uses the validated lowercase `repeat`. 

Your explanation of the `try/except` bug was well-organized. Can you explain in your own words why the original `except (TypeError, ValueError)` approach was wrong?

A couple of things inline.

heyncth:
Thanks for the second round — both inline points are addressed in the latest commit on this branch.

On your question about the original `except (TypeError, ValueError)` approach: it was wrong because `InvalidCalendar` is itself a `ValueError`. The setter deliberately raised `InvalidCalendar` for the negative-value case *inside* the `try` block, so the `except` caught that same exception and re-wrapped it as "must be an int". A negative value ended up with the wrong message, and the handler made the intent ambiguous — you couldn't tell whether it was meant for a conversion failure or was swallowing the deliberate raise. The fix removes the conversion entirely (an `isinstance` check up front, a plain comparison after), so there is nothing left to catch and each raise has exactly one meaning.

I applied the exception-type convention you pointed to: non-integers now raise `TypeError` (matching `single_utc_property` and the other setters), while the `min_value` check still raises `InvalidCalendar`.

On `bool`: I went with the stricter option and reject booleans as well (`not isinstance(value, int) or isinstance(value, bool)`), so `alarm.REPEAT = True` can no longer silently store 1.

Tests updated in `test_issue_1594.py` (TypeError cases, a bool parametrize, and a sequence case) and `test_alarm.py`. The attr, alarm, and event suites pass (1,956 tests).

src/icalendar/attr.py
heyncth: Went with the stricter option — booleans are rejected too (`not isinstance(value, int) or isinstance(value, bool)`), so `alarm.REPEAT = True` raises `TypeError` instead of silently storing 1. Added `True`/`False` parametrized tests.

src/icalendar/attr.py
heyncth: Fixed — non-integers now raise `TypeError` (with `{value!r}` in the message), matching the convention used by `single_utc_property` and the other setters. The `min_value` check still raises `InvalidCalendar`. Tests in `test_issue_1594.py` updated accordingly.

SashankBhamidi (dismissed):
Thank you, @heyncth.

---

@niccokunzmann, both points are addressed. 

`TypeError` is now raised for non-integers and booleans, `InvalidCalendar` for the `min_value` constraint, and the `Raises` section in the docstring documents both. 

I verified `alarm.REPEAT = True` now raises `TypeError`.

src/icalendar/attr.py
```diff
@@ -403,13 +403,25 @@ def fdel(self: Component):
...
+def single_int_property(
+    prop: str, default: int, doc: str, *, min_value: int | None = None
+) -> property:
     """Create a property for an int value that exists only once.
 
     Parameters:
         prop: The name of the property
         default: The default value
         doc: The documentation string
+        min_value: If set, the value must be >= this minimum
```
stevepiercy: While we're in here, parameters must end with a period, per https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure. Also let's sort them in the required and optional groups, and specify required as needed.

```suggestion
        default: Required. The default value.
        doc: Required. The documentation string.
        prop: Required. The name of the property.
        min_value: If set, the value must be greater than or equal to this minimum.
```

src/icalendar/attr.py
```diff
@@ -403,13 +403,25 @@ def fdel(self: Component):
     return property(fget, fset, fdel, doc)
 
 
-def single_int_property(prop: str, default: int, doc: str) -> property:
+def single_int_property(
```
stevepiercy, line 406: Sphinx warns about formatting of this method in https://github.com/collective/icalendar/actions/runs/31082258918/job/92553976140?pr=1598#step:7:759. I couldn't find what it chokes on. Can you find the problem and fix it?

src/icalendar/cal/alarm.py
```diff
@@ -377,7 +381,7 @@ def _apply_duration_repeat(
                     "DURATION and REPEAT must be set together or not at all"
                 )
             self.DURATION = duration
-            self.REPEAT = repeat
+            self.repeat = repeat
```
stevepiercy: Is this change correct? It's now inconsistent with the previous line:

```python
            self.DURATION = duration
```

I'd assume that one or the other is now incorrect, since they're now inconsistent.

news/1594.bugfix
```diff
@@ -0,0 +1 @@
+``Alarm.REPEAT``, ``Alarm.repeat``, ``sequence`` and ``priority`` now reject non-integers (including booleans) with ``TypeError`` and negative values with ``InvalidCalendar``.
```
stevepiercy: Let's link to the objects, and give you credit.

```suggestion
:attr:`Alarm.REPEAT <icalendar.cal.alarm.Alarm.REPEAT>`, :attr:`Alarm.repeat <icalendar.cal.alarm.Alarm.repeat>`, :attr:`Event.sequence <icalendar.cal.event.Event.sequence>`, and :attr:`Event.priority <icalendar.cal.event.Event.priority>` now reject non-integers (including booleans) with :exc:`TypeError` and negative values with :exc:`~icalendar.error.InvalidCalendar`. @heyncth
```

src/icalendar/cal/alarm.py
```diff
@@ -118,7 +118,11 @@ class Alarm(Component):
                 >>> alarm.REPEAT = 2
                 >>> alarm.REPEAT
                 2
+
+        .. versionadded:: 7.2.3
+            Negative values are no longer accepted.
```
stevepiercy: Flagging this for maintainers to update the version number, if needed, upon release.

```suggestion
        ..  versionadded:: 7.2.3
            Added the ``min_value`` parameter. Negative values are no longer accepted.
```

src/icalendar/tests/attr/test_issue_1594.py
```diff
@@ -0,0 +1,142 @@
...
+
+class NumberComponent(Component):
+    """A component with a single integer property that has a minimum."""
+
+    name = "X-NUMBER"
+    number = single_int_property("NUMBER", 0, "A number", min_value=5)
+
+
+@pytest.mark.parametrize("value", [0, 1, 10])
+def test_set_positive_value(value):
```
stevepiercy: 0 is not positive.

```suggestion
def test_set_non_negative_value(value):
```

news/1594.bugfix
angatha: Attribution is missing please add `@yourself`.

src/icalendar/tests/attr/test_issue_1594.py
```diff
@@ -0,0 +1,142 @@
...
+    alarm = Alarm()
+    with pytest.raises(InvalidCalendar, match=message):
+        alarm.REPEAT = value
+
+
+def test_set_zero():
+    """Setting zero, the boundary, works."""
+    alarm = Alarm()
+    alarm.REPEAT = 0
+    assert alarm.REPEAT == 0
```
stevepiercy: This test duplicates the currently misnamed `test_set_positive_value`. Either remove this test, or remove `0` from the misnamed `test_set_positive_value` while keeping its name so it's aligned and there's no duplication.

src/icalendar/attr.py
angatha: Should be marked as optional by new guideline. (And the three lines above should be sorted alphabetically so the order is `default`, `doc`, `prop`, `min_value`.)

angatha (changes requested):

src/icalendar/attr.py
```diff
@@ -403,13 +403,25 @@ def fdel(self: Component):
...
 
     Parameters:
         prop: The name of the property
         default: The default value
         doc: The documentation string
+        min_value: If set, the value must be >= this minimum
+
+    Raises:
+        :exc:`TypeError`: If the value is not an ``int`` (booleans are not
+            accepted either, even though ``bool`` subclasses ``int``).
```
```suggestion
        :exc:`TypeError`: If the value is not an ``int``. Booleans are
            rejected, too, even though ``bool`` subclasses ``int``.
```

src/icalendar/attr.py
```diff
@@ -403,13 +403,25 @@ def fdel(self: Component):
...
+        :exc:`~icalendar.error.InvalidCalendar`: If the value is smaller than
+            ``min_value``.
+
+    .. versionadded:: 7.2.3
+        The ``min_value`` parameter.
```
```suggestion
        Added the ``min_value`` parameter. Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -618,7 +635,11 @@ def fdel(self: Component):
         >>> event = calendar.events[0]
         >>> event.sequence
         10
+
+    .. versionadded:: 7.2.3
+        Negative values are no longer accepted.
```
```suggestion
    ..  versionadded:: 7.2.3
        Added the ``min_value`` parameter. Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -1297,7 +1318,11 @@ def _make_repeat_property() -> property:
 Defaults to ``0``, meaning the alarm fires once. Must be paired with
 :attr:`~icalendar.cal.alarm.Alarm.DURATION`. Conforms with :rfc:`5545#section-3.8.6.2`.
 The value is capped at :data:`icalendar.config.MAX_ALARM_REPEAT` on read.
+
+.. versionadded:: 7.2.3
+    Negative values are no longer accepted.
```
```suggestion
..  versionadded:: 7.2.3
    Added the ``min_value`` parameter. Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -1346,7 +1371,11 @@ def fget(self):
     Within a "VTODO" calendar component, this property specified a
     priority for the to-do.  This property is useful in prioritizing
     multiple action items for a given time period.
+
+    .. versionadded:: 7.2.3
+        Negative values are no longer accepted.
```
```suggestion
    ..  versionadded:: 7.2.3
        Added the ``min_value`` parameter. Negative values are no longer accepted.
```

stevepiercy (changes requested):
Mostly nitpicky bits, but also a malformed docstring, a potential duplicate test, and a potential inconsistency. Would you please take care? Thank you!

src/icalendar/attr.py
angatha: Sorry I was wrong. Required ones should be marked as required. Not the other way around. You may adapt this any way.

stevepiercy: We can optionally leave off "Optional." "Required." is, well, required. :wink: Should I clarify that in the docstring style guide? https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure

news/1594.bugfix
angatha: @stevepiercy also suggested this. I mark this as resolved because his request covers it.

src/icalendar/cal/alarm.py
heyncth: Good catch — that was inconsistent. I've changed it to `self.REPEAT = repeat` so both assignments use the uppercase attributes. Both setters go through the same validation, so the behavior is unchanged.

src/icalendar/attr.py
heyncth: Found it — thanks for the link. The problem was the `Raises` section of the `single_int_property` docstring: two `:exc:`-prefixed fields with multi-line bodies trip the docutils field-list parser, which emits `Field list ends without a blank line; unexpected unindent`. It only shows up on the 3.11 job because that matrix entry pins Sphinx 9.0.4, while 3.14 uses 9.1.

I reproduced it locally with Python 3.11 + Sphinx 9.0.4 and verified the fix: each `:exc:` role now sits on its own line with the description indented below — the same pattern `create_single_property` already uses in this file. The docs build is clean now, and your wording suggestions are included.

heyncth:
Thanks to you both for the careful review — all points are addressed in the latest commit:

- **Docstring**: the malformed `Raises` section is fixed. The cause was a docutils field-list parse error that only surfaced on Sphinx 9.0.4 (the 3.11 docs job) — I reproduced it with Python 3.11 + Sphinx 9.0.4 locally and confirmed the build is clean after restructuring the fields (role on its own line, matching `create_single_property`). `Parameters` are sorted (`default`, `doc`, `prop`, `min_value`), required ones are marked `Required.`, and every description ends with a period. The `versionadded` notes are unified to mention the `min_value` parameter.
- **Tests**: `test_set_positive_value` is renamed to `test_set_non_negative_value` (0 is not positive), and the duplicate `test_set_zero` was removed.
- **`_apply_duration_repeat`**: now uses `self.REPEAT = repeat` for consistency with `self.DURATION`.
- **News entry**: links to the objects and credits the author, per the suggestion.

All 843 attribute tests pass, and the docs build is green under the CI configuration.

src/icalendar/attr.py
stevepiercy: Ew, what a stinker bug! Thanks for resolving it.

stevepiercy:
I think this criteria was overlooked, as I created it after the initial issue description was published. Would you please address this one, too? Alternatively, would you create a new issue to address the omission, and then this PR can be merged?

https://github.com/collective/icalendar/issues/1594#issuecomment-5081419732

@niccokunzmann were all of your unresolved comments resolved?

src/icalendar/attr.py
```diff
@@ -403,13 +403,26 @@ def fdel(self: Component):
...
+        doc: Required. The documentation string.
+        prop: Required. The name of the property.
+        min_value: If set, the value must be greater than or equal to this minimum.
+
+    Raises:
+        :exc:`TypeError`
+            If the value is not an ``int``. Booleans are rejected, too,
+            even though ``bool`` subclasses ``int``.
+        :exc:`~icalendar.error.InvalidCalendar`
+            If the value is smaller than ``min_value``.
```
stevepiercy: Actually, let's use a simpler syntax. It renders exactly the same.

```suggestion
        TypeError: If the value is not an ``int``. Booleans are rejected,
            too, even though ``bool`` subclasses ``int``.
        ~icalendar.error.InvalidCalendar: If the value is smaller than ``min_value``.
```

stevepiercy (changes requested):
One minor simplifcation, after you found the cause of the Sphinx warning, then this is ready to merge, pending how you would like to address the omission of https://github.com/collective/icalendar/issues/1594#issuecomment-5081419732. Thank you for your work on this!

angatha (dismissed):

heyncth:
@stevepiercy good catch — I'd missed that criterion since it was added to the issue after the description was published. I've opened #1642 to track the RFC 5545 §3.8.6.2 `REPEAT`/`DURATION` pairing validation, including the current behavior and a proposed location for the check.

The docstring simplification is also pushed (`c3e9ec31`). Thanks for the close review on this one.

src/icalendar/attr.py
heyncth: Applied — the `Raises` section now uses the plain-text syntax as you suggested, and it renders as expected.

One note for future reference: under Sphinx 9.0.4 the two field lines must stay on single lines — wrapping them re-triggers the docutils field-list parse warning from before. Since the first line is now over 88 characters, it's covered by a `# noqa: E501` on the closing quotes, matching the pattern already used elsewhere in this file.

angatha (dismissed):

stevepiercy:
@niccokunzmann and @SashankBhamidi would you please resolve your unresolved comments, if they are in fact resolved? They're blocking auto-merge.

src/icalendar/attr.py
```diff
@@ -403,14 +403,24 @@ def fdel(self: Component):
...
+        doc: Required. The documentation string.
+        prop: Required. The name of the property.
+        min_value: If set, the value must be greater than or equal to this minimum.
+
+    Raises:
+        TypeError: If the value is not an ``int``. Booleans are rejected, too, even though ``bool`` subclasses ``int``.
+        ~icalendar.error.InvalidCalendar: If the value is smaller than ``min_value``.
+
+    ..  versionadded:: 7.2.3
+        Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
        Added the ``min_value`` parameter.
```

src/icalendar/attr.py
```diff
@@ -618,7 +633,11 @@ def fdel(self: Component):
         >>> event = calendar.events[0]
         >>> event.sequence
         10
+
+    ..  versionadded:: 7.2.3
+        Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
        Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -1297,7 +1316,11 @@ def _make_repeat_property() -> property:
 Defaults to ``0``, meaning the alarm fires once. Must be paired with
 :attr:`~icalendar.cal.alarm.Alarm.DURATION`. Conforms with :rfc:`5545#section-3.8.6.2`.
 The value is capped at :data:`icalendar.config.MAX_ALARM_REPEAT` on read.
+
+..  versionadded:: 7.2.3
+    Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
    Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -1346,7 +1369,11 @@ def fget(self):
     Within a "VTODO" calendar component, this property specified a
     priority for the to-do.  This property is useful in prioritizing
     multiple action items for a given time period.
+
+    ..  versionadded:: 7.2.3
+        Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
        Negative values are no longer accepted.
```

src/icalendar/cal/alarm.py
```diff
@@ -118,7 +118,11 @@ class Alarm(Component):
                 >>> alarm.REPEAT = 2
                 >>> alarm.REPEAT
                 2
+
+        ..  versionadded:: 7.2.3
+            Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
            Negative values are no longer accepted.
```

niccokunzmann (changes requested):
Thanks! I changed the docs a bit to be descriptive and not reveal how things are done.
Technically ready for merge!

src/icalendar/attr.py
stevepiercy: I couldn't reproduce that result. Locally, in `pyproject.toml`, line 58:

```toml
requires-python = ">=3.11"
```

Then:

```shell
uv lock --upgrade-package sphinx==9.0.4
make init livehtml PYTHONVERSION=3.11
```

Build succeeds with wrapping the line. Please try again, and let's see if CI fails.

stevepiercy:
@niccokunzmann I disagree with the docs changes you suggested. Both the parameter was added and the behavior changed in all instances. If you remove the parameter clause, then the admonition must be changed to `versionchanged` because the existing behavior changed.

src/icalendar/attr.py
```diff
@@ -403,14 +403,24 @@ def fdel(self: Component):
...
-        default: The default value
-        doc: The documentation string
-    """
+        default: Required. The default value.
+        doc: Required. The documentation string.
+        prop: Required. The name of the property.
+        min_value: If set, the value must be greater than or equal to this minimum.
+
+    Raises:
+        TypeError: If the value is not an ``int``. Booleans are rejected, too, even though ``bool`` subclasses ``int``.
```
stevepiercy: Per https://github.com/collective/icalendar/pull/1598#discussion_r3734363064.

```suggestion
        TypeError: If the value is not an ``int``. Booleans are rejected, too,
            even though ``bool`` subclasses ``int``.
```

stevepiercy (changes requested):

heyncth:
@stevepiercy @niccokunzmann — thanks for the last round, both are addressed in the commit just pushed:

- **`Raises` formatting** (@stevepiercy): applied your suggestion as written — the `TypeError` field is now wrapped onto two lines, and since the lines fit under 88 characters, the `# noqa: E501` on the closing quotes is gone entirely.
- **Version notes** (@niccokunzmann): applied your wording across all five docstrings. `single_int_property` keeps `versionadded` with "Added the ``min_value`` parameter.", and the SEQUENCE / REPEAT / PRIORITY / `Alarm.REPEAT` docstrings now use `versionchanged` with "Negative values are no longer accepted." — agreeing with @stevepiercy that a behavior change on existing properties warrants `versionchanged`.

Ruff (CI-pinned 0.15.2) and the full attr test suite pass locally.

@niccokunzmann @SashankBhamidi — if your remaining threads from earlier rounds are resolved, marking them resolved would let auto-merge proceed.

stevepiercy:
>     * **Version notes** (@niccokunzmann): applied your wording across all five docstrings. `single_int_property` keeps `versionadded` with "Added the `min_value` parameter.", and the SEQUENCE / REPEAT / PRIORITY / `Alarm.REPEAT` docstrings now use `versionchanged` with "Negative values are no longer accepted." — agreeing with @stevepiercy that a behavior change on existing properties warrants `versionchanged`.

I'm sorry, I think I wasn't clear. In https://github.com/collective/icalendar/pull/1598#issuecomment-5214343641, I disagreed with @niccokunzmann's suggestions and would not have applied them at all, leaving them as is. That's because both the parameter is added and the behavior changes as a consequence of adding the parameter. It's helpful to retain both sentences.

angatha (dismissed):

heyncth:
@stevepiercy thanks for the clarification, my mistake. I've reverted the version notes to the original wording at all five docstrings: `.. versionadded:: 7.2.3` with "Added the ``min_value`` parameter. Negative values are no longer accepted." — both sentences retained, since the parameter is added and the behavior changes as a consequence of adding it. The wrapped `Raises` field from your earlier suggestion is kept as-is. Committed as `da718ee4`.

src/icalendar/attr.py
stevepiercy: This looks good https://icalendar--1598.org.readthedocs.build/en/1598/reference/api/icalendar.attr.html#icalendar.attr.single_int_property

stevepiercy (approved):
I approve as is, but I'd like to give @niccokunzmann the opportunity to respond to [my comment regarding the docstrings](https://github.com/collective/icalendar/pull/1598#issuecomment-5214343641) before merging.

Also he needs to resolve any of his unresolved comments to allow auto-merge.

Thanks for your work to improve icalendar. I appreciate it!

heyncth:
@niccokunzmann , just following up on this PR. The remaining unresolved threads are currently blocking auto-merge. Could you please take a look and resolve them if everything is addressed? It’s been waiting for about 5 days now. Would really appreciate getting this merged when you have a chance. Thanks!

niccokunzmann:
@heyncth Thanks for the ping!

> @niccokunzmann I disagree with the docs changes you suggested. Both the parameter was added and the behavior changed in all instances. If you remove the parameter clause, then the admonition must be changed to `versionchanged` because the existing behavior changed.

Hm. My thought was this.

When I read the docstring, I do not care about how the negative value is implemented, I care about what it means to me.

(1) So in this case https://github.com/collective/icalendar/pull/1598#discussion_r3734231670 as someone who reads the property description, I do not see a min_value parameter anywhere. I wonder what this is for.

(2) In this case https://github.com/collective/icalendar/pull/1598#discussion_r3734208705 - the property constructor - the docstring is wrong. Negative values are still accepted. IF AND ONLY IF the `min_value` parameter is set to 0, then they are not accepted.

(1) documents the usage for someone who reads the property's description. https://icalendar--1598.org.readthedocs.build/en/1598/reference/api/icalendar.cal.event.html#icalendar.cal.event.Event.sequence There, meintioning `min_value` counts for me as technical detail of the implementation. I do not like implementation details in the docs. One could also write something like `Raises TypeError: if the isinstance(value, int) check returns False`. That would be correct but is technical detail of the implementation.

(2) In this case we document the template. It is correct that that parameter was added.

The principle is to hide complexity of how things are done and focus on what is done.

(1) To outline this: In case we rename the parameter, the docstring needs changing but how you use the property stays the same. So why does it need changing then? The `min_value` implementation is not of concern for people who get the value from the property.

Reading your @stevepiercy comment again - yes, the behaviour changed, so it should be `versionchanged` in the case where I removed the `min_value` implementation hint.

So, this is my reasoning here. Let me know if I was clear or if there are more questions.

Thanks for your patience and asking for clarification!

niccokunzmann:
Thinking again, I would expect a `Raises: ` section now for the affected properties. 

But also: We can take that into a different issue and merge - your call!

src/icalendar/cal/alarm.py
```suggestion
        ..  versionchanged:: 7.2.3
            Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -618,7 +634,11 @@ def fdel(self: Component):
         >>> event = calendar.events[0]
         >>> event.sequence
         10
+
+    ..  versionadded:: 7.2.3
+        Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
    ..  versionchanged:: 7.2.3
        Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -1297,7 +1317,11 @@ def _make_repeat_property() -> property:
 Defaults to ``0``, meaning the alarm fires once. Must be paired with
 :attr:`~icalendar.cal.alarm.Alarm.DURATION`. Conforms with :rfc:`5545#section-3.8.6.2`.
 The value is capped at :data:`icalendar.config.MAX_ALARM_REPEAT` on read.
+
+..  versionadded:: 7.2.3
+    Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
..  versionchanged:: 7.2.3
    Negative values are no longer accepted.
```

src/icalendar/attr.py
```diff
@@ -1346,7 +1370,11 @@ def fget(self):
     Within a "VTODO" calendar component, this property specified a
     priority for the to-do.  This property is useful in prioritizing
     multiple action items for a given time period.
+
+    ..  versionadded:: 7.2.3
+        Added the ``min_value`` parameter. Negative values are no longer accepted.
```
```suggestion
    ..  versionchanged:: 7.2.3
        Negative values are no longer accepted.
```

stevepiercy (changes requested):
All the suggestions are now in alignment, using either `versionchanged` for a behavior change or `versionadded` for an implementation detail.

I also agree with @niccokunzmann's suggestion to add a `Raises:` section in the affected parameters. I'd prefer it in this PR, if you're willing to take it on here, else a new issue as a follow up would be good. Your call, of course.

Would you please take care? Thank you!

heyncth:
@stevepiercy @niccokunzmann — thanks for the discussion, both points are addressed in the commit just pushed (`bf871df8`):

- **Constructor** (`single_int_property`): keeps `.. versionadded:: 7.2.3` with "Added the ``min_value`` parameter." — you're right that "Negative values are no longer accepted" was wrong there, since negative values are only rejected when `min_value` is set.
- **The four properties** (SEQUENCE, REPEAT, PRIORITY, `Alarm.REPEAT`): now use `.. versionchanged:: 7.2.3` with "Negative values are no longer accepted." (a behavior change on pre-existing properties), and each gained a `Raises:` section documenting `~icalendar.error.InvalidCalendar` for negative values.

Verified locally: ruff (CI-pinned 0.15.2) and ruff format pass, the 843 attr tests pass, and the docs build is clean of the field-list warning. The wrapped `Raises` field in `single_int_property` is unchanged.

@stevepiercy @niccokunzmann @SashankBhamidi — the threads from the earlier rounds are still open; resolving them would let auto-merge proceed. Thanks for the patience on this one.

src/icalendar/attr.py
```diff
@@ -403,13 +403,24 @@ def fdel(self: Component):
...
-        doc: The documentation string
+        default: Required. The default value.
+        doc: Required. The documentation string.
+        prop: Required. The name of the property.
+        min_value: If set, the value must be greater than or equal to this minimum.
+
+    Raises:
+        TypeError: If the value is not an ``int``. Booleans are rejected, too,
+            even though ``bool`` subclasses ``int``.
+        ~icalendar.error.InvalidCalendar: If the value is smaller than ``min_value``.
```
niccokunzmann: If both errors appear in the other places, it is correct. After all, they only happen in the property setter.


```suggestion
```

Please change the raises sections in the docs of the property. Ready for merge from my side after this!

niccokunzmann (changes requested):
Thanks for the changes!

niccokunzmann:
I think, a follow-up issue should be

- there is a value error if the defsult value is below the min value
- if the getter value is below the min value, it is set to min value - that takes on edge cases that downstream then does not need to worry about

I would not add that to this PR and also invite you @heyncth to be the first one who works on it as you pit a lot of thought into this functionality.

heyncth:
@niccokunzmann thanks for the review — both points are addressed in the commit just pushed (`953835c0`):

- **`single_int_property`**: the `Raises:` section is removed, as you suggested — the factory itself never raises, so it doesn't belong there. The two errors only ever happen in the generated property setter, so they're now documented where the user actually hits them.
- **The property docstrings** (`Event.sequence`, `Event.priority`, `Alarm.REPEAT`, and the module-level `sequence_property`/`priority_property`/`repeat_property`): the `Raises:` sections now list **both** errors — `TypeError` for non-`int` values (booleans included) and `~icalendar.error.InvalidCalendar` for negative values. I separated the two entries with a blank line so the list also parses cleanly under the newer docutils that the 3.14 documentation job uses.

And per your suggestion, I opened **[#1664](https://github.com/collective/icalendar/issues/1664)** to track the getter edge cases — `default` below `min_value`, and stored values below `min_value` being returned unclamped. I'd be happy to be the first to work on it.

Ready for merge from our side — if the remaining threads are resolved on your end, that unblocks auto-merge. Thanks again for the close review!

niccokunzmann (approved):
Thanks! This is fine with me to merge. All concerns addressed. @stevepiercy back to you. I think lots has been addressed. A bit hard to check with the app.

heyncth:
Hi! just mention @stevepiercy that its your turn

stevepiercy (approved):
Thanks for your patience and collaboration. Will auto-merge when green.

PR #1615, opened 2026-07-30
Add attachments property for ATTACH in Events, Alarm, Journal and Todo

### LINKED ISSUE #1613: add attachments property
Alarm and other classes have the ATTACH property.

Please add an attachments getter and setter in the style of the other getters and setters with the correct typing, docs and references.
If it is clear: convert bytes to vBinary and strings to vUri. Only if there is no ambiguity.
Return type is list. A note that append and such do not work must be in the docs.

Avoid long comments with AI. Write a short one yourself. That is worth more. Submit the PR with a reference to the rendered docs of the property in a class.

VedantPatel04:
## Linked issue

- Closes #1613

## Description

Adds an `attachments` getter and setter for the`ATTACH` property in the `Alarm`, `Event`, `Journal`, and `Todo` classes, matching the style of other getters/setters defined.

- Shared `attachments_property` in `attr.py`
- `str` --> `vUri`, `bytes` --> `vBinary` when unambiguous
- Existing `vUri` / `vBinary` values keep their parameters
- Setting `None` or `[]`, or deleting the property, clears attachments
- Changelog: `news/1613.feature`

## Checklist

- [x] I added a change log entry.
- [x] I followed icalendar's Artificial intelligence policy.
- [x] I added  tests in the tests/attr directory.
- [x] I ran and ensured all tests pass locally by following
- [x] I added documentation for my functions through docstrings


niccokunzmann:
Thanks for the PR! Part of it is the research: which values can be expected according to RFC 5545 and subsequent?

VedantPatel04:
Hey nicco,

I was able to dig around the docs regarding the ATTACH property for RFC 5545 and found only vUri and vBinary to be the exact defined acceptable types. Hence I defined str, binary, and vUri, vBinary.

List i included to be able to set many attachments and None to clear attachments, which can be done by passing an empty list as well, but I stuck with it.

Please let me know if changes need to be made to the PR as I'm not sure if you did want me to make changes.

Thanks,
Vedant 
:)

news/1613.feature
```diff
@@ -0,0 +1,8 @@
+Added :attr:`attachments <icalendar.cal.event.Event.attachments>` 
+to access the :rfc:`5545#section-3.8.1.1` ``ATTACH`` property of :class:`~icalendar.cal.event.Event`, 
+:class:`~icalendar.cal.todo.Todo`, 
+:class:`~icalendar.cal.journal.Journal`, 
+and :class:`~icalendar.cal.alarm.Alarm`. 
+
+I used AI to assist me with the changes made in this PR.
+@VedantPatel04
```
niccokunzmann: This must be one line in the file only

src/icalendar/attr.py
```diff
@@ -2563,11 +2564,126 @@ def fdel(self: Component):
...
+        return vUri(value)
+    if isinstance(value, bytes):
+        return vBinary(value)
+    raise TypeError(
+        f"Attachments must be str, bytes, vUri, or vBinary, not {type(value).__name__}."
+    )
+
+
+def _get_attachments(self: Component) -> list[vUri | vBinary]:
+    """ATTACH properties as a list.
```
niccokunzmann: #1244 states that we change to wsy we structure the docs, not to copy the content over, rather help people to use the functionality properly. Can you have a look? Once you have a draft, @stevepiercy can suggest improvements.

src/icalendar/attr.py
```diff
@@ -2563,11 +2564,126 @@ def fdel(self: Component):
...
+    Examples:
+        Attach a URI and inline binary data:
+
+        .. code-block:: pycon
+
+            >>> from icalendar import Event, vUri, vBinary
+            >>> event = Event()
+            >>> event.attachments
+            []
+            >>> event.attachments = "https://example.com/agenda.pdf"
```
```suggestion
            >>> event.attachments = ["https://example.com/agenda.pdf"]
```

src/icalendar/cal/event.py
```diff
@@ -422,6 +423,7 @@ def set_end(
     transparency = transparency_property
     status = status_property
     attendees = attendees_property
+    attachments = attachments_property
```
niccokunzmann, line 431: For all those additions to component classes, please add it to the new methods, too.

src/icalendar/tests/attr/test_attachments.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@pytest.mark.parametrize(
+    "values",
+    [
+        "https://example.com/one.pdf",
+        ["https://example.com/a.pdf", "https://example.com/b.pdf"],
+    ],
+)
+def test_mutating_returned_list_does_not_alter_serialization(component, values):
```
niccokunzmann: Please remove this test. We do not promise this. It just is a note.

niccokunzmann (changes requested):
Hi, thanks for the quick responnse!

The docs need adapting to the new style. Also, we already have tests for the new method in a test file. Could you add test parametrization for the parameter there?

VedantPatel04:
Hey nicco,

Thanks for the reviews, I am currently working on the changes and will get the PR updated as soon as possible.

Thanks,
Vedant

VedantPatel04:
Hey, I have a draft of the updated docstring.

I removed the old non-standard headings that I copied over from the older formatted docstrings. I also had a question regarding the placement of this new docstrings: similar to the example from #1244,  should the docstring be provided as the last arg of the attacfhments_property?

New Draft:

"""This property defines the attachments for a component.

Setting this property replaces all existing attachments. A :class:`str`
is converted to :class:`~icalendar.prop.uri.vUri`, and :class:`bytes` is
converted to :class:`~icalendar.prop.binary.vBinary`. Values that are
already :class:`~icalendar.prop.uri.vUri` or
:class:`~icalendar.prop.binary.vBinary` are stored unchanged, so their
parameters are preserved. Setting ``None`` or an empty list removes all
attachments, as does deleting the property.

This property can be used in icalendar through its Python attributes of:

- :attr:`Alarm.attachments <icalendar.cal.alarm.Alarm.attachments>`
- :attr:`Event.attachments <icalendar.cal.event.Event.attachments>`
- :attr:`Journal.attachments <icalendar.cal.journal.Journal.attachments>`
- :attr:`Todo.attachments <icalendar.cal.todo.Todo.attachments>`

The attachments property is defined by :rfc:`5545#section-3.8.1.1`. It can
be specified multiple times on "VEVENT", "VTODO", and "VJOURNAL"
components. On a "VALARM" component, how many are allowed depends on the
alarm's action: an "EMAIL" alarm may have one or more, an "AUDIO" alarm at
most one, and a "DISPLAY" alarm none.

Note:
    List modifications do not modify the component. Methods such as
    ``append()``, ``extend()``, and ``remove()``, as well as item
    assignment, act on a copy. Assign the list back to the property, or
    use :meth:`Component.add <icalendar.cal.component.Component.add>`
    with a typed value instead.

Parameters:
    attachments(str | bytes | vUri | vBinary | list | None):
        A single attachment, or a list of attachments to set. Accepts
        :class:`str`, :class:`bytes`, :class:`~icalendar.prop.uri.vUri`,
        and :class:`~icalendar.prop.binary.vBinary`, individually or
        mixed together in a list.

Example:
    Attach a URI to an event, then replace it with a URI and inline
    binary data together:

    .. code-block:: pycon

        >>> from icalendar import Event, vUri, vBinary
        >>> event = Event()
        >>> event.attachments
        []
        >>> event.attachments = ["https://example.com/agenda.pdf"]
        >>> print(event.to_ical().decode())
        BEGIN:VEVENT
        ATTACH:https://example.com/agenda.pdf
        END:VEVENT
        >>> event.attachments = [
        ...     vUri(
        ...         "https://example.com/agenda.pdf",
        ...         params={"FMTTYPE": "application/pdf"},
        ...     ),
        ...     vBinary(b"inline-bytes"),
        ... ]
        >>> len(event.attachments)
        2

See also:
    :attr:`Component.links <icalendar.cal.component.Component.links>`
"""

src/icalendar/cal/event.py
VedantPatel04: Hey, I wanted some clarification:

do you mean all new methods that have attachments as a param? Like new_email() in alarm.py

or just the new() methods found in alarm, event, journal and todo files?

niccokunzmann:
(1) The summary is cool!

(2) This should be deleted: It turns up in the docs where needed.

```
This property can be used in icalendar through its Python attributes of:

    :attr:Alarm.attachments <icalendar.cal.alarm.Alarm.attachments>
    :attr:Event.attachments <icalendar.cal.event.Event.attachments>
    :attr:Journal.attachments <icalendar.cal.journal.Journal.attachments>
    :attr:Todo.attachments <icalendar.cal.todo.Todo.attachments>

```

(3) this is last:
```
.. note::

    ... your text ....

.. seealso::

    link to the ATTACH property

```

(4) remove this:

```
The attachments property is defined by :rfc:5545#section-3.8.1.1. It can
be specified multiple times on "VEVENT", "VTODO", and "VJOURNAL"
components. On a "VALARM" component, how many are allowed depends on the
alarm's action: an "EMAIL" alarm may have one or more, an "AUDIO" alarm at
most one, and a "DISPLAY" alarm none.
```

instead, add to the note that it is only allowed once on an audio alarm

(5) This could be inspiring people to add an image:
```
vBinary(b"inline-bytes") -> vBinary(b"image-data", params={"FMTTYPE": "image/png"},),
```

niccokunzmann:
Let me know when you would like me to have a look again :)

VedantPatel04:
> Let me know when you would like me to have a look again :)

Hey Nicco, 

mind taking a look now?

I addressed your latest docstring feedback on my draft.

Much appreciated your thorough input. 

Thanks,
Vedant

VedantPatel04:
Hey @niccokunzmann were you able to review this, I can make edits if you need.
Let me know about it.

Thanks,
Vedant

:)

Solaris-star (commented):
Reviewed the current revision. The shared `attachments_property` plus the parametrized component tests are a good direction, and this is more complete than my closed #1619.

One point to clarify: the property docstring says an audio Alarm can contain at most one attachment, but the shared setter currently allows multiple values on `Alarm` just like Event/Journal/Todo. If the RFC cardinality is normative, should Alarm enforce that in its setter or in the factory path? Otherwise the note may promise a constraint the API does not enforce.

The normalize-before-delete order is good: an invalid value raises without clearing existing attachments.

VedantPatel04:
> Reviewed the current revision. The shared `attachments_property` plus the parametrized component tests are a good direction, and this is more complete than my closed #1619.
> 
> One point to clarify: the property docstring says an audio Alarm can contain at most one attachment, but the shared setter currently allows multiple values on `Alarm` just like Event/Journal/Todo. If the RFC cardinality is normative, should Alarm enforce that in its setter or in the factory path? Otherwise the note may promise a constraint the API does not enforce.
> 
> The normalize-before-delete order is good: an invalid value raises without clearing existing attachments.

Hey,

Thanks for the detailede review Solaris, I'll take a look at as soon as i'm back at my computer. Tomorrow evening.

Thanks,
Vedant
:)

src/icalendar/attr.py
```diff
@@ -2563,11 +2564,121 @@ def fdel(self: Component):
...
+        ...         params={"FMTTYPE": "application/pdf"},
+        ...     ),
+        ...     vBinary(b"image-data", params={"FMTTYPE": "image/png"},),
+        ... ]
+        >>> len(event.attachments)
+        2
+
+.. note::
+
+    An audio alarm can contain at most one attachment.
```
angatha: That is the location @Solaris-star is talnking about. Per [RFC5545 3.6.6](https://datatracker.ietf.org/doc/html/rfc5545#section-3.6.6)

> ; The following is OPTIONAL,
> ; but MUST NOT occur more than once.
> ;
> attach /

> When the action is "AUDIO", the alarm can also include one and
> only one "ATTACH" property, which MUST point to a sound resource,
> which is rendered when the alarm is triggered.

VedantPatel04: Hey, after looking through suggested changes, I wanted to see if my plan is valid: 

1. Fixed the note in my docstring to read: "An ``ACTION:AUDIO`` alarm MUST NOT contain more than one ``ATTACH`` property."

2. override the attachments property in alarm.py with a setter method that raises an error if ACTION == "AUDIO" and more than one attachment is given


I also had a question regarding an potential error that could slip through the cracks:
 I am wondering if I should add a validate() method that checks if an alarm with multiple attachments sets its ACTION to AUDIO AFTER being assigned attachments?
 
 I do not want to pollute alarm.py unnecessarily, but please do let me know if this sounds reasonable to execute.
 
 Thanks,
 Vedant

src/icalendar/cal/alarm.py
```diff
@@ -300,6 +302,18 @@ def triggers(self):
...
 
+    attachments = attachments_property
+
+    @attachments.setter
+    def attachments(self, value: ATTACHMENTS_TYPE_SETTER) -> None:
+        if value is not None and self.ACTION == "AUDIO":
+            count = len(value) if isinstance(value, (list, tuple)) else 1
+            if count > 1:
+                raise InvalidCalendar(
+                    "An AUDIO alarm MUST NOT contain more than one ATTACH property"
```
stevepiercy: Let's use the Python implementation language, and LESS SHOUTING, instead of the RFC copy-paste. Also it's helpful to be more explicit in the error message. This may require updating the test accordingly.

```suggestion
                    "An AUDIO alarm must not contain more than one attachment."
                    f"Alarm has {count} attachments."
```

src/icalendar/cal/alarm.py
```diff
@@ -594,8 +609,8 @@ def new_email(
                 :class:`~icalendar.prop.cal_address.vCalAddress` instances. A
                 single address or a sequence of addresses. At least one is
                 required.
-            attachments: Optional URI or sequence of URIs to attach to the
-                email.
+            attachments: The :attr:`attachments` of the alarm. A single value
+                or a sequence of them. URIs and binary data are both accepted.
```
stevepiercy: Better sentence structure.

```suggestion
                or a sequence of them. Both URIs and binary data are accepted.
```

src/icalendar/tests/attr/test_attachments.py
```diff
@@ -0,0 +1,217 @@
...
+    """Setter raises InvalidCalendar when >1 attachment is given and ACTION is AUDIO."""
+    alarm = Alarm.new_audio(timedelta(minutes=-5))
+    with pytest.raises(InvalidCalendar, match="AUDIO"):
+        alarm.attachments = [
+            "ftp://example.com/sound1.aud",
+            "ftp://example.com/sound2.aud",
+        ]
+
+
+def test_audio_alarm_accepts_single_attachment():
```
stevepiercy, line 194: Should there be another test to ensure that zero attachments is also acceptable?

src/icalendar/tests/attr/test_attachments.py
```diff
@@ -0,0 +1,217 @@
...
+    ical = (
+        b"BEGIN:VEVENT\r\n"
+        b"ATTACH;ENCODING=BASE64;VALUE=BINARY:!!!notbase64!!!\r\n"
+        b"END:VEVENT\r\n"
+    )
+    restored = Event.from_ical(ical)
+    assert len(restored.attachments) == 1
+
+
+def test_audio_alarm_rejects_multiple_attachments():
```
stevepiercy, line 170: I'd like to see a test of the error message, either in this test or a separate one.

stevepiercy (changes requested):
This is really nice work. Thanks for your contribution.

I have a few suggestions and minor tweaks. What do you think? Please let me know. Thank you!

VedantPatel04:
Hey Steve,

Really really appreciate your help in reviewing my changes, I reworked some of my doc string tonality and also decided to take your advice in making a test for that zero-attachment alarm, good eyes :).

Additionally, I added in a new test where now the multiple attachment error message is clearly displayed with the number of attachments passed.

Thanks again for your help,
Vedant

SashankBhamidi:
Hi @VedantPatel04, could you please check your local Git configuration (`user.name` and `user.email`)? Your commits are showing up as ghost instead of being associated with your GitHub account.

Once you've fixed the configuration, please rewrite the existing commits with the correct author information and force push the updated history to this PR.

The AI agent you're using should be able to help with rewriting the commits if you instruct it accordingly. Also, if possible, could you update the news fragment to mention the actual AI model you used? Thank you!

VedantPatel04:
Hey Sashank,

Done!

Thanks,
Vedant

src/icalendar/tests/attr/test_attachments.py
```diff
@@ -0,0 +1,240 @@
...
+    restored = type(component).from_ical(component.to_ical())
+    assert len(restored.attachments) == 2
+    assert isinstance(restored.attachments[0], vUri)
+    assert restored.attachments[0] == "https://example.com/agenda.pdf"
+    assert restored.attachments[0].params.get("FMTTYPE") == "application/pdf"
+    assert isinstance(restored.attachments[1], vBinary)
+    assert restored.attachments[1].bytes == b"\xff\x00data"
+
+
+def test_unexpected_stored_type_is_returned_as_is(component):
```
SashankBhamidi: Nicco asked to remove `test_unexpected_stored_type_is_returned_as_is` in the previous round. It is still present. The docstring note is guidance for implementors, not a tested contract. Having a test for it locks in behavior we may want to change later if we decide to reject or normalize malformed stored values.

src/icalendar/attr.py
```diff
@@ -2566,11 +2567,121 @@ def fdel(self: Component):
 )
 
 
+ATTACHMENTS_TYPE_SETTER: TypeAlias = (
```
SashankBhamidi, line 2665: `ATTACHMENTS_TYPE_SETTER` does not include `tuple`, but the setter accepts tuples via `SEQUENCE_TYPES = (list, tuple)`. A caller passing a tuple of attachments will have it accepted at runtime but rejected by the type checker. Either add `tuple[str | bytes | vUri | vBinary, ...]` to the alias or restrict `_set_attachments` to only accept lists. The rest of the codebase uses `list` as the documented input type for these setters so the simpler fix is to drop tuple support from `_set_attachments` and keep the alias as-is.

src/icalendar/cal/alarm.py
```diff
@@ -300,6 +302,19 @@ def triggers(self):
 
     repeat = repeat_property
 
+    attachments = attachments_property
+
+    @attachments.setter
```
SashankBhamidi, line 347: The AUDIO cardinality guard only checks `self.ACTION` at setter call time. That means `Alarm.new(attachments=[a, b])` bypasses it entirely since `ACTION` is not yet set when `new()` calls the setter. The caller can then set `ACTION:AUDIO` manually and end up with an invalid alarm that passes no guard. The guard should also check inside `new()` if `action` is going to be `"AUDIO"`, or the AUDIO factory should be the only way to create audio alarms. 

As-is the note in the docstring promises enforcement that the API does not fully deliver.

src/icalendar/cal/alarm.py
SashankBhamidi: `new_audio`'s `attach` parameter is typed `str | bytes | None` but since the body now delegates to `alarm.attachments`, it also accepts `vUri` and `vBinary`. The annotation and the docstring description should reflect that:

```suggestion
        attach: str | bytes | vUri | vBinary | None = None,
```

SashankBhamidi (changes requested):
Thanks for the thorough work here Vedant. 

The shared property approach is the right design, the test coverage is solid, and the AUDIO cardinality enforcement is a good addition. 

I found a few things that need addressing before this can get merged.

src/icalendar/cal/alarm.py
SashankBhamidi: Apply this suggestion locally instead of doing it from the UI, looks bugged to me. Thanks!

src/icalendar/cal/alarm.py
niccokunzmann, line 347: I am not sure. Please test that that enforcement is not replacing the original setter. I do not know it a copy is returned or the original property modified.

src/icalendar/cal/alarm.py
```diff
@@ -360,6 +382,9 @@ def new(
             refids=refids,
             concepts=concepts,
         )
+        if action is not None:
```
niccokunzmann, line 436: Please just set the ACTION property.

src/icalendar/cal/alarm.py
```diff
@@ -479,7 +504,7 @@ def new_display(
     def new_audio(
         cls,
         trigger: timedelta | datetime,
-        attach: str | bytes | None = None,
+        attach: str | bytes | vUri | vBinary | None = None,
```
niccokunzmann: This should be consistent across all alarm new methods and passed to alarm.new

src/icalendar/cal/alarm.py
```diff
@@ -553,9 +579,7 @@ def new_audio(
         alarm.add("ACTION", "AUDIO")
         alarm.TRIGGER = trigger
         if attach:
-            alarm.add(
-                "ATTACH", vBinary(attach) if isinstance(attach, bytes) else attach
-            )
+            alarm.attachments = attach
```
niccokunzmann: These should be removed and passed to new_*. That is for all new methods.

Please also add an attachments parameter. The attach parameter - @SashankBhamidi is it released? If not, it can be removed.

src/icalendar/cal/alarm.py
```diff
@@ -639,8 +663,10 @@ def new_email(
         """
         if isinstance(attendees, str):
             attendees = [attendees]
-        if isinstance(attachments, str):
-            attachments = [attachments]
+        if isinstance(attachments, Sequence) and not isinstance(
+            attachments, (str, bytes, list, tuple)
+        ):
+            attachments = list(attachments)
```
niccokunzmann: Can these be removed? The new_* mdthods must not handle this in any way that differs from new. Just pass it to new.

niccokunzmann (changes requested):
Thank you for the changes!

src/icalendar/cal/alarm.py
SashankBhamidi: I did verify this locally... `@attachments.setter` creates a new property object in `Alarm.__dict__` and does not touch the shared `attachments_property` that `Event`, `Todo`, and `Journal` use.

`Event.attachments is Alarm.attachments` is `False`. 

The override works correctly and `test_new_with_audio_action_rejects_multiple_attachments` covers the path.

src/icalendar/cal/alarm.py
SashankBhamidi:
```suggestion
        if action is not None:
            alarm.ACTION = action
```

src/icalendar/cal/alarm.py
SashankBhamidi: This reverted to `self.REPEAT` at some point. `Alarm.REPEAT` (uppercase) has no `min_value` guard,  that lives on the lowercase `repeat` property. 

The result is that `Alarm.new_display(..., repeat=-1)` silently accepts -1. I verified this locally:

```
>>> Alarm.new_display("test", timedelta(minutes=-5), duration=timedelta(minutes=1), repeat=-1).REPEAT
-1
```

```suggestion
            self.repeat = repeat
```

src/icalendar/cal/alarm.py
SashankBhamidi: `attach` was not in v7.2.2 but `new_audio` was added after that release, so it has never shipped. 

Safe to rename to `attachments` and pass it straight to `cls.new(attachments=attachments, ...)`, dropping the separate assignment at line 582.

src/icalendar/cal/alarm.py
SashankBhamidi: @VedantPatel04,

The `isinstance(attachments, Sequence)` coercion block should go. 

Just pass `attachments` to `new()` and let it handle normalisation, the same way `new_display` does.

SashankBhamidi (changes requested):
Thanks for the quick turnaround! 

The four things I flagged earlier are all addressed. 

I ran `Alarm.new(action="AUDIO", attachments=[a, b])` locally and it correctly raises.

A few more things inline.

stevepiercy:
@VedantPatel04 tests are failing, and we don't review failing PRs. Would you please resolve them locally first, then push once you've addressed the failures? See https://icalendar.readthedocs.io/en/stable/contribute/development.html.

VedantPatel04:
Hey Steve,
I looked into the logs, it seems Github jobs were interrupted while the tests were running and I think that is what may have caused them to be cancelled/failed.

I am going to push an empty commit to retrigger the CI and see if that works. I don't see the usual "Re-run jobs" button on the checks either.

 I tested the changes on my local before pushing last time so this is my best guess as to why the jobs failed.

Thanks,
Vedant

VedantPatel04:
Hey,

Is the CI pipeline broken?
I saw a lot of other pull requests were having the same issue with checks.


 :(

Thanks,
Vedant Patel

stevepiercy:
@VedantPatel04 it was GitHub being GitHub's unreliable self again, as usual.

https://www.githubstatus.com/

It's running again, we'll see if it's successful this time.

stevepiercy:
Tests all pass. @niccokunzmann @SashankBhamidi one more review on the changes, please!

src/icalendar/cal/alarm.py
```diff
@@ -377,7 +402,7 @@ def _apply_duration_repeat(
                     "DURATION and REPEAT must be set together or not at all"
                 )
             self.DURATION = duration
-            self.REPEAT = repeat
+            self.repeat = repeat
```
stevepiercy, line 456: Is this correct? This is the second PR I've seen today where AI crept into places it shouldn't and changed this same exact line!

src/icalendar/cal/alarm.py
```diff
@@ -500,11 +525,12 @@ def new_audio(
...
-                ``"ftp://example.com/pub/sounds/bell.aud"`` for a linked
-                sound file, or :class:`bytes` for inline binary audio data
-                (stored as ``VALUE=BINARY``). When ``None`` the client uses
-                its default sound.
+            attachments: Optional audio attachment. Accepts a URI as a
+                :class:`str` or :class:`~icalendar.prop.uri.vUri`, or
+                inline binary audio as :class:`bytes` or
+                :class:`~icalendar.prop.binary.vBinary`
+                (stored as ``VALUE=BINARY``). When ``None`` the client
+                uses its default sound.
```
stevepiercy: Let's drop the RFC copy.

```suggestion
                :class:`~icalendar.prop.binary.vBinary`. When ``None``,
                the client uses its default sound.
```

src/icalendar/attr.py
```diff
@@ -2566,11 +2567,121 @@ def fdel(self: Component):
...
+        ...         params={"FMTTYPE": "application/pdf"},
+        ...     ),
+        ...     vBinary(b"image-data", params={"FMTTYPE": "image/png"},),
+        ... ]
+        >>> len(event.attachments)
+        2
+
+.. note::
+
+    An ``ACTION:AUDIO`` alarm MUST NOT contain more than one ``ATTACH`` property.
```
stevepiercy: This sentence should be the Python implementation, not RFC copy paste, and MUST NOT SHOUT IN ALL CAPS.

```suggestion
    An alarm as an audio action must not contain more than one attachment.
```

stevepiercy (changes requested):
Replace RFC copy with Python implementation, and a question about AI creep. Would you please take care? Thank you!

src/icalendar/cal/alarm.py
VedantPatel04: No, this is my fault. I manually changed this and had it saved locally.

At some point between running inspections on my code, this line has gotten reversed by my AI agent to the incorrect version ... will fix!

VedantPatel04: Looking a little closer @stevepiercy , Sashank requested I make the change, I believe because there was a concern that Alarm.REPEAT has no min_value guard which lives on the repeat property.

Shouldn't this remain as self.REPEAT = repeat?

src/icalendar/cal/alarm.py
```suggestion
        alarm: Alarm = cls.new(
            action="DISPLAY",
            description=description,
            uid=uid,
            links=links,
            related_to=related_to,
            refids=refids,
            concepts=concepts,
        )
```

Now that this is possible, this is cleaner. Strictly not in scope but somehow it expanded to ACTION, too.

src/icalendar/cal/alarm.py
```diff
@@ -552,10 +577,6 @@ def new_audio(
         )
         alarm.add("ACTION", "AUDIO")
```
```suggestion
         alarm: Alarm = cls.new(
            action="AUDIO",
            attachments=attachments,
             uid=uid,
             links=links,
             related_to=related_to,
            refids=refids,
            concepts=concepts,
         )
```

Hm. On mobile, the indentation looks off.

src/icalendar/cal/alarm.py
```diff
@@ -661,9 +681,6 @@ def new_email(
         )
         alarm.add("ACTION", "EMAIL")
```
```suggestion
         alarm: Alarm = cls.new(
            action="EMAIL",
            attachments=attachments,
             summary=summary,
             description=description,
             uid=uid,
            attendees=attendees,
            links=links,
            related_to=related_to,
            refids=refids,
            concepts=concepts,
         )
```

niccokunzmann (changes requested):
Thd only think is passing action to new(). Then, it is perfect for me and ready to merge.

src/icalendar/cal/alarm.py
stevepiercy: Looks OK on desktop.

src/icalendar/cal/alarm.py
stevepiercy: Looks OK on desktop.

src/icalendar/cal/alarm.py
stevepiercy: This feels like a footgun and a bug. We don't document this discrepancy.

- https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.alarm.html#icalendar.cal.alarm.Alarm.REPEAT
- https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.alarm.html#icalendar.cal.alarm.Alarm.repeat

```pycon
>>> from icalendar import Alarm
>>> from datetime import timedelta
>>> Alarm.new_display("test", timedelta(minutes=-5), duration=timedelta(minutes=1), repeat=-1).REPEAT
-1
>>> Alarm.new_display("test", timedelta(minutes=-5), duration=timedelta(minutes=1), repeat=-1).repeat
0
```

As a design decision, @VedantPatel04 @SashankBhamidi @niccokunzmann do you think we should add the guard to `REPEAT`? I also wonder if `REPEAT` is capped, or should be, as well.

https://icalendar.readthedocs.io/en/stable/explanation/api-design.html#property-naming-convention isn't really clear on validation, but maybe "calculation" implies "validation". What do y'all think?

VedantPatel04: Hey,
I think adding a floor-guard to REPEAT makes sense here because it establishes consistency in matching the repeat() property and makes documentation simpler.

Adding a cap to REPEAT also doesn't seem bad at first, but I wonder how it affects the logic of overriding RFC's generous integer limits?

I think with good documentation, the cap is a good measure but without, not so much. 

Let me know what you think.

Thanks,
Vedant

stevepiercy (dismissed):
I approve this PR, pending:

- @niccokunzmann resolve his unresolved suggestions.
- A decision from maintainers on how to proceed from this discussion: https://github.com/collective/icalendar/pull/1615#discussion_r3732808135.

Thank you for your work on this @VedantPatel04. This is a nice improvement to icalendar.

Solaris-star (approved):
Re-reviewed the current head after the AUDIO-cardinality changes. My earlier concern is addressed: `Alarm.new()` now sets `ACTION` before assigning attachments, and the Alarm-specific setter rejects multiple attachments for AUDIO while Event/Journal/Todo retain the shared setter. I also verified the targeted attachment, alarm-factory, and component-`new()` tests locally: 409 passed. The remaining `REPEAT` discussion is separate from the attachment API in this PR.

src/icalendar/cal/alarm.py
```diff
@@ -300,6 +302,19 @@ def triggers(self):
 
     repeat = repeat_property
 
+    attachments = attachments_property
+
+    @attachments.setter
+    def attachments(self, value: ATTACHMENTS_TYPE_SETTER) -> None:
+        if value is not None and self.ACTION == "AUDIO":
+            count = len(value) if isinstance(value, list) else 1
+            if count > 1:
```
angatha, line 351:
```suggestion
            if isinstance(value, list) and value:
```

src/icalendar/cal/alarm.py
```diff
@@ -300,6 +302,19 @@ def triggers(self):
...
+                raise InvalidCalendar(
+                    "An AUDIO alarm must not contain more than one attachment.\n"
+                    f"Alarm has {count} attachments."
+                )
```
angatha, line 355: This is against some coding principle. Super classes should not contain logic for child classes. You can override properties and like this:

```python
class A:
    def __init__(self):
        self._a: int = 1

    @property
    def a(self) -> int:
        return self._a

    @a.setter
    def a(self, value: int) -> None:
        self._a = value


class B(A):

    @A.a.setter
    def a(self, value: int) -> None:
        if value < 0:
            raise ValueError
        A.a.fset(self, value)  # Calls original setter like super()


class C(B):

    @B.a.setter
    def a(self, value: int) -> None:
        if value > 10:
            raise ValueError
        B.a.fset(self, value)  # Calls setter in B super()


if __name__ == '__main__':
    A().a = 0     # fine
    A().a = -1    # fine
    A().a = 10    # fine
    A().a = 11    # fine

    B().a = 0     # fine
    B().a = -1  # bad
    # B().a = 10    # fine
    B().a = 11    # fine

    C().a = 0     # fine
    # C().a = -1  # bad
    C().a = 10    # fine
    # C().a = 11  # bad
```

angatha (changes requested):

src/icalendar/cal/alarm.py
VedantPatel04: Hey Angatha,
I wanted to be clear that the concern you are raising here is Alarm owning AUDIO specific logic correct?

I would appreciate any guidance as to how I can fix this issue without uprooting the functionality I have implemented so far because I am not able to see a clear solution here.

Any ideas?

Thanks,
Vedant

VedantPatel04: @niccokunzmann any suggestions?

Thanks,
Vedant

src/icalendar/cal/alarm.py
VedantPatel04: Wouldn't this change reject a one-element list because it checks for a non-empty list and we are trying to screen against lists with more than 1 element?

angatha: Oh sorry, you are right. I must have read `>=`

src/icalendar/cal/alarm.py
angatha: Somehow I thought that there was a subclass for AUDIO. Wasn't my best day yesterday.. Sorry again.

angatha (dismissed):
@VedantPatel04 I'm sorry for the disturbance. The solution is the best one I could see. I somehow thought that there where a subclass that can overwrite the setter.

niccokunzmann (dismissed):
Thanks so much for sticking with it and all the feedback. This is ready for merge!

niccokunzmann:
Could you resolve the merge conflicts and ping me?

niccokunzmann:
Thanks @Solaris-star for your review. I hope to see you around helping out.🙏🏽

stevepiercy:
Ping @niccokunzmann @SashankBhamidi @angatha for feedback about how to proceed on this point:

> I approve this PR, pending:
> 
>     * A decision from maintainers on how to proceed from this discussion: [#1615 (comment)](https://github.com/collective/icalendar/pull/1615#discussion_r3732808135).

src/icalendar/cal/alarm.py
niccokunzmann: It is always best to use the lowercase properties where we can. Using repeat over REPEAT here is fine.

Especially new() methods should validate input. That is not really supported by the uppercase properties.

stevepiercy: Can you elaborate why we should use the lowercase properties? Or maybe why we should avoid the uppercase properties? It's not clear in the API Design documentation.

niccokunzmann: Uppercase are basicslly straight setters and getters witha a bit of conversion. Like add() but with a little more safety and typing. Uppercase properties work independent of each other.

Lowercase can raise errors based on wrong values. They can validate and also depend on several uppercase ones in their calculation. They can have additional, non-RFC restrictions for security, see repeat.

As such adding validation in new() around an uppercase property indicates that this functionality should be moved into a lowercase property. Also, new() should validate. It is easier to do when we use lowercase properties in new() as that is already their scope. With that, we have a consistent validation living inside of the properties and no duplication of code.


So:
- add() - full flexibility, no restrictions
- Uppercase - typing, conversion, increased compatibility (e.g. convert to UTC), **RFC interface**
- lowercase - uses uppercase where possible, this is the **user friendly interface**, typing, validation, conversion, nice names (attachments vs. ATTACH), hand holding for ease

The purpose of new() is also to validate what the lowercase ones cannot. E.g. duration and repeat must occur together. That is hard to do in a property.

If you find that should be somehow documented a bit better, I agree.

stevepiercy: That's the clearest description yet. Thank you! I'll create an issue, using this as a basis, and self-assign, as this needs a little bit of wordsmithing of the current documentation.

stevepiercy: Resolving. Issue created at https://github.com/collective/icalendar/issues/1662. This can now be merged, once @VedantPatel04 resolves the merge conflicts, and we check those changes.

VedantPatel04:
Hey, I'll get this branch updated soon, will ping you when I do it.

Thanks,
Vedant

VedantPatel04:
Hey nicco(@niccokunzmann ),

I resolved any merge conflicts and re-committed.Please take a look before merging my PR.

It was fun working on this with you all , hopefully I get the chance to contribute again in the future. 

:)

Thanks,
Vedant Patel

niccokunzmann:
Hi, thanks! You are very welcome!
I would like to see you here again. It does not always have to be a PR. We value your review also and have a process to guide you to the various different ways in which you can be active here! Check it out and you are welcome to join where you like:
https://icalendar.readthedocs.io/en/latest/contribute/index.html

niccokunzmann (dismissed):
Thanks! Ready for merge from.my perspective.

@SashankBhamidi your review is pending.

src/icalendar/tests/attr/test_attachments.py
```diff
@@ -0,0 +1,236 @@
...
+    restored = type(component).from_ical(component.to_ical())
+    assert len(restored.attachments) == 2
+    assert isinstance(restored.attachments[0], vUri)
+    assert restored.attachments[0] == "https://example.com/agenda.pdf"
+    assert restored.attachments[0].params.get("FMTTYPE") == "application/pdf"
+    assert isinstance(restored.attachments[1], vBinary)
+    assert restored.attachments[1].bytes == b"\xff\x00data"
+
+
+def test_broken_attach_value_does_not_raise():
```
SashankBhamidi: `test_broken_attach_value_does_not_raise` is still here. This tests `ignore_exceptions` parsing behaviour on `Event`, which is a property of the parser, not of the `attachments` property. It would pass whether `attachments` is implemented or not. Please remove it.

src/icalendar/cal/alarm.py
SashankBhamidi: The guard fires correctly when `action="AUDIO"` is passed to `Alarm.new()`, because `alarm.ACTION = action` runs at line 385 before `alarm.attachments = attachments` at line 386. But when no action is given at construction time, `self.ACTION` is an empty string, the guard never fires, and the alarm is created with however many attachments you pass. A caller can then set `ACTION="AUDIO"` afterward:

```python
alarm = Alarm.new(attachments=["ftp://a.com/a.aud", "ftp://b.com/b.aud"])
alarm.ACTION = "AUDIO"
alarm.to_ical()  # serializes silently with two ATTACHes, no error
```

I ran this locally and confirmed it on the current branch. The setter is the right defence against post-construction assignment, but it needs a matching check in `new()`: before calling `alarm.attachments = attachments`, check if `action == "AUDIO"` and the list has more than one item, and raise `InvalidCalendar` before writing anything.

src/icalendar/cal/alarm.py
SashankBhamidi: Nicco confirmed in thread that lowercase is preferred ("Especially new() methods should validate input"). The code at line 404 is still `self.REPEAT = repeat`. `REPEAT` bypasses `_clamp_repeat`, so a negative value is stored as-is rather than clamped to 0:

```
>>> Alarm.new_display("test", timedelta(minutes=-5), duration=timedelta(minutes=1), repeat=-1).REPEAT
-1
```

```suggestion
            self.repeat = repeat
```

SashankBhamidi (changes requested):
Two bugs still reproducible on the current branch, and one test Nicco asked to remove that is still present. Notes inline.

Thank you!

VedantPatel04:
Hey Sashank (@SashankBhamidi ),

Done! Can you review it for me and see if there is anything else I need to change?

Thanks,
Vedant

SemTiOne (approved):

SashankBhamidi:
@VedantPatel04 I will review this PR by EOW. 
Could you resolve the merge conflicts in the mean time and ping me once done?

VedantPatel04:
> @VedantPatel04 I will review this PR by EOW. Could you resolve the merge conflicts in the mean time and ping me once done?

Hey @SashankBhamidi,

I have resolved all merge conflicts present before. 

Ready for review!



Thank you,
Vedant Patel

VedantPatel04:
Hey @SashankBhamidi , 
Just wanted to see if you can review these changes again :) .

Thank you,
Vedant

src/icalendar/cal/alarm.py
SashankBhamidi: The fix at line 424 guards the `action="AUDIO"` path correctly. But `Alarm.new(attachments=[a, b])` with no `action` argument still bypasses it, because `action == "AUDIO"` is `False` when `action` is `None`. A caller can then set `ACTION = "AUDIO"` after construction and hold an invalid alarm with no error raised anywhere:

I ran this on the current branch and confirmed it. The setter at line 347 is already the right place to catch post-construction assignment; it checks `self.ACTION == "AUDIO"` at that point. The gap is that `new()` sets ACTION at line 429 *after* the guard at line 424. Swapping the order; set ACTION first, then call `alarm.attachments = attachments`, would let the setter's own guard handle both paths without duplicating the check in `new()`.

SashankBhamidi (changes requested):
The REPEAT fix is actually better than what I asked for; adding `min_value=0` to the uppercase `REPEAT` property directly means both the uppercase and lowercase interfaces now validate, not just the lowercase one. The test removal is clean, and the delegation to `cls.new()` across all three factory methods is exactly what Nicco was after. 

One thing I can still reproduce inline.

VedantPatel04:
hey @SashankBhamidi ,

I see what you are saying regarding the duplicate check. I can delete the check from lines 424–428. Since`new()` already sets `ACTION` before `attachments`,  `Alarm.new(action="AUDIO", attachments=[...])` still goes through the `attachments` setter.

Regarding the block you ran: 

```python
alarm = Alarm.new(attachments=["ftp://a.aud", "ftp://b.aud"])
alarm.ACTION = "AUDIO"
``` 

Does the second assignment even go through the `attachments` setter?

I am not sure how to resolve this.
I am willing to accept help from you on this PR if you have clear solution in mind :) .

Thank you,
Vedant

SashankBhamidi:
To answer your question: no, `alarm.ACTION = "AUDIO"` doesn't go through the attachments setter; ACTION and attachments are independent properties. That's the bypass.

Fix is to override the ACTION setter the same way you overrode attachments:

```python
@ACTION.setter
def ACTION(self, value: str) -> None:
    if value == "AUDIO" and len(self.attachments) > 1:
        raise InvalidCalendar(
            "An AUDIO alarm must not contain more than one attachment.\n"
            f"Alarm has {len(self.attachments)} attachments."
        )
    self.pop("ACTION", None)
    if value is not None:
        self.add("ACTION", value)
```

And a test for this path:

```python
def test_setting_audio_action_on_alarm_with_multiple_attachments_raises():
    alarm = Alarm.new(attachments=["ftp://a.com/a.aud", "ftp://b.com/b.aud"])
    with pytest.raises(InvalidCalendar, match="must not contain more than one attachment"):
        alarm.ACTION = "AUDIO"
```

stevepiercy:
@VedantPatel04 would you please update your branch to `main`? There are merge conflicts, and reviews are stopped until they're resolved. Thank you!

VedantPatel04:
Resolved merge conflicts and added fix suggested by Sashank to alarm.py.

Thank you,
Vedant

VedantPatel04:
Fixed failing test.

Thank you,
Vedant Patel

SashankBhamidi (approved):
Thanks for sticking with this through so many rounds @VedantPatel04.

Everything is fixed. Good work getting through all of this.

PR #1619, opened 2026-07-31
feat(alarm): add attachments getter/setter/deleter for ATTACH

### LINKED ISSUE #1613: add attachments property

Solaris-star:
Adds an `attachments` property to `Alarm` following the `attendees` pattern.

- **getter**: returns a list copy of ATTACH values (`vUri` for URIs, `vBinary` for bytes). Mutating the returned list does not affect the alarm.
- **setter**: accepts a single value or a list. `bytes` are wrapped in `vBinary`, strings stored as-is. Replaces all existing ATTACH values.
- **deleter**: removes all ATTACH values.

Rendered docs: `Alarm.attachments` — see :rfc:`5545#section-3.8.1.1`.

8 tests covering empty, URI, binary, list, replace, None-clear, delete, and copy semantics.

Closes #1613

### DIFF
--- a/src/icalendar/cal/alarm.py
+++ b/src/icalendar/cal/alarm.py
@@ -207,6 +207,68 @@ class Alarm(Component):
     )
 
     @property
+    def attachments(self) -> list:
+        """The ATTACH property values of this alarm.
+
+        Returns a list of :class:`~icalendar.prop.vUri` (for URI
+        attachments) or :class:`~icalendar.prop.vBinary` (for inline
+        binary data).  The list is a copy — mutating it does **not**
+        change the alarm; use the setter to replace all attachments at
+        once.
+
+        See :rfc:`5545#section-3.8.1.1`.
+
+        Example:
+
+            >>> from icalendar import Alarm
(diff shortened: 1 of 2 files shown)


niccokunzmann:
Hi, we already have #1615 going on. Could you have a look? You could add comments for improvements or ask to collaborate or suggest changes.

Solaris-star:
Yes, I'll review #1615 and leave focused comments there. I won't open a duplicate PR.

SashankBhamidi:
Closing as discussed, #1615 covers the same ground and is further along. Thanks for looking at this and for jumping in on the #1615 review instead.

PR #1631, opened 2026-08-04
docs: add type hints to __init__ methods

ShirayukiRin:
## Linked issue

- See  #938 

## Description

Added type hints to multiple `__init__` methods.

## Checklist

## Additional information

One test failed, and it seems like it was due to a 403 Client Error when trying to connect to 
`https://stackoverflow.com/questions/79171631/how-do-i-determine-whether-a-zoneinfo-is-an-alias/79171734#79171734`

related log:
```log
(reference/api/icalendar.timezone.equivalent_timezone_ids: line   10) broken    https://stackoverflow.com/questions/79171631/how-do-i-determine-whether-a-zoneinfo-is-an-alias/79171734#79171734 - 403 Client Error: Forbidden for url: https://stackoverflow.com/questions/79171631/how-do-i-determine-whether-a-zoneinfo-is-an-alias/79171734


docs: FAIL code 2 (573.15=setup[5.42]+cmd[567.73] seconds)
```


src/icalendar/prop/dt/types.py
```diff
@@ -41,7 +41,7 @@ class vDDDTypes(TimeBase):
     params: Parameters
     dt: DT_TYPE
 
-    def __init__(self, dt, params: dict[str, Any] | None = None):
+    def __init__(self, dt, params: dict[str, Any] | None = None) -> None:
```
lcampanella98: it's cheap to add annotation for `dt` as long as we're touching the signature, and `DT_TYPE` is already defined
```suggestion
    def __init__(self, dt: DT_TYPE, params: dict[str, Any] | None = None) -> None:
```

src/icalendar/tests/conftest.py
```diff
@@ -62,7 +62,7 @@ class DataSource:
 
     def __init__(
         self, data_source_folder: Path, parser: type[Component], multiple=False
```
lcampanella98: we can add annotation on `multiple` as long as we're already touching the signature
```suggestion
        self, data_source_folder: Path, parser: type[Component], multiple: bool = False
```

src/icalendar/prop/dt/list.py
```diff
@@ -18,7 +18,7 @@ class vDDDLists:
     params: Parameters
     dts: list[vDDDTypes]
 
-    def __init__(self, dt_list, params: dict[str, Any] | None = None):
+    def __init__(self, dt_list, params: dict[str, Any] | None = None) -> None:
```
lcampanella98, line 21: `dt_list` is without annotation, but adding one would be awkward as it can accept both a single value and an iterable of values. I'd suggest leaving as-is, but curious what others think

src/icalendar/cal/component.py
lcampanella98: `mypy` reported 1 new error which traces to here. The cause of the error is now that the signature contains annotations, the body of the `__init__` method is now type-checked as well, and the `errors` field has no annotation. 

There are 2 call-sites of appending to `errors`

- [`component.errors.append((None, str(exception)))`](https://github.com/collective/icalendar/blob/622dff599439dfc4e0f6d185fd9d57a8bb41a692/src/icalendar/parser/ical/component.py#L96)
- [`self.component.errors.append((name, str(exception)))`](https://github.com/collective/icalendar/blob/622dff599439dfc4e0f6d185fd9d57a8bb41a692/src/icalendar/parser/ical/component.py#L262)

so I think we can annotate it:

```python
list[tuple[str | None, str]]
```

Not sure if it's a blocker, but worth fixing in my view.

src/icalendar/prop/dt/datetime.py
```diff
@@ -92,7 +92,7 @@ class vDatetime(TimeBase):
     default_value: ClassVar[str] = "DATE-TIME"
     params: Parameters
 
-    def __init__(self, dt, /, params: dict[str, Any] | None = None):
+    def __init__(self, dt: datetime, /, params: dict[str, Any] | None = None) -> None:
```
lcampanella98, line 95: this looks right to me. After looking at this, I believe I found a pre-existing bug, where `vPeriod` accepts `date` values but then attempts to construct `vDatetime`s from it which results in raising `AttributeError`. Filed #1633 to track it

news/938.documentation.2
```diff
@@ -0,0 +1 @@
+Added type annotations to __init__ methods. @ShirayukiRin
```
lcampanella98:
```suggestion
Added type annotations to ``__init__`` methods across the package. @ShirayukiRin
```

lcampanella98 (changes requested):
Nice PR. this covers a lot of ground for #938, and the two remaining spots I noticed are small.

`src/icalendar/tests/test_issue_1050_lazy_parsing.py:32`
`src/icalendar/tests/test_issue_798_property_parameters.py:28`

Overall this looks good to me

src/icalendar/prop/dt/list.py
stevepiercy: See the third paragraph onward in https://icalendar.readthedocs.io/en/stable/contribute/development.html#type-hints. There are several custom types in icalendar that the standard library doesn't offer.

stevepiercy (commented):
@lcampanella98's review is thorough. Once his comments are addressed, I'll approve.

ShirayukiRin:
Thank you for your review!!

I can't believe I missed so many parameters. :(
I'll fix them later.

Thanks again :)

src/icalendar/prop/dt/list.py
ShirayukiRin, line 29: I tried to add type hints for both an iterable and a single value to `dt_list`, but PyLance flagged this line as a type error because an iterable cannot be assigned to a variable typed as a single value. 

Do we have a better approach here?

stevepiercy: What did you try? Please show what you tried, and maybe we can improve it together.

Did you see my [earlier comment](https://github.com/collective/icalendar/pull/1631#discussion_r3715318667)?

SashankBhamidi: Leave `dt_list` untyped for now. The `(dt, None)` tuple-as-single-value pattern at line 25 makes it genuinely hard to express cleanly — any union broad enough to cover it overlaps with the tuple cases in `DT_TYPE` itself. Steve's pointer to the [type hints docs](https://icalendar.readthedocs.io/en/stable/contribute/development.html#type-hints) is worth reading for context, but the right fix here is probably a separate issue rather than something to block this PR on. lcampanella98 already landed on the same conclusion.

ShirayukiRin: > What did you try? Please show what you tried, and maybe we can improve it together.
> 
> Did you see my [earlier comment](https://github.com/collective/icalendar/pull/1631#discussion_r3715318667)?

Yeah I saw your comment and thanks for the link to the docs!
I tried many different kinds of type hints. The last one I tried was 
```py
DT_TYPE | vDDDTypes | tuple[DT_TYPE, None] | Iterable[DT_TYPE | vDDDTypes | tuple[DT_TYPE, None]]
```

And I found out that different static type checking tools give you different results.
MyPy did't flag the line 29, it flagged the line 32 as you cannot iterate a single value variable.
```py
tzid = None
for dt_l in dt_list: # <-- here
    dt = vDDDTypes(dt_l) if not isinstance(dt_l, vDDDTypes) else dt_l
```

ShirayukiRin: I think I will leave it untyped for now as lcampanella98 and SashankBhamidi suggested!

lcampanella98 (approved):
thanks for the changes, looks good to me, approving

src/icalendar/prop/dt/list.py
stevepiercy: Sounds reasonable. It is definitely a challenge. I edited the description in https://github.com/collective/icalendar/issues/938 to see if anyone wants to tackle it.

stevepiercy:
@lcampanella98 thanks for the approving review. I went ahead and resolved your unresolved comments, as they block auto-merge of the PR. In the future, I'd prefer that you resolve your unresolved comments when you approve the PR.

Will auto-merge when CI passes. Thank you for your work on this!

stevepiercy (approved):

lcampanella98:
Got it, thanks for the feedback, I'll resolve my comments before approving in the future

PR #1632, opened 2026-08-04
feat: Add example() to Journal and FreeBusy

Souptik96:
## Linked issue

- See #743

This addresses the **Journal** and **FreeBusy** items of the checklist. The remaining unchecked items (the class docstring examples for Event, Calendar, and Timezone) are untouched, so the issue stays open.

## Description

`Journal` and `FreeBusy` were the only two components in `icalendar.cal` without an `example()` classmethod. Every other component already has one: `alarm`, `availability`, `available`, `calendar`, `event`, `timezone`, and `todo`.

This completes all four checklist items for both components:

| Item | Journal | FreeBusy |
| --- | --- | --- |
| `example.ics` | `src/icalendar/tests/journals/example.ics` | `src/icalendar/tests/freebusy/example.ics` |
| `example()` method with default | ✅ | ✅ |
| Usage in the class docstring | ✅ | ✅ |
| Test in `test_examples.py` | ✅ | ✅ |

The implementation mirrors `Todo.example()` exactly, including the `get_example()` import placement and the method's position just before `__all__`:

```python
@classmethod
def example(cls, name: str = "example") -> Journal:
    """Return the journal example with the given name."""
    return cls.from_ical(get_example("journals", name))
```

**Example data provenance.** Both `.ics` files come from [RFC 5545](https://www.rfc-editor.org/rfc/rfc5545) itself, the same source as the existing `todos/example.ics` (which is the Section 3.6.2 VTODO example verbatim):

- `journals/example.ics` is the Section 3.6.3 journal example. Its `DESCRIPTION` is shortened to a single sentence so that the docstring example stays under the 75-octet fold and remains readable. Every other property is verbatim.
- `freebusy/example.ics` is the Section 3.6.4 free/busy example verbatim, including all three `FREEBUSY` periods.

**Test wiring.** `conftest.py` gains `journals` and `freebusy` `DataSource` fixtures so the new examples are validated by the existing `test_component_has_examples` in exactly the same way as the others, rather than by a bespoke test. Both components are also added to `test_default_example`.

The expected output in the docstrings was **generated by running the code**, not written by hand, so the property ordering and the `DTSTART;VALUE=DATE` form are what `to_ical()` actually emits.

## Checklist

## Additional information

**AI use.** Disclosed in the commit message with the model and a summary of how it was used, and in the change log entry, per the Responsible AI use policy. I opened no new issue because #743 already exists and invites exactly this work. The example data is from RFC 5545 rather than model-generated, so there is no question of reconstructed third-party content.

**Verification.**

- `pytest src/icalendar/tests/test_examples.py src/icalendar/tests/test_docstrings.py`: **1745 → 1755 passed**, 541 xfailed, 0 failed. The 10 new cases are the 2 new components × 2 `calendars` params × 2 `tzp` params in `test_component_has_examples`, plus 2 in `test_default_example`.
- The docstring examples really execute: `test_with_doctest.py::test_docstring_of_python_file[icalendar.cal.journal]` and `[icalendar.cal.free_busy]` both pass.
- **Negative control:** reverting only `journal.py` and `free_busy.py`, while keeping the tests and fixtures, gives 10 failed / 52 passed, and the 10 failures are precisely the new cases failing with `AttributeError: type object 'Journal' has no attribute 'example'`. Restoring them returns 62 passed. So the tests genuinely exercise the new code rather than passing vacuously.
- `ruff format --check` and `ruff check` are clean on all four modified Python files (ruff 0.15.2, the pinned version).
- `git ls-files --eol` shows both new `.ics` files stored as `i/lf`, matching `alarms/example.ics` and `todos/example.ics`.

**One disclosure about the checklist item "all tests pass locally".** On my machine (Windows, Python 3.13.4) the full suite has **1203 pre-existing failures and 5 errors that are unrelated to this change**, and the counts are byte-for-byte identical before and after it: 1203 failed / 16671 passed before, 1203 failed / **16681** passed after — the only delta is my 10 new passing cases. The pre-existing failures fall into three environmental groups:

1. `ValueError: the environment variable is longer than 32767 characters` in `test_parser_quadratic_unfolding.py` — a Windows environment-variable size limit, which also produces the 5 errors and the "previous item was not torn down properly" cascade.
2. `assert 'line1\r\nline2' == 'line1\nline2'` — `core.autocrlf=true` rewriting `.ics` fixtures on checkout.
3. dateutil timezone identification, for example `assert None in ('America/Noronha', 'Brazil/DeNoronha')` and `{'EST'} == {'America/New_York'}`.

I mention this rather than quietly ticking the box. If any of these are news to you I am happy to open a separate issue; I did not touch them here because they are out of scope for #743.


Souptik96:
Responding to the automated quality-check comment, since the flag is accurate and I would rather address it directly than leave it ambiguous.

I did use AI to draft this, and it is disclosed rather than hidden: the model and a summary of how I used it are in the commit message, and there is a brief disclosure in `news/743.feature`, per [Responsible AI use](https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use). I did not open a new issue because #743 already exists and explicitly invites this work.

On the two points the policy asks me to stand behind:

- **The example data is not model-generated.** `journals/example.ics` and `freebusy/example.ics` are the RFC 5545 §3.6.3 and §3.6.4 examples, the same provenance as the existing `todos/example.ics`, which is the §3.6.2 VTODO example verbatim. The only edit is shortening the journal `DESCRIPTION` to one sentence so the docstring example stays under the 75-octet fold.
- **The expected output in the docstrings was generated by running the code**, not written by hand, which is why the properties appear in emitted order and `DTSTART` carries `;VALUE=DATE`.

To make review cheap, the claim that these tests actually exercise the new code is checkable: revert only `journal.py` and `free_busy.py`, keep the tests and fixtures, and you get 10 failures, all `AttributeError: type object 'Journal' has no attribute 'example'`. Restore them and it is 62 passed. Full details are in the PR description.

Happy to split this into two PRs (Journal, then FreeBusy), drop the `conftest.py` fixtures in favour of a narrower test, or adjust the example data if you would prefer something other than the RFC samples. I am also glad to explain any part of the change.

angatha:
Code looks good.

Your PR and commit are quite long. That was the main cause for the AI flagging.

src/icalendar/cal/free_busy.py
```diff
@@ -156,5 +176,10 @@ def new(
             cls._validate_start_and_end(start, end)
         return free_busy
 
+    @classmethod
+    def example(cls, name: str = "example") -> FreeBusy:
+        """Return the free/busy example with the given name."""
```
stevepiercy:
```suggestion
        """Return the FreeBusy example with the given name."""
```

src/icalendar/tests/test_examples.py
```diff
@@ -49,10 +59,21 @@ def test_creating_calendar_with_unicode_fields(calendars, utc):
         (Event, "event_with_rsvp"),
         (Timezone, "pacific_fiji"),
         (Todo, "example"),
+        (Journal, "example"),
+        (FreeBusy, "example"),
```
stevepiercy: Please sort these in the list.

src/icalendar/tests/test_examples.py
```diff
@@ -49,10 +59,21 @@ def test_creating_calendar_with_unicode_fields(calendars, utc):
...
+    tzp,
+    calendars,
+    timezones,
+    events,
+    todos,
+    component,
+    example,
+    alarms,
+    journals,
+    freebusy,
```
stevepiercy: Please alphabetically sort these.

src/icalendar/tests/test_examples.py
```diff
@@ -75,7 +98,18 @@ def test_invalid_examples_lists_the_others():
...
+    [
+        Calendar,
+        Event,
+        Timezone,
+        Available,
+        Availability,
+        Todo,
+        Alarm,
+        Journal,
+        FreeBusy,
```
stevepiercy: Please sort these in the list.

stevepiercy (changes requested):
This is really nice work. Thank you for your contribution!

I made a few minor suggestions, one for a proper reference and three for code legibility. Would you please take care? Thank you!

stevepiercy (approved):

Issue #1620, opened 2026-07-31
Merge ``__init__`` or ``__new__`` docstrings into their class's docstring


stevepiercy:
Sphinx will automatically append the docstrings of a class's ``__init__`` or ``__new__`` methods to the class's docstring. This may render the documentation in a confusing manner to the reader.

Look for docstrings in a class's ``__init__`` or ``__new__`` methods, then move and merge their content into the class's docstring, and remove the ``__init__`` or ``__new__`` docstrings.

To close this issue:

- [ ] Write a test that ensures all ``__init__`` and ``__new__`` methods have no docstring.
- [ ] Configure [autoclass_content](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content) to `class` after all ``__init__`` and ``__new__`` methods docstrings have been migrated into their class.
- [ ] Revise the text from https://github.com/collective/icalendar/pull/1621 accordingly.
- [ ] https://github.com/collective/icalendar/blob/main/src/icalendar/prop/binary.py#L13-L32 could be improved.

See:

- https://github.com/collective/icalendar/issues/1118
- https://github.com/collective/icalendar/pull/1617#pullrequestreview-4828002594
- https://github.com/collective/icalendar/issues/1473
- https://github.com/collective/icalendar/issues/1244

niccokunzmann:
How is it with sorting of sections? Where do examples go?

stevepiercy:
Sorting of sections is already covered. Examples are one type of section. https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure

Solaris-star:
I'll take this. Will start with the test that asserts no __init__/__new__ has a docstring, then migrate them one by one.

niccokunzmann:
@Solaris-star We usually do not assign issues. So, please check of there is an open PR and help move that along. There is no problem with opening a PR if there is a bad one but if you have not checked it, it is better to check it first.

While you check it, please give it a review!

Solaris-star:
I found the existing [PR #1622](https://github.com/collective/icalendar/pull/1622), so I won’t open a duplicate. It currently has merge conflicts and is waiting on the author’s update; I’ll review that PR after the conflict resolution is pushed.

PR #1638, opened 2026-08-06
use vBinary.from_ical in Image.data and remove manual base64 decode

### LINKED ISSUE #1550: Add a property for the Base64 string view in vBinary

klouds27:
Closes #1550.

PR #1356 added `vBinary.base64data` as a getter/setter and moved internal storage to raw bytes. The one remaining manual base64 call in non-test code was `Image.data`, which called `base64.b64decode(self.b64data)` directly.

Changes:

Replaces `base64.b64decode(self.b64data)` in `Image.data` with `vBinary.from_ical(self.b64data)`. This makes validation strict and consistent with the rest of the codebase, matching the 'Not valid base 64 encoding.' error message.

Removes the dead else `base64.b64encode(value)` branch in `from_property_value`. Since `vText` inherits from `str`, `isinstance(value, str)` is always true in that branch and the else was unreachable.

Drops the now-unused `import base64` from `image.py`.

Adds a test confirming `Image.data` raises `ValueError` with the same message as `vBinary.from_ical` for invalid input.


stevepiercy:
@klouds27 you marked this PR as a draft. Is it ready for review?

news/1550.feature
stevepiercy, line 1: Never edit another contributor's news fragment. Instead create your own. See https://icalendar.readthedocs.io/en/latest/contribute/index.html#change-log-entry-file-name.

stevepiercy (changes requested):

stevepiercy (approved):
Thanks for fixing the news fragment. This LGTM. Let's get a second technical review.

angatha:
Since you want to close the issue, @stevepiercy are the docstrings nice enough? The class docstring does not document the parameters of `__init__`.

I think the "test all edge cases" part is done. and this does the last checkmark.

stevepiercy:
> Since you want to close the issue, @stevepiercy are the docstrings nice enough? The class docstring does not document the parameters of `__init__`.

I added it explicitly to the description in https://github.com/collective/icalendar/issues/1620. I'd prefer to not hold up this PR. The referenced PR did a pretty good job, but at that time, we hadn't settled on the concatenation of new and init into their class.

> I think the "test all edge cases" part is done. and this does the last checkmark.

Yup. Let's merge.

stevepiercy:
Ugh, it appears GitHub is being GitHub yet again. I got an email notification of @angatha's comment, but it doesn't show up anywhere. So weird! I'll post it as a review.

src/icalendar/prop/image.py
```diff
@@ -58,11 +57,7 @@ def from_property_value(cls, value: vUri | vBinary | vText) -> Image:
         elif isinstance(value, vBinary):
             params["b64data"] = value.base64data
         elif value_type == "BINARY":
-            params["b64data"] = (
-                value
-                if isinstance(value, str)
-                else base64.b64encode(value).decode("ascii")
-            )
+            params["b64data"] = str(value)
```
stevepiercy, line 60: From @angatha comment in an email notification.

`isinstance(value, str)` should always be false, so removing that part is fine. Can you please explain, why `base64.b64encode(value).decode("ascii")` can be replaced with `str(value)`? `value` should have type `vText` accoding to the type annotation.

stevepiercy (changes requested):

src/icalendar/prop/image.py
angatha: I removed it, becaute it was wrong. `vText` subclasses `str`.

stevepiercy (approved):

angatha (approved):

PR #1645, opened 2026-08-07
Read a PERIOD written with dates as midnight date-times

### LINKED ISSUE #1633: [Bug] Calendar with a date-only `PERIOD` parses but crashes on `to_ical()`
## Describe the bug

A calendar containing a PERIOD value written with dates instead of date-times parses without error, but writing it back out raises `AttributeError`.

[RFC 5545 section 3.3.9](https://datatracker.ietf.org/doc/html/rfc5545.html#autoid-41) defines a period as `period-explicit = date-time "/" date-time` and `period-start = date-time "/" dur-value`, so a date-only period is not valid and I would not expect it to be accepted. The problem is that it is accepted at parse time and only fails later, during serialization, with an error that doesn't say what is wrong with the input.

I hit this while reviewing type annotations on `vDatetime`, not on a real-world calendar file, so I don't know how common date-only periods are in the wild.

## To reproduce

```python

ics = """BEGIN:VCALENDAR
BEGIN:VEVENT
UID:1
DTSTART:19970101T180000Z
RDATE;VALUE=PERIOD:19970101/19970102
END:VEVENT
END:VCALENDAR
"""

calendar = Calendar.from_ical(ics)
print(calendar.to_ical().decode())
```

Output:

```pycon
  File "repro.py", line 13, in <module>
    print(calendar.to_ical().decode())
          ~~~~~~~~~~~~~~~~^^
  File ".../icalendar/cal/component.py", line 622, in to_ical
  File ".../icalendar/cal/component.py", line 611, in content_lines
  File ".../icalendar/cal/component.py", line 605, in content_line
  File ".../icalendar/parser/content_line.py", line 144, in from_parts
    values = values.to_ical()
  File ".../icalendar/prop/dt/list.py", line 46, in to_ical
    return b",".join(dts_ical)
  File ".../icalendar/prop/dt/list.py", line 45, in <genexpr>
    dts_ical = (from_unicode(dt.to_ical()) for dt in self.dts)
  File ".../icalendar/prop/dt/types.py", line 89, in to_ical
    return self.to_property_type().to_ical()
  File ".../icalendar/prop/dt/period.py", line 132, in to_ical
    return vDatetime(self.start).to_ical() + b"/" + vDatetime(self.end).to_ical()
  File ".../icalendar/prop/dt/datetime.py", line 105, in to_ical
    f"T{dt.hour:02}{dt.minute:02}{dt.second:02}"
        ^^^^^^^
AttributeError: 'datetime.date' object has no attribute 'hour'
```

The same thing happens at the property level, without going through a calendar:

```pycon
>>> from icalendar.prop import vPeriod
>>> period = vPeriod.from_ical("19970101/19970102")
>>> period
(datetime.date(1997, 1, 1), datetime.date(1997, 1, 2))
>>> vPeriod(period).to_ical()
AttributeError: 'datetime.date' object has no attribute 'hour'
```

The duration form `"19970101/P1D"` fails the same way. Periods written with date-times, such as `"19970101T180000Z/19970102T070000Z"`, round-trip correctly.

## Expected behavior

I would expect a date-only period to be rejected when it is read, with an error that names the property and says the value is not a valid period, rather than an `AttributeError` raised later from `to_ical()`.

If rejecting the value outright is too strict for existing calendars, then the alternative I would expect is for `to_ical()` to write something valid back out instead of raising.

Either way, the error a caller sees today can't be handled sensibly, since it arrives from a different call than the one that accepted the input. This may overlap with #1448, which asks that `from_ical()` only raise `ICalParsingError`.

## Environment

 - [x] Operating system: Ubuntu 22.04.5 LTS (WSL2)
 - [x] Python version: Python 3.14.5
 - [x] `icalendar` version: 7.2.2

## Additional context

 - [x] I tested it with the latest version. I reproduced it on 7.2.2 from PyPI and on `main` at 622dff5.
 - [x] I attached the `.ics` sample file, or there is no `.ics` sample file. The sample is inline above.

I did not find an existing issue for this. Related: #1448 (`from_ical()` should only raise `ICalParsingError`).

lcampanella98:
## Linked issue

- Closes #1633

## Description

A `PERIOD` written with dates was accepted when a calendar was parsed and then raised when it was written out again. Such a value is now read as midnight, per the decision on the issue to understand these calendars rather than reject them. Where the other half of the period carries a timezone, the converted half takes it.

Two commits are not the fix itself, so that they can be reviewed or dropped on their own.

- `Localize a date parsed with a TZID` fixes a pytz bug that the new period fixtures exposed through `test_equality.py`. It predates this branch and affects any date with a `TZID`. The branch is not green without it.
- `Take the timezone of the other half of a period` handles a period that mixes a date with a datetime.
- `Rewrite the vPeriod docstring as Python documentation` is the docstring work from #1244.

## Checklist

## Additional information

I ran `uv run tox -e py314,nopytz` and `make html`, all clean. I did not run the `py310` to `py313` or `pypy3` environments locally.


stevepiercy:
@lcampanella98 in the future, the PR description should be brief, especially when the original issue has all the details. I feel like I just read Tolstoy's "War and Peace", and I'm exhausted 😮‍💨.

src/icalendar/prop/dt/period.py
```diff
@@ -1,19 +1,64 @@
...
+        return tz.localize(midnight)  # type: ignore[attr-defined]
+    return midnight.replace(tzinfo=tz)
+
+
+def _to_period_datetimes(
+    start: date | datetime, end_or_duration: date | datetime | timedelta
+) -> tuple[datetime, datetime | timedelta]:
+    """Convert the dates of a period to datetimes.
+
+    :rfc:`5545` builds a period from date-times only, but calendars in the
```
stevepiercy: I think this is the precise section you intend. Please check.

Also here and elsewhere, please use the correct spelling of `datetime` or its plural in narrative documentation and docstrings. Also would you please add `date-time` to `docs/styles/config/vocabularies/icalendar/reject.txt` so that Vale rejects it in the future? We haven't been able to get Vale to process docstrings yet, but maybe some day.

```suggestion
    :rfc:`5545#section-3.3.9` builds a period from datetimes only, but calendars in the
```

src/icalendar/prop/dt/period.py
```diff
@@ -56,6 +101,11 @@ class vPeriod(TimeBase):
         the period, followed by a SOLIDUS character, followed by the
         [ISO.8601.2004] basic format for "DURATION" of the period.
```
stevepiercy: This also needs to have its RFC copy-paste replaced with its Python implementation. Would you please do the honors?

src/icalendar/prop/dt/period.py
```diff
@@ -76,6 +126,13 @@ class vPeriod(TimeBase):
...
             >>> period = vPeriod.from_ical('19970101T180000Z/19970102T070000Z')
             >>> period = vPeriod.from_ical('19970101T180000Z/PT5H30M')
+
+        A period written with dates is read as midnight:
+
+        .. code-block:: pycon
+
+            >>> vPeriod(vPeriod.from_ical('19970101/19970102')).to_ical()
+            b'19970101T000000/19970102T000000'
     """
```
stevepiercy, line 122:
```suggestion
    ..  versionchanged:: 7.2.3

        A period written with dates is read as midnight.
    """
```

src/icalendar/prop/dt/period.py
```diff
@@ -56,6 +101,11 @@ class vPeriod(TimeBase):
         the period, followed by a SOLIDUS character, followed by the
         [ISO.8601.2004] basic format for "DURATION" of the period.
 
+        A period written with dates instead of date-times does not conform
+        to this specification. Such a value is read as midnight, so that the
+        calendar can still be used, and it is written back out as a
+        date-time.
+
     Example:
```
stevepiercy:
```suggestion
    Examples:
```

src/icalendar/prop/dt/types.py
```diff
@@ -104,7 +104,9 @@ def from_ical(cls, ical, timezone=None):
             if timezone:
                 tzinfo = tzp.timezone(timezone)
                 if tzinfo is not None:
-                    return to_datetime(vDate.from_ical(ical)).replace(tzinfo=tzinfo)
+                    # ``replace`` picks the first offset a pytz timezone knows,
+                    # which is the LMT one, so localize instead.
```
stevepiercy: Licensed Massage Therapist?

```suggestion
                    # which is the local mean time one, so localize instead.
```

src/icalendar/tests/test_issue_1633.py
```diff
@@ -0,0 +1,163 @@
+"""Periods written with dates are read as midnight date-times.
+
+:rfc:`5545`, section 3.3.9, builds a period from date-times only. A period
```
stevepiercy: Tests don't render in the docs, so reStructuredText markup for links is wasted effort. In files that render to docs, I'd include the section in the markup link. Some markup, such as inline literals with backticks, is OK, as that conveys meaning to the reader.

```suggestion
RFC 5545, section 3.3.9, builds a period from date-times only. A period
```

src/icalendar/tests/test_issue_1633.py
```diff
@@ -0,0 +1,163 @@
...
+import pytest
+
+from icalendar import Event, FreeBusy
+from icalendar.prop import vPeriod
+
+UTC_MIDNIGHT = datetime(1997, 1, 1, tzinfo=timezone.utc)
+
+
+def test_calendar_from_the_issue_round_trips(calendars):
+    """The calendar in the issue serializes instead of raising."""
```
stevepiercy: I think this is more descriptive. Is this correct?

```suggestion
    """The calendar serializes the date to a datetime at midnight instead of raising."""
```

stevepiercy (changes requested):
This is thorough, thoughtful, and excellent work, @lcampanella98. Thank you for your effort and contribution.

I have a few suggestions for you to consider. Would you please take care? Thank you!

news/1633.bugfix.1
stevepiercy, line 1: I think this should be renamed to `news/1633.bugfix.1` so that it links to the original issue, as well as the other news fragment. Or merge them into a single file. Either is fine.

stevepiercy (changes requested):
One more detail.

src/icalendar/prop/dt/period.py
```diff
@@ -1,19 +1,64 @@
...
+    """
+    midnight = to_datetime(dt)
+    if tz is None:
+        return midnight
+    if is_pytz(tz):
+        return tz.localize(midnight)  # type: ignore[attr-defined]
+    return midnight.replace(tzinfo=tz)
+
+
+def _to_period_datetimes(
```
SemTiOne, line 35: I am curious how you would want this case read when the start half is already after midnight on the end date. 

For `19970102T120000Z/19970102` the converted end becomes `1997-01-02T00:00Z` and the constructor rejects the period at `to_ical` time with Start time is greater than end time, even though an end of day reading would give a valid 12 hour period. Is it worth a test that pins this limit of the midnight semantics, so the error does not sit unwanted in a different call like the original bug did?

src/icalendar/tests/test_issue_1633.py
```diff
@@ -0,0 +1,163 @@
...
+    """The end is in summer, so it takes the summer offset of the timezone.
+
+    A pytz timezone carries one offset per period of its history, so the
+    offset of the start cannot simply be copied over.
+    """
+    start = tzp.localize(datetime(1997, 7, 1, 12), "America/New_York")
+    _start, end = vPeriod((start, date(1997, 7, 2))).dt
+    assert end.utcoffset() == timedelta(hours=-4)
+
+
```
SemTiOne, line 91: The borrowed timezone branch is tested for the end date but only with UTC on the start side, so the localize path for a converted start next to a named zone is not covered. I believe this test closes that gap.

```suggestion

def test_a_converted_start_borrows_a_named_timezone(tzp):
    """The start date takes the summer offset of the other half's zone."""
    end = tzp.localize(datetime(1997, 7, 2, 12), "America/New_York")
    start, _end = vPeriod((date(1997, 7, 1), end)).dt
    assert start.utcoffset() == timedelta(hours=-4)
    
    
```

src/icalendar/tests/test_issue_1633.py
```diff
@@ -0,0 +1,163 @@
...
+    assert b"19970101T000000/19970102T000000" in event.to_ical()
+
+
+def test_dates_added_to_a_freebusy():
+    """``add`` builds the period through vPeriod itself."""
+    freebusy = FreeBusy()
+    freebusy.add("FREEBUSY", [(date(1997, 1, 1), date(1997, 1, 2))])
+    assert b"19970101T000000/19970102T000000" in freebusy.to_ical()
+
+
```
SemTiOne, line 146: I don't see a round trip test for a comma separated period list where one entry is date-only. That path combines the list splitting with the date conversion, so I believe one test would lock both in.

```suggestion

def test_a_comma_separated_period_list_round_trips():
     """A date-only period next to a normal one survives the list split."""
     event = Event()
     event.add(
         "RDATE",
         [
             (date(1997, 1, 1), date(1997, 1, 2)),
             (datetime(1997, 7, 5, 12), datetime(1997, 7, 6)),
         ],
     )
     ical = event.to_ical()
     assert b"19970101T000000/19970102T000000,19970705T120000/19970706T000000" in ical
     parsed = Event.from_ical(ical)
     assert parsed["RDATE"].dts[0].dt[0] == datetime(1997, 1, 1)
     
     
```

SemTiOne (commented):
Nice work. I have a question and some suggestions below.

src/icalendar/prop/dt/period.py
lcampanella98: Done, and `reject.txt`, which was empty until now, has `[Dd]ate-times?` in it. I made it case sensitive on purpose, so that `DATE-TIME` still passes. Say if you would rather reject it everywhere.

I left two: `types.py:165` and `period.py:270` are error strings that `rfc_7265_jcal/test_invalid_jcal.py` asserts verbatim. Happy to sweep those, and the rest of `prop/dt/`, separately.

src/icalendar/prop/dt/period.py
lcampanella98: Done for `vPeriod`, assuming we're following the guidance of #1244. I also dropped `vPeriod` from `KNOWN_BAD_HEADINGS`.

One change you did not ask for: an `Attributes:` section failed the docs build with four `duplicate object description` warnings, because autodoc already documents the annotations, so those became `#:` comments as in `cal/component.py`.

src/icalendar/prop/dt/period.py
lcampanella98: Applied, at the end of the rewritten docstring.

src/icalendar/prop/dt/period.py
lcampanella98: Applied

src/icalendar/prop/dt/types.py
lcampanella98: Fixed

src/icalendar/tests/test_issue_1633.py
lcampanella98: Done. Your line kept "date-times" in it, so I used "datetimes" there too.

src/icalendar/tests/test_issue_1633.py
lcampanella98: Applied, with dates plural, since both halves of that period are dates.

news/1633.bugfix.1
lcampanella98: Renamed to `news/1633.bugfix.1`. I kept two entries rather than merging, since they are two different bugs.

src/icalendar/tests/test_issue_1633.py
lcampanella98: You are right that the gap was there. Added.

src/icalendar/tests/test_issue_1633.py
lcampanella98: Added. I confirmed the serialized output and the first parsed period.

src/icalendar/prop/dt/period.py
lcampanella98: Right, and there is a test for it now.

Both readings are guesses, since the value does not conform either way. I kept midnight because that is what every other bare date means here, where end of day would apply only where midnight fails, making a date mean one thing or the other depending on the other half.

It is not a regression either. The same value raised `TypeError` from the same `to_ical()` call before this branch. #1448 tracks that general case.

lcampanella98:
Fair, and sorry for the PR description length. Noted for the future.

src/icalendar/prop/dt/period.py
stevepiercy: Yup, you got it. When uppercased and hyphenated as "DATE-TIME", it's an RFC 5545 value type, which is perfectly acceptable. If you see lowercase "date-time", then it's likely either copy-paste from the RFC where it's used to define syntax, or it's a jCal type identifier in a jCal file per https://datatracker.ietf.org/doc/html/rfc7265.html#section-3.6.5. "datetime" is the common programming spelling, such as in Python and SQL.

Context is everything. Would you make a new issue to sweep the remaining "date-time"s, as it's out of scope for this PR and issue? It requires thought of whether it should be "DATE-TIME", "date-time", replace the RFC docstring copy-paste with the Python implementation, or "datetime" based on its context.

news/1633.bugfix
```diff
@@ -0,0 +1 @@
+A ``PERIOD`` value written with dates instead of datetimes, such as ``RDATE;VALUE=PERIOD:19970101/19970102``, was accepted when a calendar was parsed and then raised an ``AttributeError`` when it was written out again, or a ``TypeError`` where only one half was a date. Such a value is now read as midnight, in the timezone of the other half where it has one. I used AI to assist me with this change. @lcampanella98
```
stevepiercy:
```suggestion
A ``PERIOD`` value written with dates instead of datetimes, such as ``RDATE;VALUE=PERIOD:19970101/19970102``, was accepted when a calendar was parsed and then raised an :exc:`AttributeError` when it was written out again, or a :exc:`TypeError` where only one half was a date. Such a value is now read as midnight, in the timezone of the other half where it has one. I used AI to assist me with this change. @lcampanella98
```

src/icalendar/prop/dt/types.py
```diff
@@ -147,7 +149,7 @@ def parse_jcal_value(cls, jcal: str | list) -> timedelta:
 
         Raises:
             ~error.JCalParsingError: If the value can't be parsed as either a date,
-                 time, date-time, duration, or period.
+                 time, datetime, duration, or period.
```
stevepiercy: Actually, this one should be reverted, as it's a jCal type identifier. See my comment: https://github.com/collective/icalendar/pull/1645#discussion_r3747134697

stevepiercy (changes requested):
A couple of minor tweaks and a reversion.

src/icalendar/prop/dt/period.py
stevepiercy: They also render as docstrings, a nice bonus. Example: https://icalendar--1645.org.readthedocs.build/en/1645/reference/api/icalendar.prop.dt.period.html#icalendar.prop.dt.period.vPeriod.by_duration

src/icalendar/prop/dt/period.py
lcampanella98: Thanks, that rule is clearer than what I had. Opened #1655

news/1633.bugfix
lcampanella98: Applied

src/icalendar/prop/dt/types.py
lcampanella98: Reverted, and I reverted the same change in `period.py:240` for the same reason.

src/icalendar/prop/dt/period.py
stevepiercy: Thanks. That's a well-written issue, and we already got some PRs for it. 🚀

src/icalendar/prop/dt/period.py
stevepiercy: @SemTiOne here's the referenced test.

https://github.com/collective/icalendar/pull/1645/changes#diff-aa02f64f19712053c679a39a017ad9f5b0122a5140a943af7277864eec248f77R189-R197

Would you please verify, and either resolve this suggestion or suggest an improvement?

stevepiercy (approved):
Thanks for the revisions, and also thanks to @SemTiOne for the review. Good stuff! I'd like one more review from a maintainer before merge.

niccokunzmann (approved):

src/icalendar/prop/dt/period.py
SemTiOne: LGTM, thanks.

Issue #1650, opened 2026-08-09
Replace generated docstrings with static docstrings


stevepiercy:
Three parameters and four properties generate docstrings in the objects that call them. I identified them all by searching for `doc: str` in their signatures.

This was a nifty idea, but it creates problems:

- The included docstring may be incorrect in its caller.
- The docstring gets included in an awkward place, making docstrings difficult to understand.
- Type hints may conflict with the docstring.

To close this issue, all of the following objects' `doc` parameter content should be moved into the caller and edited to be correct, then the `doc` parameter removed from the object. Only the parameters and properties are listed, and you'll need to search for its callers. You may focus on one or many objects and their callers in a pull request.

- src/icalendar/attr.py
  - [ ] main_prop
  - [ ] single_int_property
  - [ ] single_utc_property
  - [ ] create_single_property
  - [ ] multi_string_property
- src/icalendar/param.py
  - [ ] create_single_property
  - [ ] quoted_list_parameter #1651
  - [ ] boolean_parameter #1652
- [Build the docs locally](https://icalendar.readthedocs.io/en/latest/contribute/documentation/build-check.html) to ensure there are no warnings or errors.
- For each object, consider whether this is, in fact, a good idea. Would this make the reference API better or worse? AI can't answer this question. Use your judgment, not AI. See https://github.com/collective/icalendar/issues/1650#issuecomment-5237760655.
- Run `make docs-all` to ensure your changes meet all documentation quality checks.
- Run `make format` to ensure your changes meet code format quality checks.
- Follow [Docstring structure](https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html#docstring-structure) guidance.
- Follow [Contribute](https://icalendar.readthedocs.io/en/latest/contribute/index.html) guidance.
- All CI checks must pass.
- Check off each object when its PR is merged to avoid duplicate effort.

Pull requests that do not follow the above guidance may be closed without comment, as it wastes reviewer time to tell you to do this.

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/1385#discussion_r3594623845_

floze-the-genius:
I'd like to take the boolean_parameter / RSVP slice only. I'll remove the dynamic doc argument, add the static RSVP attribute documentation, and include a separate documentation news fragment; I won't overlap #1651's quoted_list_parameter work.

akkiyolo:
[...]

PR #1651, opened 2026-08-09
Replace quoted-list generated docstrings with static docstrings

cakeni:
## Linked issue

- See #1650

## Description

Moves quoted_list_parameter and its three callers from generated property docs to static Sphinx-visible attribute docstrings. Covers part of the seven-factory issue. AI use is disclosed in the commit and news fragment.

## Checklist

## Additional information

55 focused tests, targeted doctest, Ruff, format, diff, AST, and Towncrier pass; generated HTML includes them. Full Windows pytest hit an unrelated CRLF failure after 9,196 passes. Sphinx -W completed pages but hit 12 existing include warnings.


src/icalendar/param.py
```diff
@@ -208,12 +215,11 @@ def fset(self: VPROPERTY, value: str | tuple[str]):
...
 
-    return property(fget, fset, fdel, doc=doc)
+    return property(fget, fset, fdel)
 
 
-DELEGATED_FROM = quoted_list_parameter(
-    "DELEGATED-FROM",
-    """Specify the calendar users that have delegated their participation to the calendar user specified by the property.
+DELEGATED_FROM = quoted_list_parameter("DELEGATED-FROM")
+"""Specify the calendar users that delegated their participation.
```
stevepiercy: This has an interesting side effect, and I'm not sure whether it's better or worse. The docstring previously appeared in the context of:

- [vCalAddress](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.prop.cal_address.html#icalendar.prop.cal_address.vCalAddress.DELEGATED_FROM)

It got moved to:

- [icalendar.param](https://icalendar--1651.org.readthedocs.build/en/1651/reference/api/icalendar.param.html?readthedocs-diff=true&readthedocs-diff-chunk=1#icalendar.param.DELEGATED_FROM)

I think that makes it less useful because it loses its practical context. @niccokunzmann @cakeni what do you think? AI doesn't think.

Similar behavior occurred for `DELEGATED_TO`, but not `MEMBER` which appears to have been omitted by accident.

@niccokunzmann should `MEMBER` be treated the same as other CAL-ADDRESS parameters? This appears to be a bug.

stevepiercy (commented):
Although I thought this would be a good idea, I'm having second thoughts. Please see my comment.

src/icalendar/param.py
cakeni: Thanks, I dug into this more carefully, and I agree that moving the docs to icalendar.param loses useful context.
Previously, quoted_list_parameter() passed the docstring into the property itself, so Sphinx could show it under vCalAddress.DELEGATED_FROM and vCalAddress.DELEGATED_TO. With the current static attribute docstrings, Sphinx instead associates them with icalendar.param.
I think the cleaner approach is to keep the descriptor implementation in icalendar.param, but put the static documentation on the concrete vCalAddress bindings. That keeps the docs in the practical context without duplicating them.
I also checked MEMBER: it was not missed by this PR. vCalAddress simply never exposed MEMBER, unlike DELEGATED_FROM and DELEGATED_TO, so this looks like a pre-existing inconsistency. Adding it would be a new public API, though, so I don't want to include that without maintainer confirmation.
Would you prefer that I keep #1651 limited to fixing the documentation location and handle MEMBER separately, or make all three consistent here?

niccokunzmann: Member should be treated the same as the others. That is a mistake.

The change can be in here or in a follow up PR.

niccokunzmann: This is the way, variables are documented. Nice feature.

niccokunzmann:
This PR uses the same idea as the one in #1652.

Could you help me understand why you @cakeni  and @floze-the-genius have chosen this path? Did you talk to each other?

The change that I see is that now, we document in a different place.
@stevepiercy That is an opportunity to change where the docs live and they can be more specific to the classes that these attributes reside in. However, we should probably check that this is consistent.


## Before

property holds the docs

class1.PROP = class2.PROP = PROPERTY

class1.PROP and class2.PROP have the same docs

## Now

attribute is documented

class1.PROP = class2.PROP = PROPERTY

class1.PROP has documentation distinct from class2.PROP

## Evaluation

This is a documentation design decision. They use a 'newer' language feature. This increases the amount of documentation and possible duplication and inconsistency. However, it also allows tailoring the documentation of the property to the specific class that it lives in.
Moving the whole code base over is a big job (docstrings increase with the amount of occurrences in classes) but also a good first-timer issue.
We do not have to move the whole code base over.

What are your thoughts?

cakeni:
> This PR uses the same idea as the one in #1652.
> 
> Could you help me understand why you @cakeni and @floze-the-genius have chosen this path? Did you talk to each other?
> 
> The change that I see is that now, we document in a different place. @stevepiercy That is an opportunity to change where the docs live and they can be more specific to the classes that these attributes reside in. However, we should probably check that this is consistent.
> 
> ## Before
> property holds the docs
> 
> class1.PROP = class2.PROP = PROPERTY
> 
> class1.PROP and class2.PROP have the same docs
> 
> ## Now
> attribute is documented
> 
> class1.PROP = class2.PROP = PROPERTY
> 
> class1.PROP has documentation distinct from class2.PROP
> 
> ## Evaluation
> This is a documentation design decision. They use a 'newer' language feature. This increases the amount of documentation and possible duplication and inconsistency. However, it also allows tailoring the documentation of the property to the specific class that it lives in. Moving the whole code base over is a big job (docstrings increase with the amount of occurrences in classes) but also a good first-timer issue. We do not have to move the whole code base over.
> 
> What are your thoughts?

I hadn’t seen #1652 and didn’t discuss this with the author. This just seemed like the simplest way to keep the code shared while putting the docs where people would expect them. I’d keep it limited to these parameters for now.

src/icalendar/prop/cal_address.py
```diff
@@ -104,13 +104,51 @@ def email(self) -> str:
...
         LANGUAGE,
+        MEMBER,
         PARTSTAT,
         ROLE,
         RSVP,
         SENT_BY,
         VALUE,
     )
 
+    DELEGATED_FROM = DELEGATED_FROM
```
stevepiercy, line 115: I never would have thought of this technique, yet it's so obvious and simple! I'm adding this clever implementation to my repertoire.

src/icalendar/prop/cal_address.py
```diff
@@ -104,13 +104,51 @@ def email(self) -> str:
...
+    """Specify the group or list membership of a calendar user.
+
+    Description:
+        This parameter can be specified on properties with a
+        CAL-ADDRESS value type.  The parameter identifies the groups or
+        list membership for the calendar user specified by the property.
+        The parameter value is either a single calendar address in a
+        quoted-string or a COMMA-separated list of calendar addresses,
+        each in a quoted-string.  The individual calendar address
+        parameter values MUST each be specified in a quoted-string.
```
stevepiercy: All of these should remove the `DESCRIPTION` section heading, and their content dedented accordingly and rewrapped, per https://icalendar.readthedocs.io/en/stable/contribute/documentation/style-guide.html#docstring-structure. Their content is perfect and improves several grammatical mistakes.

stevepiercy (changes requested):
A minor formatting tweak, then I approve. You helped us get unstuck. ❤️

Upon final approval and merge, we should close #1652, and update the checklist in the original issue #1650.

src/icalendar/prop/cal_address.py
cakeni: Fixed the docstring formatting as requested. Thanks for the review

stevepiercy (approved):
Thank you for the changes.

stevepiercy:
@niccokunzmann this LGTM. Have all your unresolved comments been addressed?

niccokunzmann (approved):
Thanks, yes, that works!

niccokunzmann:
This is ready for merge. My concerns have been addressed. The documentation should still be updated to be more useful. Currently, it is still the RFC copy&paste and Issue #1509 addresses this.

@cakeni Thanks for your contribution!
If you would like to work further on this section of the code and improve it? E.g. #1509 shows where we really want to go. The docstrings could be nicer and we like examples in the code so people see how to use the attributes. Also no Python user is interested in knowing that this will be written to the file system with double quotes. icalendar takes care of that for everyone.

What do you think? You are very welcome!

Merge should happen automatically..

stevepiercy:
To be clear, https://github.com/collective/icalendar/pull/1509 offered a fix for the original issue https://github.com/collective/icalendar/issues/1244, but the author abandoned and closed the PR. Its content is good, and someone may pick it up complete it.

PR #1654, opened 2026-08-10
add type hints to new methods in the cal folder

ShirayukiRin:
## Linked issue

- See  #938 

## Description

Added type hints to multiple `new` methods in the `cal` folder.

## Checklist

## Additional information

One test failed, and it seems like it was due to a 403 Client Error when trying to connect to 
`https://stackoverflow.com/questions/79171631/how-do-i-determine-whether-a-zoneinfo-is-an-alias/79171734#79171734`

related log:
```log
(reference/api/icalendar.timezone.equivalent_timezone_ids: line   10) broken    https://stackoverflow.com/questions/79171631/how-do-i-determine-whether-a-zoneinfo-is-an-alias/79171734#79171734 - 403 Client Error: Forbidden for url: https://stackoverflow.com/questions/79171631/how-do-i-determine-whether-a-zoneinfo-is-an-alias/79171734


docs: FAIL code 2 (506.71=setup[7.60]+cmd[499.11] seconds)
```


news/938.documentation.3
```diff
@@ -0,0 +1 @@
+Added type annotations to ``new`` methods in the ``cal`` folder. @ShirayukiRin
```
lcampanella98: Nit: the guide asks for reStructuredText references to the changed API, and "the ``cal`` folder" won't mean much in `CHANGES.rst`.

```suggestion
Added return type annotations to the ``new()`` factory methods of the calendar components, such as :meth:`~icalendar.cal.event.Event.new`. @ShirayukiRin
```

lcampanella98 (changes requested):
Thanks for this. One suggestion: `Self` rather than the concrete class.

`new()` is a classmethod, so subclasses inherit it. `-> Calendar` means a checker narrows `LazyCalendar.new()` to `Calendar`, and the same hits anyone subclassing a component:

```python
class MyEvent(Event):
    def my_helper(self) -> str: ...

MyEvent.new().my_helper()   # error: "Event" has no attribute "my_helper"  [attr-defined]
```

`Self` is already exported from `icalendar.compatibility`, and `prop/uid.py:20` uses it on a `new()` factory.

Worth including `Component.new` (`component.py:865`) in the change. Each override assigns the `super()` result to a concrete-typed local, so mypy errors inside all eight functions here (`todo.py:393` and so on); annotating only the overrides does not clear those. It also picks up `Timezone`, `TimezoneStandard`, `TimezoneDaylight` and `LazyCalendar`, which have no `new` of their own and currently resolve to `Component`.

news/938.documentation.3
stevepiercy: I'd revise @lcampanella98's suggestion with a distinguishing label and a link to the module.

```suggestion
Added return type annotations to the ``new()`` factory methods of the :mod:`icalendar.cal` components, such as :meth:`Event.new <icalendar.cal.event.Event.new>`. @ShirayukiRin
```

stevepiercy (dismissed):
Minor tweak to news, then it LGTM.

ShirayukiRin:
> Thanks for this. One suggestion: `Self` rather than the concrete class.
> 
> `new()` is a classmethod, so subclasses inherit it. `-> Calendar` means a checker narrows `LazyCalendar.new()` to `Calendar`, and the same hits anyone subclassing a component:
> 
> ```python
> class MyEvent(Event):
>     def my_helper(self) -> str: ...
> 
> MyEvent.new().my_helper()   # error: "Event" has no attribute "my_helper"  [attr-defined]
> ```
> 
> `Self` is already exported from `icalendar.compatibility`, and `prop/uid.py:20` uses it on a `new()` factory.
> 
> Worth including `Component.new` (`component.py:865`) in the change. Each override assigns the `super()` result to a concrete-typed local, so mypy errors inside all eight functions here (`todo.py:393` and so on); annotating only the overrides does not clear those. It also picks up `Timezone`, `TimezoneStandard`, `TimezoneDaylight` and `LazyCalendar`, which have no `new` of their own and currently resolve to `Component`.

Thanks for the review and the information about `Self`!

I noticed that the `Component.new` uses the concrete class as its return type, so I followed the same pattern. 
Should I change the return type to `Self` only for classes that have subclasses, or just change the return type to `Self` for every class?
I think using `Self` for every class might be better as there is no difference when a class doesn't have any subclass, and you don't need to change the annotation after you add a subclass.

What you think?

news/938.documentation.3
ShirayukiRin: Thank you, and I’m sorry for the mistake. 
I’ll pay more attention to this next time.

lcampanella98:
I think we should change it to `Self` for every one. The test isn't whether a class has subclasses today, it's whether the method returns an instance built from `cls`. `Component.new` does `component = cls()` (`component.py:885`) and each override returns what `super().new()` handed back, so `Self` holds for all nine.

And you were right that `Component.new` set the pattern you followed. That's why it needs changing too, otherwise the eight overrides keep the errors on their `super().new()` calls.

news/938.documentation.3
lcampanella98: @stevepiercy's version is better, agreed.

@ShirayukiRin no worries, and thanks for picking it up so quickly.

ShirayukiRin:
> I think we should change it to `Self` for every one. The test isn't whether a class has subclasses today, it's whether the method returns an instance built from `cls`. `Component.new` does `component = cls()` (`component.py:885`) and each override returns what `super().new()` handed back, so `Self` holds for all nine.
> 
> And you were right that `Component.new` set the pattern you followed. That's why it needs changing too, otherwise the eight overrides keep the errors on their `super().new()` calls.

Got it and I’ve updated it accordingly.
Thanks for the clear explanation!

lcampanella98 (dismissed):
Looks good, thanks for the quick turnaround. Subclass narrowing works now and the suite passes.

Non-blocking follow-up if you want it: the locals still pin the concrete class, which widens `Self` back.

```python
todo: Todo = super().new(...)   # todo.py:394
return todo                     # todo.py:426: got "Todo", expected "Self"
```

Annotating those `Self`, or dropping the annotation since inference gives `Self` anyway, clears it in all eight files.

Happy either way, this is already an improvement on main.

ShirayukiRin:
Thanks for the review! 
I’ve updated the type hints for the local variables.

lcampanella98 (approved):
Looks good, thanks for the PR and the quick turnaround. Approving

niccokunzmann:
@lcampanella98 Thanks for the technical review and the guidance!
@stevepiercy Thanks for keeping the documentation pristine! I checked and your suggestions have been applied, so I count this as LGTM from your point of view.

@ShirayukiRin Thanks for your contribution and the nice conversation! (nice for me to look at anyway ;)
If you liked working on icalendar I hope to see you around soon. Also, we value reviews of other PRs if you like to get into that.

This is ready for merge....

PR #1657, opened 2026-08-10
Clarify RECURRENCE-ID datetime spelling

### LINKED ISSUE #1655: Sweep the remaining "date-time" spellings, deciding each by context
## Summary

``date-time`` is written in places where another form is the right one. The correct form depends on the context, per the rules stevepiercy set out in the review of #1645, https://github.com/collective/icalendar/pull/1645#discussion_r3747134697

- Where it names the value type of :rfc:`5545`, as in ``VALUE=DATE-TIME`` or in a list with ``DATE``, ``DURATION``, and ``PERIOD``, write ``DATE-TIME``.
- Where it names a jCal type identifier per :rfc:`7265#section-3.6.5`, or quotes the syntax of an RFC, write ``date-time``.
- Where it means a moment in time, so that "a timestamp" would do instead, write ``datetime``.

The docstring of ``RECURRENCE_ID`` in ``attr.py`` has two of them, four lines apart. The first is prose and should read ``datetime``, the second names the value type and is already correct.

    The value is the original start date or date-time of that instance
    The value is usually a DATE-TIME and must use the same value type

## Scope

There are 62 occurrences of ``date-time`` across ``src`` and ``docs``, 31 of them outside the tests, in 13 files, and 78 of ``DATE-TIME``, which I would leave alone unless one turns out to be prose. I expect the work to be mostly ``date-time`` becoming ``datetime``, with a good number correct already, so it is not a find and replace.

I did not find a ``date-time`` that needs to become ``DATE-TIME``. The nearest is ``cal/alarm.py:173``, "the current date-time value in UTC", which is quoted :rfc:`9074` text and belongs with the copy-paste below.

They fall into four groups.

**Correct as they are, jCal type identifiers.** ``prop/dt/types.py:152`` and ``:165``, ``prop/dt/datetime.py:171`` and ``:180``, ``prop/broken.py:90``, the jCal examples in ``cal/component.py:920`` and ``docs/how-to/jcal.rst``, and most of the 38 occurrences under ``tests/rfc_7265_jcal``.

**RFC copy-paste, which #1244 covers.** ``prop/dt/datetime.py:32`` copies the grammar rule ``date-time = date "T" time``, written in the Augmented Backus-Naur Form (ABNF) notation the RFCs use, and ``:46`` describes it. ``cal/alarm.py:173`` quotes :rfc:`9074`. These go when those docstrings are rewritten, rather than being respelled.

**Prose that should be ``datetime``.** ``attr.py:1179`` is the clearest one.

**Internal factory keys.** ``prop/factory.py`` uses ``"date-time-list"`` as a lookup key in five places. Renaming it is a code change rather than a spelling fix, and it may be relied on through ``types_factory``, so I would leave it alone unless you want it renamed.

## Notes

#1645 added ``[Dd]ate-times?`` to ``docs/styles/config/vocabularies/icalendar/reject.txt``, so Vale already rejects ``date-time`` in narrative documentation. Vale reads only ``docs/**.rst``, so it does not see docstrings, and the jCal examples in ``docs/how-to/jcal.rst`` sit inside code blocks that it skips.

I did not find an existing issue for this, open or closed.

## Related

- #1645, the pull request where this came up, and which is out of scope for it.
- #1244, replacing the RFC copy-paste in docstrings, which covers one of the groups above.
- #1007, checking style and spelling in docstrings.

atirna:
## Linked issue

- Closes #1655

## Description

Fixed spelling.

## Checklist

### DIFF
--- a/news/1655.documentation
+++ b/news/1655.documentation
@@ -1,1 +1,1 @@
-Clarified that a ``RECURRENCE-ID`` describes an occurrence's original start datetime rather than its :rfc:`5545` ``DATE-TIME`` value type. AI-assisted. @atirna
+Changed ``date-time`` to ``datetime`` in the narrative parts of the ``RECURRENCE-ID`` and ``ACKNOWLEDGED`` documentation. AI-assisted. @atirna
--- a/news/1655.documentation
+++ b/news/1655.documentation
@@ -1,0 +1,1 @@
+Clarified that a ``RECURRENCE-ID`` describes an occurrence's original start datetime rather than its :rfc:`5545` ``DATE-TIME`` value type. AI-assisted. @atirna
--- a/src/icalendar/attr.py
+++ b/src/icalendar/attr.py
@@ -1176,7 +1176,7 @@ Identify a specific occurrence of a recurring calendar object.
 
 This property is used together with ``UID`` and ``SEQUENCE`` to refer to one
 particular instance in a recurrence set. The value is the original start
-date or date-time of that instance, not the rescheduled time.
+date or datetime of that instance, not the rescheduled time.
 
 The value is usually a DATE-TIME and must use the same value type as the
 ``DTSTART`` property in the same component. A DATE value may be used for
--- a/src/icalendar/cal/alarm.py
+++ b/src/icalendar/cal/alarm.py
@@ -170,7 +170,7 @@ class Alarm(Component):
     It also allows clients to track repeating alarms or alarms on recurring events or
     to-dos to ensure that the right number of missed alarms can be tracked.
 
-    Clients SHOULD set this property to the current date-time value in UTC
+    Clients SHOULD set this property to the current datetime value in UTC
     when a calendar user acknowledges a pending alarm. Certain kinds of alarms,
     such as email-based alerts, might not provide feedback as to when the calendar user
     sees them. For those kinds of alarms, the client SHOULD set this property


niccokunzmann (commented):
That does not require a test from my view. It is ok to just correct spelling.

angatha:
For closing the issue, I would also change `./src/icalendar/cal/alarm.py:173`. @stevepiercy

atirna:
Updated alarm.py too. The documentation build, formatting, lint, and towncrier checks still pass. Thanks for catching it.

niccokunzmann:
@lcampanella98 This is for your issue.

angatha (dismissed):
Only places that should be changed eventually are copy pased grammar definitions or reverendes to a production rule.

news/1655.documentation
```diff
@@ -0,0 +1 @@
+Clarified that a ``RECURRENCE-ID`` describes an occurrence's original start datetime rather than its :rfc:`5545` ``DATE-TIME`` value type. AI-assisted. @atirna
```
lcampanella98: This reads as a semantic clarification, but the change is a spelling fix. It also sets up a contrast that is not real: the value *is* the occurrence's original start *and* its type *is* `DATE-TIME`, and the docstring still says so two lines below the edit. Could you describe the actual change instead, and include the `ACKNOWLEDGED` edit?

```suggestion
Changed ``date-time`` to ``datetime`` in the narrative parts of the ``RECURRENCE-ID`` and ``ACKNOWLEDGED`` documentation. AI-assisted. @atirna
```

lcampanella98 (changes requested):
Thanks for the PR. 

And @niccokunzmann thanks for the ping. 

Scope looks right. I checked the other 34 `date-time` occurrences outside the tests and they are all correct as they stand: jCal type identifiers, internal lookup keys in `prop/factory.py`, and ABNF grammar copy-paste. @angatha yes, agreed - those are the ABNF copy-paste, and #1655 assigns them to #1244, which rewrites those docstrings rather than respelling them. Happy for this to close the issue.

`alarm.py:173` is verbatim RFC 9074 text that #1655 had grouped with the copy-paste for #1244, but since #1244 replaces that docstring anyway, fine as changed.

One fix below.

news/1655.documentation
stevepiercy: Agree with @lcampanella98. AI is often stupid, but sounds convincing. Resist the broligarchy!

stevepiercy (changes requested):

lcampanella98 (approved):
looks good to me, approving

stevepiercy (approved):

stevepiercy:
Thank you both @lcampanella98 and @atirna for your work on this issue!

atirna:
Thank you for your review and feedback

PR #1661, opened 2026-08-12
Warn on cached lookups of guessed globally unique TZIDs

### LINKED ISSUE #1660: [Bug] Cached TZID guess swallows GloballyUniqueTZIDGuessed on repeated lookups; test suite fails without pytz (nopytz CI env installs pytz)
## Describe the bug

`TZP.timezone()` emits `GloballyUniqueTZIDGuessed` only on the *first* lookup of a globally unique TZID (RFC 5545 [section 3.2.19](https://www.rfc-editor.org/rfc/rfc5545#section-3.2.19), e.g. `/mozilla.org/20070129_1/America/New_York`). The guessed result is stored in the TZID cache under the cleaned id, and the fact that it was resolved *by guessing* is lost. Any later lookup of the same TZID on the same `TZP` instance is served from the cache and returns silently — even with `warnings.simplefilter("always")`.

This makes the warning order-dependent, and it makes `test_timezone_id_resolves` (added in #1479) fail in any environment where pytz is **not** installed:

```
FAILED src/icalendar/tests/test_issue_313_globally_unique_tzid.py::test_timezone_id_resolves[zoneinfo-/freeassociation.sourceforge.net/Europe/Berlin-Europe/Berlin] - Failed: DID NOT WARN.
FAILED src/icalendar/tests/test_issue_313_globally_unique_tzid.py::test_timezone_id_resolves[zoneinfo-/mozilla.org/20070129_1/America/New_York-America/New_York] - Failed: DID NOT WARN.
```

Chain of events when pytz is absent:

1. `test_events_with_global_tzids_are_timezone_aware` runs first and parses `calendars/issue_313_globally_unique_tzid.ics`, which contains `/freeassociation.sourceforge.net/Europe/Berlin` and `/mozilla.org/20070129_1/America/New_York`. The guesses are cached on the module-scoped shared `tzp`.
2. `test_timezone_id_resolves` then looks up the same two TZIDs. The cache hit skips the `warnings.warn(...)` branch in `TZP.timezone()` (`src/icalendar/timezone/tzp.py`), so `pytest.warns(GloballyUniqueTZIDGuessed)` fails with `DID NOT WARN`.

When pytz **is** installed, the `tzp_name` parametrization makes pytest alternate the provider (pytz ↔ zoneinfo) inside the module. Every switch calls `TZP._use()`, which resets the cache, so the zoneinfo run of `test_timezone_id_resolves` happens to start with a clean cache and the failure is masked.

### Why CI does not catch it

The `3.10 (nopytz)` job is meant to cover exactly this configuration, but `[testenv:nopytz]` in `tox.ini` inherits `dependency_groups = test` from `[testenv]`, and the `test` dependency group in `pyproject.toml` contains `pytz>=2025.2`. The job therefore runs *with* pytz installed — see the tox env listing of the latest `main` run ([run 31557281112](https://github.com/collective/icalendar/actions/runs/31557281112), job `3.10 (nopytz)`):

```
nopytz: ast-serialize==0.8.0,coverage==7.15.4,...,pytz==2026.3.post1,...
```

So the test suite is currently red on `main` for anyone who runs it in a genuinely pytz-free environment (e.g. distribution packagers), and the nopytz CI job never exercises the pytz-free code paths it was added for.

## To reproduce

In a pytz-free environment (Python 3.13, `pip install -e . pytest hypothesis`):

```console
$ python -m pytest src/icalendar/tests/test_issue_313_globally_unique_tzid.py
...
2 failed, 8 passed
$ python -m pytest src/icalendar/tests/test_issue_313_globally_unique_tzid.py::test_timezone_id_resolves
...
8 passed   # passes in isolation -> order dependent
```

Minimal reproduction of the root cause, independent of pytest and of the Python warning registry:

```python
import warnings
from icalendar.timezone.tzp import TZP

tzp = TZP("zoneinfo")
for attempt in (1, 2):
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        tzp.timezone("/mozilla.org/20070129_1/America/New_York")
    print(attempt, [w.category.__name__ for w in caught])
```

Output:

```pycon
1 ['GloballyUniqueTZIDGuessed']
2 []
```

## Expected behavior

Every lookup that is resolved through a guessed Olson identifier should emit `GloballyUniqueTZIDGuessed`, whether it is served by the provider or by the cache:

- `test_timezone_id_resolves` wraps *each* `tzp.timezone(tzid)` call in `pytest.warns(GloballyUniqueTZIDGuessed)` — the test contract does not depend on lookup history.
- The `GloballyUniqueTZIDGuessed` docstring says "Suppress this warning if the resolved timezone is correct for your data", i.e. suppression is the caller's choice via warning filters; the library should not silently drop the warning based on cache state. (Python's default warning filter already deduplicates repeated warnings for end users.)

And the nopytz CI environment should actually exclude pytz, so that pytz-free regressions like this one are caught. I am happy to send a follow-up PR for the tox/dependency-group split if wanted.

## Environment

- [x] Operating system: macOS 15 (also reproduced the CI side on ubuntu-latest logs)
- [x] Python version: Python 3.13
- [x] `icalendar` version: `main` @ 30ec6ee (0.1.dev241)

## Additional context

- [x] I tested it with the latest version (`main` @ 30ec6ee).
- [x] There is no `.ics` sample file needed; the repository's own `issue_313_globally_unique_tzid.ics` triggers the cache pollution.

Related: #313, #1479 (which added the guessing + warning), #1636 (nopytz currently pins Python 3.10).

I have a minimal fix ready (remember which cache entries were guessed and warn on cache hits too, ~10 lines in `tzp.py` plus a regression test) and will open a PR right after this issue.

*Disclosure: this report was prepared with AI assistance (Claude / Cursor agent); I reproduced the failure locally and reviewed every claim myself.*

codeAnqiang-ma:
## Linked issue

- Closes #1660

## Description

Guessed timezones were cached under the cleaned TZID, so only the first lookup of a globally unique TZID warned; later lookups hit the cache and stayed silent. The cache now remembers guesses and warns on hits too. It also fixes a test-order dependence in pytz-free runs (see the linked issue). A regression test fails before and passes after.

## Checklist

## Additional information

AI disclosure: drafted with AI (Claude via Cursor); reproduced, reviewed, and verified by me.


codeAnqiang-ma:
Amendments per the quality-check feedback: I shortened the commit message and the PR description, keeping the AI disclosure in the commit message and the change log entry per the Responsible AI use policy. The fix and its regression test are unchanged.

The earlier Documentation and Read the Docs failures were transient network errors while provisioning the build environment; all checks pass after the re-run.

niccokunzmann:
Thanks for the PR!

Is it normal practice to warn several times?

niccokunzmann:
To be clear for you: I did neither understand your issue and also not your PR description. But the code makes sense!

Could you in future just write your own, small descriptions?

codeAnqiang-ma:
> Is it normal practice to warn several times?

Yes — the usual practice is to emit on every occurrence and let the user's warning filters decide what is shown, like `DeprecationWarning`. With default filters Python deduplicates the display anyway: parsing the same calendar three times still prints one message per distinct TZID (verified locally). Emitting every time just keeps `pytest.warns` and `simplefilter("error")`/`"ignore"` working independently of cache state. If you'd prefer a single warning per TZID, I'm happy to change that.

> Could you in future just write your own, small descriptions?

Fair point — I'll keep future descriptions short.

src/icalendar/tests/test_issue_313_globally_unique_tzid.py
```diff
@@ -76,3 +76,19 @@ def test_timezone_id_resolves(tzp, tzid, resolved_tzid):
...
+    second lookup is served from the cache and must still emit
+    GloballyUniqueTZIDGuessed: the result is a guess no matter where it
+    comes from. See https://github.com/collective/icalendar/issues/1660
+    """
+    tzid = "/mozilla.org/20070129_1/America/New_York"
+    with pytest.warns(GloballyUniqueTZIDGuessed):
+        first = tzp.timezone(tzid)
+    with pytest.warns(GloballyUniqueTZIDGuessed):
+        second = tzp.timezone(tzid)
+    assert first is second
```
lcampanella98, line 94: I think it'd be good to pin the message as well which changed from using `lookup_id` to `guessed_id` which are different on cache hit

```suggestion
    tzid = "/mozilla.org/20070129_1/America/New_York"
    guess = "guessing it means 'America/New_York'"
    with pytest.warns(GloballyUniqueTZIDGuessed, match=guess):
        first = tzp.timezone(tzid)
    with pytest.warns(GloballyUniqueTZIDGuessed, match=guess):
        second = tzp.timezone(tzid)
    assert first is second
```

lcampanella98 (changes requested):
Nice PR. 

One suggestion on the test below, and one question on scope.

This says `Closes #1660`, but the issue reports two problems and this fixes one. `[testenv:nopytz]` still inherits `dependency_groups = test`, which lists `pytz>=2025.2`, so the `3.10 (nopytz)` job keeps running with pytz and would not catch a repeat of this. Fine with me to leave that fix out of this PR, but then better not to auto-close the issue in my opinion.

Solaris-star (commented):
Verified the current head `0213619845d5ff2525de409d70dd40c0fe1d5f1f` locally.

- `PYTHONPATH=src uv run --no-project --python 3.13 --with pytest --with python-dateutil --with tzdata --with typing-extensions --with pytz python -m pytest src/icalendar/tests/test_issue_313_globally_unique_tzid.py -q` — **22 passed**.
- The same command without `--with pytz` — **11 passed**.
- `git diff --check 0213619845d5ff2525de409d70dd40c0fe1d5f1f^ 0213619845d5ff2525de409d70dd40c0fe1d5f1f` — clean.

The cache-warning fix and regression test behave correctly in both provider configurations. I did not run the full tox/docs matrix. The remaining merge blocker appears to be scope/issue linkage rather than this code path: #1660 also tracks the `nopytz` tox dependency problem, while this PR fixes the cache-warning behavior only. Changing `Closes #1660` to a non-closing reference (and adjusting the news wording if needed) would align with the collaborator feedback.

niccokunzmann:
@Solaris-star Thanks for your review and continuous engagement with the project!
In future, when people contribute, it can be overwhelming to read a technical comment like this. Doing this has worked for me:

- Thank for the PR - we are all volunteers here, so our time is a gift!
- Give new contributors a hand in understanding the project, assess where they are at. E.g. for your comment - I do not know what to do with it - are changes needed? Do you think the issue is addressed?
- Which type of comment helps get the PR merged, not just the code but what does the person need?
- Ask a question at the end so that contributors know how to engage with your review. E.g. broadly "What do you think?"

So, how is that for you?

Solaris-star:
Thanks for the reminder — noted! Let me condense my conclusion into an actionable single point: the code at head `0213619845d5ff2525de409d70dd40c0fe1d5f1f` does fix the warning behavior for cached globally-unique TZIDs. However, #1660 also contains a separate tox configuration problem where `[testenv:nopytz]` still inherits pytz. So I would prefer this PR not auto-close all of #1660; changing it to `Part of #1660` (or `Relates to #1660`) would be more accurate, with the nopytz config handled separately.

I re-verified the current head myself:

- `PYTHONPATH=src uv run --no-project --python 3.13 --with pytest --with python-dateutil --with tzdata --with typing-extensions --with pytz python -m pytest src/icalendar/tests/test_issue_313_globally_unique_tzid.py -q` → **22 passed** (12 warnings).
- The same command without `--with pytz` → **11 passed** (6 warnings).
- `git diff --check 0213619845d5ff2525de409d70dd40c0fe1d5f1f^ 0213619845d5ff2525de409d70dd40c0fe1d5f1f` → clean.
- Tracked secret scan covered **708** tracked files with **0 hits**.

I did not run the full tox suite, docs build, or the full-matrix CI, so those checks should still be confirmed against remote CI. Going forward I'll keep reviews to short conclusions, blockers, and concrete next steps to avoid extra reading burden for contributors.

niccokunzmann:
@Solaris-star 谢谢您的回复！能否请您将其翻译成英语，并修改一下您的评论？

Thank you for the response! Could you translate it to English and edit your comment, please?

src/icalendar/timezone/tzp.py
```diff
@@ -48,6 +48,7 @@ def use_zoneinfo(self) -> None:
     def _use(self, provider: TZProvider) -> None:
         """Use a timezone implementation."""
         self.__tz_cache = {}
+        self.__guessed_ids = {}
```
SemTiOne, line 51: Good separate cache for guessed IDs. What if same cleaned ID comes from different vendor prefixes? Current logic stores first guess. Is that intended?

SemTiOne (approved):
This is a good PR. I don't see any bugs. Issue #1660 is real also.

I have 1 question below.

src/icalendar/timezone/tzp.py
niccokunzmann: Since we guess vendor id -> timezone, that should be alright. The minimal solution that covers the tests is preferable.

Does that fit for what you intended to ask?

niccokunzmann:
@codeAnqiang-ma your PR receives a lot of attention. Let us know how you would like to proceed. If you have any questions or if we can clarify sometjing, please let us know!

src/icalendar/timezone/tzp.py
SemTiOne: Yup. That's what I figured. This comment can be resolved. Once @codeAnqiang-ma applies @lcampanella98's suggestion, this is good to merge.

news/1660.bugfix
```diff
@@ -0,0 +1 @@
+Emit :class:`~icalendar.error.GloballyUniqueTZIDGuessed` on every lookup of a globally unique TZID (:rfc:`5545#section-3.2.19`), including lookups served from the TZID cache. Previously only the first lookup on a ``TZP`` instance warned: the guessed timezone was cached under the cleaned id, so repeated lookups returned silently, which also made ``test_timezone_id_resolves`` order-dependent and fail in pytz-free environments. Drafted with AI assistance; reproduced, reviewed, and verified by the author. @codeAnqiang-ma
```
stevepiercy, line 1: Minor spelling tweak.

```suggestion
Emit :class:`~icalendar.error.GloballyUniqueTZIDGuessed` on every lookup of a globally unique TZID (:rfc:`5545#section-3.2.19`), including lookups served from the TZID cache. Previously only the first lookup on a ``TZP`` instance warned: the guessed timezone was cached under the cleaned ID, so repeated lookups returned silently, which also made ``test_timezone_id_resolves`` order-dependent and fail in pytz-free environments. Drafted with AI assistance; reproduced, reviewed, and verified by the author. @codeAnqiang-ma
```

stevepiercy (changes requested):
@codeAnqiang-ma would you please take a look at the feedback and respond in kind? This is nice work. Thank you!

niccokunzmann:
@codeAnqiang-ma This PR had a bit of review. Do you have time to complete it? If not, let us know - we can still collaborate on merging it.

I think, the suggestions can just be applied and it can be merged - not much to do.

PR #1672, opened 2026-08-17
Add REQUEST-STATUS property

### LINKED ISSUE #1666: Add REQUEST-STATUS property
REQUEST-STATUS is defined in RFC 5545 section 3.8.8.3: https://www.rfc-editor.org/rfc/rfc5545#section-3.8.8.3

In the codebase it is stored under the key `RSTATUS`. It is a TEXT property that can appear in VEVENT, VTODO, VJOURNAL, and VFREEBUSY. Each value is semicolon-separated: statcode;statdesc[;extdata]. Multiple values per component are allowed.

Currently it is listed in the `multiple` tuple of Event, Todo, Journal, and FreeBusy, but there is no Python property in attr.py. The parser handles it generically. Please add a getter, setter, and deleter in the style of the other properties. Return type should be list[str]. A note that append and such do not work must be in the docs. The `multi_string_property` helper in attr.py (line 2493) already handles the get/set/del pattern for multi-value TEXT properties. See `refids_property` for an example.

Avoid long comments with AI. Write a short one yourself.

SemTiOne:
## Linked issue

- Closes #1666

## Description

Add REQUEST-STATUS property (RFC 5545 §3.8.8.3) to Event, Todo, Journal, FreeBusy via `multi_string_property`. Includes `new()` param, parametrized tests.

## Checklist

- [X] I added a change log entry, following the instructions in [Change log entry format](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-entry-format).


src/icalendar/cal/event.py
```diff
@@ -530,6 +534,7 @@ def new(
         event.transparency = transparency
         event.contacts = contacts
         event.status = status
+        event.rstatus = rstatus
```
niccokunzmann: The convention is to not shorten it.
It would be request_status.

Could you change this?

src/icalendar/tests/test_issue_843_new_method_for_components.py
```diff
@@ -961,6 +962,42 @@ def assert_component_attribute_has_value(
         True,
         "set two values",
     ),
+    (
+        COMPONENTS_RSTATUS,
+        "rstatus",
+        "RSTATUS",
```
niccokunzmann: This is the actual RFC property and dictionary key.
   Property Name:  REQUEST-STATUS
```suggestion
        "REQUEST-STATUS",
```

Same for the other occurrences.

Thanks for adding tests here and not writing new ones!

src/icalendar/tests/test_issue_843_new_method_for_components.py
```diff
@@ -961,6 +962,42 @@ def assert_component_attribute_has_value(
...
+        "2.0;Success",
+        ["2.0;Success"],
+        True,
+        "set a value",
+    ),
+    (
+        COMPONENTS_RSTATUS,
```
niccokunzmann:
```suggestion
        COMPONENTS_REQUEST_STATUS,
```

For constency, could you raname it to `REQUEST_STATUS`. It is easier to reference the RFC explicitly. We do not have a consistent shortening convention.

src/icalendar/attr.py
```diff
@@ -2561,6 +2561,42 @@ def fdel(self: Component):
 """,
 )
 
+rstatus_property = multi_string_property(
+    "RSTATUS",
```
```suggestion
REQUEST_STATUS = multi_string_property(
    "REQUEST-STATUS",
```

We have a convention for how we name the properties.
https://icalendar.readthedocs.io/en/stable/explanation/api-design.html#property-naming-convention
Because no calculation is taking place, no value checking and no conversion, this property must be an UPPER case property.

Please rename the occurrences.

I do not know how this changes the test file, yet.

src/icalendar/attr.py
```diff
@@ -2561,6 +2561,42 @@ def fdel(self: Component):
...
+        A single status string, or a list of status strings to set.
+
+Example:
+    Add a request status to an event:
+
+    .. code-block:: pycon
+
+        >>> from icalendar import Event
+        >>> event = Event.new(rstatus="2.0;Success")
+        >>> event.rstatus == ["2.0;Success"]
```
```suggestion
        >>> event = Event.new(request_status="2.0;Success")
        >>> event.REQUEST_STATUS == ["2.0;Success"]
```

niccokunzmann (changes requested):
Thanks for the PR! I left some comments. It is mostly about the naming convention that we use.

src/icalendar/attr.py
niccokunzmann: It would be good to research if anyone actually uses icalendar in this way. I think, other standards handle the update mechanism. But generally, we like to make the usage of properties nice.

E.g. for the property `request_status`, one could change the return type to a named tuple or add attributes to the vRequestStatus(vText) for message and return type. But since this is not really used, I believe, this kind of attention is more useful for other properties.

src/icalendar/cal/event.py
SemTiOne: 1

src/icalendar/cal/free_busy.py
SemTiOne: 2

src/icalendar/cal/journal.py
SemTiOne: 3

src/icalendar/cal/todo.py
SemTiOne: 4

SemTiOne (commented):
I initialized the property as `rstatus` or `RSTATUS` to be consistent with `RSTATUS` in the `multiple` tuple of todo.py, journal.py, event.py, and free_busy.py. Should these be changed to `REQUEST-STATUS` as well? @niccokunzmann

niccokunzmann:
> I initialized the property as `rstatus` or `RSTATUS` to be consistent with `RSTATUS` in the `multiple` tuple of todo.py, journal.py, event.py, and free_busy.py. Should these be changed to `REQUEST-STATUS` as well? @niccokunzmann

Ah I understand now where it comes from!

```
rstatus    = "REQUEST-STATUS"
```

So that is part of the grammar definition, so RFC internal. People do not use the grammar definitions when they use the icalendar files (That is only important for the parser, not for the people, so icalender hidden knowledge that we should not expose when not required). So, I do not want grammar variables in the code. This way, it is only the parameter names like "REQUEST-STATUS" that people see and worry about. They do not need the depth of RFC grammar logic then and it takes one more thing off their minds when they want to create valid calendars. To be friendly to readers and be explicit about what the property is for, REQUEST_STATUS is the right name.

In short: People see this:

image

And they do not need to read on to understand that rstatus and REQUEST-STATUS are the same if the property is called REQUEST_STATUS.

`"RSTATUS"` is a bug in todo.py and in the other files. It does not exist and should be corrected to "REQUEST-STATUS". Thanks for finding this in the first place. Can you make that change here, too?

```sh
$ grep -r RSTATUS src/icalendar/
src/icalendar/cal/free_busy.py:        "RSTATUS",
src/icalendar/cal/event.py:        "RSTATUS",
src/icalendar/cal/journal.py:        "RSTATUS",
src/icalendar/cal/todo.py:        "RSTATUS",
```

It must have been there for more than 10 years!

src/icalendar/cal/todo.py
niccokunzmann: Should be corrected to "REQUEST-STATUS", see https://github.com/collective/icalendar/pull/1672#issuecomment-5317883738

src/icalendar/cal/journal.py
niccokunzmann: Should be corrected to "REQUEST-STATUS", see https://github.com/collective/icalendar/pull/1672#issuecomment-5317883738

src/icalendar/cal/free_busy.py
niccokunzmann: Should be corrected to "REQUEST-STATUS", see https://github.com/collective/icalendar/pull/1672#issuecomment-5317883738

src/icalendar/cal/event.py
niccokunzmann: Should be corrected to "REQUEST-STATUS", see https://github.com/collective/icalendar/pull/1672#issuecomment-5317883738

SemTiOne:
I believe all addressed.

news/1666.feature
```diff
@@ -0,0 +1 @@
+Add :attr:`REQUEST_STATUS` property to Event, Todo, Journal, and FreeBusy with ``request_status`` parameter in ``new()`` constructors per :rfc:`5545#section-3.8.8.3`. I used AI to assist with this change. @SemTiOne
```
stevepiercy:
```suggestion
Added :attr:`REQUEST_STATUS` property to :class:`~icalendar.event.Event`, :class:`~icalendar.todo.Todo`, :class:`~icalendar.journal.Journal`, and :class:`~icalendar.free_busy.FreeBusy` with ``request_status`` parameter in ``new()`` constructors per :rfc:`5545#section-3.8.8.3`. I used AI to assist with this change. @SemTiOne
```

stevepiercy (changes requested):
Minor tweak to the news fragment to use past tense and link to the classes, then I approve. The new docstring looks great! Thank you for taking care!

niccokunzmann (dismissed):
Thank you very much for the PR!

news/1666.feature
SemTiOne: Done!

stevepiercy (approved):

Issue #1493, opened 2026-06-22
[Bug] missing file funding.json in pypi source tarball


0-wiz-0:
## Describe the bug

pkgsrc uses the pypi source tarball for packaging icalendar.
When running the test-suite, a number of tests fail because the file `funding.json` is missing.
Please add it to the pypi source tarball.

## To reproduce
```
pytest
```
gives a number of failing tests due to https://github.com/collective/icalendar/issues/731 (would be nice to get that fixed) and then
```
ERROR src/icalendar/tests/test_funding_json.py::test_some_validity_of_funding_json[$schema] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/time/py-icalendar/work/icalendar-7.1.3/funding.json'
ERROR src/icalendar/tests/test_funding_json.py::test_some_validity_of_funding_json[version] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/time/py-icalendar/work/icalendar-7.1.3/funding.json'
ERROR src/icalendar/tests/test_funding_json.py::test_some_validity_of_funding_json[entity] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/time/py-icalendar/work/icalendar-7.1.3/funding.json'
ERROR src/icalendar/tests/test_funding_json.py::test_some_validity_of_funding_json[projects] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/time/py-icalendar/work/icalendar-7.1.3/funding.json'
ERROR src/icalendar/tests/test_funding_json.py::test_some_validity_of_funding_json[funding] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/time/py-icalendar/work/icalendar-7.1.3/funding.json'
ERROR src/icalendar/tests/test_funding_json.py::test_consistent_format - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/time/py-icalendar/work/icalendar-7.1.3/funding.json'
```
## Expected behavior

All tests pass.

## Environment

 - [x] Operating system:       NetBSD 11.99.6/x86_64  
 - [x] Python version:           3.14.6 
 - [x] `icalendar` version:     7.1.3 

## Additional context

- [x] I tested it with the latest pypi version.
- [x] There is no `.ics` sample file.

Hishamkhashman1:
Opened PR #1494 for this.

[...]

PR #1680, opened 2026-08-18
Test inclusion of funding.json file

niccokunzmann:
## Linked issue

- Contributes to #1493

## Description

This adds CI tests for the inclusion of the funding.json file.

## Checklist


niccokunzmann:
Zizmor was complaining about an unpinned reference: https://github.com/collective/icalendar/pull/1680#discussion_r3803946651

I wonder: Should we pin all references?

stevepiercy:
> I wonder: Should we pin all references?

Yes, assuming that dependabot gives us automatic routine updates. It's a best security practice.

.github/workflows/tests.yml
```diff
@@ -86,22 +86,17 @@ jobs:
...
   test-distribution:
-    name: Check built package
+    name: Check built packages
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v7
-    - name: Set up Python
-      uses: actions/setup-python@v7
+    - name: Install uv and Python
+      uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
```
stevepiercy, line 94: FYI, Dependabot won't update the comment, only the hash. Even though Dependabot is trusted, we still need to manually verify that any upgrade is safe and won't break something, like dropping Python support. It's a trade-off for better security. We'll also need to remember to update the comment.

news/1493.bugfix.1
```diff
@@ -0,0 +1 @@
+Test release files to include funding.json. @niccokunzmann
```
stevepiercy:
```suggestion
Test that the released files include :file:`funding.json`. @niccokunzmann
```

src/icalendar/tests/test_create_release.sh
```diff
@@ -1,32 +1,50 @@
...
 cd "`dirname \"$0\"`"
 cd "../../.."
 
+# setup clean environment
+
 rm -rf dist
-pip3 install build
-python3 -m build
+make dev
+uv pip install twine
```
stevepiercy: Avoid `uv pip`. It's merely a wrapper around pip and doesn't gain anything.

Instead add twine to the `dev` dependency group.

```shell
uv add --group dev twine
```

Then the above line is unnecessary.

```suggestion
```

This will also help when someone new to releasing icalendar by installing everything they need to make a release.

stevepiercy: Actually, I see below that you use:

```shell
uvx run twine check dist/*
```

[`uvx twine check dist/*`](https://docs.astral.sh/uv/reference/cli/#uv-tool-run) means you don't need to add twine as a dependency or install it. See comment below.

src/icalendar/tests/test_create_release.sh
```diff
@@ -1,32 +1,50 @@
...
   exit 1
 fi
 
+if ! echo "$FILES" | grep -q '/funding.json$'; then
+  echo "ERROR: Funding files need to be included in the release files."
+  echo "       See https://github.com/collective/icalendar/issues/1493"
+  exit 1
+fi
+
+uv run twine check dist/*
```
stevepiercy: See comment above.

```suggestion
uvx run twine check dist/*
```

stevepiercy (changes requested):
Does documentation need to be updated to include a step to run the test? If so, at which step?

https://icalendar.readthedocs.io/en/latest/contribute/maintenance.html#new-releases

.github/workflows/tests.yml
niccokunzmann: Ok. More manual work!
I just wonder: we do not review the code of the actions anyway - so - I guess that can go into a discussion. So, maybe we update when (a) there is a vulnerability discovered (b) when the action's version does not work any more.

news/1493.bugfix.1
niccokunzmann: Cool. The button appeared now in the main PR view, too. 
image

src/icalendar/tests/test_create_release.sh
niccokunzmann: Ok. Let's try this. I wonder if it is also installed with uv... It seems like it.

src/icalendar/tests/test_create_release.sh
niccokunzmann: uvx is good. I had the problem that pip was not inside of the .venv - updates to how to work :)

niccokunzmann:
Hm. I would like to have a make target `make test-dist`. @stevepiercy What do you think? Should I add it here? The Ci can use that, too.

.github/workflows/tests.yml
stevepiercy: In packages I maintain, I rarely read dependency code. When I'm notified of an update, either by Dependabot or `uv tree --outdated --depth 1`, I _always_ read the dependency's release note. This practice has saved me a few times from releasing a broken package, and it helps our ecosystem. For example, from 3 days ago:

https://github.com/daniperez/vale-python-package/issues/71

stevepiercy:
> Hm. I would like to have a make target `make test-dist`. @stevepiercy What do you think? Should I add it here? The Ci can use that, too.

What would it do?

src/icalendar/tests/test_create_release.sh
```diff
@@ -1,32 +1,49 @@
...
+set -euo pipefail
 cd "`dirname \"$0\"`"
 cd "../../.."
 
+# setup clean environment
+
 rm -rf dist
-pip3 install build
-python3 -m build
+make dev
```
stevepiercy: It just occurred to me that `make init` may be the better command here. Then in the Makefile, edit the `init` target.

```
.PHONY: init
init: clean clean-python clean-dist .venv  ## Clean docs build directory, Python virtual environment, the release dist directory, and initialize Python virtual environment
```

And add a `clean-dist` target.

```
.PHONY: clean-dist
clean-dist: clean
	rm -rf dist/
```

Also consider adding `build/`. I'm not sure of your workflow to release.

stevepiercy (changes requested):
Does documentation need to be updated to include a step to run the test? If so, at which step?

https://icalendar.readthedocs.io/en/latest/contribute/maintenance.html#new-releases

https://github.com/collective/icalendar/pull/1680#pullrequestreview-4965783300

Holding off approval until these items are discussed and resolved.

.github/workflows/tests.yml
niccokunzmann: Do we need to do anything to resolve this?

src/icalendar/tests/test_create_release.sh
niccokunzmann: I wonder if this command is even needed now. uv is required to run the script but maybe, it does everything. uvx runs twine... I will check.

I would not want to do make init because it deletes so much. The script is scoped to the release files only and working on anything else will create more side effects.

niccokunzmann: Tested: This line can be removed.

[388031d](https://github.com/collective/icalendar/pull/1680/commits/388031d071ea3123ec9b510d1b26d879949e9097)

I do not understand the rest of the suggested ideas but I don't think their exploration or implementation is required for merge.

.github/workflows/tests.yml
stevepiercy: Possibly. It's not mentioned in docs to read the release notes of a bumped dependency to verify whether it's compatible. Where would that go?

- [Development](https://icalendar.readthedocs.io/en/stable/contribute/development.html)
- [Maintenance](https://icalendar.readthedocs.io/en/stable/contribute/maintenance.html)
- [Contribute](https://icalendar.readthedocs.io/en/stable/contribute/index.html)

I'm leaning toward Development, since that's when the work would happen. Contribute is more for the essential requirements to contribute, and Maintenance is too late in the workflow.

```
Bump a dependency
-----------------

When updating a dependency for icalendar to a higher version, it's a good practice to read its release notes to verify whether it's compatible with the current version of icalendar and its supported versions of Python.
The dependency might also benefit from reporting potential issues in its issue tracker.

Dependabot often notifies maintainers of dependency updates in both icalendar's source code requirements and workflows, but not in development dependencies.
For the latter, it's a good idea to periodically check them.
The following command will check all dependencies in :file:`pyproject.toml`, and return a list of their pinned versions and if there is an update available.

..  code-block:: shell

    make outdated

```

Then in Makefile:

```
.PHONY: outdated
outdated:  ## List the outdated pinned dependencies in pyproject.toml
	@uv tree --outdated --depth 1
```

What do you think?

stevepiercy: This could also be a new issue.

stevepiercy: Also does [Maintenance - New releases section](https://icalendar.readthedocs.io/en/stable/contribute/maintenance.html#new-releases) need to be updated to reflect any changes made in this PR? @niccokunzmann please answer.

stevepiercy (approved):
Changes LGTM. Check unresolved comments for any pending matters.

src/icalendar/tests/test_create_release.sh
stevepiercy: The other suggested ideas depended on running `make dev` (now obsolete), and whatever idea you had for `make test-dist`. What would it do?

niccokunzmann: I will track that in a new issue. #1706 Please add your thoughts (you can edit).

.github/workflows/tests.yml
niccokunzmann: > does [Maintenance - New releases section](https://icalendar.readthedocs.io/en/stable/contribute/maintenance.html#new-releases) need to be updated

No. This is handled by CI. We rely on the CI to check everything.

> make outdated
> ...

I like having a command for that. I think, it is worth tracking that in a separate issue.
It could later also include zizmor/workflow hashes.
Let's collect and discuss in an issue what is required to keep the package updated and how to check and maintain that.

It is kind-of a maintenance task, I feel. Could you open an issue and I add my thoughts?
If this resolves it for you, have a look if that can be merged, please.

stevepiercy: Issue created https://github.com/collective/icalendar/issues/1710. Resolving conversation.

PR #1696, opened 2026-08-19
Add RESOURCES property, update REQUEST-STATUS and REFIDS property docstring

### LINKED ISSUE #1667: Add RESOURCES property
RESOURCES is defined in RFC 5545 section 3.8.1.10: https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.10

It is a TEXT property that can appear in VEVENT and VTODO. Each value is a comma-separated list of resources.
Conformance says it should appear once per component, but multiple property parameters are allowed (e.g. for language variants).

Currently it is listed in the `multiple` tuple of Event and Todo, but there is no Python property in attr.py. The parser handles it generically. Please add a getter, setter, and deleter in the style of the other properties. Return type should be list[str]. A note that append and such do not work must be in the docs. The `multi_string_property` helper in attr.py (line 2493) already handles the get/set/del pattern for multi-value TEXT properties. See `refids_property` for an example.

Avoid long comments with AI. Write a short one yourself.

SemTiOne:
## Linked issue

- Closes #1667 
- Contributes to #1666 
- See #1672

## Description

Add RESOURCES property (RFC 5545 §3.8.1.10) to Event and Todo via  `multi_string_property`. Includes new() param, parametrized tests. Also updated REQUEST-STATUS property docstring.

## Checklist

- [X] I added a change log entry, following the instructions in [Change log entry format](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-entry-format).


src/icalendar/attr.py
```diff
@@ -2638,13 +2638,50 @@ def fdel(self: Component):
...
+
+Note:
+    List modifications do not modify the component. Methods such as
+    ``append()``, ``extend()``, and ``remove()``, as well as item
+    assignment, act on a copy. Assign the list back to the property, or
+    use :meth:`Component.add <icalendar.cal.component.Component.add>`
+    with a typed value instead.
+
+Parameters:
+    resources(str | list[str] | None):
```
niccokunzmann, line 2678: Can you have a look at the documentation? A Parameters Section might not be needed here.

src/icalendar/tests/test_issue_843_new_method_for_components.py
```diff
@@ -998,6 +999,42 @@ def assert_component_attribute_has_value(
...
+    ),
+    (
+        COMPONENTS_RESOURCES,
+        "RESOURCES",
+        "RESOURCES",
+        [],
+        [],
+        False,
+        "setting nothing",
+    ),
```
niccokunzmann:
```suggestion
```

This is duplicated

niccokunzmann (changes requested):
Technically sound! Thanks!

I will wait for the documentation review.

src/icalendar/attr.py
SemTiOne: Yes you're right. RESOURCES_property is not a function/method.

So REQUEST-STATUS_property and categories_property should not have a `Parameters:` section either? Should those be in a follow up?

All properties in attr.py don't have a `Parameters:` section except those 3.

news/1667.feature
```diff
@@ -0,0 +1 @@
+Added :attr:`RESOURCES` property to :class:`~icalendar.event.Event` and :class:`~icalendar.todo.Todo` with ``resources`` parameter in ``new()`` constructors per :rfc:`5545#section-3.8.1.10`. I used AI to assist with this change. @SemTiOne
```
stevepiercy: Let's make this less tech-ish, more user-ish, and fix links.

```suggestion
Added a ``RESOURCES`` property to both :meth:`Event.new <icalendar.cal.event.Event.new>` and :meth:`Todo.new <icalendar.cal.todo.Todo.new>` constructors through the use of a ``resources`` parameter per :rfc:`5545#section-3.8.1.10`. I used AI to assist with this change. @SemTiOne
```

src/icalendar/tests/test_issue_843_new_method_for_components.py
```diff
@@ -998,6 +999,33 @@ def assert_component_attribute_has_value(
...
+        ["EASEL"],
+        True,
+        "set a value",
+    ),
+    (
+        COMPONENTS_RESOURCES,
+        "RESOURCES",
+        "RESOURCES",
+        ["EASEL", "PROJECTOR", "VCR"],
+        ["EASEL", "PROJECTOR", "VCR"],
```
stevepiercy, line 1092: If there's a "VCR", I totally want to see a "TELEGRAPH". 😜

src/icalendar/tests/attr/test_resources.py
```diff
@@ -0,0 +1,65 @@
...
+def test_add_one_resource(component: ET):
+    """Add one resource."""
+    component.add("RESOURCES", "EASEL")
+    assert component.RESOURCES == ["EASEL"]
+
+
+def test_add_multiple_resources(component: ET):
+    """Add several resources."""
+    component.add("RESOURCES", ["EASEL", "PROJECTOR", "VCR"])
+    assert component.RESOURCES == ["EASEL", "PROJECTOR", "VCR"]
```
stevepiercy, line 37: What happens if you do multiple successive adds?

```python
    component.add("RESOURCES", "EASEL")
    component.add("RESOURCES", "TELEGRAPH")
```

src/icalendar/tests/attr/test_resources.py
```diff
@@ -0,0 +1,65 @@
...
+    component.RESOURCES = ["EASEL", "PROJECTOR", "VCR"]
+    assert component.RESOURCES == ["EASEL", "PROJECTOR", "VCR"]
+
+
+def test_delete_resources(component: ET):
+    """Delete resources."""
+    component.RESOURCES = ["EASEL"]
+    del component.RESOURCES
+    assert "RESOURCES" not in component
+    assert component.RESOURCES == []
```
stevepiercy, line 58: Can you delete one of many resource items? I assume that you can't do it by key, but only by index, if at all. The docstring for RESOURCES implies it's not possible, but not what would happen with something like this?

```python
    component.RESOURCES = ["EASEL", "TELEGRAPH"]
    del component.RESOURCES[0]
    assert component.RESOURCES == ???
```

src/icalendar/tests/attr/test_resources.py
```diff
@@ -0,0 +1,65 @@
+"""This tests the RESOURCES property.
+
+See
+- https://github.com/collective/icalendar/issues/1667
+- https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.10
```
stevepiercy, line 5: I noticed that you use https://www.rfc-editor.org/ instead of https://datatracker.ietf.org/. The former appears to be the "official home of RFCs", whereas all of our RFC links auto-generate to the latter. It wouldn't be the first time that the hostname has changed. Anyway, just a comment, nothing to do here. I might open an issue with Sphinx.

stevepiercy (changes requested):
Excellent work! I suggested a rewrite of the news fragment, and asked a couple of questions about tests and functionality. Would you please take care? Thank you!

src/icalendar/tests/attr/test_resources.py
SemTiOne: They work too. Multiple successive add() calls accumulate, so you'd get ["EASEL", "TELEGRAPH"].

src/icalendar/tests/attr/test_resources.py
SemTiOne: Yes, you can delete by index. Deleting by key would raise KeyError. After assignment or successive adds, `del component.RESOURCES[0]` modifies the component directly. I think it's worth 2 tests here such as successive_adds test and delete_by_index test. Also yes the docstring is inaccurate.

stevepiercy: Yes, please, tests covering both my questions would be welcome and helpful. Is the docstring still correct about `remove()` and others?

src/icalendar/attr.py
niccokunzmann: This is your choice, just let me know of your decision - this PR or another or an issue. All are fine.

SemTiOne: I'll add the fix for those two in this PR since it's a small change.

src/icalendar/tests/test_resources.py
```diff
@@ -0,0 +1,65 @@
...
+    component.RESOURCES = ["EASEL", "PROJECTOR", "VCR"]
+    assert component.RESOURCES == ["EASEL", "PROJECTOR", "VCR"]
+
+
+def test_delete_resources(component: ET):
+    """Delete resources."""
+    component.RESOURCES = ["EASEL"]
+    del component.RESOURCES
+    assert "RESOURCES" not in component
+    assert component.RESOURCES == []
```
SemTiOne: Yes, `remove()` and other list mutations do modify the component when the property holds a list. The old Note was inaccurate.

src/icalendar/tests/attr/test_resources.py
niccokunzmann: Hm. I think there was a comment.

Here, it is true that the docstring is wrong if the value can be changed. That should be documented and tested.
Other properties also use the same constructor - what do their tests and doc strings state?
Hm. I wonder though if that is good to not operate on a copy. The problem is the conversion to and from ical. What happens to escaping - is vText used? Is the string just encoded and copied in directly?
I remember that with categories, there is mutability of the list.
The problem is with the types, too. Promised is list[str] and not list[vText] which is ok.

SemTiOne: > Yes, please, tests covering both my questions would be welcome and helpful. Is the docstring still correct about remove() and others?

No, I just noticed that the docstring is inaccurate on the REQUEST-STATUS too. List mutations (append, extend, remove, item assignment, del) should modify the component when the property holds a list. Thanks for pointing that out. I'll edit them.

stevepiercy: @niccokunzmann should a new issue be created to answer your questions, or should they be answered before merging this PR?

niccokunzmann: Since this PR does not touch the code that needs investigating, these questions should be addressed in a separate issue.

I will have a technical review now.

src/icalendar/attr.py
```diff
@@ -2638,13 +2631,46 @@ def fdel(self: Component):
...
+facilities, or other things that the component needs. Each item of the
+list becomes its own RESOURCES property value. See
+:rfc:`5545#section-3.8.1.10` for the specification.
+
+Note:
+    After assigning a list or adding several values, the returned list
+    is the same object that stores the values. Modifying it (``append()``,
+    ``extend()``, ``remove()``, item assignment, or ``del``) changes the
+    component. If the property holds a single value, the returned list is
+    a new wrapper and modifications are lost.
```
niccokunzmann: Thank you! This seems like an accurate description of the behavior. I think, it is worth opening an issue that addresses the inconsistency.

niccokunzmann (dismissed):
A small request: Could you move the test file into tests/attr?

I try to gather all tests there - if you feel like it, that is a contribution for a new PR to move these tests for the attr.py content into that directory.

Also: It seems to me like you are really good at writing documentation... I invite you to have a look at the code in attr.py: #1244 clearly says that the way I approached documentation by copy&paste from the RFC is now obsolete and the way you write it now is much better.

SemTiOne:
Done

src/icalendar/attr.py
stevepiercy: @niccokunzmann what's the inconsistency? What would need to be done to resolve an issue?

Honestly, I don't understand this note, and I read it several times.

src/icalendar/attr.py
stevepiercy: I don't understand why Parameters was removed. @niccokunzmann can you elaborate?

It's OK to keep them, even if they're not directly user-accessible through published HTML documentation, because it helps explain to the reader of code what types are acceptable.

src/icalendar/attr.py
```diff
@@ -705,9 +705,6 @@ def _del_categories(component: Component) -> None:
     At present, icalendar doesn't take the LANGUAGE parameter as defined
     in :rfc:`5545#section-3.2.10` into account.
 
-Parameters:
-    categories(list[str]): A list of categories as strings.
-
```
stevepiercy, line 712: See https://github.com/collective/icalendar/pull/1696#discussion_r3836801120

src/icalendar/attr.py
```diff
@@ -2616,15 +2613,11 @@ def fdel(self: Component):
...
 Note:
-    List modifications do not modify the component. Methods such as
-    ``append()``, ``extend()``, and ``remove()``, as well as item
-    assignment, act on a copy. Assign the list back to the property, or
-    use :meth:`Component.add <icalendar.cal.component.Component.add>`
-    with a typed value instead.
-
-Parameters:
-    request_status(str | list[str] | None):
-        A single status string, or a list of status strings to set.
```
stevepiercy, line 2624: See https://github.com/collective/icalendar/pull/1696#discussion_r3836801120.

src/icalendar/attr.py
```diff
@@ -2616,15 +2613,11 @@ def fdel(self: Component):
...
+    After assigning a list or adding several values, the returned list
+    is the same object that stores the values. Modifying it (``append()``,
+    ``extend()``, ``remove()``, item assignment, or ``del``) changes the
+    component. If the property holds a single value, the returned list is
+    a new wrapper and modifications are lost.
```
stevepiercy: I'm struggling to comprehend this note.

- Assigning a list to what?
- Why is `Parameters` removed? Without it, I wouldn't be able to easily see what are the acceptable parameter types, which in turn makes this paragraph more difficult to comprehend.
- @niccokunzmann mentioned an inconsistency, but I can't determine between what two things.
- What changes in the component after modifying it? The second sentence is hard for me to parse, as I'm not clear what is being modified: the list, property, or the component.

I wish I could provide an alternative suggestion, but I'm lost.

src/icalendar/attr.py
```diff
@@ -2638,13 +2631,46 @@ def fdel(self: Component):
 """,
 )
 
+RESOURCES_property = multi_string_property(
```
stevepiercy, line 2657: Just a note that my comments above also apply to this property.

stevepiercy (changes requested):

src/icalendar/attr.py
SemTiOne: 1. To this property, for example `event.RESOURCES = [...]`. I can make it explicit: “When you assign a list to this property (for example `event.RESOURCES = [...]`)”.

2. Google style reserves `Parameters:` for function and method arguments (style-guide.rst line 60 → “following the Google Python Style Guide” + line 312 → “Each parameter should consist of its name and a brief description” with the example `def some_function(creq, breq…:326)`. These objects are property descriptors, while `sphinx_autodoc_typehints` (style-guide.rst line 66) is described at line 264 as documenting “return value types of functions”. If you want, I can move the accepted types into the description itself like this: “You can assign a single string, a list of strings, or None to this property. The property stores a str as-is.” or I can restore the `Parameters:` block if you prefer it for code readers.

3. Old doc at branch main, attr.py line 2619 said “List modifications do not modify the component … act on a copy.” The current `multi_string_property` (attr.py line 2529) shares the list object, so `append()` etc. do modify that component instance. That mismatch surfaced while testing `del`/`remove()` for RESOURCES.

4. The returned list is what you modify (for example `event.RESOURCES.append("EASEL")`), and that modifies the `Event`'s stored property values. If the property holds a single value, the returned list is a temporary copy, so modifying it does not affect the `Event`.

What do you think? Let me know.

src/icalendar/attr.py
stevepiercy: After further review, I found that we should preserve the Parameters section. It's very important, as without it, the user won't know what to pass into the `*.categories` property.

- [Available.categories](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.available.html#icalendar.cal.available.Available.categories)
- [Availability.categories](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.availability.html#icalendar.cal.availability.Availability.categories)
- [Calendar.categories](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.calendar.html#icalendar.cal.calendar.Calendar.categories)
- [Event.categories](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.event.html#icalendar.cal.event.Event.categories)
- [Journal.categories](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.journal.html#icalendar.cal.journal.Journal.categories)
- [Todo.categories](https://icalendar.readthedocs.io/en/stable/reference/api/icalendar.cal.todo.html#icalendar.cal.todo.Todo.categories)

src/icalendar/attr.py
stevepiercy: This section heading should also be preserved, but edited. I can't make suggestions on deletions, but I'd do this:

```rst
Parameters:
    request_status(str | list[str] | None):
        Either a single status string, a list of status strings, or ``None`` to set the component's status code returned for a scheduling request.
```

src/icalendar/attr.py
```diff
@@ -2616,15 +2613,11 @@ def fdel(self: Component):
 The return status components are defined in :rfc:`5545#section-3.8.8.3`.
```
stevepiercy: This helps cross-reference the property across components, and corrects an omission.

```suggestion
The value consists of a short return status code component, a longer
return status description component, and optionally a status-specific
data component, separated by semicolons (statcode;statdesc[;extdata]).
The return status components are defined in :rfc:`5545#section-3.8.8.3`.

The REQUEST-STATUS property can be specified in the following icalendar components as ``REQUEST_STATUS``.

-   :attr:`Event.REQUEST_STATUS <icalendar.cal.event.Event.REQUEST_STATUS>`
-   :attr:`FreeBusy.REQUEST_STATUS <icalendar.cal.free_busy.FreeBusy.REQUEST_STATUS>`
-   :attr:`Journal.REQUEST_STATUS <icalendar.cal.journal.Journal.REQUEST_STATUS>`
-   :attr:`Todo.REQUEST_STATUS <icalendar.cal.todo.Todo.REQUEST_STATUS>`
```

src/icalendar/attr.py
stevepiercy: 1. Explicit is better.
2. See above suggestions for more precise wording for `Parameters`, and an explanation of how the docstrings of properties get displayed in their components.
3. OK. I'm not sure if it's been resolved. Is there a separate issue to correct the mismatch, or has that already been resolved in this PR? Either option is good.
4. Let's see what you come up with. I might need to see this in code examples. It's not sinking into my brain as is. Wordsmithing is a challenge here.

stevepiercy (changes requested):
Thanks for your collaboration. I have clarity on `Parameters` now and offered suggestions. I'm still murky on its behavior, but let's see if your next draft will sink in.

src/icalendar/attr.py
```diff
@@ -2613,18 +2610,24 @@ def fdel(self: Component):
 The value consists of a short return status component, a longer
 return status description component, and optionally a status-specific
 data component, separated by semicolons (statcode;statdesc[;extdata]).
```
stevepiercy: Somehow this wasn't replaced by the ensuing three lines. I blame GitHub's suggestion UI which is as buggy as a dog turd.

```suggestion
```

src/icalendar/attr.py
```diff
@@ -2613,18 +2610,24 @@ def fdel(self: Component):
...
+The REQUEST-STATUS property can be specified in the following icalendar components as ``REQUEST_STATUS``.
+
+-   :attr:`Event.REQUEST_STATUS <icalendar.cal.event.Event.REQUEST_STATUS>`
+-   :attr:`FreeBusy.REQUEST_STATUS <icalendar.cal.free_busy.FreeBusy.REQUEST_STATUS>`
+-   :attr:`Journal.REQUEST_STATUS <icalendar.cal.journal.Journal.REQUEST_STATUS>`
+-   :attr:`Todo.REQUEST_STATUS <icalendar.cal.todo.Todo.REQUEST_STATUS>`
 
-Parameters:
-    request_status(str | list[str] | None):
-        A single status string, or a list of status strings to set.
```
stevepiercy, line 2624: Please restore these lines. They're necessary.

src/icalendar/attr.py
```diff
@@ -2638,13 +2641,46 @@ def fdel(self: Component):
...
+list becomes its own RESOURCES property value. See
+:rfc:`5545#section-3.8.1.10` for the specification.
+
+Note:
+    After assigning a list or adding several values, the returned list
+    is the same object that stores the values. Modifying it (``append()``,
+    ``extend()``, ``remove()``, item assignment, or ``del``) changes the
+    component. If the property holds a single value, the returned list is
+    a new wrapper and modifications are lost.
+
```
stevepiercy, line 2690: There should be a `Parameters` section heading here, too.

src/icalendar/attr.py
```diff
@@ -2638,13 +2641,46 @@ def fdel(self: Component):
...
+    """This property defines resources for a calendar component.
+
+Setting this property replaces all existing RESOURCES values. A :class:`str`
+is stored as-is. Setting ``None`` or an empty list removes all
+RESOURCES values, as does deleting the property.
+
+The value is a comma-separated list of resources, such as equipment,
+facilities, or other things that the component needs. Each item of the
+list becomes its own RESOURCES property value. See
+:rfc:`5545#section-3.8.1.10` for the specification.
```
stevepiercy, line 2668:
```suggestion
:rfc:`5545#section-3.8.1.10` for the specification.

The RESOURCES property can be specified in the following icalendar components as ``resources`` using the component's ``new()`` constructor.

-   :attr:`Event.RESOURCES <icalendar.cal.event.Event.RESOURCES>`
-   :attr:`Todo.RESOURCES <icalendar.cal.todo.Todo.RESOURCES>`
```

stevepiercy (changes requested):
Another round of docstring amendments.

I would still like to see a rewrite of the `Note` sections for clarity per the discussion in https://github.com/collective/icalendar/pull/1696#discussion_r3836804393.

src/icalendar/attr.py
```diff
@@ -2606,25 +2608,36 @@ def fdel(self: Component):
...
-    List modifications do not modify the component. Methods such as
-    ``append()``, ``extend()``, and ``remove()``, as well as item
-    assignment, act on a copy. Assign the list back to the property, or
-    use :meth:`Component.add <icalendar.cal.component.Component.add>`
-    with a typed value instead.
+    When you assign a list to this property, the returned list
+    is the same object stored in the component. Modifying it (``append()``,
+    ``extend()``, ``remove()``, item assignment, or ``del``) changes what
+    the component stores. If the property holds a single value, the returned
+    list is a temporary copy, and changes to it do not affect the component.
```
stevepiercy: Now I get it! Basically, a single value is an immutable string, and a list is a mutable collection of elements.

I think two paragraphs will help my brain pause and process each of these logical points. Space is important to comprehension. I also structured the second paragraph to follow the pattern of the first, so the comparison is more obvious.

Also apply this to the RESOURCES property.

Thanks for taking the time to work through this with me. I *really* appreciate it.

```suggestion
    When you assign a list to this property, the returned list
    is the same object stored in the component. Modifying it (``append()``,
    ``extend()``, ``remove()``, item assignment, or ``del``) changes what
    the component stores.

    However, if you assign a single string value to this property, the returned
    list is a temporary copy, and changes to this list don't affect the component.
```

stevepiercy (changes requested):
This is much clearer to me. Thank you! Please take a look, and let me know.

src/icalendar/attr.py
```diff
@@ -2638,13 +2653,58 @@ def fdel(self: Component):
...
+    resources(str | list[str] | None):
+        Either a single resource string, a list of resource strings,
+        or ``None`` to set the component's resources.
+
+Note:
+    When you assign a list to this property, the returned list
+    is the same object stored in the component. Modifying it (``append()``,
+    ``extend()``, ``remove()``, item assignment, or ``del``) changes what
+    the component stores. If the property holds a single value, the returned
+    list is a temporary copy, and changes to it do not affect the component.
```
stevepiercy: Copy from above note.

```suggestion
    When you assign a list to this property, the returned list
    is the same object stored in the component. Modifying it (``append()``,
    ``extend()``, ``remove()``, item assignment, or ``del``) changes what
    the component stores.

    However, if you assign a single string value to this property, the returned
    list is a temporary copy, and changes to this list don't affect the component.
```

stevepiercy (changes requested):
Copy the note, and I approve. Thank you!

src/icalendar/attr.py
SemTiOne: 1. Done
2. Done
3. Only REQUEST_STATUS_property and refids_property is inconsistent. REQUEST-STATUS is done. Now REFIDS is done too.
4. Code example:
```python
from icalendar import Event
event = Event()
event.RESOURCES = ["EASEL", "PROJECTOR"]
event.RESOURCES.append("VCR")
event.RESOURCES == ["EASEL", "PROJECTOR", "VCR"]
Output: True

event2 = Event()
event2.RESOURCES = "EASEL"
event2.RESOURCES.append("PROJECTOR")
event2.RESOURCES == ["EASEL"]
Output: True
```

niccokunzmann: Thanks for the code example!

We should make this consistent. Nobody expects the behavior of a list to be different when it is one element and when it has two elements. I count this as a bug. I will open an issue for that. This PR is about adding the properties and not fixing the bug, so let's keep it like this.

Would you be interested in working on the other issue, too?

Issue: https://github.com/collective/icalendar/issues/1722

SemTiOne: Yes I am interested.

niccokunzmann: @SemTiOne Could you join the discussion in #1722?

For this PR, it is ok to keep the inconsistent behavior but add/correct what is already there.

niccokunzmann:
@SemTiOne Let us know when we should have a look again!

src/icalendar/attr.py
stevepiercy: I added the code examples to the description in #1722. They're worth a thousand words.

src/icalendar/attr.py
```diff
@@ -687,6 +687,8 @@ def _del_categories(component: Component) -> None:
 
 Within the calendar components, specify categories as a list of strings.
 You can get, set, and delete categories for a component.
+Assign a list of strings, a single string, or ``None`` to this property;
+assigning ``None`` or an empty list removes the categories.
```
stevepiercy: I'd remove this addition. It conflicts with both the previous sentence and the `Parameters` section below, it's feature creep, and we have a new issue to handle this buggy behavior.

```suggestion
```

stevepiercy (changes requested):
Good to copy the note to other `multi_string_property` properties.

However, the addition to `categories` should be handled in #1722.

stevepiercy (dismissed):
Thank you!

stevepiercy:
@niccokunzmann would you please do a final review?

stevepiercy (approved):

niccokunzmann (approved):
Great! Ready for merge 🥳

PR #1698, opened 2026-08-19
Chore/type checking imports

Priyanshu-pulak:
Used ChatGPT-5.6 to assist with drafting the PR
- Refs #1707

## Description

This PR refactors the type-checking imports in `icalendar.cal`.

The imports are updated to use the appropriate `icalendar.cal` package exports for type checking, while keeping the imports inside the `TYPE_CHECKING` block.

## Checklist

## Additional information

No additional information.


stevepiercy:
I think this PR has nothing to do with #938. It's a refactor.

I also think this change makes the code less obvious where the imported class actually resides.

@Priyanshu-pulak can you please elaborate why you think this change is an improvement?

Priyanshu-pulak:
Thanks for pointing this out. I agree that this is a refactor rather than a chore.

The motivation for the change was to make the `TYPE_CHECKING` imports use the public exports of `icalendar.cal`, which are defined in `__init__.py`, rather than importing each class directly from its implementation module.

Also there is a trade-off in readability, as importing directly from the implementation module makes the location of the class more explicit. However, using the public exports avoids coupling these imports to the internal module structure and keeps them consistent with the public API of `icalendar.cal`. This could make future internal reorganization easier without requiring changes to these imports.

stevepiercy:
Thanks for the explanation. I'm not sure which option is more valuable. That's a @niccokunzmann or @SashankBhamidi design question.

In the meanwhile, would you please remove the mention of #938 from the description, since it's unrelated to this PR, even if it inspired it? Also rename the news fragment to an orphan, per https://icalendar.readthedocs.io/en/stable/contribute/index.html#change-log-entry-file-name? Thank you!

niccokunzmann:
Thanks for the PR!

I think the change is fine. If you find it is an improvement to type checking readability, that is fine!

I was concerned about import order but since this is in the type checking block and not before that, this should work out.

Also, I tend to think like this: If it does not break functionality and you are willing to put the time in, it is probably valuable to you.

So, it can be merged from my point of view. Could you change the issue description and the news fragment as suggested?

Priyanshu-pulak:
@niccokunzmann I have made the changes which have mentioned. Can you please tell me why `check news fragment filenames` tests fails even i am using the naming notation which is mentioned in change log entry file name for `orphan change log`?
Is it good if I open an issue for this as there are many modules in which we can refactor it?

niccokunzmann:
If you like to track changes in one place and create more PRs, an issue is a good idea.

niccokunzmann:
I think, .refactor is not an ending that is used. .chore would be it.

Priyanshu-pulak:
Thanks for specifying the reason and i rename the file according to this. In my opinion if `.refactor` is also used in ending in change log.

niccokunzmann:
.refactor is not listed: https://icalendar.readthedocs.io/en/latest/contribute/#change-log-types
I think, the cosmetic changes are good for developers of icalendar and readers that only use icalendar do not need to know that as they are not affected. So, I vote for .chore.

niccokunzmann:
Would you like to open an issue now and link this as the first PR?

Priyanshu-pulak:
I had open and issue & link this PR

niccokunzmann (dismissed):
Thank you! A clear code contribution. Thanks for opening the issue to reference the changes and track conversation in one place.

news/1707.chore
```diff
@@ -0,0 +1 @@
+Simplified type-checking imports in :mod:`icalendar.cal.calendar`. @Priyanshu-pulak
```
stevepiercy:
```suggestion
Simplified type checking imports in :mod:`icalendar.cal.calendar`. @Priyanshu-pulak
```

stevepiercy (changes requested):
Thanks for the PR and following our guidelines. LGTM, with this one edit that I'll commit, then I'll auto-merge.

stevepiercy (approved):

stevepiercy:
@Priyanshu-pulak does this PR close or merely reference #1707?

Priyanshu-pulak:
This PR is reference #1707.
There are still some modules in which there is need of refactor of TYPE_CHECK imports

stevepiercy:
@Priyanshu-pulak would you please add a checklist of remaining modules that need to be refactored in #1707, including this item checked off? That would be super helpful for us to track when it's complete.

In the description, edit it with a check list:

```md
- [ ] todo 1
- [ ] todo 2
- [ ] todo N
```

Thank you!

PR #1723, opened 2026-08-24
Reject control characters in TEXT values on parse

abhijeet117:
﻿## Summary
RFC 5545 forbids control characters in TEXT values, but vText.from_ical accepted them, so a NUL in a property value round-tripped into serialized output where C-based consumers silently truncate it. TEXT values are now validated against the CONTROL production on parse: invalid values raise ValueError, which error-tolerant components like Event collect in errors as vBroken, matching how invalid parameter values are already handled.

## Testing
Reproduced with Event.from_ical of a VEVENT whose SUMMARY contained A<NUL>B: the NUL passed through to to_ical(). Added regression tests covering forbidden and valid characters, error collection for tolerant components, strict raising for Calendar, and escaped newline handling; they fail without the fix.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

### DIFF
--- a/news/1712.bugfix
+++ b/news/1712.bugfix
@@ -1,1 +1,1 @@
-TEXT property values now reject :rfc:`5545` control characters such as NUL on parse instead of passing them through to serialized output. Invalid lines are collected in ``errors`` or raise, like invalid parameter values. @abhijeet117 (`Issue #1712 <https://github.com/collective/icalendar/issues/1712>`_)
+TEXT property values now reject :rfc:`5545` control characters such as NUL on parse instead of passing them through to serialized output. Invalid lines are collected in ``errors`` or raise, like invalid parameter values. @abhijeet117
--- a/news/1712.bugfix
+++ b/news/1712.bugfix
@@ -1,0 +1,1 @@
+TEXT property values now reject :rfc:`5545` control characters such as NUL on parse instead of passing them through to serialized output. Invalid lines are collected in ``errors`` or raise, like invalid parameter values. @abhijeet117 (`Issue #1712 <https://github.com/collective/icalendar/issues/1712>`_)
--- a/src/icalendar/prop/text.py
+++ b/src/icalendar/prop/text.py
@@ -9,11 +9,20 @@ from icalendar.parser import Parameters, _escape_char
 from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE, to_unicode
 
 # :rfc:`5545#section-3.3.11` defines TEXT as
-# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` and
-# ``CONTROL = %x00-08 / %x0A-1F / %x7F``, so no control character except the
-# horizontal tab may appear in a TEXT value. The line feed is additionally
-# accepted here because it is the result of the escaped sequences ``\N``
-# and ``\n``.
+# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` where TSAFE-CHAR is in turn
+# defined by the following grammar in :rfc:`5545#section-3.1`.
+#
+# ..  code-block:: text
+#
+#     TSAFE-CHAR = WSP / %x21 / %x23-2B / %x2D-39 / %x3C-5B /
+#              %x5D-7E / NON-US-ASCII
+#        ; Any character except CONTROLs not needed by the current
+#        ; character set, DQUOTE, ";", ":", "\", ","
+#
+# CONTROL is defined in the same section as ``%x00-08 / %x0A-1F / %x7F``, so no
+# control character except the horizontal tab may appear in a TEXT value.
+# The line feed, ``\x0a``, is additionally accepted here because it is the
+# result of the escaped sequences ``\N`` and ``\n``.
 UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
 
 
--- a/src/icalendar/prop/text.py
+++ b/src/icalendar/prop/text.py
@@ -1,11 +1,20 @@
 """TEXT values from :rfc:`5545`."""
 
+import re
 from typing import Any, ClassVar
 
 from icalendar.compatibility import Self
 from icalendar.error import JCalParsingError
 from icalendar.parser import Parameters, _escape_char
 from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE, to_unicode
+
+# :rfc:`5545#section-3.3.11` defines TEXT as
+# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` and
+# ``CONTROL = %x00-08 / %x0A-1F / %x7F``, so no control character except the
+# horizontal tab may appear in a TEXT value. The line feed is additionally
+# accepted here because it is the result of the escaped sequences ``\N``
+# and ``\n``.
+UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
 
 
 class vText(str):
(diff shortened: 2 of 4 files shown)


src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -0,0 +1,69 @@
+r"""TEXT values must reject control characters on parse.
+
+:rfc:`5545#section-3.3.11` defines TEXT as
+``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` with
+``CONTROL = %x00-08 / %x0A-1F / %x7F``, so control characters other than
+the horizontal tab cannot appear in a valid TEXT value. A NUL byte in a
```
stevepiercy: Let's be explicit and not omit the grammars. It helps me understand the test, without having to jump around the RFC, although references are present to verify.

Please rewrap as needed.

```suggestion
:rfc:`5545#section-3.3.11` defines TEXT as
``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` where TSAFE-CHAR is defined by the following grammar.

..  code-block:: text

    TSAFE-CHAR = WSP / %x21 / %x23-2B / %x2D-39 / %x3C-5B /
             %x5D-7E / NON-US-ASCII
       ; Any character except CONTROLs not needed by the current
       ; character set, DQUOTE, ";", ":", "\", ","

In turn, CONTROL is defined by the following grammar in :rfc:`5545#section-3.1`.

..  code-block:: text

    CONTROL       = %x00-08 / %x0A-1F / %x7F
    ; All the controls except HTAB

Although "the current character set" may be one of many, it is UTF-8 by default
per :rfc:`5545#section-3.1.4`. Thus TEXT should reject all the CONTROLs except
the horizontal tab, %x09. Additionally the new line, %x0A, is needed by the current character set because it is the result of the escaped sequences ``\N`` and ``\n``.

A NUL byte in a
```

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -0,0 +1,69 @@
...
+from icalendar import Calendar, Event, vText
+from icalendar.prop import vBroken
+
+FORBIDDEN_VALUES = [
+    "A\x00B",  # NUL from the issue report
+    "\x00",
+    "\x01",
+    "\x08",
+    "\x0b",  # vertical tab
+    "\x0c",  # form feed
```
stevepiercy: I'd like to see the full range of disallowed CONTROLs. Maybe a Python generator would work to get the wider range, deleting the HTAB, and adding the others as onesies?

Also add a note that the new line `\x0a` is not forbidden.

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -0,0 +1,69 @@
...
+    "a\rb",  # lone carriage return
+    "\x7f",  # delete
+]
+
+VALID_VALUES = [
+    "Hello World!",
+    'quotes " and colons : and semicolons ; and commas ,',
+    r"escapes \; \, \\ stay verbatim here",
+    "horizontal tab\tinside",
```
stevepiercy, line 55: I'm curious whether its hex equivalent will pass.

```suggestion
    "horizontal tab\tinside",
    "horizontal tab as hex\x09inside",
```

src/icalendar/prop/text.py
```diff
@@ -1,12 +1,21 @@
...
 
 from icalendar.compatibility import Self
 from icalendar.error import JCalParsingError
 from icalendar.parser import Parameters, _escape_char
 from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE, to_unicode
 
+# :rfc:`5545#section-3.3.11` defines TEXT as
+# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` and
+# ``CONTROL = %x00-08 / %x0A-1F / %x7F``, so no control character except the
+# horizontal tab may appear in a TEXT value. The line feed is additionally
```
stevepiercy:
```suggestion
# horizontal tab may appear in a TEXT value. The line feed, ``\x0a``, is additionally
```

src/icalendar/prop/text.py
```diff
@@ -1,12 +1,21 @@
...
+import re
 from typing import Any, ClassVar
 
 from icalendar.compatibility import Self
 from icalendar.error import JCalParsingError
 from icalendar.parser import Parameters, _escape_char
 from icalendar.parser_tools import DEFAULT_ENCODING, ICAL_TYPE, to_unicode
 
+# :rfc:`5545#section-3.3.11` defines TEXT as
+# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` and
```
stevepiercy:
```suggestion
# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)`` where TSAFE-CHAR is in turn defined as
```

news/1712.bugfix
```diff
@@ -0,0 +1 @@
+TEXT property values now reject :rfc:`5545` control characters such as NUL on parse instead of passing them through to serialized output. Invalid lines are collected in ``errors`` or raise, like invalid parameter values. @abhijeet117 (`Issue #1712 <https://github.com/collective/icalendar/issues/1712>`_)
```
stevepiercy: Towncrier by virtue of the filename will automatically link to the issue.

```suggestion
TEXT property values now reject :rfc:`5545` control characters such as NUL on parse instead of passing them through to serialized output. Invalid lines are collected in ``errors`` or raise, like invalid parameter values. @abhijeet117
```

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -0,0 +1,69 @@
...
+    "line\nbreak from the escaped sequences",
```
stevepiercy: A line break is `CRLF` per https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.1. A newline is encoded as `\n` or `\N` per https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.3.11.

```suggestion
    "newline\nfrom the escaped sequences",
```

stevepiercy (changes requested):
Thank you for your contribution. Technically, this looks good to me. There are some comments that I'd like to see expanded. Would you please take care?

abhijeet117:
Thank you for the thorough review! I applied all of the suggestions:

- The test module docstring now spells out the full TSAFE-CHAR and CONTROL grammars with the UTF-8 default note.
- The forbidden characters are now generated from the RFC ranges, so every CONTROL is covered except the horizontal tab, and a comment notes that the new line is allowed because of the ``\N`` and ``\n`` escape sequences.
- Added the hex-escaped horizontal tab to the valid values.
- Reworded the comments in prop/text.py along the same lines.
- Removed the manual issue link from the news entry so towncrier links it by filename.

abhijeet117:
@stevepiercy Thanks for the detailed suggestions. They are all applied in the latest commit: the test docstring spells out both grammars, the forbidden characters come from the RFC ranges, the hex tab and escaped newline cases were added, the comments in prop/text.py follow your wording, and the news entry lets towncrier link the issue. I also rebased the branch on current main so it merges cleanly.

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -0,0 +1,92 @@
...
+import pytest
+
+from icalendar import Calendar, Event, vText
+from icalendar.prop import vBroken
+
+#: Every CONTROL from :rfc:`5545#section-3.1`. The horizontal tab, ``%x09``,
+#: is not a CONTROL, and the new line, ``%x0A``, is not forbidden because the
+#: escaped sequences ``\N`` and ``\n`` produce it.
+FORBIDDEN_CONTROL_CHARS = [
+    chr(control) for control in range(0x20) if control not in (0x09, 0x0A)
```
stevepiercy, line 42: Confirming this works. Nice!

```pycon
>>> FORBIDDEN_CONTROL_CHARS = [
...     chr(control) for control in range(0x20) if control not in (0x09, 0x0A)
... ] + ["\x7f"]
>>> print(FORBIDDEN_CONTROL_CHARS)
['\x00', '\x01', '\x02', '\x03', '\x04', '\x05', '\x06', '\x07', '\x08', '\x0b', '\x0c', '\r', '\x0e', '\x0f', '\x10', '\x11', '\x12', '\x13', '\x14', '\x15', '\x16', '\x17', '\x18', '\x19', '\x1a', '\x1b', '\x1c', '\x1d', '\x1e', '\x1f', '\x7f']
```

stevepiercy (approved):
Thanks for the updates. I learned a few new things along the way, too.

Issue #1707, opened 2026-08-20
Refactor Type Checking Imports


Priyanshu-pulak:
## Description

Refactor the `TYPE_CHECKING` imports in `icalendar` to use the public exports from `icalendar.cal` instead of importing classes directly from their implementation modules.

Currently, some `TYPE_CHECKING` imports depend on the internal module structure of `icalendar.cal`. Since these classes are already publicly exported through `icalendar.cal.__init__.py`, the imports can be updated to use those public exports.

## Motivation

Using the public exports keeps `TYPE_CHECKING` imports consistent with the public API of `icalendar.cal` and reduces coupling to the internal module structure.

This could also make future internal reorganization easier, as changes to the implementation modules would not necessarily require updating these imports.

## Trade-off

There is a readability trade-off with this approach. Importing directly from the implementation module makes the location of the class more explicit. However, using the public exports avoids coupling these imports to the current internal module layout and keeps the imports aligned with the public API.

## Proposed Changes

* Update relevant `TYPE_CHECKING` imports to import classes from `icalendar.cal`.
* Remove unnecessary direct imports from internal implementation modules where appropriate.
* Ensure there are no runtime import changes, since these imports are only used for type checking.
* Run the project's existing tests and linting/type-checking checks to verify that the refactor does not introduce any issues.

## Proposed Modules to Refactor
- [ ] `src/icalendar/cal/alarms.py`
- [x]  `src/icalendar/cal/availability.py` (#1751)
- [x] `src/icalendar/cal/calendar.py` (#1698)
- [ ] `src/icalendar/cal/event.py`
- [ ] `src/icalendar/cal/free_busy.py`
- [ ] `src/icalendar/cal/journal.py`
- [x]  `src/icalendar/cal/lazy.py` (#1751)
- [x] `src/icalendar/cal/timezone.py` (#1751)
- [ ] `src/icalendar/cal/todo.py`
- [ ]  `src/icalendar/timezone/zoneinfo.py`
## Expected Outcome

The `TYPE_CHECKING` imports should consistently use the public exports where appropriate, reducing dependency on the internal module structure while preserving existing behaviour.

stevepiercy:
@Priyanshu-pulak would you please add a checklist of remaining modules that need to be refactored here, including the item from #1698 checked off? That would be super helpful for us to track when it's complete.

In the description, edit it with a check list:

```md
- [ ] todo 1
- [ ] todo 2
- [ ] todo N
```

Thank you!

[...]

PR #1730, opened 2026-08-27
refactor: use public icalendar.cal exports in TYPE_CHECKING imports

TrueFurina:
## Summary

Refactors `TYPE_CHECKING` imports in three `icalendar.cal` modules to use the **public exports** from `icalendar.cal` instead of importing classes from internal implementation modules, per #1707:

- `src/icalendar/cal/availability.py`: `from icalendar.cal.available import Available` → `from icalendar.cal import Available`
- `src/icalendar/cal/timezone.py`: `from icalendar.cal.calendar import Calendar` → `from icalendar.cal import Calendar`
- `src/icalendar/cal/lazy.py`: `from .component import Component` → `from icalendar.cal import Component`

## Motivation

Using the public exports keeps `TYPE_CHECKING` imports consistent with the public API of `icalendar.cal` and reduces coupling to internal module structure, which also makes future internal reorganization easier.

## Trade-off

The location of the class is slightly less explicit, but the imports now align with the public API — same trade-off accepted in #1698 for `calendar.py`.

## Changes

- 3 files, TYPE_CHECKING imports only; no runtime import changes (these imports are only used for type checking).

## Verification

- Syntax check passed for all three files
- No runtime import changes (guarded by `TYPE_CHECKING`)

### DIFF
--- a/src/icalendar/cal/availability.py
+++ b/src/icalendar/cal/availability.py
@@ -39,7 +39,7 @@ if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
     from datetime import date
 
-    from icalendar.cal.available import Available
+    from icalendar.cal import Available
     from icalendar.compatibility import Self
     from icalendar.enums import BUSYTYPE, CLASS
     from icalendar.prop import vCalAddress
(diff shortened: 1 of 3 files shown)


angatha:
Why does this touches all lines of the three files?

stevepiercy:
Closed, blocked, and reported.

PR #1682, opened 2026-08-18
Calculate new release version

niccokunzmann:
## Linked issue

- Contributes to #1681 #1595

## Description

- Calculate the new version reliably
- Add this to the release process docs
- add a scripts directory to collect all the scripts and Makefile helpers

## Checklist

- [x] I added a change log entry, following the instructions in [Change log entry format](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-entry-format).
- [x] I followed icalendar's [Artificial intelligence policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) and disclosed my [Responsible AI use](https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use) in my commit messages, if applicable.
- [x] I added or updated tests, if applicable.
- [x] I ran and ensured all tests pass locally by following [Run tests](https://icalendar.readthedocs.io/en/latest/contribute/development.html#run-tests).
- [x] I added or edited documentation as necessary, both as docstrings to be rendered in the API documentation and narrative documentation, following the [Style guide](https://icalendar.readthedocs.io/en/latest/contribute/documentation/style-guide.html).

Issue #1595, opened 2026-07-25
8.0 Checklist


niccokunzmann:
Add items here that MUST be done before/with the next release:

- [x] https://github.com/collective/icalendar/pull/1450#discussion_r3406818767
- [x] https://github.com/collective/icalendar/pull/1466/changes#diff-d4883f185e2ae8ad286fff113fc20cb00df73772d21fe65559a1e37f8ad1544aR297 needs a `versionadded` admonition in its docstring. #1699 
- [x] https://github.com/collective/icalendar/pull/1356/changes#diff-5d29b31b790f118197fa5347be902b762e5457bfad20b3fb4bf8b429db0081d4R80 #1699 
- [x] https://github.com/collective/icalendar/pull/1356/changes#diff-5d29b31b790f118197fa5347be902b762e5457bfad20b3fb4bf8b429db0081d4R121 #1699 
- [x] https://github.com/collective/icalendar/pull/1603/changes#r3661578638 #1699 
- [x] https://github.com/collective/icalendar/pull/1603/changes#r3661585809 #1699 
- [x] https://github.com/collective/icalendar/pull/1603/changes#r3661593347 #1699 
- [ ] Add or update any necessary `versionadded` or `versionchanged` admonitions in the docstrings according to the release version number, including `7.2.3`, `7.3.0`, or `8.0.0`. #1699 
- [ ] remove functions that are deprecated for version 8
- [ ] adjust issue for deprecation for version 8 for version 9
- [ ] write update guide
- [ ] update test matrix tests for 8.x (drop 7.x)

These are necessary items for the 8.0 release. Nice to have items must not be on the list.

stevepiercy:
@niccokunzmann please use milestones. "Next" does not designate the explicit version number for a release and whether to include breaking changes.

stevepiercy:
@niccokunzmann re-ping! https://github.com/collective/icalendar/issues/1595#issuecomment-5081344686 is important.

niccokunzmann:
#1681 is at a stage where we can say: Let's prepare the main branch for an 8.x release.
This also affects #1692.

stevepiercy:
Added "Manage GHSA" to the checklist.

PR #1741, opened 2026-08-30
Sanitize RFC 5545 control characters in TEXT on output (#1712)

### LINKED ISSUE #1712: [Bug] NUL bytes in TEXT property values pass through to serialized output
## Describe the bug

RFC 5545 forbids control characters in TEXT values, but `Event.from_ical` accepts NUL in a property value and `to_ical()` emits it verbatim. The parameter path handles this correctly: a NUL there rejects the line into `component.errors`. The value path has no check. Consumers written in C typically truncate at NUL, so a round-tripped file can silently change meaning for them.

## To reproduce

```python
from icalendar import Event

e = Event.from_ical(b"BEGIN:VEVENT\r\nSUMMARY:A\x00B\r\nEND:VEVENT\r\n")
print(ascii(str(e["SUMMARY"])))
print(b"\x00" in e.to_ical())
```

## Output:

```pycon
'A\x00B'
True
```

## Expected behavior

Reject control characters on parse (collect in errors like the parameter path does), or escape them on output.

## Environment

 - [X] Operating system: Windows 11
 - [X] Python version: 3.13.13
 - [X] `icalendar` version: 7.3.0 

## Additional context

Related to #1471

- [X] I tested it with the latest version. For example, I used the command `pip3 install --upgrade https://github.com/collective/icalendar.git`.
- [X] I attached the `.ics` sample file, or there is no `.ics` sample file.
      To remove private information, use icalendar-anonymizer either through its [website](https://icalendar-anonymizer.com/) or [install it locally](https://docs.icalendar-anonymizer.com/latest/installation.html).

Dmao233:
## Linked issue

- Closes #1712
- See #1723

## Description

#1723 rejects control characters on parse. That still lets round-trip emit NUL, and it fails callers on junk files. nicco asked to keep parse open and emit valid TEXT. This strips leftover CONTROLs on output. Parameter NULs still go into errors.

## Checklist

## Additional information

I used AI to help draft the patch. I read the issue thread and checked the tests, including the reporter repro.


Dmao233:
Quality bot failed on description length, missing checklist boxes, commit author not matching, and a first commit over 500 chars.

Description is back on the template. Commits rewritten as me: `d242400f`, `e3cf05ae`.

It also flags the username Dmao233 for consecutive digits. I cannot change that.

stevepiercy:
@Dmao233 there's no need to repeat anything from the comments made by automation or in pushed commits. Maintainers receive an email notification for each comment and pushed commit, as well as view the results of CI.

We need to do a better job of clarifying that in our contributing documentation, as many new contributors don't understand that AI is incredibly noisy and states the obvious.

And there's no need to reply to this comment, as that would be yet more noise, but you can use reactions. Thank you!

stevepiercy:
@Dmao233 the CI fuzzing check fails. We don't review pull requests until CI checks pass. Would you please take care? Thank you!

Dmao233:
Fuzzing is green now.

src/icalendar/tests/fuzzed/__init__.py
```diff
@@ -35,6 +35,17 @@
...
+    "unsupported ",  # dateutil tzical: unsupported TZID/TZOFFSET/TZNAME parm
+    "empty offset",  # dateutil tzical
+    "invalid offset",  # dateutil tzical
+    "empty property name",  # dateutil tzical
+    "mandatory TZID",  # dateutil tzical
+    "mandatory DTSTART",  # dateutil tzical
+    "mandatory TZOFFSETFROM",  # dateutil tzical
+    "no timezones defined",  # dateutil tzical
+    "more than one timezone available",  # dateutil tzical
+    "Unsupported DTSTART param",  # dateutil tzical
```
angatha: Why did thouse became a problem now?

src/icalendar/prop/text.py
```diff
@@ -1,30 +1,12 @@
...
-#     TSAFE-CHAR = WSP / %x21 / %x23-2B / %x2D-39 / %x3C-5B /
-#              %x5D-7E / NON-US-ASCII
-#        ; Any character except CONTROLs not needed by the current
-#        ; character set, DQUOTE, ";", ":", "\", ","
-#
-# CONTROL is defined in the same section as ``%x00-08 / %x0A-1F / %x7F``, so no
-# control character except the horizontal tab may appear in a TEXT value.
-# The line feed, ``\x0a``, is additionally accepted here because it is the
-# result of the escaped sequences ``\N`` and ``\n``.
-UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
```
angatha, line 26: This is a breaking change. Please keep it.

src/icalendar/parser/string.py
```diff
@@ -5,6 +5,13 @@
...
 
+# :rfc:`5545#section-3.3.11` TEXT is
+# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)``. TSAFE-CHAR excludes
+# CONTROL except HTAB (see :rfc:`5545#section-3.1`). After the line-break
+# escapes below, any remaining CONTROL other than HTAB is stripped so
+# serialized output never contains a raw control character.
+_UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
+
 
 def _escape_char(text: str | bytes) -> str:
```
angatha, line 9: Method name and description does not suggest character filtering. Additionally there is `_unescape_char` in `string.py`, which should do the inverse. I would prefere another method for that, or you move this to `vText.__init__`. Then it would also catch `to_jcal`.

src/icalendar/parser/string.py
```diff
@@ -5,6 +5,13 @@
 from icalendar.compatibility import deprecate_for_version_8
 from icalendar.parser_tools import DEFAULT_ENCODING, to_unicode
 
+# :rfc:`5545#section-3.3.11` TEXT is
+# ``*(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)``. TSAFE-CHAR excludes
+# CONTROL except HTAB (see :rfc:`5545#section-3.1`). After the line-break
+# escapes below, any remaining CONTROL other than HTAB is stripped so
+# serialized output never contains a raw control character.
+_UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
```
angatha: Please rever this and use the constant defined in `text.py`

src/icalendar/tests/fuzzed/__init__.py
```diff
@@ -61,3 +75,10 @@ def fuzz_v1_calendar(
         if any(m in str(e) for m in _value_error_matches):
             return -1
         raise
+    except AssertionError as e:
+        # Contentline.__new__ fails loudly on a raw LF (issue #1445). That is
+        # expected for non-TEXT values such as a malformed RRULE; do not treat
+        # it as a fuzzer crash. Other asserts still propagate.
+        if _CONTENTLINE_NEWLINE_ASSERT in str(e):
+            return -1
+        raise
```
angatha: Why is this now a problem?

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -19,33 +19,34 @@
...
+#: is not a CONTROL, and the new line, ``%x0A``, is escaped rather than
+#: stripped because the sequences ``\N`` and ``\n`` represent it.
 FORBIDDEN_CONTROL_CHARS = [
     chr(control) for control in range(0x20) if control not in (0x09, 0x0A)
 ] + ["\x7f"]
 
 FORBIDDEN_VALUES = [
     *FORBIDDEN_CONTROL_CHARS,
     "A\x00B",  # NUL from the issue report
-    "a\rb",  # lone carriage return
```
angatha, line 48: Why was this removed?

angatha (changes requested):
Thank you for your contribution. I have some suggestions and questions.

Dmao233:
Addressed in 278d2670 and cada72d3.

- Kept public `UNSAFE_TEXT_CHARS`; filter moved to `vText.__new__` (covers to_ical and to_jcal). `_escape_char` is escape-only again.
- Restored the control-char cases that still apply under sanitize-on-construct.
- Fuzzer: only kept proven ignores (`unsupported property`, Contentline raw-newline assert for #1445). Dropped the broad dateutil ones.
- Also coerce non-str TEXT inputs before strip so `vUid.new` / `Event.new` (UUID) keep working on 3.14.

src/icalendar/prop/text.py
```diff
@@ -166,4 +196,4 @@ def parse_jcal_value(cls, jcal_value: Any) -> Self:
         return cls(str(jcal_value))
 
 
-__all__ = ["vText"]
```
angatha, line 186: Oh I'm sorry, it was not public. Why didn't you say so?

src/icalendar/prop/text.py
angatha: It actually wasn't. I was not aware of `__all__`.

src/icalendar/prop/text.py
angatha: Then I would prefere to keep it internal. Please rename it to `_UNSAFE_TEXT_CHARS`. Then you can still use it within tests.

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -59,29 +65,80 @@
 ]
 
 
+def _contains_forbidden_control(data: str | bytes) -> bool:
+    if isinstance(data, bytes):
+        return any(ch.encode("latin-1") in data for ch in FORBIDDEN_CONTROL_CHARS)
+    return any(ch in data for ch in FORBIDDEN_CONTROL_CHARS)
+
+
+def test_unsafe_text_chars_is_public():
```
angatha: This is not actually tested by this. The import requires that. But I like the idea of testing it. Can you use `FORBIDDEN_CONTROL_CHARS` to test all ascii characters?

angatha:
Now fuzzing fails again. Please let us have a conversation about that, so we understand why that happen. You already had problems with it and somehow solfed them. Can you please explain, why that was or was not a viable solution?

Dmao233:
Fuzzing crash is dateutil, not the TEXT filter.

`ValueError: mandatory DTSTART not found` from `tzical._parse_rfc`. Corpus case glues DTSTART into TZNAME (`TZNAME:CET\x8bDTSTART:243T030000`), so dateutil never sees DTSTART. Stack is `from_ical` → `cache_timezone_component` → `to_tz`.

Parse stays open for #1712, so junk VTIMEZONE now reaches `to_tz()`. The TEXT strip does not raise this.

The ignore list was viable for CI. It was not a library fix. I am putting those dateutil tzical messages back in the fuzzer, commented. Not swallowing them in icalendar.

Renaming to `_UNSAFE_TEXT_CHARS` next. Tests will walk ASCII via `FORBIDDEN_CONTROL_CHARS`.

src/icalendar/prop/text.py
Dmao233: Renamed in 115b0f4e: `_UNSAFE_TEXT_CHARS`, `__all__ = ["vText"]` only.

src/icalendar/tests/prop/test_control_chars_in_text_values.py
Dmao233: Done in 115b0f4e: parametrize ASCII 0–127 against `FORBIDDEN_CONTROL_CHARS` / `_UNSAFE_TEXT_CHARS`.

src/icalendar/tests/fuzzed/__init__.py
```diff
@@ -35,9 +35,27 @@
...
+    # dateutil.tz.tzical ValueErrors on malformed VTIMEZONE after parse stays
+    # open for #1712; not TEXT-filter bugs. The ignore list is how this
+    # project's fuzzer treats expected third-party parse failures (same
+    # pattern as "not enough values to unpack").
+    "unsupported ",  # dateutil tzical: unsupported TZID/TZOFFSET/TZNAME parm
+    "empty offset",
+    "invalid offset",
+    "empty property name",
+    "mandatory TZID",
+    "mandatory DTSTART",  # CIFuzz crash-ba601c46 on this PR
```
angatha: @niccokunzmann maybe you can help with this. I see that this error showed up in https://github.com/collective/icalendar/actions/runs/33458752743/job/99704224183 but

1. I can not reproduce it locally. From [line 9764](https://github.com/collective/icalendar/actions/runs/33458752743/job/99704224183#step:5:9765), i would asusme that this is the input `QkVHaU46VkVWRU5UClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6ClJzVlA6` and `Calendar.from_ical` was called, but I `ValueError: Found no components where exactly one is required:[...]`
2. The real problem seems to be [at line 9829](https://github.com/collective/icalendar/actions/runs/33458752743/job/99704224183#step:5:9829), because this is the sample that was uploaded.

As I understand it, @Dmao233 want's to propagate the exception from `dateutil`. I would say, that we miss some error handling/conversion.

Dmao233: Agree on conversion, not growing the fuzzer ignore list.

`ZONEINFO.create_timezone` already catches `ValueError` for X- / DATE-DTSTART cleanup, then lets the second `tzical` failure bubble as dateutil's message. I will wrap that final failure as `Invalid iCalendar VTIMEZONE: …` and drop the dateutil-only ignore strings.

On the line-9764 sample: that path is a different parse failure (`Found no components…`). The uploaded crash at 9829 is the DTSTART-in-TZNAME one.

Dmao233: Done in d8238a64.

`ZONEINFO.create_timezone` now wraps the leftover dateutil `ValueError` as `Invalid iCalendar VTIMEZONE: …`. Fuzzer no longer ignores those dateutil strings; `not enough values to unpack` and the Contentline raw-newline assert for #1445 stay.

angatha:
Now I understand how to reproduce the fuzzer crash. Wrapping the error like this is fine. Thank you for that.

src/icalendar/tests/fuzzed/__init__.py
```diff
@@ -61,3 +64,8 @@ def fuzz_v1_calendar(
         if any(m in str(e) for m in _value_error_matches):
             return -1
         raise
+    except AssertionError as e:
+        # CIFuzz: Contentline AssertionError on raw LF in a non-TEXT RRULE key (#1445).
+        if _CONTENTLINE_NEWLINE_ASSERT in str(e):
+            return -1
+        raise
```
angatha: This would be right but we had #1559 which replaced the assert statement with a `raise ValueError`. This simplifies this by much. Revert all and add the string to `_value_error_matches`. Then the fuzzing part should be done.

src/icalendar/tests/prop/test_control_chars_in_text_values.py
```diff
@@ -59,29 +65,79 @@
...
+    return any(ch in data for ch in FORBIDDEN_CONTROL_CHARS)
+
+
+@pytest.mark.parametrize("codepoint", range(128))
+def test_unsafe_text_chars_matches_forbidden_ascii(codepoint):
+    ch = chr(codepoint)
+    if ch in FORBIDDEN_CONTROL_CHARS:
+        assert _UNSAFE_TEXT_CHARS.search(ch)
+    else:
+        assert _UNSAFE_TEXT_CHARS.search(ch) is None
```
angatha, line 80: Thank you :)

angatha (changes requested):
Only the one task for fuzuing remains (because there where changes in main).

@niccokunzmann this changes how null bytes are handled. Before this, there was an exception. Now thouse are silently ignored. Can we merge this in 7.x or is this something for 8.0?

src/icalendar/tests/fuzzed/__init__.py
Dmao233: Done in c0788669. Reverted AssertionError handler; string is in `_value_error_matches` now (#1559).

angatha:
Thank you for the quick response. I think this is ready to merge. I would wait for @niccokunzmann s answer about the targeted release.

src/icalendar/prop/text.py
```diff
@@ -23,7 +23,26 @@
...
 # The line feed, ``\x0a``, is additionally accepted here because it is the
 # result of the escaped sequences ``\N`` and ``\n``.
-UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
+_UNSAFE_TEXT_CHARS = re.compile(r"[\x00-\x08\x0b-\x1f\x7f]")
+
+
+def _strip_unsafe_text_chars(value: object) -> str:
+    r"""Remove CONTROL characters that :rfc:`5545#section-3.3.11` forbids in TEXT.
+
+    ``\\r\\n`` and a lone ``\\r`` become ``\\n`` first so an intentional line
```
stevepiercy: Because you use `r` on the docstring, you don't need to escape inline literals. Although private methods' docstrings don't get published, we should be consistent with the published ones.

```suggestion
    ``\r\n`` and a lone ``\r`` become ``\n`` first, so an intentional line
```

src/icalendar/prop/text.py
```diff
@@ -41,9 +60,11 @@ class vText(str):
 
     Use the LANGUAGE property parameter to set the language of the text.
 
-    When the TEXT object is serialized to an icalendar stream, certain
-    characters are escaped or changed.
-    These characters include the COMMA, SEMICOLON, BACKSLASH, and line breaks.
+    When the TEXT object is created, CONTROL characters other than HTAB
+    are removed so both :meth:`to_ical` and :meth:`to_jcal` stay valid
+    :rfc:`5545` TEXT. Parsing does not raise; the value is corrected.
```
stevepiercy:
```suggestion
    :rfc:`5545#section-3.3.11` TEXT. Parsing does not raise; the value is corrected.
```

src/icalendar/prop/text.py
stevepiercy, line 75: Always include a version* directive when changing, adding, deprecating, or removing functionality to the API.

```suggestion

..  versionchanged:: 0.0.0

    Remove CONTROL characters other than HTAB.

```

src/icalendar/prop/text.py
```diff
@@ -111,26 +132,18 @@ def __repr__(self) -> str:
...
         r"""Parse a TEXT value from its iCalendar representation.
 
-        Raises:
-            ValueError: If the value contains a control character that
-                :rfc:`5545#section-3.3.11` does not allow in TEXT values.
-                The line feed is exempt because it can only enter a parsed
-                value through the escaped sequences ``\N`` and ``\n``.
+        Control characters that :rfc:`5545#section-3.3.11` does not allow
+        in TEXT do not raise. :meth:`__new__` removes them so the parsed
+        object can still be read and later serialized.
```
stevepiercy: This needs a versionchanged directive, and it refers to removed behavior. Also we don't need to mention implementation details in the public docstring. Let's use this instead.

```suggestion
        Control characters that :rfc:`5545#section-3.3.11` does not allow
        in TEXT get removed.

        ..  versionchanged:: 0.0.0

            Don't raise for unsafe characters in TEXT, but instead remove them.
            The parsed object can now be read and later serialized.

            ..  seealso::

                -   :issue:`1712`
                -   :pr:`1723`
```

stevepiercy (changes requested):
Thanks for your work on this and persevering. I have some docstring suggestions for you to consider. AI is not good at addressing the audience or writing technical documentation, even if you tell it, "Address the end users of the API, and omit implementation details." Humans ftw!

niccokunzmann:
@angatha  the next release will be an 8.x one. So, this can be merged and we can be more at ease with breaking changes.
Null bytes are not supposed to be in text, we can accept the change and if anyone thinks the behaviour is wrong (unlikely) then they can open an issue.

src/icalendar/prop/text.py
Dmao233: Done in 45023b8e.

src/icalendar/prop/text.py
Dmao233: Done in 45023b8e.

src/icalendar/prop/text.py
Dmao233: Done in 45023b8e.

src/icalendar/prop/text.py
Dmao233: Done in 45023b8e.

Dmao233:
Docstring suggestions from @stevepiercy applied in 45023b8e.

src/icalendar/prop/text.py
angatha: This should be `8.0.0`

src/icalendar/prop/text.py
```diff
@@ -111,26 +136,27 @@ def __repr__(self) -> str:
...
 
-        Raises:
-            ValueError: If the value contains a control character that
-                :rfc:`5545#section-3.3.11` does not allow in TEXT values.
-                The line feed is exempt because it can only enter a parsed
-                value through the escaped sequences ``\N`` and ``\n``.
+        Control characters that :rfc:`5545#section-3.3.11` does not allow
+        in TEXT get removed.
+
+        ..  versionchanged:: 0.0.0
```
angatha, line 149: `8.0.0` again

src/icalendar/prop/text.py
stevepiercy: @angatha it's OK to leave the version as it is, per https://github.com/collective/icalendar/pull/1682. @niccokunzmann needs to resolve the merge conflict and merge it. We adopted the convention of `0.0.0` for an unknown release number, so that we don't hold up pull requests or miss updating the version, and the automation should handle it. In fact, I'd prefer to leave `0.0.0` in place as a practical test of that script.

src/icalendar/prop/text.py
angatha: Please ignore as steve explained.

angatha (approved):

angatha:
This currently claims to close #1712. I asked there, if the current behavior is enough to close the issue.

news/1712.bugfix.1
```diff
@@ -0,0 +1 @@
+Serialized :rfc:`5545` TEXT values no longer emit control characters such as NUL. Parsing stays open so problematic files can still be read; :class:`~icalendar.prop.text.vText` removes leftover CONTROLs when constructed so both :meth:`~icalendar.prop.text.vText.to_ical` and :meth:`~icalendar.prop.text.vText.to_jcal` stay valid. I used AI to assist me with this change. @Dmao233
```
angatha, line 1: Please rename this to `1712.breaking` (or with `.x`, if already present)

angatha (changes requested):

src/icalendar/prop/text.py
```diff
@@ -140,8 +144,17 @@ def from_ical(cls, ical: ICAL_TYPE) -> Self:
         r"""Parse a TEXT value from its iCalendar representation.
 
         Control characters that :rfc:`5545#section-3.3.11` does not allow
-        in TEXT do not raise. :meth:`__new__` removes them so the parsed
-        object can still be read and later serialized.
+        in TEXT get removed.
```
stevepiercy, line 147: Looks beautiful!

https://icalendar--1741.org.readthedocs.build/en/1741/reference/api/icalendar.prop.text.html#icalendar.prop.text.vText.from_ical

stevepiercy (dismissed):
Thanks for the improvement to icalendar.

Would you please respond to @angatha's remarks? Thank you!

Dmao233:
@angatha on #1712: this PR is nicco's strip-on-construct path, not reject/report. Fine to keep Closes #1712 and open a follow-up for error reporting / accept-list. Or I drop Closes from the description — say which.

Renaming `news/1712.bugfix.1` → `1712.breaking` next.

news/1712.breaking
```diff
@@ -0,0 +1 @@
+Serialized :rfc:`5545` TEXT values no longer emit control characters such as NUL. Parsing stays open so problematic files can still be read; :class:`~icalendar.prop.text.vText` removes leftover CONTROLs when constructed so both :meth:`~icalendar.prop.text.vText.to_ical` and :meth:`~icalendar.prop.text.vText.to_jcal` stay valid. I used AI to assist me with this change. @Dmao233
```
Dmao233: Done in c9119399.

angatha:
> @angatha on #1712: this PR is nicco's strip-on-construct path, not reject/report. Fine to keep Closes #1712 and open a follow-up for error reporting / accept-list. Or I drop Closes from the description — say which.
> 
> Renaming `news/1712.bugfix.1` → `1712.breaking` next.

I would prefere to give @niccokunzmann some time (maybe another day) to respond. I will come back to you.

src/icalendar/prop/text.py
```diff
@@ -101,7 +126,7 @@ def __new__(
         /,
         params: dict[str, Any] | None = None,
     ) -> Self:
-        value = to_unicode(value, encoding=encoding)
+        value = _strip_unsafe_text_chars(to_unicode(value, encoding=encoding))
```
SemTiOne, line 129: I think better approach here is to guard the hot path so clean text skips the work.
```suggestion
        value = to_unicode(value, encoding=encoding)
        if _UNSAFE_TEXT_CHARS.search(value) or "\r" in value:
            value = _strip_unsafe_text_chars(value)
```

SemTiOne (commented):
The codes make sense, 1 nit:

src/icalendar/prop/text.py
angatha: I disagree. You would repeate logic from `_strip_unsafe_text_chars` and create a second spot that needs to be updated, when someone changes `_strip_unsafe_text_chars`. You may replace `replace` in `_strip_unsafe_text_chars` with something that needs only one swipe through the string. But thats seems like something one should measure.

angatha:
@SemTiOne does ths close the issue despite not tracking the errors?

src/icalendar/prop/text.py
SemTiOne: Fair enough, I withdraw my suggestion.

SemTiOne:
> @SemTiOne does ths close the issue despite not tracking the errors?

@angatha It fixes the symptom but I agree we should track the errors. So almost, but not completely.

angatha:
I think we should include this 8.x and create a follow up issue for error recording. Recording errors does not break the API and can therefore easily added. This PR however removes an exception that is thrown for `\n` and therefore changes behavior.

stevepiercy:
@angatha that sounds like a good plan, and is in line with @niccokunzmann's earlier comment.

Added to [8.0.0 milestone](https://github.com/collective/icalendar/milestone/8) and https://github.com/collective/icalendar/issues/1595.

angatha (approved):

angatha:
> > @SemTiOne does ths close the issue despite not tracking the errors?
> 
> @angatha It fixes the symptom but I agree we should track the errors. So almost, but not completely.

I created the follow up issue #1781. Can this be merged (also @stevepiercy)

Issue #692, opened 2024-07-09
Test with dateutil timezones [Python, guided by maintainer]


niccokunzmann:
At the moment, we check icalendar with zoneinfo and pytz.
We can also make sure that `dateutil` works.

- [ ] create a TimzoneProvider for PyICU
- [ ] parametrize the tests to work with PyICU (conftest.py)
- [ ] add possible additional tests

devdanzin:
For anyone trying to tackle this, building PyICU can be a hassle, so I suggest either using Anaconda or trying a wheel from https://github.com/cgohlke/pyicu-build/releases.

h4syy:
Hi @niccokunzmann, can I take this one? Happy to be assigned.

I've already started on it. So far I added `use_pyicu()` to `TZP` and `timezone/__init__.py`, and I'm writing the `PYICU` provider in `timezone/pyicu.py` based on how the pytz and zoneinfo ones work. Next I want to hook it into `conftest.py` so the existing tests also run under pyicu, and add a few pyicu-specific tests.

One thing I could use your input on before I go too far: for `create_timezone`, would you rather I parse the VTIMEZONE with ICU's `VTimeZone`, or build it from the transitions like the pytz provider does?

niccokunzmann:
Hi,

nice that you started. Please open a draft pull request as soon as you do some work - then no assignment is needed.

I would like it if PyICU's VTimeZone parsing is used. When you open the PR, ping me on the lines where you would like input. I am happy to review your PR.

PR #1755, opened 2026-09-02
Add a PyICU timezone provider (#692)

h4syy:
## Linked issue

- Contributes to #692

## Description

Adds a **PyICU timezone provider** — a `PYICU` implementation of `TZProvider`, selectable via `use_pyicu()` and wired through `TZP` and the `icalendar.timezone` package. This is **part 1 of #692** and is opened as a **draft / work in progress**.

Working so far (verified locally against the repo's VTIMEZONE fixtures — 37/38 `create_timezone` succeed):
- timezone lookup via `icu.ICUtzinfo`, UTC handling, and `knows_timezone_id`
- unknown ids return `None`, so the Windows→Olson and globally-unique-TZID fallbacks in `TZP` still work
- `create_timezone` parses the `VTIMEZONE` via `icu.VTimeZone.createVTimeZone` (DST transitions correct); a `VTIMEZONE` that ICU cannot parse raises `ValueError` instead of segfaulting the interpreter

Still to do (follow-up commits): parametrize `conftest.py` across the pyicu backend, add PyICU-specific tests, declare the optional PyICU test dependency (gated so it skips when ICU is absent), resolve `uses_pyicu()`, and lint/type-hint cleanup.

Questions I'd love input on, @niccokunzmann (I'll also add inline comments on the lines):
1. `create_timezone` uses `icu.VTimeZone.createVTimeZone` as you suggested. One fixture — `issue_321_assert_dst_offset_is_not_false.ics` (`Europe/Berlin`) — is rejected by ICU (empty id) where zoneinfo/pytz succeed. Prefer a fallback for those, or an `xfail` for pyicu?
2. `uses_pyicu()`: add it to the `TZProvider` ABC + `TZP` proxy for symmetry with `uses_pytz`/`uses_zoneinfo`, or keep providers reporting only pytz/zoneinfo?
3. PyICU is hard to install (especially in CI). How would you like the optional dependency and test-skipping handled?

## Checklist

## Additional information

Draft / WIP — tests, docs, and the conftest parametrization are coming in follow-up commits.

AI disclosure: I used Claude (Opus 4.8, via Claude Code) to explain the timezone-provider architecture and to verify the provider's behaviour by running it; the code is my own. This is also noted in the commit message and the `news/692.feature` entry.

### DIFF
--- a/news/692.feature
+++ b/news/692.feature
@@ -1,0 +1,1 @@
+Added a PyICU timezone provider (``use_pyicu()``) implementing the timezone interface using ICU's :rfc:`5545` ``VTIMEZONE`` support. I used AI to explain the timezone architecture and to verify its behaviour; the code is my own. @h4syy
--- a/src/icalendar/timezone/__init__.py
+++ b/src/icalendar/timezone/__init__.py
@@ -15,6 +15,9 @@ def use_zoneinfo():
     """Use zoneinfo as the implementation that looks up and creates timezones."""
     tzp.use_zoneinfo()
 
+def use_pyicu():
+    """Use pyicu as the implementation that looks up and creates timezones."""
+    tzp.use_pyicu()
 
 __all__ = [
     "TZP",
@@ -25,4 +28,5 @@ __all__ = [
     "tzp",
     "use_pytz",
     "use_zoneinfo",
+    "use_pyicu",
 ]
--- a/src/icalendar/timezone/pyicu.py
+++ b/src/icalendar/timezone/pyicu.py
@@ -1,0 +1,81 @@
+"""Use PyICU timezones."""
+from __future__ import annotations
+from datetime import datetime, tzinfo
+import threading
+from typing import TYPE_CHECKING
+import icu  # the PyICU module
+from .provider import TZProvider
+
+if TYPE_CHECKING:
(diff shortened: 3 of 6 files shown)


niccokunzmann:
Hi, thanks for the PR!

Some answers:

uses_pyicu() - yes, they all should have the same interface.

issue_321_assert_dst_offset_is_not_false.ics - if a fix cannot be done because it is too much for this PR, then skipping is a good way to work this this (and preferably adding a TODO in the code or an issue if it is important). Best would be to somehow fix this - either in icalendar or in PyICU itself - report this upstream. Generally all files should work with all timezones. The reason is that we collect the different quirks and wild (mis-)interpretations of the RFC standard and want to take the burden off developers to deal with these. So, that special case is best handeled in a consistent way with the others by calendar, so there is no deviation. If you think, this should be handled by PyICU instead of icalendar, create an issue in PyICU to fix it and add a skip with a reference to that issue so it can be removed when the issue is closed.

If it is hard to install, then, I would say the tests should check - maybe like with pytz - wether pyicu is installed. It is also a question of documentation but some kind of switch is required, so they can be switched on. Is it hard to install PyICU on linux? What is the easiest way to install it?

Let me know when you would like me to review this.

PR #1761, opened 2026-09-03
Enable uv cache for test jobs

### LINKED ISSUE #1757: Allow setup-uv to use cache where it's safe
The key phrase was "in this context," specifically lines 45-49 within the `run-tests` job. Here, I think it's OK to use `enable-cache: true` as it has nothing whatsoever to do with the release-shaped parts.

The release-shaped parts in this workflow are the following jobs:

- `test-distribution` on lines 88-99, which uses `setup-uv` and has `enable-cache: false`. This is good in this context.
- `deploy-tag-to-pypi` on lines 101-147, which uses `setup-python` which doesn't enable its `cache` setting by default. See https://github.com/actions/setup-python#caching-packages-dependencies
- `deploy-github-release` uses https://github.com/ncipollo/release-action which doesn't include artifacts by default

_Originally posted by @stevepiercy in https://github.com/collective/icalendar/pull/1688#discussion_r3913081030_

Ryota-Di:
AI assistance: GPT-5.6 Sol was used to review the issue context, explain cache-poisoning risks, identify the intended workflow change, and help validate the zizmor exception. I reviewed and applied the change myself.

## Linked issue

- Closes #1757

## Description

Enable the `setup-uv` cache for the `run-tests` job, where cached runtime artifacts do not feed into release artifacts.

The cache remains disabled for `test-distribution` and the release-related jobs. A targeted `zizmor` suppression documents that cache use in `run-tests` is intentional because that job does not produce release artifacts.

## Checklist


news/1757.chore
```diff
@@ -0,0 +1 @@
+Enabled the ``uv`` cache for regular test jobs while keeping it disabled for release-related jobs. Prepared with AI assistance. @Ryota-Di
```
stevepiercy: uv is not an inline literal, but its actual brand name.

```suggestion
Enabled the uv cache for regular test jobs while keeping it disabled for release-related jobs. Prepared with AI assistance. @Ryota-Di
```

.github/workflows/tests.yml
```diff
@@ -50,7 +50,8 @@ jobs:
       if: ${{ !matrix.skip }}
       uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
       with:
-        enable-cache: false
+        # zizmor: ignore[cache-poisoning] - run-tests does not produce release artifacts
```
stevepiercy, line 53: Can you explain why this comment is necessary? I'm curious if zizmor will flag it if it's not present.

stevepiercy (changes requested):
Thank you for first contribution to icalendar! I have minor format tweak and a question. Would you please take care?

.github/workflows/tests.yml
Ryota-Di: Yes. The comment includes the inline `zizmor: ignore[cache-poisoning]` directive, so removing it would cause zizmor to flag the enabled cache in this job.

The rationale after the directive documents why the suppression is intentional here: `run-tests` does not produce release artifacts, while caching remains disabled for the release-related jobs.

stevepiercy: Thanks for confirming my intuition.

stevepiercy (approved):
Thank you for this contribution! Will auto-merge when CI passes.

PR #1785, opened 2026-09-14
feat: add enum for alarm actions

### LINKED ISSUE #1784: Create ACTION enum
Alarm.ACTION and AlarmTime.action should use a str enum. They have a certain set of values that are expected to work on them.

```
# Action is missing
>>> from icalendar import ACTION
ImportError: cannot import name 'ACTION' from 'icalendar' (/home/blackberry/icalendar/src/icalendar/__init__.py)

# partstat works for example
>>> from icalendar import PARTSTAT
>>> PARTSTAT.ACCEPTED
<PARTSTAT.ACCEPTED: 'ACCEPTED'>
```

To solve this issue:

- [ ] The properties for ACTION should use the enum
- [ ] `Alarm.new*()` should add they enum to their type hints
- [ ] The enum is documented and contains the values described by the RFC

Related:

- #1783

1cbyc:
## Summary
- add the public `ACTION` string enum for RFC 5545 alarm actions
- use enum members in Alarm factory construction and the `Alarm.new` type hint
- add public API, factory, and docstring regression coverage
- add a towncrier feature note

## Validation
- `uv run --offline pytest -q` (18,668 passed, 24 skipped, 541 xfailed)
- `uv run --offline ruff check` (changed files)
- `uv run --offline ruff format --check` (changed files)
- `git diff --check`

Closes #1784


Mmc1xs:
Thanks for working on this! I tested the current behavior while studying #1784.

It looks like the PR adds and exports `ACTION`, and uses it in `Alarm.new()` and the factory methods. However, `Alarm.ACTION` is still backed by `single_string_property`, so reading a standard action returns `vText` rather than the corresponding `ACTION` member. `AlarmTime.action` also still has a `str` return annotation.

The current assertion `alarm.ACTION == ACTION.DISPLAY` passes because the values compare equal, even when the returned object is not an `ACTION` member. A stricter regression check would be:

```python
alarm = Alarm()
alarm.ACTION = "DISPLAY"
assert alarm.ACTION is ACTION.DISPLAY
```

Would it make sense to convert recognized RFC values to `ACTION` members and update `AlarmTime.action`, while preserving unknown IANA and `X-` extension values as strings?

1cbyc:
Addressed the typed-access feedback in 0a9958f6.

- `Alarm.ACTION` now returns `ACTION.AUDIO`, `ACTION.DISPLAY`, or `ACTION.EMAIL` for the RFC-defined values.
- Unknown IANA and `X-` extension values remain strings so parsing and serialization stay lossless.
- `AlarmTime.action` now exposes the same `ACTION | str` type.
- Added strict identity and extension round-trip regression tests.

Validation: `uv run --offline pytest -q` — 18,669 passed, 24 skipped, 541 xfailed.

niccokunzmann:
Hi, I will give this a review now, too. Thanks for the other review and opening this PR!

One note already: Please test setting ACTION to None if you can. Also check what happens if ACTION is defined several times, see the code coverage report:

image

src/icalendar/cal/alarm.py
```diff
@@ -28,8 +28,11 @@
 )
 from icalendar.cal.component import Component
 from icalendar.cal.examples import get_example
+from icalendar.enums import ACTION
```
```suggestion
from icalendar.enums import ACTION as AlarmAction
```

src/icalendar/cal/alarm.py
```diff
@@ -355,18 +358,23 @@ def attachments(self, value: ATTACHMENTS_TYPE_SETTER) -> None:
...
-    ACTION = single_string_property(
-        "ACTION",
-        """The action invoked when the alarm triggers.
-
-        Typical values defined by :rfc:`5545#section-3.8.6.1` are
-        ``AUDIO``, ``DISPLAY``, and ``EMAIL``. The empty string is
-        returned when no ``ACTION`` property is present.
-        """,
-    )
+    @property
```
niccokunzmann: Instead of duplicating the code, could you please keep the tests and use the 
single_string_enum_property?

niccokunzmann (changes requested):
Thank you! This looks good. I also appreciate the thorough testing and finding all these places where the typing helps.

I have a few suggestions.

niccokunzmann:
@1cbyc Please let me know when I should have another look.

I wonder about the coverage changes... We only recently introduced them. Let's figure out together what they mean here.

PR #1794, opened 2026-09-16
Reject invalid UTF-8 calendar bytes instead of silent replacement (#1793)

### LINKED ISSUE #1793: Non-UTF-8 calendar files are silently corrupted with replacement characters
`Calendar.from_ical()` accepts `bytes`. When those bytes are not valid UTF-8, the library silently replaces every bad byte with `U+FFFD` instead of raising.

```python

# \xe9 = é in cp1252/Latin-1, not valid UTF-8
ics = b"BEGIN:VCALENDAR\r\nVERSION:2.0\r\nBEGIN:VEVENT\r\nSUMMARY:R\xe9union\r\nDTSTART:20240101T100000Z\r\nEND:VEVENT\r\nEND:VCALENDAR\r\n"
cal = Calendar.from_ical(ics)
print(cal.events[0].get("SUMMARY"))  # R�union (silent corruption)
```

`Contentlines.from_ical` passes the raw bytes to `to_unicode()` in `parser_tools.py`, which catches `UnicodeDecodeError` and falls back to `decode('utf-8-sig', 'replace')`. No warning, no exception.

[RFC 5545 §3.1.4](https://datatracker.ietf.org/doc/html/rfc5545#section-3.1.4) requires UTF-8, so these files are non-compliant. In practice, Lotus Notes, old Outlook, and similar clients produce cp1252 or Latin-1 files without declaring it.

**Options:**

1. Raise on non-UTF-8 input. RFC-correct. Breaks real-world legacy files.
2. Add an `encoding` parameter to `from_ical`. Callers who know the source can pass `encoding='cp1252'`. Default stays UTF-8 but fails loudly.
3. Heuristic fallback to cp1252/Latin-1. Guesses wrong on short or ambiguous input.

Option 2 is the right call for a library. Option 1 is at minimum better than the current behaviour.

Santhoshcoder001:
Fixes collective/icalendar#1793.

Calendar byte input with invalid UTF-8 sequences now raises `InvalidCalendar` instead of silently replacing bytes with `U+FFFD`.

### Changes

- Added strict decoding by default.
- Added `encoding=` and `errors=` support to calendar, component, content-line, and lazy parsing APIs.
- Preserved `utf-8-sig` BOM handling.
- Documented legacy encoding and explicit replacement recovery.
- Added regression tests for invalid UTF-8, CP1252 decoding, and `errors="replace"`.

### Migration

```python
Calendar.from_ical(raw_bytes, encoding="cp1252")
```

To explicitly restore replacement behavior:

```python
Calendar.from_ical(raw_bytes, errors="replace")
```

### Tests

187 focused tests pass.


niccokunzmann:
@Santhoshcoder001 Let us know if you need any help. Please also follow the PR template that lists all the important points.

PR #1804, opened 2026-09-16
Fix typing of TZP.localize and TZP.localize_utc (#1802)

### LINKED ISSUE #1802: [TYPING] TZP.localize and TZP.localize_utc
TZP.localize and TZP.localize_utc have outdated type hints.
Basically, they both should take a date, a datetime or a time object.

date -> datetime
datetime -> datetime
time -> time

and set the timezone.
At the moment, they are neither consistent nor do they accept all of these types.
We should be sure that they work with all of these combinations and that the typing is correct.

Found during #1789

SUJAYSAKHARE22:
## Linked issue

- Closes #1802
- Contributes to #1789

## Description

Fixed type hints of `TZP.localize` and `TZP.localize_utc` in `src/icalendar/timezone/tzp.py`.

The annotations referenced `datetime.date`, `datetime.datetime` and `datetime.tzinfo`, but only the `datetime` class is imported — these resolved to attributes of the `datetime` class rather than to the `date` and `tzinfo` classes. Imported `date` and `tzinfo` and annotated them directly.

Both methods now accept `date`, `datetime` and `time` consistently:

- `date` → `datetime`
- `datetime` → `datetime`
- `time` → `time`

**Changes:**

- `localize_utc` gains overloads and `time` support; previously it passed a `time` through `to_datetime`'s unchecked cast straight to the provider.
- `localize` gains a `date` overload.
- `localize(date, None)` no longer raises `TypeError`; the conversion to `datetime` now happens before the `tz is None` branch, since `date.replace()` has no `tzinfo` parameter.

Found during #1789.

## Checklist

- [ ] I added a change log entry, following the instructions in all subsections under [Change log requirements](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-requirements).
- [ ] I followed icalendar's [Artificial intelligence policy](https://icalendar.readthedocs.io/en/latest/contribute/index.html#artificial-intelligence-policy) and disclosed my [Responsible AI use](https://icalendar.readthedocs.io/en/latest/contribute/index.html#responsible-ai-use) in my commit messages, if applicable.


src/icalendar/tests/test_tzp_localize_1802.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def test_localize_datetime_with_none_removes_timezone(tzp):
+    aware = tzp.localize(datetime(2024, 1, 15, 10, 30), "Europe/Berlin")
+    result = tzp.localize(aware, None)
+    assert isinstance(result, datetime)
+    assert result.tzinfo is None
+
+
+def test_localize_time_with_none_removes_timezone(tzp):
```
niccokunzmann, line 67: This is actually a special case that should be considered for date, too. What happens with (date, None) ?

src/icalendar/tests/test_tzp_localize_types.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+# ---------------------------------------------------------------- localize
+
+def test_localize_date_returns_datetime(tzp):
+    result = tzp.localize(date(2024, 1, 15), "Europe/Berlin")
+    assert isinstance(result, datetime)
+    assert result.tzinfo is not None
+    assert (result.year, result.month, result.day) == (2024, 1, 15)
+    assert (result.hour, result.minute, result.second) == (0, 0, 0)
```
niccokunzmann: I suggest something like this:

```suggestion
    assert get_tzid(result.tzinfo) == ...
    assert result.replace(tzinfo=None) == ...
```

get_tzid is in icalendar, I think.
And this way, all the other tests can be of consistent form and smaller.

src/icalendar/tests/test_tzp_localize_types.py
```diff
@@ -0,0 +1,110 @@
...
+    result = tzp.localize_utc(berlin)
+    assert result.utcoffset().total_seconds() == 0
+    assert result.hour == 9  # Berlin is UTC+1 in January
+
+
+def test_localize_and_localize_utc_agree_on_accepted_types(tzp):
+    """Both methods must accept the same three input types."""
+    for value in (date(2024, 1, 15), datetime(2024, 1, 15, 10, 30), time(10, 30)):
+        tzp.localize(value, "UTC")
+        tzp.localize_utc(value)
```
niccokunzmann:
```suggestion
```

The other tests already cover this.

src/icalendar/timezone/tzp.py
```diff
@@ -66,39 +69,68 @@ def use_default(self) -> None:
...
+    @overload
+    def localize_utc(self, dt: time) -> time: ...
+
+    @overload
+    def localize_utc(self, dt: date) -> datetime: ...
+
+    def localize_utc(self, dt: date | datetime | time) -> datetime | time:
+        """Return the value in UTC.
+
+        If the value has no timezone, set UTC as its timezone.
```
niccokunzmann, line 84:
```suggestion
        If the value has no timezone, set UTC as its timezone.
        Otherwise, the same date/time in UTC is returned.
```

niccokunzmann (changes requested):
Thank you! This looks good and addresses the issue.

Could you also please:

- add a changelog entry
- rename the test file to include the issue like it is with the others

src/icalendar/tests/test_tzp_localize_1802.py
SUJAYSAKHARE22: Good catch — this is already covered. I added `test_localize_date_with_none_returns_naive_datetime` in the latest commit, which verifies `tzp.localize(date(2024, 1, 15), None)` returns a naive datetime instead of raising TypeError.

src/icalendar/tests/test_tzp_localize_types.py
SUJAYSAKHARE22: Done — I found the actual exported function is `tzid_from_tzinfo` (not `get_tzid`), from `icalendar.timezone`. I've updated all the tests to use `tzid_from_tzinfo(result.tzinfo) == ...` and `result.replace(tzinfo=None) == ...` for a consistent, smaller shape.

src/icalendar/tests/test_tzp_localize_types.py
SUJAYSAKHARE22: Removed — you're right, the individual tests already cover both methods and all three input types.

src/icalendar/timezone/tzp.py
SUJAYSAKHARE22: Added — the localize_utc docstring now includes "Otherwise, the same date/time in UTC is returned."

niccokunzmann:
@SUJAYSAKHARE22 You added tests and somehow the coverage went down. Strange ... Could you investigate this?

I will do a review now. Thanks for all the updates!

src/icalendar/tests/test_tzp_localize_1802.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+# ------------------------------------------------------------ localize_utc
+
+
+def test_localize_utc_date_returns_datetime(tzp):
+    result = tzp.localize_utc(date(2024, 1, 15))
+    assert isinstance(result, datetime)
+    assert tzid_from_tzinfo(result.tzinfo) == "UTC"
+    assert result.replace(tzinfo=None) == datetime(2024, 1, 15, 0, 0, 0)
```
niccokunzmann, line 83: Could you refactor this into a function and use it throughout the tests? That makes it much more readable.


```suggestion
    assert_equals_dt(result, datetime(2024, 1, 15, 0, 0, 0), "UTC" | None)
```

This should then be applied to all test cases and makes them even shorter and easier to review.

src/icalendar/timezone/tzp.py
```diff
@@ -66,39 +69,69 @@ def use_default(self) -> None:
...
+        Otherwise, the same date/time in UTC is returned.
+
+        Returns:
+            -   A localized :class:`datetime.datetime` in UTC when a
+                :class:`datetime.datetime` is given.
+            -   A localized :class:`datetime.datetime` at midnight in UTC
+                when a :class:`datetime.date` is given.
+            -   A localized :class:`datetime.time` in UTC when a
+                :class:`datetime.time` is given.
```
niccokunzmann, line 93: Yes! This is a much clearer description.

niccokunzmann (changes requested):
Thanks!

Please disclose AI assistance for this change in the changelog entry, please.

The documentation looks good!

Image

I have a suggestion for the tests because I find them really hard to read and verify for completeness.

PR #1811, opened 2026-09-18
Type shared attr helpers with HasStartEndDuration protocol

### LINKED ISSUE #897: component._get_start_end_duration called but not defined on Component
Component._get_start_end_duration() is called here:

https://github.com/collective/icalendar/pull/896/files/b44b57b1cbbf9b1aed9535bb38035fa21de37358#diff-46392468ae661de0550d017d8667adaa696b9d0d32e5fa0ad746623844d2a315R1570

But it is not defined on the component.

To solve this issue:

- [ ] change the typing so that it is defined on all instances.

https://github.com/collective/icalendar/pull/896/files/b44b57b1cbbf9b1aed9535bb38035fa21de37358#diff-46392468ae661de0550d017d8667adaa696b9d0d32e5fa0ad746623844d2a315R1555

_Originally posted by @niccokunzmann in https://github.com/collective/icalendar/pull/896#discussion_r2355077712_

SemTiOne:
## Linked issue

- Closes #897

## Description

Type the shared attr helpers with `HasStartEndDuration`, the contract only `Event` and `Todo` satisfy.

## Checklist

- [X] I added a change log entry, following the instructions in all subsections under [Change log requirements](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-requirements).


src/icalendar/attr.py
```diff
@@ -34,6 +34,39 @@
...
+    def end(self) -> date | datetime: ...
+    @property
+    def duration(self) -> timedelta: ...
+
+    def _get_start_end_duration(
+        self,
+    ) -> tuple[date | datetime | None, date | datetime | None, timedelta | None]: ...
+    def __contains__(self, key: object) -> bool: ...
+    def __getitem__(self, key: object) -> Any: ...
+    def pop(self, key: object, default: Any = None) -> Any: ...
```
angatha, line 67: I don't want to review this. The original code is already quite messi with e.g. two wais to access the duration and this reflects in this protocol. It just wraps everything needed but wraps up the mess in one place.

Two ways to access start and duration. Callers of `get_start_end_duration_with_validation` still need to pass the correct name of the end property. What has `__contains__`, `__getitem__` and `pop` todo with something that has `HasStartEndDuration`. Maybe some documentation could help.

So in all, technically fine but I do not want to touch it.

src/icalendar/attr.py
```diff
@@ -34,6 +34,39 @@
     from icalendar.cal import Component
 
 
+class HasStartEndDuration(Protocol):  # pragma: no cover
```
niccokunzmann, line 37: Instead of adding a protocol here, it is probably enough to use the Union type. That would make type checks happen here nicely, I believe. What do you think about that?

src/icalendar/attr.py
```diff
@@ -1866,10 +1901,11 @@ def get_end_property(component: Component, end_property: str) -> date | datetime
         msg = f"No {end_name} or DURATION+DTSTART given."
         raise IncompleteComponent(msg)
 
+    assert end is not None  # narrowed: end or duration given, duration is None
```
niccokunzmann, line 1904: Usually, people come and replace assert statements that we have in production code - is it work raising some kind of error here?

niccokunzmann (changes requested):
Hi, thanks for cleaning up typing.

I suggest we replace the protocol class with a Union type. That would be enough.

I left some comments. Let me know what you think.

src/icalendar/attr.py
SemTiOne: Agreed. This keeps the typing explicit to the two actual callers and lets the type checker validate the member access directly.

src/icalendar/attr.py
SemTiOne: Also agreed. Replaced the assert with `IncompleteComponent` error. This also keeps the check active when Python runs with optimizations enabled.

PR #1812, opened 2026-09-18
Return earliest timestamp for duplicated singleton UTC properties

### LINKED ISSUE #1795: [Bug] DTSTAMP and other single properties cannot cope with duplication (broken calendar)
## Describe the bug

Calendars sometimes have a duplicated entries where there should be only one.
icalendar should guard against these oddities.

## To reproduce

```python
>>> c = Calendar.from_ical("""
... BEGIN:VCALENDAR
... BEGIN:VEVENT
... DTSTAMP:20210205T101751Z
... UID:20200516T060000Z-123401@example.com
... DTSTAMP:20200516T060000Z
... SUMMARY:Do the needful
... DTSTART:20220517T060000Z
... DTEND:20220517T230000Z
... RECURRENCE-ID:20220517T060000Z
... END:VEVENT
... BEGIN:VEVENT
... DTSTAMP:20210205T101751Z
... UID:20200516T060000Z-123401@example.com
... DTSTAMP:20200516T060000Z
... SUMMARY:Do the needful
... DTSTART:20230517T060000Z
... DTEND:20230517T230000Z
... RECURRENCE-ID:20230517T060000Z
... END:VEVENT
... END:VCALENDAR""")
>>> c.events[-1].stamp
  File "<python-input-4>", line 1, in <module>
    c.events[-1].stamp
  File "/home/blackberry/icalendar/src/icalendar/cal/component.py", line 751, in stamp
    return self.DTSTAMP
           ^^^^^^^^^^^^
  File "/home/blackberry/icalendar/src/icalendar/attr.py", line 467, in fget
    raise InvalidCalendar(f"{name} must be a datetime in UTC, not {value}")
icalendar.error.InvalidCalendar: DTSTAMP must be a datetime in UTC, not [vDDDTypes(2021-02-05 10:17:51+00:00, Parameters({})), vDDDTypes(2020-05-16 06:00:00+00:00, Parameters({}))]
```

## Expected behavior

I would expect the earliest timestamp in this list to be returned.
It should be tested that they are comparable (all UTC) and tests should throw UTC, other timezone, date, floating datetime in there to see that it works.

See also here:
https://github.com/niccokunzmann/python-recurring-ical-events/blob/2ba2510ea0cbd4dd956f6923e1fcf330d6b82b96/recurring_ical_events/util.py#L91

It might be useful to move the code from recurring-ical-events over to icalendar in order to allow that type of comparing.

## Environment

 - [x] Operating system:       ubuntu  
 - [x] Python version:        3.14   
 - [x] `icalendar` version:     7.3.0 

## Additional context

      To remove private information, use icalendar-anonymizer either through its [website](https://icalendar-anonymizer.com/) or [install it locally](https://docs.icalendar-anonymizer.com/latest/installation.html).

This is sourced from https://github.com/pycalendar/icalendar_compatibility/issues/3#issuecomment-3992707487

SemTiOne:
## Linked issue

- Closes #1795

## Description

Return the earliest timestamp for duplicated singleton UTC properties instead of raising, keeping the raise for non-datetime values.

## Checklist

- [X] I added a change log entry, following the instructions in all subsections under [Change log requirements](https://icalendar.readthedocs.io/en/latest/contribute/#change-log-requirements).


src/icalendar/tests/test_issue_1795_duplicate_single_properties.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def test_issue_example_returns_earliest_dtstamp(tzp):
+    """The exact calendar from the issue returns the earliest DTSTAMP.
+
+    The earliest timestamp is not the first one encountered,
+    proving earliest wins over first encountered.
+    """
+    c = Calendar.from_ical(
+        """BEGIN:VCALENDAR
```
niccokunzmann, line 24: This must be moved to a file, always.

src/icalendar/tests/test_issue_1795_duplicate_single_properties.py
```diff
@@ -0,0 +1,110 @@
...
+    event.add("CREATED", datetime(2021, 2, 5, 10, 17, 51, tzinfo=UTC))
+    event.add("CREATED", datetime(2020, 5, 16, 6, 0, tzinfo=UTC))
+    assert event.CREATED == datetime(2020, 5, 16, 6, 0, tzinfo=UTC)
+
+
+def test_invalid_item_still_raises(tzp):
+    """An item that cannot be a datetime keeps raising InvalidCalendar."""
+    event = Event()
+    event.add("DTSTAMP", datetime(2021, 2, 5, 10, 17, 51, tzinfo=UTC))
+    event["DTSTAMP"] = [event["DTSTAMP"], 42]
```
niccokunzmann, line 99: I would like to change this:

- [ ] If there is one valid value, that should be used.
- [ ] If you can, use add()

src/icalendar/attr.py
```diff
@@ -445,6 +445,18 @@ def fdel(self: Component):
     return property(fget, fset, fdel, doc)
 
 
+def _decode_single_utc_value(name: str, value: object) -> date | datetime:
+    """Extract and check one datetime value of a singleton UTC property."""
+    if isinstance(value, (vText, vUnknown)):
+        # we might be in an attribute that is not typed
+        value = vDDDTypes.from_ical(value)
```
niccokunzmann, line 452: Is this case tested and ever used?
I think, it can be removed. Which tests fail when you remove this case?

niccokunzmann (changes requested):
Thanks for the PR! I had a look and left some comments. Let me know what you think...

src/icalendar/attr.py
SemTiOne: > Is this case tested and ever used? 

Yes. The parser can produce `vText` and `vUnknown` for explicit `VALUE=TEXT` and `VALUE=UNKNOWN`.

> I think, it can be removed. 

I disagree. I think it should stay because it preserves that existing decoding behavior.

> Which tests fail when you remove this case?

Removing it makes both cases of `test_explicit_value_type_decodes_like_typed` fail.

src/icalendar/attr.py
```diff
@@ -458,14 +470,19 @@ def fget(self: Component) -> datetime | None:
...
+            values = []
+            for item in dt:
+                try:
+                    values.append(_decode_single_utc_value(name, item))
+                except (InvalidCalendar, ValueError):  # noqa: PERF203  # duplicate lists are tiny
+                    continue
+            if not values:
+                raise InvalidCalendar(f"{name} must be a datetime in UTC, not {dt}")
+            return min(tzp.localize_utc(value) for value in values)
+        return tzp.localize_utc(_decode_single_utc_value(name, dt))
```
niccokunzmann, line 485: Can this potentially raise a ValueError?

src/icalendar/attr.py
```diff
@@ -445,6 +445,18 @@ def fdel(self: Component):
     return property(fget, fset, fdel, doc)
 
 
+def _decode_single_utc_value(name: str, value: object) -> date | datetime:
```
niccokunzmann, line 448: Could you add a Raises: section here?

src/icalendar/attr.py
niccokunzmann: Ah. Ok. Thanks.

niccokunzmann (changes requested):
Thanks for the adjustments! I think, we are almost there, just documentation.

src/icalendar/attr.py
SemTiOne: Yes. Scalar `ValueError` from malformed `vText`/`vUnknown` already existed before this PR, while the duplicate path catches it so a malformed sibling does not block a valid value.

